PR #140, opened 2013-08-01
Add documentation ArrayDeclarationSniff

Konafets:
I updated this documentation to get an overview for myself about what that monster sniff actually do. 
The order of the violations in the documentation is the same order as in the code.

Please have also a look at this issue which I think its a bug in the sniff: https://pear.php.net/bugs/bug.php?id=20026


gsherwood:
I'm never going to get around to adding docs for this sniff before I break it out into many sniffs, so I'm closing this. Thanks for taking the time to write the docs and sorry I never got around to it.

Issue #311, opened 2014-11-07
Suppression of function prototype breaks checking of lines within function


abulford:
More generally, it appears that if the _@codingStandardsIgnoreEnd_ line is after a function's prototype and before the closing _}_ but the _@codingStandardsIgnoreStart_ is above the function, everything following the _@codingStandardsIgnoreEnd_ is treated as if it's not in the function.

This can be a problem if just the function prototype needs suppressing, but not the entire function, currently the only workaround is to suppress the entire function.

The following code demonstrates the issue:

```
<?php

namespace A;

// @codingStandardsIgnoreStart
class lowerCamelCaseClass
// @codingStandardsIgnoreEnd
{
    private $memberVariable;

    // @codingStandardsIgnoreStart
    public function _functionWithAnUnderscorePrefix()
    // @codingStandardsIgnoreEnd
    {
        $this->memberVariable = 'string';
        $this->memberVariable .= PHP_EOL;
        return $this->memberVariable;
    }
}
```

While the class name can be suppressed to prevent the _Class name "lowerCamelCaseClass" is not in camel caps format_ error, attempting to suppress the function name causes the three lines within the function to display a _Visibility must be declared on property "$this"_ error.

PR #312, opened 2014-11-07
Implementation of @codingStandardsIgnoreLine

abulford:
If this suppression is used, only the following line will be suppressed, without the need for a closing _@codingStandardsIgnoreEnd_ line.  This allows the suppression to occupy a little less space in the code, but was specifically implemented as a result of issue squizlabs/PHP_CodeSniffer#311.  Use of this suppression may also discourage large blocks of code being suppressed, or accidental suppression to the end of the file in the case where an _@codingStandardsIgnoreEnd_ is misspelt or forgotten.

Commit includes updated unit tests.  The first attempt at this resulted in _@codingStandardsIgnoreLine_ breaking any _@codingStandardsIgnoreStart_/_@codingStandardsIgnoreEnd_ suppressions it was nested within, so there's a specific test for this case, too.

Alternative names for the suppression include _@codingStandardsIgnore_ and _@codingStandardsIgnoreNext_.

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -417,11 +417,20 @@ class PHP_CodeSniffer_File
 
         $foundCode = false;
         $ignoring  = false;
+        $ignoreLine = false;
 
         // Foreach of the listeners that have registered to listen for this
         // token, get them to process it.
         foreach ($this->_tokens as $stackPtr => $token) {
             // Check for ignored lines.
+            // First check if this line should be ignored, due to a @codingStandardsIgnoreLine comment.
+            if ($ignoreLine) {
+                $ignoreLine = false;
+                $this->_ignoredLines[$token['line']] = true;
+                continue;
+            }
+
+            // Now check for other @codingStandards comments.
             if ($token['code'] === T_COMMENT || $token['code'] === T_DOC_COMMENT) {
                 if (strpos($token['content'], '@codingStandardsIgnoreStart') !== false) {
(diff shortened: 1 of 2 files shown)


Fuco1:
Would it make sense to extend the `_ignoredLines` array to not only hold values `true` but also the type of a warning/error we want to ignore? Ignoring all warnings is too coarse.

We have a bunch of sniffs that check cases where 99% of the time you should not do something, but sometimes you need to.  However, multiple warnings can be produced by one line and we only want to suppress the specific one.

The syntax could be `@codingStandardsIgnore code_of_warning code_of_another_warning`, the codes being optional.

Then in `_addWarning` we would not only check if the line is present, but also if the code of the warning matches some on the list.  If the value is present, we ignore it.  If no argument was specified, the value can stay `true` in which case we ignore all warnings.

It would also make sense (maybe) to add scope-dependent switches you could place on a class or a method.

aik099:
> We have a bunch of sniffs that check cases where 99% of the time you should not do something, but sometimes you need to. However, multiple warnings can be produced by one line and we only want to suppress the specific one.

Then I recommend you creating custom sniff, that would only match needed cases. What's the point in automatic detection fixing if sniff result needs to be manually ignored by human in some cases anyway.

Fuco1:
Because the decision is not one you can make automatically.  Halting problem and all.  Plus CS is very limited in it's abilities, it can't observe state of the entire project, just one file.

PR #370, opened 2014-12-14
Add configurable PHP namespace prefix for sniffs

klausi:
I want to use PHP namespaces for the Drupal coding standard to have reliable PSR-4 autoloading during test runs and when the standard is used as library.

PHPCS currently forces me to use the namespace "Drupal\Sniffs\Commenting", but I actually want "Drupal\coder\Drupal\Sniffs\Commenting" because the sniffs live in the coder library on drupal.org.

So I invented a nsPrefix config option so that I can specify "Drupal\coder\" in my ruleset.xml. Otherwise instantiating the sniff classes fails with a fatal error because there is the namespace mismatch. Is this the right approach?

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -1258,7 +1258,7 @@ class PHP_CodeSniffer
         // For sniff classes using a PHP namespace a prefix can be defined. Example:
         // 'Acme\' for a namespace 'Acme\MyStandard\Sniffs\Commenting'.
         $cliValues = $this->cli->getCommandLineValues();
-        $nsPrefix = isset($cliValues['nsPrefix']) ? $cliValues['nsPrefix'] : '';
+        $nsPrefix  = isset($cliValues['nsPrefix']) === true ? $cliValues['nsPrefix'] : '';
 
         foreach ($files as $file) {
             // Work out where the position of /StandardName/Sniffs/... is
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -1255,6 +1255,11 @@ class PHP_CodeSniffer
     {
         $listeners = array();
 
+        // For sniff classes using a PHP namespace a prefix can be defined. Example:
+        // 'Acme\' for a namespace 'Acme\MyStandard\Sniffs\Commenting'.
+        $cliValues = $this->cli->getCommandLineValues();
+        $nsPrefix = isset($cliValues['nsPrefix']) ? $cliValues['nsPrefix'] : '';
+
         foreach ($files as $file) {
             // Work out where the position of /StandardName/Sniffs/... is
             // so we can determine what the class will be called.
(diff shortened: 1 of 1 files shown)


gsherwood:
In the 3.0 branch, I've refactored everything to use namespaces, which also forces custom sniffs to pick a namespace. But it can be anything you want, as long as it ends with: STANDARD\Sniffs\CATEGORY

I haven't finished the migration guide for version 3.0 yet, so it might be a little too early to test. But I could help convert over your custom sniffs if you can point me to the repo they exist in.

I don't want to merge in any extra namespace code to 2.x though as it adds another thing that needs to be undone with 3.0. I'd rather see if 3.0 is working okay with your standard and push for a release once it is cleaned up.

aik099:
Would existing custom standards work by default with PHPCS 3.0 version if they don't use namespaces?

VasekPurchart:
@gsherwood does this mean that there will be no 2.4 release? I would really appriciate a 2.4 release with current fixes and improvements such as the new array format sniffs.

gsherwood:
> Would existing custom standards work by default with PHPCS 3.0 version if they don't use namespaces?

No. They must use namespaces and they must do some minor updating as functions and classes have changed. It's not a lot of work (I did it for hundreds of Sniffs in the core) but I'll have a guide with the details long before a stable release.

> does this mean that there will be no 2.4 release?

There will be a 2.4 release. And a 2.5, 2.6 etc. It takes the best part of a year to get a new major version out when it has such as big BC break.

klausi:
Thanks, PHPCS 3.0 probably also means I should create a new branch for Drupal Coder, which will be most likely 8.x-3.x.

The repo is on drupal.org, you can find all info on https://packagist.org/packages/drupal/coder

Do you have a script to convert the sniffs? Anyway, you can either submit a patch to https://drupal.org/project/issues/coder or a pull request to https://github.com/klausi/coder

gsherwood:
A new branch is a good idea. I've got to do a 2.x release this week, but I'll convert over that Drupal standard after that.

There is no script to convert them, but there are just a couple of things to look for. No major BC breaks like with commenting sniffs in 2.x; you just need to know what namespace all the classes are in now. Because I can drop PHP < 5.4, I finally got a chance to reorganise everything.

aik099:
> Because I can drop PHP < 5.4, I finally got a chance to reorganise everything.

That's something new. So since PHPCS 3.0 the PHP 5.3 won't be supported anymore?

gsherwood:
> That's something new. So since PHPCS 3.0 the PHP 5.3 won't be supported anymore?

That's the plan.

> And if server (where I run phpcs) has PHP 5.4, but code that is analyzed still uses PHP 5.3 it will be supported, right?

Yes.

klausi:
Ok, so let's wait for 3.0 to convert to namespaces.

PR #381, opened 2014-12-22
[FEATURE] Add markdown generator

Konafets:
Introduce a generator which produces documentation in Markdown format.


gsherwood:
Thanks for the contribution. I did make some changes to remove a bit more HTML and use markdown syntax, but that's inside the code comparison blocks. Nothing was being escaped in there so the formatting was a bit weird. But you can't highlight empty blocks using markdown (HTML does this using empty spans) so I just changed the code comparison blocks into PREs instead, using indents.

PR #388, opened 2014-12-24
[CLEANUP] Remove legacy code for PHP4 constructors

Konafets:
The removed code was used to compare the class name with the method name to skip PHP4 constructors.


gsherwood:
Thanks. I obviously missed that code when I removed the check.

PR #391, opened 2014-12-24
[FEATURE] Add a .editorconfig file

Konafets:
EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs.

Have a look at http://editorconfig.org/ for more information


gsherwood:
This isn't something I use, so I'm not going to include it in the project. If it becomes natively included in most IDEs and there is demand for it, I'll consider it again. Thanks for bringing it to my attention though.

Issue #393, opened 2014-12-26
 Generic.WhiteSpace.DisallowSpaceIndentSniff allows multiple whitespaces after tab in doc comments


Konafets:
Given the following code, which is mainly taken from the UnitTest:

```php 
// Doc comments are indent with tabs and one space
// [tab]/**
// [tab][space]*
// [tab][space][space]*
// [tab][space]*/

    /**
     *
      *
     */
```
Is this behavior intended? If yes, can we introduce a variable to define the allowed whitespaces? ```$multipleWhitespacesAllowed = true|false;```

And what is the reason that this sniff ignores file/class level doc comments?

gsherwood:
You probably want to look at the Squiz.Commenting.DocCommentAlignment sniff. Running that over your sample code produces:

```
$ phpcs temp.php --standard=Squiz --sniffs=Squiz.Commenting.DocCommentAlignment

FILE: temp.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 4 | ERROR | [x] Expected 1 space(s) before asterisk; 2 found
--------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------

Time: 8ms; Memory: 1.75Mb
```

aik099:
It turned out, that I was using that sniff in my standard as well, but totally forgot about this. I should have recommended it myself as well.

Konafets:
Thanks for recommending this sniff. We will use this sniff from now on too.

PR #395, opened 2014-12-26
[BUGFIX] Disallow multiple spaces in indented doc comments

Konafets:
Resolves: #393


gsherwood:
Please see my comment on #393

There is already a sniff to do this, and the one that checks if spaces are used for indenting should not be checking doc comment alignment. That is not the job of this particular sniff and would break a standard like PSR2, which is supposed to ignore comment blocks.

PR #398, opened 2014-12-27
Replace "%s" in "Squiz.ControlStructures.ForLoopDeclaration.SpacingBeforeClose" error

aik099:
The "SpacingBeforeClose" error in the "Squiz.ControlStructures.ForLoopDeclaration" sniff has non-replaced "%s" in it:
```
Expected %s spaces before closing bracket; %s found 
```
This PR replaces them with actual values.


gsherwood:
Thanks. You can tell I don't use that option :)

aik099:
I also started to use that sniff only recently when I realized that sniff that check for no whitespace before `;` doesn't check special case when there should be whitespace after `;` inside `for` loops.

PR #403, opened 2014-12-28
[FEATURE] Improve the doc comment detection.

Konafets:
Since the sniff registers the T_DOC_COMMENT_OPEN_TAG there is no need to use the findNext() method to get the end of the comment. The comment end tag is stored into the open tag token.


gsherwood:
Thanks for that change.

PR #409, opened 2014-12-29
[TASK] Configure Travis for faster builds

oliverklee:
Set sudo: false. This allows Travis use their new and faster
container-based infrastructure.


gsherwood:
Thanks. Made a bit of difference to the build times.

PR #417, opened 2014-12-30
[TASK] Fix typos in error messages

Konafets:


gsherwood:
Thanks for taking the time to do this. Even though we spell "tokenize" as "tokenise" in AU, I've accepted the change because of how PHP spells it. A sad day though :disappointed:

PR #427, opened 2015-01-08
fix #348 add option to treat newline as whitspace

### LINKED ISSUE #348: The "Squiz.WhiteSpace.OperatorSpacing" and new lines issue
When code around mathematical operation or any other operator is split across several lines then new line symbol or any other whitespace (except comment) should be considered as valid whitespace.

```php
$original_amount = $order->GetDBField('SubTotal') + $order->GetDBField('ShippingCost') +
			$order->GetDBField('VAT') + $order->GetDBField('ProcessingFee') + $order->GetDBField('InsuranceFee') -
			$order->GetDBField('GiftCertificateDiscount');
```

Currently one space specifically is asserted to be present.

euoia:
Fixes #348 by adding an option to treat newlines as whitespace.

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
@@ -48,6 +48,7 @@ class Squiz_Sniffs_WhiteSpace_OperatorSpacingSniff implements PHP_CodeSniffer_Sn
      */
     public $treatNewlineAsWhitespace = false;
 
+
     /**
      * Returns an array of tokens this test wants to listen for.
      *
@@ -142,8 +143,8 @@ class Squiz_Sniffs_WhiteSpace_OperatorSpacingSniff implements PHP_CodeSniffer_Sn
                     $found = $tokens[($stackPtr - 1)]['length'];
                 }
 
-                if ($this->treatNewlineAsWhitespace === false ||
-                    ($found !== 'newline' && $this->treatNewlineAsWhitespace === true)
+                if ($this->treatNewlineAsWhitespace === false
+                    || ($found !== 'newline' && $this->treatNewlineAsWhitespace === true)
                 ) {
                     $phpcsFile->recordMetric($stackPtr, 'Space before operator', $found);
                     if ($found !== 1) {
@@ -175,8 +176,8 @@ class Squiz_Sniffs_WhiteSpace_OperatorSpacingSniff implements PHP_CodeSniffer_Sn
 
                 $phpcsFile->recordMetric($stackPtr, 'Space after operator', $found);
                 if ($found !== 1) {
-                    if ($this->treatNewlineAsWhitespace === false ||
-                        ($found !== 'newline' && $this->treatNewlineAsWhitespace === true)
+                    if ($this->treatNewlineAsWhitespace === false
+                        || ($found !== 'newline' && $this->treatNewlineAsWhitespace === true)
                     ) {
                         $error = 'Expected 1 space after "&" operator; %s found';
                         $data  = array($found);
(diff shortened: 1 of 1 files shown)


gsherwood:
I can't merge this as-is because it is not recording metrics correctly with this change. A few of those IF statements need restructuring as well. I'll make the necessary corrections when I get to request #348.

euoia:
Thanks for looking at this. I have fixed the style errors reported by phpcs in Travis CI. I'm not sure what changes are required to record the metrics correctly, but I'm happy to make them.

gsherwood:
Sorry, I should have explained the metrics stuff. Those metrics lines are now being recorded conditionally, but they need to always be recorded as they describe the code rather than enforce specific standards.

euoia:
I made the change to always record the metrics. It seems to me though that perhaps the metrics *ought* to be recorded conditionally, since the condition in question changes what is considered to be a "space" (and the metric itself refers to a space).

aik099:
Doesn't look right. I think that we need to record actual used char (space or newline) as metric value. Currently it seems to write "space" all the time.

euoia:
I'm happy to make this change but I can't figure out how to get phpcs to report the metrics. Could you tell me how you're testing this? Thanks.

euoia:
Sorry I didn't mean click "close and comment".

aik099:
> but I can't figure out how to get phpcs to report the metrics. Could you tell me how you're testing this? Thanks.

There is special new report type `info` in PHPCS, that does that. See https://github.com/squizlabs/PHP_CodeSniffer/wiki/Reporting

euoia:
Ah got it, thanks. That report type doesn't show up in the help:
```
$ php scripts/phpcs --help
...
        <report>      Print either the "full", "xml", "checkstyle", "csv"
                      "json", "emacs", "source", "summary", "diff"
                      "svnblame", "gitblame", "hgblame" or "notifysend" report
                      (the "full" report is printed by default)
```

aik099:
> That report type doesn't show up in the help:

I think you should report that as a separate issue.

euoia:
> I think you should report that as a separate issue.
For sure, I'll do that.

I have made the improvements you suggested to recording the metrics. I've also made the message "Expected 1 space" say "Expected 1 space or newline" when $treatNewlineAsWhitespace is true.

euoia:
Well that wasn't good. I have tried to fix that error up. I'm having trouble running the tests locally.

aik099:
Error cause of the build: https://travis-ci.org/squizlabs/PHP_CodeSniffer/jobs/46759299#L138

Looks like `$data` parameter isn't passed to `->addError/->addWaring/->addFixableError/->addFixableWarning`, but `%s` is used in error message text.

euoia:
I fixed the error - it occurred because when you add a $data argument to add
FixableError the $error is passed through sprintf. In some of the tests the '%' operator is used and is treated as a format character when passed to sprintf.

gsherwood:
It looks like this metric stuff is confusing you. Metrics are the complete opposite of error messages. They are also incredibly important as I use them to generate my coding standard report data here: http://squizlabs.github.io/PHP_CodeSniffer/analysis/

So I'll look into this myself to ensure the data is being generated correctly and then regenerated for all the repos I collect data about, if needed.

euoia:
See my most recent commit. I'm now matching the behaviour on master where newlines result in a call to recordMetric with a third argument of "newline" instead of the number of spaces.

I also fixed a bug I introducted where a "space after operator" was being reported as a "space before operator".

Maybe there's something else I'm missing...

http://squizlabs.github.io/PHP_CodeSniffer/analysis/ is very nice!

gsherwood:
Thanks for all the effort on this PR, but it was getting more complicated than I thought the change needed to be. I especially didn't want the sniff to suggest that a newline was acceptable, and I can't have the metrics changed in the way you did them because of the way the info report works.

My commit is here: https://github.com/squizlabs/PHP_CodeSniffer/commit/bfbdc6a14bc643ab3d2d95e5d7228be8872f0e37

PR #430, opened 2015-01-14
Minor correction of CLI usage message

Pittiplatsch:


gsherwood:
I actually like the way it is printed now because I think it it is clear that there are 3 levels. How do you see this change as a correction?

Pittiplatsch:
It's the usual way to document optional cumulation of flags, which I've already seen in other *nix commands / programs - and especially in the case of verbosity, when the same flag is just multiplied as you do. IMO the `v[v[v]]` is a little more correct as each `v` is fixed on its position, where in `v[v][v]` it is ambiguous which `v` is the second - you can leave out either the 2nd or 3rd one.
Sure, functionally both notations are identically in this case of just multiplying the same flag, and granted: it's nitpicking ;-)

iammichiel:
I confirm the behaviour of the suggested change. For example, you can check the official php.net documentation which uses this style also.

Konafets:
:-1:  I am voting against this change. The current form is much more readable (and for me understandable) than the proposed one.

aik099:
:-1: also. For example default help message for verbosity flag of Symfony Console application looks like this:

```
--verbose (-v|vv|vvv) Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug
```

gsherwood:
I couldn't find any programs that use the -v[v[v]] syntax, or my -v[v][v] syntax. I really don't think it matters as it seems fairly rare.

All I want to convey is that there are 2 debug levels, and it seems like that is well understood.

If someone can show me a pattern of programs using this other syntax, I'll look at it again. But for now, I'm going to leave the docs as they are.

PR #436, opened 2015-01-16
Removing notices for undefined indexes in TODO/FIXME sniffs.

iammichiel:
When a TODO/FIXME message is empty, the sniff raises a notice when trying to access an undefined index of the array.

And adding a second commit removing a NOTICE send on empty function `@throws` comment.

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/Commenting/FixmeSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Commenting/FixmeSniff.php
@@ -71,11 +71,16 @@ class Generic_Sniffs_Commenting_FixmeSniff implements PHP_CodeSniffer_Sniff
         if (preg_match('/(?:\A|[^\p{L}]+)fixme([^\p{L}]+(.*)|\Z)/ui', $content, $matches) !== 0) {
             // Clear whitespace and some common characters not required at
             // the end of a fixme message to make the error more informative.
-            $type         = 'CommentFound';
-            $fixmeMessage = trim($matches[1]);
-            $fixmeMessage = trim($fixmeMessage, '-:[](). ');
-            $error        = 'Comment refers to a FIXME task';
-            $data         = array($fixmeMessage);
+            $fixmeMessage = '';
+            if (false === empty($matches)) {
+                $fixmeMessage = trim($matches[1]);
+                $fixmeMessage = trim($fixmeMessage, '-:[](). ');
+            }//end if
+
+            $type  = 'CommentFound';
+            $error = 'Comment refers to a FIXME task';
+            $data  = array($fixmeMessage);
+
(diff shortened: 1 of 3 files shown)


CodeSniffer/Standards/Generic/Sniffs/Commenting/FixmeSniff.php
```diff
@@ -72,7 +72,7 @@ public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
             // Clear whitespace and some common characters not required at
             // the end of a fixme message to make the error more informative.
             $type         = 'CommentFound';
-            $fixmeMessage = trim($matches[1]);
+            $fixmeMessage = trim(@$matches[1]);
```
aik099: Hiding errors is bad practice especially within a library, that enforces good practice.

I recommend doing proper array key existence check.

CodeSniffer/Standards/Generic/Sniffs/Commenting/TodoSniff.php
```diff
@@ -70,7 +70,7 @@ public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
             // Clear whitespace and some common characters not required at
             // the end of a to-do message to make the warning more informative.
             $type        = 'CommentFound';
-            $todoMessage = trim($matches[1]);
+            $todoMessage = trim(@$matches[1]);
```
aik099: Hiding errors is bad practice especially within a library, that enforces good practice.

I recommend doing proper array key existence check.

iammichiel:
Did a proper check and validated PHPCS checkstyle.

gsherwood:
I dont get notices for empty FIXME and TODO comments. Can you please provide the sample code you are testing on that produces the notices. I might also need to know the PHP version.

iammichiel:
Well you can reproduce it on any code snippet with empty content. 

```
<?php 

/**
 * @throws
 */
function testFunction() {
    throw new \Exception(); 
    // TODO
    // FIXME
}
```

Tested on two different computers (linux/mac) 
PHP version : 5.5.20
However error, reporting was defined to `E_ALL`

iammichiel:
My bad. When trying to find a working example I actually found the **real** error. 
In fact, the problem is the use of the preg_match expression. As defined in the official PHP documentation : 

> preg_match() returns 1 if the pattern matches given subject, 0 if it does not, or FALSE if an error occurred.

However the condition used in the sniffs is : 

```
 if (preg_match('/(?:\A|[^\p{L}]+)fixme([^\p{L}]+(.*)|\Z)/ui', $content, $matches) !== 0) {
```

Which will in case of FALSE, enter the block code. Which it should not. I will submit a new PR correcting this behaviour.

PR #440, opened 2015-01-16
- checking if regexp matches before using it, because I encountered an e...

lechbaczynski:
...rror Undefined offset: 1 in /usr/share/php/PHP/CodeSniffer/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php on line 247

The diff looks ugly, but I only added if () around that block of code

Encountered error:

PHP Code Sniffer
phpcs: PHP Notice: Undefined offset: 1 in /usr/share/php/PHP/CodeSniffer/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php on line 247
PHP Stack trace:
PHP 1. {main}() /usr/bin/phpcs:0
PHP 2. PHP_CodeSniffer_CLI->runphpcs() /usr/bin/phpcs:25
PHP 3. PHP_CodeSniffer_CLI->process() /usr/share/php/PHP/CodeSniffer/CLI.php:91
PHP 4. PHP_CodeSniffer->processFiles() /usr/share/php/PHP/CodeSniffer/CLI.php:789
PHP 5. PHP_CodeSniffer->processFile() /usr/share/php/PHP/CodeSniffer.php:615
PHP 6. PHP_CodeSniffer->_processFile() /usr/share/php/PHP/CodeSniffer.php:1534

at comment:

  * @param $nrWniosku

using
phpcs 2.0.RC1 beta
and PHPStorm

### DIFF
--- a/CodeSniffer/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/CodeSniffer/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
@@ -225,47 +225,48 @@ class PEAR_Sniffs_Commenting_FunctionCommentSniff implements PHP_CodeSniffer_Sni
             $comment   = '';
             if ($tokens[($tag + 2)]['code'] === T_DOC_COMMENT_STRING) {
                 $matches = array();
-                preg_match('/([^$&]+)(?:((?:\$|&)[^\s]+)(?:(\s+)(.*))?)?/', $tokens[($tag + 2)]['content'], $matches);
-
-                $typeLen   = strlen($matches[1]);
-                $type      = trim($matches[1]);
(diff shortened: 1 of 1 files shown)


aik099:
Does it work in PHP_CodeSniffer 2.1.0 version?

gsherwood:
I can't reproduce a failing case on any PHPCS version, Can you please post the function comment and the function definition that it causing this issue so I can build a unit test with it?

lechbaczynski:
The problem is that we cannot reproduce it either :) Sorry.
Anyway, checking if the string matches might be a godd idea before usung $matches[1].
On the other hand, now I reased that my fix was wron, as it might silently overlook "wrong" comments that shoud be sniffed.

Issue #439, opened 2015-01-16
Extend the sniff PEAR.Commenting.ClassComment to Interfaces and Traits


Konafets:
Does it makes sense to do this? If yes, I can add a patch.

aik099:
I think it does. I've also seen some sniffs ignoring traits and this was also fixed in previous releases.

Konafets:
Ah, yes, it listen to interfaces but not for traits and the error messages say "Missing class doc comment" instead of "Missing class/interface/trait doc comment".

aik099:
You can use `$tokens[$stackPtr]['content']` to get `class`, `interface` or `trait` word in your error message.

Konafets:
I do in my custom sniff. Just thought it might be a good idea for the the shipped sniff as well.

gsherwood:
If you submit a PR for this, I will merge it. I think this sniff can do all 3.

gsherwood:
Closing as their is a PR for this, although it can't yet be merged in.

PR #441, opened 2015-01-16
[FEATURE] Extend the sniff PEAR.Commenting.ClassComment to Traits

Konafets:
Resolves: #439


Konafets:
LOL. Its failing because of the inexistence of Traits in PHP 5.3.

gsherwood:
Sorry, finally got time to look at this.

First thing, recordMetric doesn't accept an array of error data, so all those recordMetric calls are going to be recording incorrect string values. You can see this by running `--report=info` over your code. Instead, just do the string sub yourself:
```php
$phpcsFile->recordMetric($stackPtr, ucfirst($type).' has doc comment', 'no');
```
Which leads to the second point; the metric currently collected doesn't bother differentiating between classes and interfaces. Adding the checking of traits will now change the way this metric is recorded - splitting it into 2 different metrics. This is more a reminder to me that I need to regen all all data for another project that uses this metric data.

gsherwood:
Closing due to changes not being implemented. It's been over a year, so I think this is done.

PR #446, opened 2015-01-19
A better version of https://github.com/squizlabs/PHP_CodeSniffer/pull/440

lechbaczynski:
A better version of https://github.com/squizlabs/PHP_CodeSniffer/pull/440 

I understand that you may not want to check whether the pattern matches, as in that place it probably should match.

Also, checking if pattern matche will need changes on many other places too (like: 

protected function processThrows(PHP_CodeSniffer_File $phpcsFile, $stackPtr, $commentStart)
    {
        $tokens = $phpcsFile->getTokens();

        $throws = array();
        foreach ($tokens[$commentStart]['comment_tags'] as $tag) {
            if ($tokens[$tag]['content'] !== '@throws') {
                continue;
            }

            $exception = null;
            $comment   = null;
            if ($tokens[($tag + 2)]['code'] === T_DOC_COMMENT_STRING) {
                $matches = array();
                preg_match('/([^\s]+)(?:\s+(.*))?/', $tokens[($tag + 2)]['content'], $matches);
                $exception = $matches[1];

### DIFF
--- a/CodeSniffer/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/CodeSniffer/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
@@ -267,6 +267,10 @@ class PEAR_Sniffs_Commenting_FunctionCommentSniff implements PHP_CodeSniffer_Sni
                         $phpcsFile->addError($error, $tag, 'MissingParamName');
                     }//end if
                 }
+                else {
+                    $error = 'Parameter syntax wrong';
+                    $phpcsFile->addError($error, $tag, 'MissingParamType');
+                }
             } else {
                 $error = 'Missing parameter type';
                 $phpcsFile->addError($error, $tag, 'MissingParamType');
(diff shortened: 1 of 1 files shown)


aik099:
This particular error with PEAR FunctionComment sniff and missing `@param` type was fixed in 2.1.0 release if I'm not mistaken. Or at least in `master` branch.

Can you please check that instead?

iammichiel:
Hey, I have encoutered the same mistake as you describe here and in your previous request. The problem is actually quite simple. Checkout this line : 

```
if (isset($matches[2]) === true) {
    $comment = $matches[2];
}
```

The isset function checks if the variable is defined, not its content. So when the string is empty, the isset function will return true. However, the following line will fail as the $comment variable will be empty : 
```
$firstChar = $comment{0};
```

The problem was corrected by the following commit : 7012042683ead777fc96cd2525e7d419b173b6c3

gsherwood:
Seems this is fixed.

PR #447, opened 2015-01-19
Changing the regex check in the FIXME and TODO sniffs.

iammichiel:
The problem is the use of the preg_match expression. As defined in the official PHP documentation : 

> preg_match() returns 1 if the pattern matches given subject, 0 if it does not, or FALSE if an error occurred.

However the condition used in the sniffs is : 

```
 if (preg_match('/(?:\A|[^\p{L}]+)fixme([^\p{L}]+(.*)|\Z)/ui', $content, $matches) !== 0) {
```

Which will in case of FALSE, enter the block code. Which it should not. This generates a lot of useless PHP notices : 

```
Notice: Undefined offset: 1 in /apps/test/vendor/squizlabs/php_codesniffer/CodeSniffer/Standards/Generic/Sniffs/Commenting/TodoSniff.php on line 73
PHP Notice:  Undefined offset: 1 in /apps/test/vendor/squizlabs/php_codesniffer/CodeSniffer/Standards/Generic/Sniffs/Commenting/TodoSniff.php on line 73
```

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/Commenting/FixmeSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Commenting/FixmeSniff.php
@@ -68,7 +68,7 @@ class Generic_Sniffs_Commenting_FixmeSniff implements PHP_CodeSniffer_Sniff
 
         $content = $tokens[$stackPtr]['content'];
         $matches = array();
-        if (preg_match('/(?:\A|[^\p{L}]+)fixme([^\p{L}]+(.*)|\Z)/ui', $content, $matches) !== 0) {
+        if (preg_match('/(?:\A|[^\p{L}]+)fixme([^\p{L}]+(.*)|\Z)/ui', $content, $matches) === 1) {
             // Clear whitespace and some common characters not required at
             // the end of a fixme message to make the error more informative.
             $type         = 'CommentFound';
--- a/CodeSniffer/Standards/Generic/Sniffs/Commenting/TodoSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Commenting/TodoSniff.php
@@ -66,7 +66,7 @@ class Generic_Sniffs_Commenting_TodoSniff implements PHP_CodeSniffer_Sniff
 
         $content = $tokens[$stackPtr]['content'];
         $matches = array();
-        if (preg_match('/(?:\A|[^\p{L}]+)todo([^\p{L}]+(.*)|\Z)/ui', $content, $matches) !== 0) {
+        if (preg_match('/(?:\A|[^\p{L}]+)todo([^\p{L}]+(.*)|\Z)/ui', $content, $matches) === 1) {
             // Clear whitespace and some common characters not required at
             // the end of a to-do message to make the warning more informative.
             $type        = 'CommentFound';
--- a/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -181,10 +181,10 @@ class Squiz_Sniffs_Commenting_FunctionCommentSniff extends PEAR_Sniffs_Commentin
                 }
             }
 
-            if ($exception === null) {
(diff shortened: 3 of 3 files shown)


aik099:
What `$content` causes error to happen?

iammichiel:
So far, I have not identified a pattern in the failing strings. However, a recurring element is wrongly encoded chars. Seems to be a mixin of encodings...

aik099:
Are there any other similar `preg_match` checks in this/other sniffs that are checked for `!== 0` and might lead to similar issues?

iammichiel:
For the `preg_match` check I didn't find any other occurences. (Tested on a legacy app with several millions LOC). However, another error appeared on the `@throws` annotation. It has been corrected in the last commit.

aik099:
Maybe https://github.com/iammichiel/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php#L184 line should also be fixed to use `empty` because if exception is there but empty (I have no idea what code might produce such result), then null check won't work either.

iammichiel:
You are right. I am adding it to the commit.

gsherwood:
I can't merge this in without failing test cases. The todo/fixme sniffs I'm not so worried about. But I absolutely need to find test cases for the FunctionCommentSniff. I'll have a play around and see if I can get some failing code.

gsherwood:
I've tried a mix of lots of different encodings and still can't get any failures. I'll need to leave this PR here until someone can find a bit of failing code for these changes.

aik099:
Maybe you can try putting some Russian text in UTF-8 MySQL column and then reading this text using connection to MySQL server, where latin1 is set as encoding.

gsherwood:
> Maybe you can try putting some Russian text in UTF-8 MySQL column and then reading this text using connection to MySQL server, where latin1 is set as encoding.

If the content can't be hand-written, it isn't going to end up as a comment in a PHP file. So I'm looking for some content that the preg_match function fails to process. preg_match mostly returns FALSE when the regex itself is broken, or if you hit some sort of internal limit. But I don't think that is the case here.

Presumably, a problem must exist somewhere for it to be reported. What I'm looking for are some test cases to go with the fix. If only one person can get this to fail on one project, the test cases are probably going to have to come from there.

iammichiel:
I finally isolated the problem. Took me a while. The problem appears when the file it is trying to analyze has different line-endings. Mixing both LF and CRLF. The preg_match function will behave strangely. 

I'll try to produce a test-case.

gsherwood:
> The problem appears when the file it is trying to analyze has different line-endings. Mixing both LF and CRLF.

I've had a try with mixed line endings after various comments are still can't get it to fail. I also having the EOL chars after the comments different from the first EOL char (so PHPCS might leave the `\r` or the `\n` in the comment) but that isn't causing any failures either.

I can get PHPCS to produce this core warning for me:
` 1 | WARNING | [ ] File has mixed line endings; this may cause incorrect results`

So I know PHPCS has detected the mixed line endings, but I'm still stuck trying to replicate a problem. Hopefully you have more luck.

slitvyachenko:
http://www.filedropper.com/pregmatchreturnsfalse
Hi! You can download a zip-archive in which there are 2 files that are causing an error:

An error occurred during processing; checking has been aborted. The error message was: "Notice: Undefined offset: 1 in squizlabs\php_codesniffer\CodeSniffer\Standards\Generic\Sniffs\Commenting\FixmeSniff.php line 75" at squizlabs\php_codesniffer\CodeSniffer\Standards\Generic\Sniffs\Commenting\FixmeSniff.php:75

I can send these files on your email if the above link is broken.
Thank you!

gsherwood:
@slitvyachenko Thanks for sending those files through. I have committed a fix for the todo and fixme sniffs, but as soon as I open those files in an editor, it changes the file and fixes the problem. So I havent been able to use them to create unit tests or check other issues.

Do these files have a specific encoding that I should use to open them? What is that character that is causing the issue here?

slitvyachenko:
@gsherwood I didn't investigate this moment. And my editor (phpstorm) also fixes those files when I open them. I had the previous error a few times on different files.
But now all works as expected. Thank you!

PR #449, opened 2015-01-19
use json_encode for encoding strings

pscheit:
this is a quickfix for encoding messages with tabs and newlines. I don't understand enough of this part of code, why you couldn't use json_encode for all the string concatenation?

### DIFF
--- a/CodeSniffer/Reports/Json.php
+++ b/CodeSniffer/Reports/Json.php
@@ -61,11 +61,9 @@ class PHP_CodeSniffer_Reports_Json implements PHP_CodeSniffer_Report
         foreach ($report['messages'] as $line => $lineErrors) {
             foreach ($lineErrors as $column => $colErrors) {
                 foreach ($colErrors as $error) {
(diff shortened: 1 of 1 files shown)


aik099:
I guess because the `json_encode` function was introduced only since PHP 5.2.0 and before it was only available with optional PECL extension (see http://php.net/manual/en/function.json-encode.php).

Libraries that were created before that time usually have own json encoding system. Since this library states that it's compatible with PHP 5.1.2 (from `composer.json`) then I think that some kind of `json_encode` polyfill needs to be created for cases, when `json_encode` function is missing.

aik099:
@gsherwood , while we're at it maybe we can raise minimal required PHP version to 5.3.2?

gsherwood:
I can't use json_encode because it is not available in the min PHP version. I will not be raising the min PHP version until version 3.0, which has no release date, so I can't accept this PR.

Once I do raise the min version, I'll be changing this, and a lot more.

PR #462, opened 2015-01-28
PHP_CodeSniffer should depends on the XMLWriter extension

mathroc:
As a few reporters uses the XMLWriter extension. I suggest to add this in composer.json to warn users as soon as possible

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,8 @@
     },
     "require": {
         "php": ">=5.1.2",
-        "ext-tokenizer": "*"
+        "ext-tokenizer": "*",
+        "ext-xmlwriter": "*"
     },
     "bin": [
         "scripts/phpcs",


gsherwood:
Thanks

jian-wu:
Shouldn't this rather be using suggest instead?  This change immediately breaks users' composer install/update if they don't have xmlwriter in their require block.

```
"suggest": {
    "ext-xmlwriter": "Required if you are using XML reporter."
},

**Comment by gsherwood** [comment]
> Shouldn't this rather be using suggest instead? This change immediately breaks users' composer install/update if they don't have xmlwriter in their require block.

I'm not a big composer users, so excuse my ignorance, but why would you need to include PHP_CodeSniffer's required packages in your own `require` block if PHPCS includes it in it's `require` block? You don't need to include ext-tokenizer in your 'require' right now for PHPCS to work correctly, so is there a difference with ext-xmlwriter?

**Comment by jian-wu** [comment]
You are correct about composer's inheritance of the require packages.  However, I was getting this error when I switched to use dev-master from 2.2.0.
```
Problem 1
    - Installation request for squizlabs/php_codesniffer dev-master -> satisfiable by squizlabs/php_codesniffer[dev-master].
    - squizlabs/php_codesniffer dev-master requires ext-xmlwriter * -> the requested PHP extension xmlwriter is missing from your system.
```
I'm not sure if that was caused by some other package's dependencies or a onetime hiccup.  I apparently cannot reproduce the same error.

Issue is resolved since it doesn't exist anymore.  I apologize for causing any inconvenience.  If it happens again, I'll definitely write down the steps to reproduce it.

Thanks.

gsherwood:
Thanks for testing it out and getting back to me.

PR #464, opened 2015-01-29
New features in operator spasing

kalinin-k-a:
* Some refactoring in the place of processing unary minus.
* Appended boolean operators into the list of operators wich allow use the unary minus after them (true || -1 == $b) was wrong. But now it's ok.
* Appended flag for allowing use multispaces for allignment (see details in phpdoc near allowMultispaceForAssignmentAlignment property in Squiz_Sniffs_WhiteSpace_OperatorSpacingSniff)
* Appended options to determine after/before wich of operators is newline allowed (see details in phpdoc near propeties allowNewLineAfter and allowNewLineBefore)

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
@@ -112,7 +112,7 @@ class Squiz_Sniffs_WhiteSpace_OperatorSpacingSniff implements PHP_CodeSniffer_Sn
      * Here is the unary minus. It must not require space after himself.
      * @var array
      */
-    protected $tokensThatMayGoBeforeUnaryMinus = [];
+    protected $tokensThatMayGoBeforeUnaryMinus = array();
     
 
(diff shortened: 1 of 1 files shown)


gsherwood:
I can't merge this change in. It doesn't follow the coding standards and has no tests for something so complicated. I really don't think I want to add these features to this core sniff either. These are better off being in your own custom sniff. I think they are far too complex for this one.

It also looks like there is a bug fix in here. From what I can tell of the description, I think this is the bit of code that is failing:
```php
if (true || -1 == $b) {
}
```

Is this the only code that you have found errors with?

gsherwood:
I've committed a fix for the code in my previous comment.

kalinin-k-a:
Sorry for long-time offline. 

> I think this is the bit of code that is failing:
> if (true || -1 == $b) {

Yes, this bugfix was a part of commit.
> I've committed a fix for the code in my previous comment.
Thank you.

> These are better off being in your own custom sniff. I think they are far too complex for this one.
Ok, I'll put it into a separate sniff. Thank you for feedback

gsherwood:
Thanks for getting back to me.

PR #469, opened 2015-02-02
Fix Inspection Error

zebraf1:
The code should follow same syntax in the whole file. You should not use PHP variables inside double quoted string.
Using this syntax  "$param{" causes an inspection error:
Expected: right double quote | Expecting statement

a) Concatenate string
b) Escape bracket: "$param\\{"

I have made a PR for version a

### DIFF
--- a/CodeSniffer/Reports/Json.php
+++ b/CodeSniffer/Reports/Json.php
@@ -54,7 +54,7 @@ class PHP_CodeSniffer_Reports_Json implements PHP_CodeSniffer_Report
         $filename = str_replace('\\', '\\\\', $report['filename']);
         $filename = str_replace('"', '\"', $filename);
         $filename = str_replace('/', '\/', $filename);
-        echo "\"$filename\":{";
+        echo '"'.$filename.'":{';
         echo '"errors":'.$report['errors'].',"warnings":'.$report['warnings'].',"messages":[';
 
         $messages = '';


aik099:
Can you please add some context. For example:

1. PHP file caused the issue
2. standard used to validate the code
3. output before your change
4. output after your change

Right it's kind of hard to see whole picture with only pieces of information grabbed from all places.

aik099:
And looking at file you've changed (the JSON inspection result reporter) the issue title clearly doesn't correlate with that.

zebraf1:
Sorry for my inadequate explanation.

First of all the code works and there is no bug. The issue is that my IDE (PHPStorm) reports that there is a code error inside this file Json.php on this line:
```php
echo "\"$filename\":{";
```
I'm using PHP 5.5 interpreter and it doesn't like that the variable is used in the string with the bracket { unescaped. Brackets are used for escaping variables in double quoted strings: "{$filename}".

The rest of the code uses single quotes for string, I am only suggesting to use the same coding standard on that line also.

In addition I noticed that this file has duplicate PHP doc for the PHP_CodeSniffer_Reports_Json class.

aik099:
Interesting why double quotes are used here in first place. But it seems that your fix doesn't break anything or output of reports isn't covered by any tests.

My rule of thumb regardless single quotes, double quotes and escaping is: it's better to change double quotes into single quotes instead of escaping every double quote inside double quotes.

gsherwood:
There is nothing wrong with the syntax of this code. PHPStorm obviously has a bug, that maybe you should report.

Still, I'll merge this in because it's easier to read with your change. Thanks for submitting it.

zebraf1:
Yeah, I guess there is a bug with the inspector in PHPStorm. Thanks for merging though.

PR #481, opened 2015-02-08
[BUGFIX] Fix the aligning of doc comment start after rewrite

Konafets:
The trailing comment closer will be pushed to the next line but
a whitespace at the beginning is missing.

Current behavior:
```
-/** @var Database $mockedDatabase */
+/**
+ * @var Database $mockedDatabase
+*/
```

New behavior
```
-/** @var Database $mockedDatabase */
+/**
+ * @var Database $mockedDatabase
+ */
```

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/Commenting/DocCommentSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Commenting/DocCommentSniff.php
@@ -95,7 +95,10 @@ class Generic_Sniffs_Commenting_DocCommentSniff implements PHP_CodeSniffer_Sniff
             $error = 'The close comment tag must be the only content on the line';
             $fix   = $phpcsFile->addFixableError($error, $commentEnd, 'ContentBeforeClose');
             if ($fix === true) {
+                $phpcsFile->fixer->beginChangeset();
+                $phpcsFile->fixer->addContentBefore($commentEnd, ' ');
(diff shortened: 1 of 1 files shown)


aik099:
:+1: 

In meanwhile you should include https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Sniffs/Commenting/DocCommentAlignmentSniff.php sniff in the standard to get auto-alignment fix.

gsherwood:
Most sniffs don't fix alignment when they fix code so that blanket rules can be enforced by other sniffs. As a general rule: if the sniff doesn't check it, it shouldn't fix it. It can end up conflicting with other sniffs, including custom sniffs.

In this case, putting a single space in front of the closer is not going to achieve alignment in many cases. You would need to actually implement the fixing technique in the DocCommentAlignment sniff to ensure everything looks correct. At that point, it's easier to just include the DocCommentAlignment sniff in your standard and let that do the fixing for you.

Konafets:
I just wonder that on line 219 the sniff takes care about indention by itself and adds a space indention to the new line.

https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Generic/Sniffs/Commenting/DocCommentSniff.php#L219

gsherwood:
When I break a line up, I sometimes try and keep the same alignment, especially if I know other sniffs may not fix it.

gsherwood:
Forgot to close this.

PR #482, opened 2015-02-09
Check if Sniff is abstract before registering it

### LINKED ISSUE #476: Ignore abstract sniff classes
It appears that if you define your own coding standard, then *all* classes, including abstract classes, in the `/Sniffs` directory that end with 'Sniff' are improperly instantiated when running `phpcs`.

There should be a check for whether a class is abstract before it is instantiated, even if its name ends in 'Sniff'. This could be useful if someone wants to maintain a naming convention in that namespace, but still make a class abstract (i.e. `BaseMyStandardSniff`) rather than changing it to `BaseMyStandardHelper` or something of the sort.

Decave:
Closes #476

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -1319,6 +1319,12 @@ class PHP_CodeSniffer
                 $className = $classNameNS;
             }
 
+            // Skip abstract classes.
+            $reflection = new ReflectionClass($className);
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for the contribution.

PR #487, opened 2015-02-11
Fix erroneus whitespace error when token contents have value "false"

rp4:
This error can be reproduced with accented letters, with PSR2 standards:

```
<?php
/**
 * @author Tést mé
 */
```

/vendor/bin/phpcs -s --standard=psr2 --encoding=utf-8 <target>

```
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
 4 | ERROR | [x] Whitespace found at end of line
   |       |     (Squiz.WhiteSpace.SuperfluousWhitespace.EndLine)
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
```

This it seems to happen when the content value is "false", but rtrim() casts it to string and returns an empty string.

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
@@ -218,7 +218,7 @@ class Squiz_Sniffs_WhiteSpace_SuperfluousWhitespaceSniff implements PHP_CodeSnif
                         $phpcsFile->fixer->replaceToken($stackPtr, rtrim($tokenContent).$phpcsFile->eolChar);
                     }
                 }
-            } else if (!empty($tokens[($stackPtr - 1)]['content'])
+            } else if (empty($tokens[($stackPtr - 1)]['content']) === false
                 && $tokens[($stackPtr - 1)]['content'] !== rtrim($tokens[($stackPtr - 1)]['content'])
                 && $tokens[($stackPtr - 1)]['line'] === $tokens[$stackPtr]['line']
             ) {
--- a/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
@@ -218,7 +218,8 @@ class Squiz_Sniffs_WhiteSpace_SuperfluousWhitespaceSniff implements PHP_CodeSnif
                         $phpcsFile->fixer->replaceToken($stackPtr, rtrim($tokenContent).$phpcsFile->eolChar);
                     }
                 }
-            } else if ($tokens[($stackPtr - 1)]['content'] !== rtrim($tokens[($stackPtr - 1)]['content'])
+            } else if (!empty($tokens[($stackPtr - 1)]['content'])
+                && $tokens[($stackPtr - 1)]['content'] !== rtrim($tokens[($stackPtr - 1)]['content'])
                 && $tokens[($stackPtr - 1)]['line'] === $tokens[$stackPtr]['line']
             ) {
                 $fix = $phpcsFile->addFixableError('Whitespace found at end of line', ($stackPtr - 1), 'EndLine');


aik099:
But then there is another problem in tokenizer when token's `'content'` parameter value can be `false`, which doesn't naturally happen. What you're fixing only solves that problem side effect.

rp4:
Thanks for the comment. I will try to investigate the root cause.

Also, it might help to know that this is happening on PHP 5.3.3 but not on PHP 5.4.36-0.

gsherwood:
I cant replicate this issue on PHP5.6, 5.5 or 5.3.29 (I couldn't install 5.3.3 specifically). My comment is always tokenized like this:

```
*** START COMMENT TOKENIZING ***
Create comment token: T_DOC_COMMENT_OPEN_TAG => /**
Create comment token: T_DOC_COMMENT_WHITESPACE => \n
Create comment token: T_DOC_COMMENT_WHITESPACE => ·
Create comment token: T_DOC_COMMENT_STAR => *
Create comment token: T_DOC_COMMENT_WHITESPACE => ·
Create comment token: T_DOC_COMMENT_TAG => @author
Create comment token: T_DOC_COMMENT_WHITESPACE => ·
Create comment token: T_DOC_COMMENT_STRING => Tést·mé
Create comment token: T_DOC_COMMENT_WHITESPACE => \n
Create comment token: T_DOC_COMMENT_WHITESPACE => ·
Create comment token: T_DOC_COMMENT_CLOSE_TAG => */
*** END COMMENT TOKENIZING ***
```

And tokens should never have empty content, so this would need to be a tokenizer error.

Can you run `/vendor/bin/phpcs -s --standard=psr2 --encoding=utf-8 -vv` and paste the output of your tokenized comment?

rp4:
Hi,

this is the output of the tokenised comment:
                *** START COMMENT TOKENIZING ***
                Create comment token: T_DOC_COMMENT_OPEN_TAG => /**
                Create comment token: T_DOC_COMMENT_WHITESPACE => \n
                Create comment token: T_DOC_COMMENT_WHITESPACE => ·
                Create comment token: T_DOC_COMMENT_STAR => *
                Create comment token: T_DOC_COMMENT_WHITESPACE => ·
                Create comment token: T_DOC_COMMENT_TAG => @author
                Create comment token: T_DOC_COMMENT_WHITESPACE => ·
                Create comment token: T_DOC_COMMENT_STRING => Tést·mé
                Create comment token: T_DOC_COMMENT_WHITESPACE => \n
                Create comment token: T_DOC_COMMENT_STRING => ·*/
                Create comment token: T_DOC_COMMENT_CLOSE_TAG =>
                *** END COMMENT TOKENIZING ***

I have patched it successfully using iconv_* functions instead of normal string functions, but I see they are handled with care in the code, so I am not too sure about the viability of the solution - especially since this bug seems to be fixed in later versions of PHP.

aik099:
@rp4 you can add 3 tilde symbols on new line before/after the code to use monospaced font. It's called Fenced code blocks on GitHub.

gsherwood:
Thanks for posting your output. You can see that tokenized output is different, so the error lies in the comment tokenizer and not the sniff. Merging this change in wont actually fix the underlying error, so I'll need to do a bit more investigation to see if there is any way the tokenizer can cause this.

gsherwood:
I can't see a reason why using the iconv functions in the comment tokenizer would make a difference. The fact that the `substr` and `strlen` functions are counting multi-byte characters shouldn't cause any issues in this bit of code because it doesn't really care about accurate lengths here.

Given that I can't replicate this behaviour in the latest 3.3 version, or any newer version of PHP, I don't think I can properly fix this. Adding all the iconv checks is only going to slow things down for all comment checking, even though the problem doesn't occur for others.

So I think I'll have to close this without a code resolution. The fix is really to upgrade to a newer supported version of PHP, or a later version of 5.3.

If you have any more info to help me replicate or fix this issue, please continue to comment and I'll reopen if anything new comes up.

rp4:
Thanks for taking your time looking at this issue.
~~~substr~~ and ~~~strlen~~~ *should* count multi-byte characters, but for some reasons in one environment I am running they yield different results.

I have tried to set-up a clean VM with PHP 5.3.3 (and same php.ini) and I cannot reproduce this, so it must be something very specific to the configuration I am running.

gsherwood:
I'm not sure why that would happen. The only thing I can think of is if you are using `mbstring.func_overload` on that particular PHP install, although I don't know why that would actually break anything.

Thanks for taking the time to do so much testing.

PR #488, opened 2015-02-11
switch back to the original error_reporting level

gimler:
Instate of switching to a high error reporting level switch back to last error reporting level

### DIFF
--- a/CodeSniffer/DocGenerators/HTML.php
+++ b/CodeSniffer/DocGenerators/HTML.php
@@ -188,12 +188,13 @@ class PHP_CodeSniffer_DocGenerators_HTML extends PHP_CodeSniffer_DocGenerators_G
     {
         // Turn off errors so we don't get timezone warnings if people
         // don't have their timezone set.
+        $errorLevel = error_reporting();
         error_reporting(0);
         echo '  <div class="tag-line">';
         echo 'Documentation generated on '.date('r');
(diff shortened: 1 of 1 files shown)


CodeSniffer/DocGenerators/HTML.php
```diff
@@ -188,12 +188,13 @@ protected function printFooter()
     {
         // Turn off errors so we don't get timezone warnings if people
         // don't have their timezone set.
+        $errorLevel = error_reporting();
         error_reporting(0);
```
aik099: Just do `$errorLevel = error_reporting(0);` instead of adding extra line because `error_reportring` function, if I'm not mistaken always returns previous error level.

aik099:
:+1: , I wonder how you're able to find this place. The only reason would be, when PHP_CodeSniffer's PHP files are included in other project that had some errors in it.

CodeSniffer/DocGenerators/HTML.php
gimler: you are tight i have changed this

gsherwood:
If you include CLI.php to actually run PHPCS, error reporting is set for you. CLI is required for any sort of checking or reporting functionality, so it would be pretty hard to get the error reporting level set to anything other than `E_ALL | E_STRICT`.

But I'll still merge the code in because it wont break anything. I assume PHP_CodeSniffer must be being used in a different way in this case, causing a problem.

PR #491, opened 2015-02-13
Generic InlineControlStructureSniff can correct with missing semicolon

donatj:
When using `phpcbf` with `Generic.ControlStructures.InlineControlStructure` on the following code:

```php
<label><input type="radio" <?php if($series) echo 'checked="checked"' ?> > By Series</label>
```

it was transformed into the following which is ***not valid*** PHP

```php
<label><input type="radio" <?php if($series) { echo 'checked="checked"' } ?> > By Series</label>
```

I have altered the correction to include a semicolon if the end is not one.

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/ControlStructures/InlineControlStructureSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/ControlStructures/InlineControlStructureSniff.php
@@ -171,6 +171,10 @@ class Generic_Sniffs_ControlStructures_InlineControlStructureSniff implements PH
                 if ($endLine !== $end) {
                     $phpcsFile->fixer->addContent($endLine, '}');
                 } else {
+                    if (in_array($tokens[$end]['code'], array(T_SEMICOLON, T_CLOSE_CURLY_BRACKET)) === false) {
+                        $phpcsFile->fixer->addContent($end, ';');
+                    }
+
                     $phpcsFile->fixer->addContent($end, ' }');
                 }
             } else {
--- a/CodeSniffer/Standards/Generic/Tests/ControlStructures/InlineControlStructureUnitTest.inc.fixed
+++ b/CodeSniffer/Standards/Generic/Tests/ControlStructures/InlineControlStructureUnitTest.inc.fixed
@@ -76,5 +76,5 @@ if ($a) {
 
 ?>
 <div style="text-align: right;">
-    <?php if ($model->scenario == 'simple') { $widget->renderPager() } ?>
+    <?php if ($model->scenario == 'simple') { $widget->renderPager(); } ?>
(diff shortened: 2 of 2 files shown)


donatj:
The weird part is you were even testing this behaviour despite the test containing invalid PHP. You should lint your '.fixed' files as part of your testing.

CodeSniffer/Standards/Generic/Tests/ControlStructures/InlineControlStructureUnitTest.inc.fixed
```diff
@@ -76,5 +76,5 @@ if ($a) {
 
 ?>
 <div style="text-align: right;">
-    <?php if ($model->scenario == 'simple') { $widget->renderPager() } ?>
```
donatj: This was testing for not valid php. :smirk:

CodeSniffer/Standards/Generic/Sniffs/ControlStructures/InlineControlStructureSniff.php
```diff
@@ -171,6 +171,10 @@ public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
                 if ($endLine !== $end) {
                     $phpcsFile->fixer->addContent($endLine, '}');
                 } else {
+                    if (in_array($tokens[$end]['code'], array(T_SEMICOLON, T_CLOSE_CURLY_BRACKET)) === false) {
```
aik099: I'm not sure, that this covers all possible cases.

CodeSniffer/Standards/Generic/Tests/ControlStructures/InlineControlStructureUnitTest.inc.fixed
aik099: You really can write code like this, where `;` is omitted before `}`?

gsherwood:
Thanks for the fix.

donatj:
Ooh great. I've been offline all weekend until now. I'm glad to see this merged :)

CodeSniffer/Standards/Generic/Tests/ControlStructures/InlineControlStructureUnitTest.inc.fixed
donatj: No, that was the line I was correcting because it was throwing an error.

CodeSniffer/Standards/Generic/Sniffs/ControlStructures/InlineControlStructureSniff.php
donatj: If there are, I couldn't come up with them.  Either way it's an improvement over it's current state.

PR #492, opened 2015-02-14
Use statements don't increase the scope indent

klausi:
See http://php.net/manual/en/language.oop5.traits.php#language.oop5.traits.conflict

Only containing a failing test case right now, not sure where the correct place is to fix this.


aik099:
In other words: `use traitName {` doesn't increase indentation, but it should.

gsherwood:
I've fixed this issue now. Use statements were not seen as scope openers at all, but they now are.

I couldn't merge in the failing unit test because you removed the parse error (this is a special test for a previous bug fix) and the `.1` and `.2` files had different code in them (they only diff should be spaces vs tabs for indent). But thank you for submitting the report and failing test anyway.

pfrenssen:
Thanks!

PR #493, opened 2015-02-16
Fix: PSR1_Sniffs_Methods_CamelCapsMethodNameSniff and magic method detections

localheinz:
This PR

* [x] adds a failing test demonstrating that `PSR1_Sniffs_Methods_CamelCapsMethodNameSniff` incorrectly determines that `o_toString` is a magic method while it shouldn't be
* [x] fixes the determination whether a string is a magic method

### DIFF
--- a/CodeSniffer/Standards/PSR1/Sniffs/Methods/CamelCapsMethodNameSniff.php
+++ b/CodeSniffer/Standards/PSR1/Sniffs/Methods/CamelCapsMethodNameSniff.php
@@ -62,11 +62,13 @@ class PSR1_Sniffs_Methods_CamelCapsMethodNameSniff extends Generic_Sniffs_Naming
         }
 
         // Ignore magic methods.
-        $magicPart = strtolower(substr($methodName, 2));
-        if (isset($this->magicMethods[$magicPart]) === true
-            || isset($this->methodsDoubleUnderscore[$magicPart]) === true
-        ) {
-            return;
+        if (preg_match('|^__|', $methodName) !== 0) {
+            $magicPart = strtolower(substr($methodName, 2));
+            if (isset($this->magicMethods[$magicPart]) === true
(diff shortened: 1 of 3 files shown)


aik099:
In other words the `PSR1.Methods.CamelCapsMethodName` sniff is considering `o_toString` a magic method and reports it as an error?

localheinz:
@aik099 

Exactly!

CodeSniffer/Standards/PSR1/Sniffs/Methods/CamelCapsMethodNameSniff.php
```diff
@@ -98,4 +95,24 @@ protected function processTokenOutsideScope(PHP_CodeSniffer_File $phpcsFile, $st
...
 
 
+    /**
+     * Returns true if the specified method name refers to a magic method.
+     *
+     * @param string $methodName The name of the method that should be checked.
+     *
+     * @return bool
+     */
+    private function _isMagicMethod($methodName)
```
localheinz: Would this be worth pulling up into `Generic_Sniffs_NamingConventions_CamelCapsFunctionNameSniff`?

aik099:
Do we have any other magic method related code in other sniffs, that needs fixing in the same manner?

localheinz:
@aik099 

Doesn't look like it, but there is a lot of repetition - maybe it's worth pulling this into a utility class?

aik099:
What I was thinking if there are similar magic method code in other sniffs. I haven't checked myself.

localheinz:
@aik099 

See https://github.com/squizlabs/PHP_CodeSniffer/search?utf8=✓&q=preg_match%28%27%7C%5E__%7C%27%2C&type=Code

aik099:
But they look solid by matching `__` just at the beginning of the method.

gsherwood:
Thanks for the fix and the test.

PR #494, opened 2015-02-16
Fix: Remove duplicate key-value-pair from magic methods array

localheinz:
This PR

* [x] removes a duplicate key-value-pair from the array of magic method names


aik099:
:+1:

gsherwood:
Thanks

Issue #483, opened 2015-02-09
Short array syntax sniffer


o5:
Hi,

sorry but I'm not able to find sniff, who check using of short array syntax (PHP >= 5.4). I've removed an old array syntax from my project with [PHP-CS-Fixer](https://github.com/fabpot/php-cs-fixer) and I would like to add rule into my standarts.xml. 

I looked on [Squiz.Arrays.ArrayDeclaration](https://github.com/squizlabs/PHP_CodeSniffer/tree/master/CodeSniffer/Standards/Squiz/Sniffs/Arrays) but it makes something different things what I need.

Thank you!

aik099:
I think there are no sniff, that enforces PHP 5.4 short array syntax and would transform regular array syntax to short form. If you'd like to create such sniff, then please place it in https://github.com/squizlabs/PHP_CodeSniffer/tree/master/CodeSniffer/Standards/Generic/Sniffs/Formatting folder.

All existing array-related sniffs tolerate both regular and short array notations.

o5:
Thanks for reply! I'm a totally tokenizer noobie but I'll try it and maybe, I'll send PR :)

aik099:
I recommend looking at the simplest sniff there can be: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Generic/Sniffs/Formatting/SpaceAfterCastSniff.php , which:

1. finds a cast token
2. checks for next token after it to be a single space
3. if it's not reports that
4. in fixing mode adds missing space

Your sniff would:

1. listen for T_ARRAY tokens
2. report them as an error
3. when fixing:
 * the `array` word itself is T_ARRAY token so it can be dropped
 * replace the opening parenthesis `(` (can be obtained via `$tokens[$stackPtr]['parenthesis_opener']`) with an `[`
 * replace the closing parenthesis `)` (can be obtained via `$tokens[$stackPtr]['parenthesis_closer']`) with an `[`

Then write a test case to confirm that behavior and a fixture + fixed file.

aik099:
I guess you should also write the opposite sniff to convert short array to regular ones.

aik099:
If you're totally stuck with implementation let me know and I see what I can do this week.

o5:
Your suggested steps don't look hard. I'll try it.

gsherwood:
If you get stuck, I'm happy to help as well, or accept a "nearly there" PR.

[...]

PR #495, opened 2015-02-18
Disallow Long Array Syntax Sniff & Disallow Short Array Syntax Sniff 

xalopp:
Based on the discussion regarding Short array syntax sniffer #483  this PR adds  
- DisallowLongArraySyntaxSniff
- DisallowShortArraySyntaxSniff

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/Formatting/DisallowLongArraySyntaxSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Formatting/DisallowLongArraySyntaxSniff.php
@@ -1,0 +1,90 @@
+<?php
+
+/**
+ * Generic_Sniffs_Formatting_DisallowLongArraySyntaxSniff.
+ *
+ * PHP version 5
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Xaver Loppenstedt <xaver@loppenstedt.de>
+ * @copyright 2013-2015 Xaver Loppenstedt
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ * @link      http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+/**
+ * Generic_Sniffs_Formatting_DisallowLongArraySyntaxSniff.
+ *
+ * Disallow long array syntax.
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Xaver Loppenstedt <xaver@loppenstedt.de>
+ * @copyright 2013-2015 Xaver Loppenstedt
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ * @version   Release: @package_version@
+ * @link      http://pear.php.net/package/PHP_CodeSniffer
+ */
+class Generic_Sniffs_Formatting_DisallowLongArraySyntaxSniff implements PHP_CodeSniffer_Sniff
+{
+
+
+    /**
+     * Registers the tokens that this sniff wants to listen for.
+     *
+     * @return array(int)
+     */
+    public function register()
+    {
+        return array(T_ARRAY);
+
+    }//end register()
+
+
+    /**
+     * Called when one of the token types that this sniff is listening for
+     * is found.
+     *
+     * @param PHP_CodeSniffer_File $phpcsFile The PHP_CodeSniffer file where the
+     *                                        token was found.
+     * @param int                  $stackPtr  The position in the PHP_CodeSniffer
+     *                                        file's token stack where the token
+     *                                        was found.
+     *
+     * @return void
+     */
+    public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+        $token  = $tokens[$stackPtr];
+        $opener = $token['parenthesis_opener'];
+        $closer = $token['parenthesis_closer'];
+
+        $errorMessage = 'Short array syntax must be used';
+
+        if (($opener !== null) && ($closer !== null)) {
+            $fix = $phpcsFile->addFixableError($errorMessage, $stackPtr);
+
+            if ($fix === true) {
+                $phpcsFile->fixer->beginChangeset();
+                $phpcsFile->fixer->replaceToken($stackPtr, '');
+
+                $phpcsFile->fixer->replaceToken($opener, '[');
+                for ($i = ($stackPtr + 1); $i < $opener; $i++) {
+                    $phpcsFile->fixer->replaceToken($i, '');
+                }
+
+                $phpcsFile->fixer->replaceToken($closer, ']');
+
+                $phpcsFile->fixer->endChangeset();
+            }
+        } else {
+            // Don't fix erroneous arrays with!
+            $phpcsFile->addError($errorMessage, $stackPtr);
+        }
+
+    }//end process()
+
+
+}//end class
--- a/CodeSniffer/Standards/Generic/Sniffs/Formatting/DisallowShortArraySyntaxSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Formatting/DisallowShortArraySyntaxSniff.php
@@ -1,0 +1,87 @@
+<?php
+
+/**
+ * Generic_Sniffs_Formatting_DisallowShortArraySyntaxSniff
+ *
+ * PHP version 5
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Xaver Loppenstedt <xaver@loppenstedt.de>
+ * @copyright 2015 Xaver Loppenstedt
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ * @link      http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+/**
+ * Generic_Sniffs_Formatting_DisallowShortArraySyntaxSniff.
+ *
+ * Disallow short array syntax.
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Xaver Loppenstedt <xaver@loppenstedt.de>
+ * @copyright 2015 Xaver Loppenstedt, All rights reserved.
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ * @version   Release: @package_version@
+ * @link      http://pear.php.net/package/PHP_CodeSniffer
+ */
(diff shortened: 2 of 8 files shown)


CodeSniffer/Standards/Generic/Sniffs/Formatting/DisallowLongArraySyntaxSniff.php
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Generic_Sniffs_Formatting_DisallowLongArraySyntaxSniff.
+ *
+ * PHP version 5
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Xaver Loppenstedt <xaver@loppenstedt.de>
+ * @copyright 2013-2015 Xaver Loppenstedt
```
aik099: I think copyright note should be one, that is used in existing sniffs.

CodeSniffer/Standards/Generic/Sniffs/Formatting/DisallowLongArraySyntaxSniff.php
```diff
@@ -0,0 +1,88 @@
...
+        );
+
+        if ($fix === true) {
+            $tokens = $phpcsFile->getTokens();
+            $token  = $tokens[$stackPtr];
+
+            $phpcsFile->fixer->beginChangeset();
+            $phpcsFile->fixer->replaceToken($stackPtr, '');
+
+            if ($token['parenthesis_opener'] !== null) {
```
aik099: In which cases this (and `parenthesis_closer`) can be `null`? I bet for malformed PHP file these keys will absent a all and this check would trigger a notice. Please check that this doesn't happen.

Usually it's a good idea to check for `parenthesis_opener` right at the beginning of the sniff (even before fixing is started) and exit early.

CodeSniffer/Standards/Generic/Sniffs/Formatting/DisallowShortArraySyntaxSniff.php
```diff
@@ -0,0 +1,77 @@
...
+            'Long array syntax must be used',
+            $stackPtr
+        );
+
+        if ($fix === true) {
+            $tokens = $phpcsFile->getTokens();
+            $token  = $tokens[$stackPtr];
+
+            $phpcsFile->fixer->beginChangeset();
+            $phpcsFile->fixer->replaceToken($token['bracket_opener'], 'array(');
```
aik099: Maybe amount of spaces between `array` and `(` should be configurable via public sniff property. I'm telling this because opposite sniff (that transforms into short arrays) also checks for extra spaces between `array` and `(`.

CodeSniffer/Standards/Generic/Tests/Formatting/DisallowLongArraySyntaxUnitTest.fail.inc
```diff
@@ -0,0 +1,20 @@
+<?php
```
aik099: Fixture filename must match used sniff name or otherwise it might not be discovered by test runner.

Correct name would be `DisallowLongArraySyntaxUnitTest.inc`. And then fixed filename would be `DisallowLongArraySyntaxUnitTest.inc.fixed`.

CodeSniffer/Standards/Generic/Tests/Formatting/DisallowLongArraySyntaxUnitTest.pass.inc
```diff
@@ -0,0 +1,20 @@
+<?php
```
aik099: No need to create separate fixture files for pass and fail cases. You can just have one file.

CodeSniffer/Standards/Generic/Tests/Formatting/DisallowLongArraySyntaxUnitTest.php
```diff
@@ -0,0 +1,82 @@
...
+     * The key of the array should represent the line number and the value
+     * should represent the number of errors that should occur on that line.
+     *
+     * @param string $testFile test file
+     *
+     * @return array(int => int)
+     */
+    protected function getErrorList($testFile = '')
+    {
+        switch ($testFile) {
```
aik099: Having single fixture file will also eliminate the need for the `switch` here.

aik099:
These (see above) would be the corrections I believe are required.

CodeSniffer/Standards/Generic/Sniffs/Formatting/DisallowLongArraySyntaxSniff.php
xalopp: Unfortunately in didn't find a concrete copyright/copyleft guideline.
That's why I looked a several files like  OneClassPerFileSniff and ClosingPHPTagSniff and decided to put a copyright in similar form.

CodeSniffer/Standards/Generic/Sniffs/Formatting/DisallowLongArraySyntaxSniff.php
xalopp: Actually, the values for keys "parenthesis_opener" and "parenthesis_closer" are null, if there are no parenthesis in the array statement.

xalopp: I agree that an early exit on error is a good strategy.
I this particular case I had the choice between fixing automatically a bogus "array;" statement without parenthesis  or keeping the old code unchanged.

In case you prefer leaving arrays with missing parenthesis untouched, it would be fine for me to change the code.

CodeSniffer/Standards/Generic/Sniffs/Formatting/DisallowShortArraySyntaxSniff.php
xalopp: Sounds like a nice feature

CodeSniffer/Standards/Generic/Tests/Formatting/DisallowLongArraySyntaxUnitTest.fail.inc
xalopp: The current implementation of  `AbstractSniffUnitTest::testSniff` will consider test files matching `FooUnitTest.*` for a sniff called `FooSniff`.

CodeSniffer/Standards/Generic/Tests/Formatting/DisallowLongArraySyntaxUnitTest.pass.inc
xalopp: If that's the preferred why, it's just fine for me.

CodeSniffer/Standards/Generic/Tests/Formatting/DisallowLongArraySyntaxUnitTest.fail.inc
aik099: Still it's better not to create separate fixture files for pass/fail cases.

xalopp:
@aik099 is there anything left to be done, to complete this pull request?

aik099:
Except that copyright change all seems to be ok, but only @gsherwood has merge rights.

mmoll:
@aik099 which"copyright _change_" do you mean?

aik099:
That one I've mentioned in inline comment earlier, don't you see it?

mmoll:
I'm not sure if I did get that correct. You're suggesting, that "Squiz Pty Ltd" also gets the copyright of this added file?

aik099:
@mmoll , the @xalopp is PR author so I'm not sure what you're not happy with here. If you look at the library files, then copyright is the same for all sniffs, but original sniff author (name + e-mail) is also mentioned via `@author` tag in file/class DocBlock. I see no problems with that.

aik099:
@mmoll , I've looked at some other sniffs and they indeed have their own copyright, so I guess we can keep it as-is.

xalopp:
@aik099 great to hear that code is ready for merge as-is :dancer:

aik099:
@gsherwood , any estimates on when this will be merged?

gsherwood:
> any estimates on when this will be merged

I'm pretty busy at the moment. I haven't even had a chance to look at it. So no estimates.

mmoll:
#noestimates :laughing:

gsherwood:
I finally got around to looking at this, but I already had sniffs in the works for this exact problem as they were suggested a while ago on issue #483, so I just completed those and committed them here: https://github.com/squizlabs/PHP_CodeSniffer/commit/081f49aa967e2cd7c579442ffa05ef0193c6018a

I did take a look through this PR though to see how different our solutions were. The things that popped out at me were:

1. The sniff should be able to fix arrays without braces. I hadn't considered this parse-error case before I looked at your tests, but it was easy enough to fix it. If there is a reason you specifically don't, please let me know.

2. Comments can be preserved during fixing. I hadn't considered this case specifically, but the way I decided to fix things (without the FOR loop) made it work regardless. I've made changes recently in other sniffs to ensure comments are not removed while fixing, so I wanted to make sure it works here too.

3. There is no need to worry about array formatting when you are fixing the code. These sniffs only do one thing, so they can fix the arrays with zero spaces, or 1 space, or whatever they like. Other sniffs should be used to enforce and fix all arrays, including auto fixed ones. This is why the fixer loops over a file until all fixable errors are gone.

4. You must provide a code for all errors messages so they can be overridden in ruleset.xml files. This lets people turn the errors into warnings, reduce their severity, or translate them. I just used `Found` for my codes because I do that elsewhere for very simple sniffs.

5. I added metric recording to these sniffs, in line with the way the Squiz array formatting sniffs does it. This lets you use `--report=info` and see how many arrays are using long and short array syntax. It's an optional feature, but something I've been adding to various sniffs.

Thanks for submitting the PR. The test cases you had were particularly useful for testing the sniffs I committed.

aik099:
> The sniff should be able to fix arrays without braces.

Is that even possible to have array declaration without braces.

gsherwood:
> Is that even possible to have array declaration without braces.

No, it is a syntax error. But there was a special case checking for it in this PR, so I added it to the core sniff as well. Doesn't hurt.

aik099:
Are you certain that `parenthesis_opener` and `parenthesis_closer` keys from token information are present at all when there is a syntax error. For example the `scope_opener` and `scope_closer` keys are absent on syntax error.

gsherwood:
> Are you certain that parenthesis_opener and parenthesis_closer keys from token information are present at all when there is a syntax error.

I ran some tests and the array indexes were there, but empty. Some arrays are not even tokenized as T_ARRAY tokens if the syntax error is great enough, or if the tokenizer get confused by a short array close by.

I guess we'll find out if people with syntax errors start reporting bugs. But the unit test has a case for this in it and it is being detected and fixed properly.

aik099:
Nice, thanks.

PR #504, opened 2015-03-01
Adding back "branch-alias" setting that points to "master" branch

aik099:
When in 32dd78e5bc2d512e89aaaf1ac0870404df854041 commit the `branch-alias` setting was removed from `composer.json` the versions like `~2.0@dev` stopped working.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,11 @@
         "wiki": "https://github.com/squizlabs/PHP_CodeSniffer/wiki",
         "source": "https://github.com/squizlabs/PHP_CodeSniffer"
     },
+    "extra": {
+        "branch-alias": {
+            "dev-master": "2.0.x-dev"
+        }
+    },
     "autoload": {
         "classmap": [
             "CodeSniffer.php",


gsherwood:
Why do you need to alias dev-master when there is only one active branch? (serious question, I don't use composer)

I thought people just require dev-master to get the latest code. If you are using an old syntax to get the pre-release code, wouldn't you just change it to 2.x or dev-master now that the version is stable? Then I add a new branch alias when I'm starting 3.x.

aik099:
Let's consider following example:

1. LibraryA has `~2.0@dev` version for phpcs in it's composer.json = everybody using LibraryA directly will get `master` branch content then
2. LibraryB has `1.0` version for LibraryA = since `@dev` part in version isn't retained, when using library indirectly (phpcs used indirectly through LibraryA by LibraryB), then `2.2` tag is used.

This way with same version constraint it's possible for LibraryA developer to always use latest code from `master` branch but for LibraryA users always get latest stable version of phpcs.

aik099:
In general `branch-alias` needs to be used to allow regular branches, that naturally don't map to any version to be used in version comparisons.

gsherwood:
Thanks for the explanation.

Issue #470, opened 2015-02-02
Basepath option to show relative path names in report output


Hounddog:
I am currently running php_codesniffer on jenkins with a vagrant-aws plugin.
When i run phpcs on the vagrant box and generate the report i get a fullpath starting at /vagrant/src/...

Once i copy this back to jenkins however the path doesnt match anymore.

Is there a way to generate reports with a base path or relative path?

aik099:
how come, that you generate report on one machine and use it on another machine?

Hounddog:
As mentioned i use jenkins. I set up the test environment using vagrant-aws and also generate all the reports on the "vagrant" machine and then just syncronize the reports back to the host. This way i dont have to install any dependencies on jenkins and can keep my whole development/test environment boxed.

gsherwood:
There is no base path setting. Have you tried just passing a relative path to phpcs?

`phpcs temp.php  --report-file=./report.txt`

gsherwood:
Oh sorry, I think I misunderstood what you are asking.

If you mean: can the report output contain a relative path instead of an absolute path to the files

The answer is: no

Your best option right now is to generate the file and then run a command over it to replace the `/vagrant/src/...` bit with whatever prefix you need for the other machine.

Hounddog:
i figured as much. Would be great to generate it relative or absolute or with an own base path. Thanks for the quick reply though

gsherwood:
Thanks for clarifying for me. Seems like a good feature request.

muffl0n:
@Hounddog like mentioned in #522 the violations plugin in jenkins works perfectly with absolute paths and even refuses to display the file content when I use relative paths. Maybe you have to modify your reports with a script after you moved them from your VM to your jenkins machine?

aik099:
Maybe instead you need to have some kind of report file transformation script that would fix paths as needed by machine that reads the report.

gsherwood:
I've added this option in. You can use `--basepath=<basepath>` to have a path stripped from the front of file paths output to reports and in debug output. The path can be absolute or relative to the current dir.

A handy settings is: `--basepath=.`

GaryJones:
@gsherwood Did this functionality get removed again?

gsherwood:
> @gsherwood Would it be possible to add this to the 2.x branch?

[...]

PR #522, opened 2015-03-20
Add cli-option '--report-relative-path'

muffl0n:
This option enables the output of the filenames to be relative to the cwd. Currently this only works for "checkstyle" reports.

fixes squizlabs/PHP_CodeSniffer#470

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -326,7 +326,7 @@ class PHP_CodeSniffer_CLI
             // Use function defaults.
             $defaults['reportWidth'] = null;
         }
-        
+
         $reportRelativePath = PHP_CodeSniffer::getConfigData('report_relative_path');
         if ($reportRelativePath !== null) {
             $defaults['reportRelativePath'] = $reportRelativePath;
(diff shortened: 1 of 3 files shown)


muffl0n:
This PR lacks support for all other report types than "checkstyle". If this PR is going to be accepted all other reports would have to be adjusted.

aik099:
Relative path is when you have `../` in url. What you do here is just remove some common base path from paths in the file. I'm not sure how this shortened path can be matched to actual file path in the file by tool, that reads that report.

muffl0n:
According to [Wikipedia] (http://en.wikipedia.org/wiki/Path_%28computing%29#Unix_style) a relative path is a "path relative to the working directory of the user or application". That can but does not have to include "../"

This is a sample output of pylint:
```
python/test.py:1: [C0111(missing-docstring), ] Missing module docstring
```

As @Hounddog mentions in #470 this is used by the Violations-plugin in Jenkins to link to the source files.

aik099:
But you also need for all other tools used on Jenkins to use relative paths, nope? Does PHPUnit support that?

muffl0n:
Sorry, I'm not sure about PHPUnit.
I was just testing pylint (which works as described above) and now phpcs with the violations plugin in jenkins. I then found the issue #470 where @Hounddog describes exactly what I also need.

muffl0n:
I'm sorry, I think I just had another quirk in the setup. :( The violations-plugin works perfectly with the absolute paths that are supplied by phpcs.
Seems like I should just go home for today. :)

PR #546, opened 2015-04-13
Update ForbiddenFunctionsSniff.php

caipre:
Minor correction to error message


aik099:
:+1:

gsherwood:
The type there is used to generate the unique error message code, not the message text itself. Changing this turns the error code from Generic.ForbiddenFunctions.Found (.Found is used across many sniffs) to Generic.ForbiddenFunctions.Forbidden, which also makes the friendly text version change from `Forbidden functions found` to `Forbidden functions forbidden`.

So this is a major BC break, a consistency break, and it makes the readable version of the code less readable.

So for these reasons, I wont merge this in.

caipre:
@gsherwood : Thanks for the thorough explanation; I didn't read the code closely enough!

gsherwood:
@caipre Not a problem. Thanks for taking the time to contribute.

PR #562, opened 2015-04-21
Add support for allowing string concatenation when the concatenated string would result in a line being to long

stefanlenselink:
Default setting is to ignore the limit and always complain, lineLimit is configurable in an AnnotatedRuleset to match the PSR-2 120 chars for example.

Could not write a UnitTest case because I didn't find a way to set the configurable property from the TestSuit, when the default would be set to 80 or 120 I can make a unit-test for that case. Didn't set any default to keep the unconfigured behaviour the same.

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/Strings/UnnecessaryStringConcatSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Strings/UnnecessaryStringConcatSniff.php
@@ -46,6 +46,11 @@ class Generic_Sniffs_Strings_UnnecessaryStringConcatSniff implements PHP_CodeSni
      */
     public $error = true;
 
+    /**
+     * If set do allow string concatenation if
+     * without the line would become too long
+     */
+    public $lineLimit = 0;
 
     /**
      * Returns an array of tokens this test wants to listen for.
@@ -103,6 +108,16 @@ class Generic_Sniffs_Strings_UnnecessaryStringConcatSniff implements PHP_CodeSni
                     $nextChar = $tokens[$next]['content'][1];
                     $combined = $prevChar.$nextChar;
                     if ($combined === '?'.'>' || $combined === '<'.'?') {
+                        return;
+                    }
+                }
+
+                if ($this->lineLimit) {
+                    $end = $tokens[$prev]['column']
+                           + $tokens[$prev]['length']
+                           + $tokens[$next]['length'];
+
+                    if ($end > $this->lineLimit) {
                         return;
                     }
                 }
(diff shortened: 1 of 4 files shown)


aik099:
There is no such standard-wide line length limit. There is a sniff, that checks line length and it has a parameter. Having to specify a length 2 times (in that sniff and in this one) sounds incorrect.

stefanlenselink:
I must agree that specifying the length twice sounds incorrect, but we run both sniffs in our ruleset. This result in having 1 sniff telling make the line shorter, while the other telling make the line longer....Fixing one resolves into breaking the other...

What would be your suggestion to go having 1 length define?

aik099:
> Fixing one resolves into breaking the other...

Ah, conflicting sniffs. That sounds familiar. If that's a custom standard then you can copy sniff in it and update as needed. Not sure if that change needs to be pushed to upstream of PHPCS.

stefanlenselink:
If we 'fork' that sniff there is no need for a push upstream offcourse, but this 'fix' could also be beneficial for other users as wel.

So it comes down to duplicate configuration vs. duplication of code (force a user to fork the sniff). I'll leave it up to you :)

For now we run our copy-and-past forked sniff :)

aik099:
We'll just wait for @gsherwood opinion on this one.

gsherwood:
Personally I think that if you've got a really long string, for some good reason, you should just ignore the long line warning completely. But if you want to break it up, you can obviously use a concat split over multiple lines, or you could just split the concat into different statements:
```php
$string = 'This is a really long string that is being used as an error message';

$string = 'This is a really long string '
        . 'that is being used as an error message';

$string  = 'This is a really long string ';
$string .= 'that is being used as an error message';
```
None of the sniffs are going to complain about that last code block, so that's how the sniff wants you to write code.

If you prefer to write code using the second code block, you might need a completely different sniff to ensure that split concat statements are formatted correctly, or you can end up with inconsistent code like:
```php
$string = 'This is a really long string '
        . 'that is being used as an error message';

$string = 'This is a really long string ' .
          'that is being used as an error message';

$string = 'This is a really long string '
. 'that is being used as an error message';

$string = 'This is a really long string '
    . 'that is being used as an error message';
```

Having said all that, this is still an interesting feature for people who do want to split up a single concat statement over multiple lines. It could really use that extra formatting sniff to go with it, but it is a good first step.

But I wouldn't implement it by using a line length setting. I'd would assume that if you have split a concat over multiple lines, you are obviously trying to fit within a line limit or some other type of readability (possible to make sentences line up:
```php
$string = 'This is a really long string. '
        . 'It is being used for errors. '
        . 'The message is not translated.';
```
 So a simple option in the sniff to allow multi-line concat statements is all that this really needs.

stefanlenselink:
@gsherwood Swapped implementation for multi-line approach. By default nothing changes, option to allow multiline concats.

gsherwood:
Im really sorry for taking so long to get back to this, but it is merged in now. Thanks a lot for making those changes and contributing to PHPCS.

sebastienbarre:
Just dropping by to say thanks, took me a bit to find that `allowMultiline` feature, but this is really useful.

PR #565, opened 2015-04-23
Squiz.WhiteSpace.OperatorSpacing - minus sign (negative number) in short array should not be reported

VasekPurchart:
Since `T_OPEN_SQUARE_BRACKET` is replaced with `T_OPEN_SHORT_ARRAY` during tokenization this should be also being detected.

Also there was previously (afaik) no test for normal `array()` (but this worked thanks to the long array using `T_OPEN_PARENTHESIS`).


gsherwood:
Thanks a lot for this fix. I also applied one to the Squiz.Formatting.OperatorBracket sniff at the same time.

PR #569, opened 2015-04-24
Generic ScopeIndentSniff throws PHP notices in JS files

klausi:
Not sure what this triggers yet, but here is at least a test case as a start.


klausi:
Ok, so I found the problem with the JS object scope property. Yep, that test case is quite an ugly bit and the fixed version does not look much better :-D

gsherwood:
The smallest bit of code I could get this to fail on was:
```js
for (var i = 0; i < 10; i++) {
    var foo = {foo:{'a':'b',
    'c':'d'}}
}
```

The error was actually caused by the `$openScopes` array, which was causing the token *after* the object closer to be checked as a scope closer as well, but only for JS objects. This was because the array index was not set. It appears that the tokenizer is working okay.

I've committed a fix for this now that works with my test code and your test file. Please have a try if you can.

gsherwood:
I still think there is something wrong with this `$openScopes` array, and I may need your code to fix it as well. I'll keep looking.

gsherwood:
I've just pushed a new change where I removed that `$openScopes` line instead of modifying it. Closers for JS objects are handled by specific code already and so don't make use of that generic scope array. Let me know if that change doesn't work for you.

klausi:
Thanks, I merged the ScopeIndentSniff changes into Coder: http://cgit.drupalcode.org/coder/commit/?id=37fc805

Issue #541, opened 2015-04-10
PHP_CodeSniffer could/should try to use $PATH in case if *_path is not configured


morozov:
The actual problem consists of a few parts:

1. Currently, `ClosureLinterSniff` and similar sniffs silently skip in case if the corresponding command line application path (e.g. `gjslint_path`) is not specified.
2. The path to the `CodeSniffer.conf` file is hard-coded as `dirname(__FILE__).'/CodeSniffer.conf'`.
3. I have an application which is distributed as a PHAR archive and contains PHP_CodeSniffer inside.

As the result, once the application is packaged and then installed, there's no way to specify `gjslint_path` and the others. In fact, it becomes non-portable.

The solution may be to replace the following parts:
```php
$lintPath = PHP_CodeSniffer::getConfigData('gjslint_path');
if ($lintPath === null) {
    return;
}
```
with someghing like
```php
$lintPath = PHP_CodeSniffer::findBinary('gjslint');
if ($lintPath === null) {
    return;
}
```
which would try to use `PHP_CodeSniffer::getConfigData($binary.'_path');` first and then fallback to finding the path using `$PATH`.

Another solution (which doesn't conflict with the previous one) is to have `CodeSniffer.conf` searched in `$HOME`, so that the same instance of PHP_CodeSniffer might have different configuration for different users, or otherwise, many instances of PHP_CodeSniffer (or other apps based on in) might have the same configuration for the same user.

morozov:
@gsherwood, @aik099, please let me know if the above makes sense. I'll try to implement it then.

gsherwood:
I wont be accepting any changes to how config files are sourced in the current master branch, but would consider it for the 3.0 branch, where some config file changes are already being made.

But for your case, you have 2 options.

1. Use --runtime-set to set a config var for a single run: `phpcs.phar --runtime-set gjslint_path /path/to/gjslint /path/to/code ...`
https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#setting-configuration-options

2. Use a custom coding standard and define the paths in there by creating a ruleset.xml file:
```xml
<?xml version="1.0"?>
<ruleset name="Custom Standard">
  <config name="gjslint_path" value="/path/to/gjslint"/>
  <!-- Include a reference to the standard or sniffs you use here -->
</ruleset>
```
Then run phpcs with that standard: `phpcs.phar --standard=/path/to/ruleset.xml /path/to/code`
[...]

PR #571, opened 2015-04-24
Automatic discovery of executable paths

morozov:
See https://github.com/squizlabs/PHP_CodeSniffer/issues/541. Alternatively, the approach used in [symfony/process](https://github.com/symfony/Process/blob/master/ExecutableFinder.php) could be used, however it's much more complex than just calling a shell command.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -242,6 +242,13 @@ class Config
      */
     private static $configData = null;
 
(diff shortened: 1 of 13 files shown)


gsherwood:
Great addition. Thanks a lot.

PR #572, opened 2015-04-24
Add warning/error toggle to UnconditionalIfStatementSniff

chpatton013:
This pull adds to ability to specify that the `Generic.CodeAnalysis.UnconditionalIfStatement` sniff should trigger an error, rather than a warning (which is its current behavior).

This option can be customized in the phpcs config file by referencing the `treatAsError` property:

```xml
<rule ref="Generic.CodeAnalysis.UnconditionalIfStatement">
   <properties>
      <property name="treatAsError" value="1"/>
   </properties>
</rule>
```


gsherwood:
You can already change warnings to errors, and errors to warnings, using a ruleset.xml file. In this case, you'd use:
```xml
<rule ref="Generic.CodeAnalysis.UnconditionalIfStatement.Found">
    <type>error</type>
</rule>
```
Does this do what you want?

chpatton013:
Your suggestion worked great! I missed that while I was looking through the docs. Thanks!

PR #575, opened 2015-04-26
Generic Sniffs for CodeAnalysis and Methods

jpgerdeman:
The following sniffs are added:

- **Generic_Sniffs_CodeAnalysis_FunctionLengthSniff**  warns resp. errors on very long functions.
- **Generic_Sniffs_CodeAnalysis_NumberOfMethodsInClassSniff** warns resp. errors, when a class contains a lot of methods.
- **Generic_Sniffs_Functions_SpaceBeforeParenthesisSniff** checks the spacing for a parenthesis in a functions signature.
- **Generic_Sniffs_Methods_ForceVisibilitySniff** checks if the visibility of a class method is defined.
- **Generic_Sniffs_Methods_MethodSignatureSniff** checks if a class methods signature matches a specific pattern (order of tokens).


aik099:
Please don't add sniffs, that are in fact implementation PHP Mess Detector rules (http://phpmd.org/rules/index.html) as PHP_CodeSniffer sniffs.

> Generic_Sniffs_CodeAnalysis_FunctionLengthSniff

Matches to PHP Mess Detector's ExcessiveMethodLength rule (see http://phpmd.org/rules/codesize.html#excessivemethodlength)

> Generic_Sniffs_CodeAnalysis_NumberOfMethodsInClassSniff

Matches to PHP Mess Detector's TooManyMethods rule (see http://phpmd.org/rules/codesize.html#toomanymethods)

> Generic_Sniffs_Functions_SpaceBeforeParenthesisSniff

Instead of checking just a space I recommend using existing sniffs, that validate complete method signature.

> Generic_Sniffs_Methods_ForceVisibilitySniff 

The `Squiz.Scope.MethodScope` is doing that.

> Generic_Sniffs_Methods_MethodSignatureSniff

Are you certain that existing FunctionComment and similar sniffs in different standards doesn't check that.

aik099:
Thanks for the contribution, but before making any large changes it's recommended to discuss first. It's also possible to use Gitter channel to ask questions: https://gitter.im/squizlabs/PHP_CodeSniffer

jpgerdeman:
I didn't know about https://gitter.im/squizlabs/PHP_CodeSniffer.
Will head there.

aik099:
The #540 PR adds proper button to repo homepage, but it's not yet merged.

gsherwood:
While PHP_CodeSniffer is easier able to check things like number of methods in a class, or the length of functions, I agree that these sniffs don't belong in the core because they confuse the purpose of the tool. For that reason, I don't want to merge in more of those checks in. The current ones are quite old, but I'd still like to remove them in the future as well.

The other sniffs replicate the behaviour of existing core sniffs, but either in a slightly different way or by doing a single check instead of multiple. I don't want to merge these in because it might become confusing as to which sniff to use, and I also then need to maintain two sets of checks.

If you want to make use of the core sniffs, it sounds like you need one of two things:

1. Extra config options for sniffs, such as the ability to change the order of method keywords in this sniff (assuming you don't like the order PSR2 defines): https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/PSR2/Sniffs/Methods/MethodDeclarationSniff.php

2. The ability to include a specific error code in a ruleset.xml file instead of an entire sniff. A sniff is basically a bundle of related checks that you are likely to want to use together (especially if you don't already have a well defined standard). Being able to pull specific checks out of that bundle and have PHP_CodeSniffer automatically ignore all other checks in that sniff class is an idea I've been playing around with. It's probably something I'll look at for the 3.0 version.

PR #581, opened 2015-05-01
Allow DocBlocks on require|include(_once)

GaryJones:
PSR-5 (draft) [defines](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#3-definitions) `include()`, `include_once()`, `require()` and `require_once()` as Structural Elements which can therefore have DocBlocks.

_Since this was loosening off of a potential false positive, I was unsure if / how unit tests needed to be updated for this._


gsherwood:
Sorry it took me so long to merge this in. Thanks a lot for letting me know about this.

For reference, the structural elements are documented here: http://phpdoc.org/docs/latest/glossary.html#term-structural-element

PR #585, opened 2015-05-03
Long CSS file performance issue because array_values() is very slow

klausi:
I have this 8000 lines generated CSS file and seems to send PHPCS into an endless loop. Not sure what is causing this, but here is a test case as a start. So this should send the Travis CI test into some kind of timeout.

### DIFF
--- a/CodeSniffer/Standards/Squiz/Tests/CSS/ClassDefinitionClosingBraceSpaceUnitTest.css
+++ b/CodeSniffer/Standards/Squiz/Tests/CSS/ClassDefinitionClosingBraceSpaceUnitTest.css
@@ -46,6 +46,11 @@
 
 .GUITextBox.container:after {}
 
+/* Now just add a very large portion of CSS to test the performance of the CSS
+ * tokenizer, we don't care about coding standards in that.
+ */
+
+/* @codingStandardsIgnoreStart */
 /*! normalize.css v3.0.1 | MIT License | git.io/normalize */
 /* line 9, ../../../../../../usr/local/rvm/gems/ruby-1.9.3-p547@global/gems/bootstrap-sass-3.2.0.2/assets/stylesheets/bootstrap/_normalize.scss */
 html {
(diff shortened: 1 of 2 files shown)


klausi:
Interesting, so PHP runs into the timeout as expected, but HHVM completes the test suite. So this might be a PHP core bug :-(

klausi:
After having some fun debugging the PHP_CodeSniffer_Tokenizers_CSS class I did some timing analysis on the array_values() calls there. They are very expensive on large arrays. In my testing I had a token array with more than 20,000 entries and one array_values() takes 11ms! And that of course accumulates to very long runtimes when parsing bigger CSS files.

aik099:
Why you're analyzing generated CSS file at all? Shouldn't you be analyzing source CSS files used for generation, because error reporting in there is better since only there you can fix reported errors.

klausi:
It looks like HHVM has a far superior array_values() implementation, since the Travis test run took just the usual 1 minute time. Incredible!

klausi:
@aik099 Unfortunately we cannot distinguish yet between generated CSS and handwritten CSS in Drupal projects, so we just run PHPCS on all CSS files. But yeah, we should probably skip CSS file checking once we found a SCSS file in a project.

Nevertheless PHPCS should work on large CSS files also, so we should fix this critical performance issue.

klausi:
Yay, so Travis CI is now able to work through the test suite again. I fixed the test case now and added an explanation.

gsherwood:
Thanks a lot for looking into this and getting that performance improvement. I also added one more that I noticed when running your sample file: https://github.com/squizlabs/PHP_CodeSniffer/commit/c0d26900997e4d781cb546054551b0386372f153

Just letting you know that I removed your 8000 line CSS test file for 2 reasons: unit tests aren't where I want to put performance tests (they need to run as fast as possible because I run them all the time) and my local phpunit install was still crashing out on that file even though I can process it in about 5 secs now (probably memory related).

Thanks again.

PR #593, opened 2015-05-12
Commented code

fpietka:
Right now, when commenting an url, `CommentedOutCode` sniffs will detect a `goto` token.
This pull request is about avoiding that false positive for a list of protocols.


gsherwood:
Detecting commented out code isn't an exact science, and doesn't have to be, so I've applied a fix for this in a different way. I don't want to maintain a list of protocols in this sniff just for checking URLs, so I'm just ignoring comments and goto labels, which is what a URL pattern looks like.

Thanks for reporting the issue. Commit is here: https://github.com/squizlabs/PHP_CodeSniffer/commit/980ce82a133d31a47f5c6d720fc458f31e451d46

fpietka:
Even better. Thanks

PR #595, opened 2015-05-16
Colons in CSS selectors within media queries throw false positives

klausi:
Apparently :hover exists in CSS and confuses a couple of CSS sniffs. Test case only for now.


klausi:
The CS tokenizer does not seem to know about At-rules in CSS, so I implemented a flag for that case. That stops the tokenizer from classifying selectors in media query blocks as T_STYLE types.

gsherwood:
A similar bug report was posted on the PEAR tracker: https://pear.php.net/bugs/bug.php?id=20542

This fix works for it as well.

gsherwood:
Thanks for fixing this. Nicely done.

PR #599, opened 2015-05-22
Fix division by zero warning in test report

loonies:
Running sniff tests that doesn't produce any errors will raise
"Division by zero" PHP warning while calculating fixable percentage.
This can happen if we filter out too much tests, or if sniff adds
warnings only, but not errors.

I've found that this is [addressed in 3.0] by suppressing message. This
fix will still print the message, but with zeros.

[addressed in 3.0]: https://github.com/squizlabs/PHP_CodeSniffer/blob/3.0/tests/TestSuite.php#L56-L60

### DIFF
--- a/tests/TestSuite.php
+++ b/tests/TestSuite.php
@@ -48,9 +48,11 @@ class PHP_CodeSniffer_TestSuite extends PHPUnit_Framework_TestSuite
         $result = parent::run($result, $filter);
         spl_autoload_unregister(array('PHP_CodeSniffer', 'autoload'));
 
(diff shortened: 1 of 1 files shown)


aik099:
:+1:

loonies:
If this gets accepted, and is desirable behavior, I'll port this to 3.0
as well.

gsherwood:
As you said, I already fixed this in the 3.0 branch, so no porting of anything is required there. Instead of using your fix, I just ported that change back to master, but made one extra change to both branches; I still print a confirmation of zero error codes if no errors were found.

PR #607, opened 2015-05-31
Don't fatal due to Phar::running() on older HHVM versions

legoktm:
Until https://github.com/facebook/hhvm/issues/4323 was resolved, older
HHVM versions had the Phar extension installed, but it did not support
the Phar::running() function, causing fatal errors.

For other reasons, Wikimedia's CI installation is still running HHVM
3.3.1, causing fatal errors: https://phabricator.wikimedia.org/T100544.

The code now checks whether the Phar::running() function it plans to
call exists instead of just the class.

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -2300,7 +2300,7 @@ class PHP_CodeSniffer
     {
         if ($temp === false) {
             $path = '';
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for the PR

PR #608, opened 2015-05-31
Add a `brew install` method

sarahmonster:
This one was easier than the previously listed "easiest way to get started" :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ The SVN pre-commit hook requires PHP version 5.2.4 or greater due to its use of 
 Installation
 ------------
 
+On OS X you can install PHP\_CodeSniffer using [Homebrew](http://brew.sh/):
+
+    brew install php-code-sniffer
+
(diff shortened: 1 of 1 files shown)


TomasVotruba:
:+1:

aik099:
Are you sure, that this installation way still works? I can't seem to find Homebrew formula for that on https://github.com/Homebrew/homebrew/blob/master/Library/Formula/php-code-sniffer.rb url.

aik099:
The `brew search sniffer` returns following formulas and non of them seems to be official:

* josegonzalez/php/drupal-code-sniffer
* josegonzalez/php/php-code-sniffer
* homebrew/php/drupal-code-sniffer
* homebrew/php/php-code-sniffer
* Caskroom/cask/logicsniffer

gsherwood:
It's nice to know that there is a brew formula for PHPCS out there, but I only want the PHPCS docs to list install methods that I personally support. Then, if something is out of date or not working, they can talk to the people who maintain that project rather than asking here and not getting any real help.

Issue #530, opened 2015-04-03
Caching between runs for better performance


JDGrimes:
First let me thank you for this great tool. :+1: 

I've been using this on my PHP projects, and I've found that it can take a while to sniff the code, especially on larger projects with complex configurations. Performance will naturally be determined largely by how well the sniffs used are written. However, I think that performance could be increased by caching the hash signatures of the files being sniffed. Then only those files which have changes since the last sniff was conducted would need to be sniffed (there are some caveats which I'll get to in a moment). This wouldn't improve the performance of the initial sniffing (and might even degrade it slightly), but would drastically improve performance for latter sniffings.

As I noted above, there are some caveats:

- **The cache needs to be invalidated when the configuration changes.** This would be when the XML config file is edited, for example, and possibly at other times as well. This could be facilitated by saving a hash signature of these configuration settings to the cache, and checking whether this matches the current configuration when the cache is being loaded.
- **If a file has errors, the errors still need to get reported on subsequent sniffings even if the file hasn't changed.** This could be done by not caching files with errors, or by caching a list of the errors found and replaying them to the user without sniffing those files.

There are probably other things I haven't thought of, maybe regarding interactive mode, reports, or automatic fixing, all of which I am unfamiliar with. And there would probably need to be an easy way for the user to bypass the cache as needed.

There are probably also other things that could be cached between runs on a project as well.

Exactly how the cache is saved is up to you. I was thinking of a `.phpcs-cache` in the root of the project being sniffed that would contain the cache represented as a JSON object.

If this is something that you think could be done, I'd be happy to work up a PR if you'll give me a little guidance on how you'd like this implemented.

gsherwood:
I am still undecided as to if PHPCS itself should be doing things like this or if a wrapper script should be used, like what is done for hooking into a VCS.

Of course, it's possible, and probably fairly easy to get going, but I would not introduce it into the current 2.x versions regardless. I'd start with the 3.0 version, which is still in heavy development here: https://github.com/squizlabs/PHP_CodeSniffer/tree/3.0

How it is done is 1 of 2 options: something in the core and some command line values, or a new script like the SVN pre-commit hook or phpcbf. Either way, it's going to be easier/cleaner to do in the 3.0 version because of the refactoring I've done there, but I'm not ready to accept any new features on that branch at the moment.

JDGrimes:
[...]

PR #620, opened 2015-06-08
Implemented RecursiveFilterIterator to improve exclude file performance

oligriffiths:
The current exclude mechanism is very slow for a large number of excludes.
For each file to test, the entire excludes array is iterated over and a regex is performed. 
For our codebase, this takes over a minute for a single test (we have 1800 excludes).

The reason we have so many excludes is we want to switch from one coding standard to another, so we include all the current files in one standard and exclude in another.

This implementation uses a RecursiveFilterIterator and an improved exclude feature to speed up the process. It now takes just over 1 second instead of 1 minute to run out test suite.


oligriffiths:
Note that the tests pass, but the test suite doesn't pass on master.

gsherwood:
I really don't want to refactor how the directory is scanned in the 2.x version because it has been stable for too long. You might want to look at the 3.x branch, where I've already done some refactoring around how directories are searched.

For 2.x, it would be best if you could just tell me which bit of your code made the most difference, or re-submit a PR with the smallest changeset possible to achieve the performance increase.

On a more general note, if you have 1800 exclude patterns, then you are clearly in need of a completely different feature to help you manage things. If you give me more of an idea of what you are doing, maybe we can design a feature together that makes life easier for you.

Also, the reason the build is failing is because you have coding standard errors in the PR. It's not a big deal - I would need to add comments in your new class if I merged this PR anyway.

oligriffiths:
Sure thing, ill look into the 3.0 dev branch.

So let me explain the use case here.

We have a codebase that currently uses a subset of the PSR1/2 standard, with a bunch of exclusions and custom sniffs, snake case instead of camel case for example. We're talking about moving to a full PSR2 standard without the exclusions. Obvious, things would blow up if we just switched, so we want a way to run one rules across the codebase as it currently stands, and run the new ruleset against new files only.

In order to do that, we need to produce an index of all the current files, and use that for one ruleset, and exclude it from the other ruleset. Currently, the only way to do this is to include those files in a ruleset using <files> and exclude it from a ruleset using <exclude-pattern>.

The optimizations to the `ignoreFile()` method, checks the file using an index based lookup against a set of standard patterns: `(dir/path.php, ./dir/path.php, ^dir/path.php, ^dir/path.php$, ./dir/path.php$, dir, dir/, dir/*, ^dir/*, */dir/*)`, which is way faster than iterating the list. In the case of our new ruleset, we must exclude all the current files, so this was very slow.

The other issue with this approach, is that you can't process the entire codebase in one pass, you have to run phpcs with 2 different commands, one using the legacy ruleset, and one using the new ruleset.

Now if we're talking about adding features, for 3.0, a really useful feature would be to be able to have 2 separate standards, and be able to run each over a different set of files. An alternative approach would be to allow the file to contain a dockblock directive that would inform phpcs which standard/ruleset to use.

Thoughts?

oligriffiths:
Additionally, the new class, optimizes the filter iteration performance, if a file is to be excluded, then the iterator never iterates over it, additionally, if a directory is to be excluded, ALL subdirectories are then skipped, the current implementation will iterate over all subdirectories and perform a regex match check over them, for every file in that subdirectory, which is pointless if the directory is being ignored.

gsherwood:
Thanks for describing your use case. It makes sense why you've optimised things the way you have, but it is only optimising for a very specific case and I'm sure we can figure out a better way of doing things for you.

One obvious suggestion would be to tell PHPCS which files to process instead of which files to exclude. If you're already generating an excluded file list, outputting the opposite seems like it might be a simply thing to do. Writing a wrapper around PHPCS is another option, so you can inject the file list (presumably stored next to the rulesets) instead of including the files in a ruleset.

Ultimately, you're probably going to want some sort of Filter class that you can write and supply to PHPCS. I've described that concept briefly in another conversation about result caching in 3.0 here: https://github.com/squizlabs/PHP_CodeSniffer/issues/530

The idea being that once PHPCS has a list of files to process, a series of built-in filters can be applied (things like regex exclude patterns, modification date before/after x, git status) and/or custom filters can be applied for very custom filtering rules.

If this sounds like it might be something that could work for you, let's use this issue to figure something out and we can get some code going. The caching stuff on that other issue is pretty much done, so that's a lot of refactoring that's already been done.

oligriffiths:
The only reason I'd shy away rom producing a file list is that once we switch standards, the excludes will stay fixed, all new files will then conform to the new standard. I don't want to have to have scripts that we have to run every time a new file is added, that just won't work for our org. 

We basically want an include list for the old rule set and an exclude list for the old rule set. The current implementation is very wasteful when doing non-pattern based matches. 

Additionally, the current method will test all files in a subdirectory against a pattern match, even if the directory is excluded. The filter ensures that files in a directory that is excluded aren't event processed. Much more efficient.

oligriffiths:
Any update on this?

gsherwood:
Finding you a good solution is still on my todo, but I don't have time at the moment.

oligriffiths:
@gsherwood Any progress on this?

gsherwood:
> Any progress on this?

None. Still on my todo list though.

gsherwood:
Related commit: https://github.com/squizlabs/PHP_CodeSniffer/commit/4315292970fb1e48cf14b4a9012c6cf78f22df68

oligriffiths:
@gsherwood looking sharp, the git diff one is good, we currently use this via a custom script and comments back on github!

The reason we wanted this all in the first place, is to apply different sniffs to different files. For example, we want to apply one siff to new files, and a different one for old files, so we made an index of all current files, and exclude those from the new sniff, but can't do the inverse.

gsherwood:
@oligriffiths Sorry, I didn't have time to explain the change yesterday and left that comment as a note for myself.

I am building that filtering feature with the ability to have your own custom filter classes, like you can with reports. One of the options we discussed in this PR is for you to have a custom filter so you can do whatever you need to do.

For your specific case, you'll probably not make use of any PHPCS ignore patterns and instead use a filter (one for each of your standards) where you can do exact match checking on the file paths, or create something completely custom that doesn't even extend the PHPCS base Filter class. That allows PHPCS filtering to still be pretty basic but for you to do something more efficient for your specific case.

But I haven't done any dev on custom filters yet. Shouldn't take too long though.

gsherwood:
Another related commit: https://github.com/squizlabs/PHP_CodeSniffer/commit/f90e08c0c284df4a6b368af7a1af0d19ce27853a

oligriffiths:
Awesome. Sounds good. I like the direction.

gsherwood:
Closing this in favour of custom filter classes in 3.0. 3.0.0a1 is now released, so it can be tested.

Issue #627, opened 2015-06-19
Update your wiki page 'Coding Standard Tutorial'


christian-weiss:
Wiki page to update:
https://github.com/squizlabs/PHP_CodeSniffer/wiki/Coding-Standard-Tutorial

See diff for details:
https://github.com/christian-weiss/PHP_CodeSniffer_wiki/commit/f80ba1db45bfd7aadadfb201946ddf9035d39c51

gsherwood:
I made a slight change that I think makes the sub-directory requirement even more clear in the tutorial, without beating people over the head with it.

https://github.com/squizlabs/PHP_CodeSniffer/wiki/Coding-Standard-Tutorial/_compare/194f0525ba774c3a418426b46314c99ff6d3a0b3...ba8dd0ae9a72c1d1d80f91046a17363105683aaf

PR #628, opened 2015-06-19
enforce class naming convention for sniff classes, to …

christian-weiss:
...prevent subsequent E_NOTICE.

There are many places in this project where we access arrays by index without prior checking if the array do have that index key. In many places it is expected to have exact 4 elements in class names.

An E_NOTICE can make you IDE integration fail.

This issue is related to issue #627 .

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -1317,6 +1317,11 @@ class PHP_CodeSniffer
 
             include_once $file;
 
+            $elements = explode('_', $className);
+            if (count($elements) <> 4) {
+                throw new Exception('Sniff: class name '.$className.' do not fit to class naming convention. The class name must be with 4 elements separated by underscores.');
+            }
+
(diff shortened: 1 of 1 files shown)


aik099:
Please post a link to code line, that uses 1st, 2nd, 3rd and 4th parts of class name.

christian-weiss:
e.g https://github.com/christian-weiss/PHP_CodeSniffer/blob/master/CodeSniffer.php#L1371

christian-weiss:
and e.g. https://github.com/christian-weiss/PHP_CodeSniffer/blob/master/CodeSniffer.php#L1400
...maybe in other places, too.

gsherwood:
I've pushed a slightly different bit of code to fix the issue. Does this work for you?

christian-weiss:
Code looks ok. I think you have tested it, so we should go with that.

PR #630, opened 2015-06-20
Add a public member var so we can choose to customize...

photodude:
...elements of the sniff


aik099:
I usually just copy paste the sniff into my standard because all the sniff is doing is defining a patterns.

photodude:
I think it would be preferable to be able to use the ruleset override methods. Unfortenetly It looks like we `Cannot override final method PHP_CodeSniffer_Standards_AbstractPatternSniff::register()`

So without some change in AbstractPatternSniff::register() this change for defining a pattern that can be overridden likely isn't possible. (same with all other items that don't contain a public member var like register tokens)

aik099:
The `getPatterns` method defined patterns, that the `AbstractPatternSniff` uses. The `register` method you mentioned defined tokens to listen for. They are not related because `register` method listens for many common tokens to make it's work.

PR #631, opened 2015-06-21
Document the return value of the process() method on the sniff interface

klausi:
Just discovered this, so let's document it.


gsherwood:
I finally did something about this, although I changed the docs a bit. But at least it is documented now. Thanks for getting me to do it.

PR #634, opened 2015-06-26
Extended Squiz.Operators.ComparisonOperatorUsage to for, while and do-while

aboks:
When enforcing strict and explicit comparison operators for if-statements, I think it would be logical to also enforce these rules for for-statements, while-statements and do-while statements.


aboks:
I also wanted to add a unit test for javascript, but ran into a problem. I think the rule should also work for javascript without any modifications, but at the moment it does not. I suspect something is wrong in the JS tokenizer, but I do not have enough knowledge of that component to pinpoint the exact problem.

aboks:
Just ignore my last comment about problems with the JS tokenizer; I just hadn't noticed that the `ImplicitTrue`-error does not apply to JS files.

gsherwood:
Sorry for sitting on this for so long. It's a great change. Thanks a lot.

PR #639, opened 2015-07-09
Silence errors from iconv_strlen()

willemstuursma:
The return value of iconv_strlen() is checked, thus there is not need to have an error (E_NOTICE) when an illegal byte sequence is encountered in a file that is being checked.


gsherwood:
Makes sense. Thanks for submitting.

PR #642, opened 2015-07-16
Method params incorrectly detected when default value uses short array syntax

joshdavis11:
*When code sniffing for doc blocks, PHPCS broke when the PHP 5.4 short array syntax was introduced.

*To fix this, I added a check for a bracket_opener and then skipped to the bracket_closer similar to the check for a parenthesis_opener/closer.

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -2717,14 +2717,21 @@ class PHP_CodeSniffer_File
         $typeHint        = '';
 
         for ($i = ($opener + 1); $i <= $closer; $i++) {
-            // Check to see if this token has a parenthesis opener. If it does
-            // its likely to be an array, which might have arguments in it, which
-            // we cause problems in our parsing below, so lets just skip to the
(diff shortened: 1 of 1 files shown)


aik099:
Can you please add code example where it fails and used code standard?

joshdavis11:
This is what was failing:
```php
/**
 * Complete a step
 *
 * @param string $status status of step to complete
 * @param array  $array  array
 * @param string $note   optional note
 *
 * @return void
 */
public function completeStep($status, array $array = [Class1::class, 'test'], $note = '') {
...
}
```

```
FILE: /Users/jdavis/repos/main/test.php
 8 | ERROR | Doc comment for parameter $note does not match actual
   |       | variable name $array
```

aik099:
Now I clearly can see the problem. Thanks.

joshdavis11:
One other note: it only seemed to break when `$array` was not the last parameter. If it was the last parameter, it was fine (because it wasn't trying to parse the next parameter).

gsherwood:
Thanks a lot for fixing this.

PR #658, opened 2015-07-29
Squiz.Functions.FunctionDeclarationSpacing error for multi-line declarations with required spaces greater than zero

JDGrimes:
This fixes a bug in the `Squiz.Functions.FunctionDeclarationSpacing`
sniff when the `requiredSpacesAfterOpen` option was set to anything
other than 0.

When the first argument in a multiline function declaration had a
type-hint, this error would be given:

> Expected 1 spaces between opening bracket and type hint "array"; 0
found

I don’t think this check was ever intended to affect multiline
functions. The logic was such that it never does when
`requiredSpacesAfterOpen` is 0, and because this is the default, the
bug was never noticed.


gsherwood:
Thanks for the test and the fix.

PR #660, opened 2015-07-30
Syntax checks can fail on Windows with PHP5.6

hu2008yinxiang:
…PHP5.6

On Windows & PHP5.6, the "php -l" output an empty leading line, and make this sniff down, add flag "m" to make it work again.


gsherwood:
I recently made a change to fix HHVM support for this sniff in the 3.0 branch. I just ended up trimming the output before using it, so I've done the same thing in the master branch now so they are consistent.

PR #662, opened 2015-07-31
Add code coverage whitelist filter to phpunit config

munkie:
I would suggest to use code coverage tracking services like [coveralls.io](http://coveralls.io) or [codecov.io](http://codecov.io).
I would also suggest to rename _phpunit.xml_ to _phpunit.xml.dist_, but it will introduce BC break is guess.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -5,4 +5,12 @@
             <file>tests/AllTests.php</file>
         </testsuite>
     </testsuites>
+    <filter>
+        <whitelist>
+            <directory>./CodeSniffer</directory>
+            <exclude>
+                <directory>./CodeSniffer/Standards/*/Tests</directory>
+            </exclude>
+        </whitelist>
+    </filter>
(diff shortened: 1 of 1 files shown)


gsherwood:
I don't use code coverage reporting so I don't see a need for this change.

munkie:
@gsherwood I don't think this extra config in phpunit.xml will hurt somebody ((
Since code sniffer is about code quality, why don't you want to know code coverage percent of you library (which code quality metric too) ?

gsherwood:
I don't like including changes that I don't use as they will sit there untested over time. It should be easy enough for you to create your own config file if you'd like to do your own code coverage testing of PHPCS itself.

munkie:
I would suggest to rename phpunit.xml to phpunit.xml.dist and add phpunit.xml to .gitignore in order for contributors to be able to have their own version of phpunit.xml without risk of committing it to git

PR #668, opened 2015-08-03
Fix endless loop in PRS1 SideEffects sniffer

thomasjfox:
findNext() might return false, so we have to check for that.
Otherwise we overwrite our loop variable and get stuck.

Detected while playing around with the horde source tree.


aik099:
Haven't you noticed similar semicolon-related issues in other sniffs?

thomasjfox:
I have PSR1, PSR2, PEAR and "Generic" enabled for testing purposes.
They all seem to run fine.

thomasjfox:
I'll reopen the pull request with the typo "PRS1" <-> "PSR1" fixed.

PR #669, opened 2015-08-03
Convert HHVM's T_HASHBANG to T_INLINE_HTML (fixes #625)

### LINKED ISSUE #625: Consider working around T_HASHBANG in HHVM 3.5.x and 3.6.x
Some versions of HHVM's `token_get_all()` function will parse a hashbang like `#!/usr/bin/php` as T_HASHBANG instead of the T_INLINE_HTML that PHP5 uses. 

This has been fixed in HHVM 3.7 (https://github.com/facebook/hhvm/pull/5023), but does it make sense to add a workaround for this in PHPCS? https://gerrit.wikimedia.org/r/219394 is my fix for MediaWiki's custom sniff.

3.6 (an affected version) is an HHVM LTS that will be supported until January 28, 2016.

legoktm:
HHVM 3.5-3.6 tokenize "#!/usr/bin/php" as a non-standard T_HASHBANG
instead of PHP5's T_INLINE_HTML, so convert it to the expected
T_INLINE_HTML.

### DIFF
--- a/CodeSniffer/Tokenizers/PHP.php
+++ b/CodeSniffer/Tokenizers/PHP.php
@@ -673,6 +673,27 @@ class PHP_CodeSniffer_Tokenizers_PHP
             }//end if
 
             /*
+                HHVM 3.5 and 3.6 tokenizes a hashbang like #!/usr/bin/php
+                as T_HASHANG instead of T_INLINE_HTML,so fix that.
+            */
+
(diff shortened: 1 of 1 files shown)


legoktm:
Bump?

gsherwood:
Thanks for the PR. Sorry it took so long - been a crazy few months.

PR #670, opened 2015-08-04
Endless loop in PSR1 SideEffects sniffer if no semicolon after last statement

thomasjfox:
findNext() might return false, so we have to check for that.
Otherwise we overwrite our loop variable and get stuck.

Detected while playing around with the horde source tree.

Update v2: Fixed a "PRS1" <-> "PSR1" typo in my branch name / commit message.


gsherwood:
Thanks a lot for the fix, and special thanks for that unit test.

Issue #675, opened 2015-08-07
phpcbf not checking if patch is installed


manquer:
If patch is not installed phpcbf fails with the following error
````
sh: patch: command not found
Array
(
)
````

aik099:
Please use `--no-patch` option of `phpcbf` to prevent usage of `patch` program. This way diff will be created using pure PHP code without Unix-specific program usage.

manquer:
of course thanks for the info ..
shouldn't the fall back happen automatically if patch does not exist ? alternatively, the error message can suggest the --no-patch option ?

aik099:
Maybe. Theoretically Cygwin can install patch command on Windows as well and then it could work.

manquer:
of course, also for me this actually happened on a *nix machine which for some reason did not have patch installed.

gsherwood:
I've decided to remove the use of `diff` and `patch` for PHPCBF in the 3.0 version, and have committed that change to the 3.0 branch already. I don't want to make any changes to the process in the existing 2.x version for backwards compatibility reasons.

For 2.x versions, please use the `--no-patch` command line argument. This will replicate the default behaviour of the 3.0 version.

PR #677, opened 2015-08-09
PHPCBF not checking if patch or diff installed

Konafets:
Closes: #675

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -135,6 +135,14 @@ class PHP_CodeSniffer_CLI
         $cliValues['reportFile']   = null;
         $cliValues['reports']      = array();
 
+        if ((isset($cliValues['no-patch']) === false)
+            && ($this->_commandExists('diff') === false || $this->_commandExists('patch') === false)
+        ) {
+            echo 'The commandline tool "diff" or "patch" are not available on the system.'.PHP_EOL;
+            echo 'Use --no-patch option or install the programs.'.PHP_EOL;
+            exit(2);
+        }
+
         $suffix = '';
         if (isset($cliValues['suffix']) === true) {
             $suffix = $cliValues['suffix'];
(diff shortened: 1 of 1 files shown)


CodeSniffer/CLI.php
```diff
@@ -135,6 +135,14 @@ public function runphpcbf()
         $cliValues['reportFile']   = null;
         $cliValues['reports']      = array();
 
+        if ((isset($cliValues['no-patch']) === false)
+            && ($this->_commandExist('diff') === false && $this->_commandExist('patch') === false)
```
aik099: Maybe we should use `||` instead, because if user have diff installed, but patch isn't installed this ends up in same error as before this PR.

CodeSniffer/CLI.php
```diff
@@ -1276,4 +1284,23 @@ private function _validateReportWidth($width)
...
+    /**
+     * Checks if the given command exists on the system
+     *
+     * @param string $cmd The name of the command
+     *
+     * @return bool
+     */
+    private function _commandExist($cmd)
+    {
+        $returnVal = shell_exec("which $cmd  2> /dev/null");
```
aik099: Ha, now `which` only available on Unix-like systems. Instead you should use OS detection and for Windows use `where` command (see http://stackoverflow.com/questions/304319/is-there-an-equivalent-of-which-on-the-windows-command-line).

CodeSniffer/CLI.php
```diff
@@ -1276,4 +1284,23 @@ private function _validateReportWidth($width)
...
 
 
+    /**
+     * Checks if the given command exists on the system
+     *
+     * @param string $cmd The name of the command
+     *
+     * @return bool
+     */
+    private function _commandExist($cmd)
```
aik099: Please add `s` to the method name (to have `_commandExists`), because it would be grammatically correct.

CodeSniffer/CLI.php
```diff
@@ -1276,4 +1284,23 @@ private function _validateReportWidth($width)
...
+    {
+        $returnVal = shell_exec("which $cmd  2> /dev/null");
+        if (empty($returnVal) === true) {
```
aik099: You can just return inverted IF condition.

gsherwood:
I've decided to remove the use of `diff` and `patch` for PHPCBF in the 3.0 version, and have committed that change to the 3.0 branch already. I don't want to make any changes to the process in the existing 2.x version for backwards compatibility reasons.

For 2.x versions, please use the `--no-patch` command line argument. This will replicate the default behaviour of the 3.0 version.

Issue #624, opened 2015-06-18
Expected type hint "MyClass[]"; found "array"


NikolaySl:
**Example code:**
```php
    /**
     * Does foo with MyClass objects.
     *
     * @param MyClass[] $list List of objects.
     */
    function foo( array $list ) {}
```

**Expected result:**
This should be OK.

**Actual result:**
```
 4 | ERROR | Expected type hint "MyClass[]"; found "array" for $list
```

gsherwood:
Duplicate of #601.

6d70:
Why was this closed? The comments in #601 clearly mentioned to create a new issue? 

I have the same issue, PHPStan is happy this comments but this rule is conflicting. What should the advice be when we PHPCS and PHPStan?

PR #685, opened 2015-08-21
Allow specified arrays in docblock type hints

### LINKED ISSUE #601: Expected type hint int[]; found array in Squiz FunctionCommentSniff
**Example code:**

```php
	/**
	 * Does foo with IDs.
	 *
	 * @param int[] $ids List of IDs.
	 */
	function foo( array $ids ) {}
```

**Expected result:**
This should be OK.

**Actual result:**

```
 4   | ERROR | Expected type hint "int[]"; found "array" for $ids
     |       | (Squiz.Commenting.FunctionComment.IncorrectTypeHint)
 ```

---

The error comes from https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php#L386.

### PR DESCRIPTION
**Pull request by nicwortel**
Example:

```php
/**
 * @param Item[] $items
 */
public function foo(array $items)
{
    // do something
}
```

Should be fine but currently returns an error: Expected type hint "Item[]"; found "array" for $items

This fixes #601 and #624

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -359,7 +359,7 @@ class Squiz_Sniffs_Commenting_FunctionCommentSniff extends PEAR_Sniffs_Commentin
                 } else if (count($typeNames) === 1) {
                     // Check type hint for array and custom type.
                     $suggestedTypeHint = '';
-                    if (strpos($suggestedName, 'array') !== false) {
+                    if (strpos($suggestedName, 'array') !== false || strpos($suggestedName, '[]') !== false) {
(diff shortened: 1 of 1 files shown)


aik099:
Nice catch.

There might be other sniff, that verify function parameter types against DocBlock as well. Please check them if you find any.

nicwortel:
To be honest, I'm not very familiar with the internals of code sniffer yet. Do you mean other sniffs within the Squiz standard? Or in other standards?

aik099:
All sniffs in all standards. I guess you can search for `array` or `callable` find find these places.

gsherwood:
This issue is being fixed in PR #708 , along with another issue. The fix is almost the same, but just checks if the last two chars are `[]` rather than looking for them anywhere in the type hint. Both ways looks to work fine.

Thanks a lot for sending in the PR.

PR #706, opened 2015-09-16
Fix neglect of indentation in JS arrays

dryabkov:
Fix neglect of indentation in arrays typical for ExtJs, like  

Ext.define('AddressBook.view.matrix.base.BaseToolbar', {
    extend: 'Ext.toolbar.Toolbar',

    cls: 'addressbook-matrix-base-toolbar',
    height : 27,

    initComponent: function() {
        var me = this;

        me.items = [
            '->',
            {
                tooltip: me.viewHelpTooltip,
                text: '?',
                cls: 'matrix-viewhelp-button',
                listeners: {
                    click: 'onViewHelpButtonClick'
                }
            }
        ];

        me.callParent();
    }
});


gsherwood:
This sniff specifically doesn't try to indent arrays (and a range of other multi-line definitions) because coding standards tend to have different indent standards when single-line statements go across multiple lines.

So the indent of JS arrays (and PHP arrays) is neglected on purpose.

I'd be willing to look at that JS tokenizer change in a new PR though, but only for a major version bump as it breaks backwards compatibility. It also looks like the kind of change that could be made in the main tokenizer method rather than in processAdditional(). If that's possible, it makes the debug output a bit easier to understand.

gsherwood:
Closing due to lack of feedback.

PR #707, opened 2015-09-16
Squiz MethodScopeSniff doesn't handle nested functions

oligriffiths:
Currently, this sniff incorrectly fails for locally scoped named method like:

```php

public function test()
{
    function my_sort()
    {
    // do stuff
    }

    $a = array();
    usort($a, 'my_sort');
}
```

This patch checks if the method is defined within the scope of another method


gsherwood:
I couldn't merge this PR in because there was already a core function to help with this (see the commit that fixes the problem) so I used a different method of fixing it. But thank you for submitting this PR and reporting the problem.

PR #710, opened 2015-09-22
Use Composer's facility of autoloading files

morozov:
If PHP_CodeSniffer is a composer dependency of another application, then instead of only requiring `vendor/autoload.php` (which is a common practice when using Composer), it's also needed to require `vendor/squizlabs/php_codesniffer/autoload.php` additionally.

Adding `autoload.php` to composer's autoload.files section eliminates this need.

### DIFF
--- a/bin/phpcbf
+++ b/bin/phpcbf
@@ -8,6 +8,6 @@
  * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
  */
 
-require_once __DIR__.'/../autoload.php';
+require_once __DIR__.'/../vendor/autoload.php';
 $runner = new PHP_CodeSniffer\Runner();
 $runner->runPHPCBF();
--- a/bin/phpcs
+++ b/bin/phpcs
@@ -8,6 +8,6 @@
  * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
  */
 
-require_once __DIR__.'/../autoload.php';
+require_once __DIR__.'/../vendor/autoload.php';
 $runner = new PHP_CodeSniffer\Runner();
 $runner->runPHPCS();
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "psr-4": {
             "PHP_CodeSniffer\\": "src/",
             "PHP_CodeSniffer\\Tests\\": "tests/"
-        }
+        },
+        "files": ["autoload.php"]
     },
     "require": {
         "php": ">=5.4.0",
--- a/tests/AllTests.php
+++ b/tests/AllTests.php
@@ -30,7 +30,7 @@ if (defined('PHP_CODESNIFFER_VERBOSITY') === false) {
     define('PHP_CODESNIFFER_VERBOSITY', 0);
 }
 
-require_once __DIR__.'/../autoload.php';
+require_once __DIR__.'/../vendor/autoload.php';
 
 $tokens = new Tokens();
 
--- a/tests/Core/AllTests.php
+++ b/tests/Core/AllTests.php
@@ -30,7 +30,7 @@ if (defined('PHP_CODESNIFFER_VERBOSITY') === false) {
     define('PHP_CODESNIFFER_VERBOSITY', 0);
 }
 
-require_once __DIR__.'/../../autoload.php';
+require_once __DIR__.'/../../vendor/autoload.php';
 
 $tokens = new Tokens();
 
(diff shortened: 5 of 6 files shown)


aik099:
Why you need to have direct access to PHP_CodeSniffer classes?

bin/phpcbf
```diff
@@ -8,6 +8,6 @@
  * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
  */
 
-require_once __DIR__.'/../autoload.php';
+require_once __DIR__.'/../vendor/autoload.php';
```
aik099: This (and other changes below) are BC break because this PATH would only exist, when PHP_CodeSniffer is located in `/vendor` folder, but not when it's installed in PEAR or as standalone package.

morozov: No, it works on the out-of-the box PHP_CodeSniffer. Instead of requiring code sniffer's autoload file, we delegate autoloading to Composer and thus make sure all other dependencies will be loaded as well.

If at any point in time PHP_CodeSniffer will depend on anything, it will require this change anyways.

morozov:
> Why you need to have direct access to PHP_CodeSniffer classes?

Because I build my application on top of PHP_CodeSniffer as a composer dependency.

aik099:
Can you please share a link to that application? I wonder what application might have PHP_CodeSniffer as it's dependency. To be the PHP_CodeSniffer is pure development-related tool.

morozov:
It is a pure development-related tool. Please take a look at https://github.com/morozov/diff-sniffer-pre-commit and https://github.com/morozov/diff-sniffer-pull-request.

aik099:
Sounds familiar. I recall some discussions about that in past. It all sounds good in theory, but in real life there always would be cases that prevent code from being PHPCS error/warning free. Blocking commit based on error/warning presence is too harsh IMO.

morozov:
You can always add `--no-verify` to the commit command if you think the hook is too picky.

aik099:
I prefer post validation. This way nobody evades from their code being checked and check results are publicly visible on build server.

Example from my life:

* there is a rule, that checks that type hint in method declaration matches one in DocBlock (e.g. if there is `array` in DocBlock then there must be `array` type hint before parameter declaration as well)
* there is an interface that violates that rule (in vendor package, that isn't validated)
* there is a class that implements that interface (in checked code)
* if you attempt to fix PHPCS warning by adding array type hint, then this would result in fatal error because method signature no longer matches the interface

morozov:
> I prefer post validation. This way nobody evades from their code being checked and check results are publicly visible on build server.

Pull request validator implements post validation.

> Example from my life: …

This is when you'd have to use `--no-verify` but it's more of an exception rather than a rule.

aik099:
> This is when you'd have to use --no-verify but it's more of an exception rather than a rule.

But there might some other non-related violations that needs to be checked.

Anyway great package as replacement to current `pre-commit` validation hook that PHP_CodeSniffer has. Maybe it's worth mentioning on documentation somewhere.

morozov:
> But there might some other non-related violations that needs to be checked.

Sure. Usually I read the hook output and try to fix all its concerns that make sense to me. Then I commit with `--no-verify`.

Coding standards is one of those rules which developers should either follow voluntarily or not follow at all. If you decide to follow them, you'll try to help your tools to help you, not to cheat them :-).

gsherwood:
Sorry, just getting around to working on 3.0 again.

I can't merge this PR in as it doesn't allow PHPCS to work when not installed via composer. Indeed, it makes composer a dependency of PHPCS (a regular git clone no longer works), which is something I don't want to do.

Is the composer.json change enough to do what you want?

morozov:
@gsherwood the changes in `composer.json` should be enough.

Regarding the dependency on composer, I agree that introducing a new dependency is not a good thing in general. But in my understanding, at some point dependency on composer is inevitable. As soon as you start breaking the app into components (and in fact, PHPCF and PHPCBF are two separate application which share some code base), you will have to use a dependency management tool which composer is the de-facto standard for.

Not trying to teach you anything, but if the dependency is avoided in order to make the app easier to install, the approach which PHPUnit uses seems the right way to me.

gsherwood:
> @gsherwood the changes in composer.json should be enough.

Thanks.

> But in my understanding, at some point dependency on composer is inevitable.

That's not my understanding. I have been able to keep PHPCS working from a checkout or clone (CVS, SVN, now Git) even while still keeping it installable by PEAR, Composer and OS package managers. It doesn't have any dependencies, so it's pretty easy.

> As soon as you start breaking the app into components (and in fact, PHPCF and PHPCBF are two separate application which share some code base), you will have to use a dependency management tool which composer is the de-facto standard for.

If I ever do that, then maybe I will, yes. Although PHPCBF/PHPCS is not a good example because they actually share the entire code base. The only difference is the init code, so I would never separate them.

PR #712, opened 2015-09-23
Fix whitespaces

KorDum:
Убраны правила валидации пробелов у закрывающих скобок блоков - }, endif и тд. - для YL1Strict


aik099:
@KorDum you don't need to fork whole PHP_CodeSniffer repository to add your own standard. You just need to install PHP_CodeSniffer globally (via Composer or PEAR) or locally (via Composer) and then just specify full path to your standard folder when running `phpcs` or `phpcbf` command instead of just standard name.

PR #714, opened 2015-09-23
Fixes suppression of errors using docblocks

akarmazyn:
Currently the only acceptable way of suppressing errors is by either using single line comments:
```
// @codeStandardsIgnoreStart
```
Or doc block comments:
```
/*
 * @codeStandardsIgnoreStart
 */
```

This commit makes it possible to suppress errors by using PHPDoc comments:
```
/**
 * @codeStandardsIgnoreStart
 */
```

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -485,7 +485,7 @@ class PHP_CodeSniffer_File
         foreach ($this->_tokens as $stackPtr => $token) {
             // Check for ignored lines.
             if ($token['code'] === T_COMMENT
-                || $token['code'] === T_DOC_COMMENT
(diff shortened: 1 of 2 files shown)


aik099:
Before this PR following code was working, but I think after this PR it won't:
```php
/** @codeStandardsIgnoreStart */
```

akarmazyn:
My original comment has a typo on every annotation, it should be `@codingStandardsIgnoreStart` instead of `@codeStandardsIgnoreStart`.
The commit of course uses the correct annotation name.

gsherwood:
Thanks for the fix and tests. Looks like a problem that came from the switch to the new comment tokenizer.

PR #716, opened 2015-09-24
JSON report is invalid when messages contain newlines or tabs

pfrenssen:
When any of the error or warning messages contain tabs or newlines and a JSON report is generated then the resulting JSON will be invalid.

A good solution for this would be to use `json_encode()` to generate the JSON, but since we need to support old PHP versions this is not possible until PHP CodeSniffer 3 (see #449 for more information).

For the moment we can work around the issue by replacing the offending characters before outputting the JSON report.

### DIFF
--- a/CodeSniffer/Reports/Json.php
+++ b/CodeSniffer/Reports/Json.php
@@ -65,6 +65,7 @@ class PHP_CodeSniffer_Reports_Json implements PHP_CodeSniffer_Report
                     $error['message'] = str_replace('"', '\"', $error['message']);
                     $error['message'] = str_replace('/', '\/', $error['message']);
                     $error['message'] = str_replace("\n", '\n', $error['message']);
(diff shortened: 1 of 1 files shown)


CodeSniffer/Reports/Json.php
```diff
@@ -64,6 +64,8 @@ public function generateFileReport(
                     $error['message'] = str_replace('\\', '\\\\', $error['message']);
                     $error['message'] = str_replace('"', '\"', $error['message']);
                     $error['message'] = str_replace('/', '\/', $error['message']);
+                    $error['message'] = str_replace("\n", '\n', $error['message']);
```
aik099: Maybe also add `\r` to the list.

pfrenssen:
@aik099 great suggestion, thanks!

I have added a line that converts carriage returns to the harmless '\r' character sequence.

gsherwood:
Thanks a lot for the fix.

PR #717, opened 2015-09-24
Remove default full report when requiring a single report.

icharlie:
Remove the default full report since it is a single report.

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -663,6 +663,7 @@ class PHP_CodeSniffer_CLI
                     }//end if
                 } else {
                     // This is a single report.
+                    unset($this->values['reports'['full']])
                     $report = substr($arg, 7);
                     $output = null;
                 }//end if
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -663,7 +663,7 @@ class PHP_CodeSniffer_CLI
                     }//end if
                 } else {
                     // This is a single report.
-                    unset($this->values['reports'['full']])
+                    unset($this->values['reports']['full']);
                     $report = substr($arg, 7);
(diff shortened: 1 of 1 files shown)


CodeSniffer/CLI.php
```diff
@@ -663,6 +663,7 @@ public function processLongArgument($arg, $pos)
                     }//end if
                 } else {
                     // This is a single report.
+                    unset($this->values['reports'['full']])
```
aik099: This results in syntax error on PHP 5.3.

icharlie: My mistake. Thanks!

gsherwood:
I don't know what this change is designed to fix and it actually breaks a pretty key feature of PHPCS, which is the ability to print multiple reports at a time. For example, this change breaks this command:

`phpcs --report=full --report=summary /path/to/code`

But because of the ordering, would still allow:

`phpcs --report=summary --report=full /path/to/code`

PHPCS doesn't hard-coded a default report format before processing the CLI argument (it uses the full report only if no reports were specified) so maybe your full report is being set via a config setting or inside a custom ruleset? If this is the case, you should probably remove it from there as it would be forcing that report format at all times.

icharlie:
@gsherwood Thanks! After, I use `phpcs --config-delete` and it works correctly. But I can't find `ruleset.xml` or `phpcs.xml`. How can I find the configuration file used by phpcs?

gsherwood:
> How can I find the configuration file used by phpcs?

Run the `phpcs` command with the `-vv` command line argument and the first chunk of output will show you how it is setting all its rules. Run it on a small file though because the `-vv` output can be pretty long.

icharlie:
@gsherwood Thank you!!

Issue #704, opened 2015-09-15
ScopeIndent can fail when an opening parenthesis is on a line by itself


oligriffiths:
Run this sniff over the following:

```php
$a =
    (
        $b = 1
    );
$c = 2;
```

You get the error `7 | ERROR | [x] Line indented incorrectly; expected at least 4 spaces, found 0`

This is correctly indented (and an abbreviated example). The code for this sniff is quite complex which is why I've not submitted a PR.

Any ideas?

gsherwood:
I think I figured it out. Thanks for reporting.

oligriffiths:
Anytime. 

Got a couple of others that I'll submit PRs for too.

PR #730, opened 2015-10-04
ScopeIndent can fail when a short array opening square bracket is on a line by itself

klausi:
Only a test case for now, not sure about the correct fix.


gsherwood:
Thanks for the test. Basically the same fix as #704.

PR #732, opened 2015-10-06
PHP Notice if @package name is made up of all invalid characters

3aBXo3:
If `@package` name will be empty in comment block, e.g. `@package ''`, then PHP throw notice: 
`PHP Notice:  Uninitialized string offset: 0 in .composer/vendor/squizlabs/php_codesniffer/CodeSniffer/Standards/PEAR/Sniffs/Commenting/FileCommentSniff.php on line 350`,
because `$firstBit[0]` is not set, but used.

### DIFF
--- a/CodeSniffer/Standards/PEAR/Sniffs/Commenting/FileCommentSniff.php
+++ b/CodeSniffer/Standards/PEAR/Sniffs/Commenting/FileCommentSniff.php
@@ -347,6 +347,12 @@ class PEAR_Sniffs_Commenting_FileCommentSniff implements PHP_CodeSniffer_Sniff
             $newContent = preg_replace('/[^A-Za-z_]/', '', $newContent);
             $nameBits   = explode('_', $newContent);
             $firstBit   = array_shift($nameBits);
+            
+            if(!isset($firstBit{0})){
+				// Package is empty.
+				continue;
+			}
+            
             $newName    = strtoupper($firstBit{0}).substr($firstBit, 1).'_';
             foreach ($nameBits as $bit) {
                 if ($bit !== '') {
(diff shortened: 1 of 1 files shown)


CodeSniffer/Standards/PEAR/Sniffs/Commenting/FileCommentSniff.php
```diff
@@ -347,6 +347,12 @@ protected function processPackage(PHP_CodeSniffer_File $phpcsFile, array $tags)
             $newContent = preg_replace('/[^A-Za-z_]/', '', $newContent);
             $nameBits   = explode('_', $newContent);
             $firstBit   = array_shift($nameBits);
+            
+            if(!isset($firstBit{0})){
```
aik099: There is something wrong with indentation here and in other changed place.

3aBXo3: 1. Sorry about this. I'm editing file online, from github.  
2. If using just `$firstBit[0]`, and `@package` name will be empty in comment block, e.g. `@package ''`, then PHP throw notice: 
`PHP Notice:  Uninitialized string offset: 0 in .composer/vendor/squizlabs/php_codesniffer/CodeSniffer/Standards/PEAR/Sniffs/Commenting/FileCommentSniff.php on line 350`,
because `$firstBit[0]` is not set, but used.
3. Fixed.

aik099: But what `$firstBit{0}` technically does? I thought that it only can be used on objects.

3aBXo3: Characters within strings may be accessed and modified by specifying the zero-based offset of the desired character after the string using square array brackets, as in `$str[42]`. Think of a string as an array of characters for this purpose.
Strings may also be accessed using braces, as in `$str{42}`, for the same purpose. 

More on [php.net](http://php.net/manual/en/language.types.string.php#language.types.string.substr).

aik099: OK, then `$firstBit{0}` represents 1st letter in a string.

3aBXo3: Yes, it is right.

gsherwood:
The package name is not empty in this test case - it's just made up entirely of invalid characters. So an error must be thrown, but with a different message and error code.

Further testing shows that this isn't a problem for subpackage tags. If they are empty, they aren't processed.

gsherwood:
I've committed a fix for this issue that also adds the new error message and unit tests. Thanks for reporting it.

Issue #735, opened 2015-10-07
`switch` related problem when using PSR2 standard


raincious:
So, this happened after I committed my code:
![](https://cloud.githubusercontent.com/assets/4244947/10332045/eec535e2-6d0c-11e5-8e09-8c5d132e1dfb.png) 
Notice the CI (running `dev-master` version of the PHP_CodeSniffer) tells me my code had problem, and my local `phpcs` (stable version `2.3.4`) tell my everything is fine.

And I don't think there should be 16 spaces right? :P

Please investigate.

Code example:
https://github.com/raincious/facula/blob/4b1bc7bfe754b5031f015f094ed4a567e455d518/src/Facula/Base/Tool/Misc/ConfigParser.php#L129

CI Result:
https://travis-ci.org/raincious/facula/jobs/84041382

Minimized example code:

    function blabla()
    {
        switch ($dataType) {
            case static::DATA_LIST:
            case static::DATA_LIST_FORMATSTR:
            case static::DATA_HASHTABLE:
                return array();
            case static::DATA_VECTOR:
                return '';
        }

        throw new Exception\FailedGetDataTypeBase(
            $dataType
        );
    }

Result:

    $ ./vendor/bin/phpcs /tmp/phpcstest.php --standard=PSR2

    FILE: /tmp/phpcstest.php
    ----------------------------------------------------------------------
    FOUND 1 ERROR AFFECTING 1 LINE
    ----------------------------------------------------------------------
     10 | ERROR | [x] Line indented incorrectly; expected 12 spaces,
        |       |     found 8
    ----------------------------------------------------------------------
    PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
    ----------------------------------------------------------------------

And if I change above code to:

    function blabla()
    {
        switch ($dataType) {
            case static::DATA_LIST:
            case static::DATA_LIST_FORMATSTR:
            case static::DATA_HASHTABLE:
                return; // Notice this line. bool, integer, string etc will also work.
            case static::DATA_VECTOR:
                return '';
        }

        throw new Exception\FailedGetDataTypeBase(
            $dataType
        );
    }

It passes. 

I heavily suspect it related to `()[]` symbol matching.

aik099:
@raincious , any more details you can provide about this problem except that it's `switch` and PSR-2 related?

raincious:
Hi @aik099, 

[...]

PR #736, opened 2015-10-08
More complex switch-return structure (Affiliate to #735)

raincious:
Previously:
https://github.com/squizlabs/PHP_CodeSniffer/issues/735

The commit 9badf27cc7 did fix the most of the problems, but there are still a missed hit.

**I have no idea on how to fix the problem itself**, but I did added the test case for it. In short, it's something like:

	switch ($foo) {
		.....
		case 7:
			return functionCall(
				$withMultiLineParam[0],
				array(),
				$functionReturningString(
					$withMultiLineParam[1]
				)
			);
		.....
	}

It's also effecting something like ...

	switch ($foo) {
		.....
		case 7:
			return functionCall(
				array()
			);
		.....
	}

Or just ...

	switch ($foo) {
		.....
		case 7:
			return array(
				array(),
			);
		.....
	}


gsherwood:
Thanks for the additional tests. They are passing now.

raincious:
@gsherwood Thank you!

PR #744, opened 2015-10-17
PSR2.ControlStructures.SwitchDeclaration now produces a distinct code when body is not on line following statement.

mikesherov:


mikesherov:
I searched for a way to test this, but it seems the current unit test infra how no way to verify the errorCode produces by an error. Please let me know how to proceed if you feel this needs a test. Thanks!

mikesherov:
Nevermind, discovered how to change settings inline which will allow me to write an effective test. Updating now!

gsherwood:
You don't need to check the specific error code in the unit tests. The testing infrastructure doesn't care and trying to trick it with things like suppressing specific error codes isn't worth doing. 

Thanks for reporting the duplicate error code. It was obviously a copy/paste error I have made as it should clearly be a distinct code.

mikesherov:
Hi @gsherwood, thanks for merging! Any idea when we might be able to expect a patch release containing this change?

gsherwood:
> Any idea when we might be able to expect a patch release containing this change?

No idea, although the bug list is just about cleared, so I imagine it wont be too long.

PR #748, opened 2015-10-23
Auto fix for Squiz.Commenting.BlockComment.WrongEnd is incorrect

JDGrimes:
The `$stackPtr` was being used instead of `$end`, leading to the
comment opener being replaced with `*/`. This would cause a syntax
error in the “fixed” PHP file.

The sniff did not have a test for the fixing component, which is why
this wasn’t caught sooner. I’ve added the `.fixed` file, but it is just
a copy of the `.inc` file, because I didn’t know for sure whether all
of the fixing behavior is correct. So currently the tests are
intentionally failing.

See
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/478


gsherwood:
Thanks a lot for finding and fixing that.

PR #754, opened 2015-10-28
Update .travis.yml

paladox:
Add testing for php 7.


gsherwood:
"nightly" is php7. I've been testing on PHP7 since Feb when they made it available. "7" is just an alias at the moment, and I'll change it when it is officially released with a version number.

PR #756, opened 2015-10-28
Lowering php version down to 5.3

paladox:


gsherwood:
I'll make a call on 5.3 when I am closer to a release. There will also be pressure for me to support 5.6 as the min, but I can't do that because I can't even use 5.6 is production yet.

PR #769, opened 2015-11-15
Incorrect detection of variable reference operator when used with short array syntax

klausi:


gsherwood:
Thanks a lot for the fix.

Issue #747, opened 2015-10-21
Enhancement to the Squiz_Sniffs_Commenting_FileCommentSniff Sniff


techxplorer:
Hi,

Would you consider making the author and copyright values in the [Squiz_Sniffs_Commenting_FileCommentSniff](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FileCommentSniff.php) sniff properties that could be overridden by configuration? 

I'm using many of the other comment related tests for my own development, and would like to be able to use this one as well. However I don't want to create my own standard for just one sniff. 

If you would consider it, and would prefer a pull request, please let me know.

gsherwood:
I've been meaning to remove them for a long time. I'd be happy to make them configurable instead. If you'd like to submit a PR, please do so. It would make things move a lot faster.

techxplorer:
Hi,

I've been working on this issue, the progress commit is [available here](https://github.com/techxplorer/PHP_CodeSniffer/commit/3f9b3548ab437e918806d3a2a0bd127319a1c703).

The unit tests are failing for the author tag sniff, which is odd as my changes are not substantial. I've had a quick look and it appears that with my changes in place the fixer is not enabled. 

Could you investigate and let me know what I'm getting wrong?

gsherwood:
I got this when running your new sniff: PHP Notice:  Undefined property: Squiz_Sniffs_Commenting_FileCommentSniff::$expectedCopyright in /Users/gsherwood/Sites/Projects/PHP_CodeSniffer/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FileCommentSniff.php on line 159

You're using the wrong member var name inside your code. Fixing that gets the tests passing again.

Just a quick related comment: I don't think having the regex as a member var is a good idea. It looks very error prone, and you've hard-coded the fact that the years should be in the fixed version as well with `str_replace('xxxx-xxxx', '', $this->expectedCopyright)`.

I think the sniff should enforce the copyright years bit, but allow you to decide what string goes after that. Then you don't need a regex sitting in a ruleset.xml file.

It might also be worth considering breaking author into a name and email. This, along with the copyright change, would allow the sniff to enforce a specific format for these tags, but not make assumptions about the values that go in there.

So I'd have member vars like: authorName, authorEmail, copyrightName, copyrightMinYear

Then figure out the formats and regexes from those values.

[...]

PR #774, opened 2015-11-19
Enhance Squiz/Sniffs/FileCommentSniff #747

techxplorer:
Makes the content of the author and copyright tags available via
properties so that they can be overridden in a custom standard.

Addresses the enhancement that I lodged in item #747


gsherwood:
I know it's been a long time, but I've changed my mind on this one and don't think these settings should be configurable. I prefer the idea that these settings are removed, or the sniffs rewritten in some way, as doc comment standards are a real pain to produce.

Thanks for your time making this PR.

PR #775, opened 2015-11-19
Ensure $values['standard'] is set when generator is specified on the command line.

kenguest:
This is to fix the following from happening:

    $ phpcs --generator=Markdown .
    PHP Warning:  Invalid argument supplied for foreach() in /usr/share/php/PHP/CodeSniffer/CLI.php on line 789
    PHP Stack trace:
    PHP   1. {main}() /usr/bin/phpcs:0
    PHP   2. PHP_CodeSniffer_CLI->runphpcs() /usr/bin/phpcs:25
    PHP   3. PHP_CodeSniffer_CLI->process() /usr/share/php/PHP/CodeSniffer/CLI.php:95

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -782,6 +782,9 @@ class PHP_CodeSniffer_CLI
 
         if ($values['generator'] !== '') {
             $phpcs = new PHP_CodeSniffer($values['verbosity']);
+            if (is_null($values['standard'])) {
+                $values['standard'] = $this->validateStandard(null);
+            }
             foreach ($values['standard'] as $standard) {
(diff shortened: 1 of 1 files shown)


aik099:
Strange, that you're getting that error considering, that by default `PSR-2` standard is used, when no `--standard` argument is specified.

kenguest:
They weren't pulling back the default standard, and checking if one was set before going on to that foreach loop on line 789, which is what my Pull Request addresses.

gsherwood:
Thanks for the fix

gsherwood:
Just a note that the 3.0 branch sets the default standard much higher up and didn't suffer from this problem.

PR #776, opened 2015-11-19
Add CLI parameter for listing available generators.

kenguest:


aik099:
By generators you mean what you can specify via `--report` parameter?

kenguest:
I mean a doc. generator, not a report . Pertaining to this:
        <generator>   The name of a doc generator to use
                      (forces doc generation instead of checking)

which you'll see if you run "phpcs --help" at the command line :-)

aik099:
Ah. I never knew it even existed. I saw the folder in each standard with docs, but I thought it's for reading by human only, not for automatic documentation generation.

kenguest:
Yes, it seems to be a way of generating documentation of the "Do's" and "Dont's" for an installed Standard. Very nifty.

gsherwood:
I don't really think a CLI option to list generators is needed. There are so few of them (and no plans for any more) that I don't really want to reserve a CLI arg for it. An easier solution is just to include them in the help content, like I do for report types.

Something like this:
```
<generator>   Uses either the "HMTL", "Markdown" or "Text" generator
              (forces documentation generation instead of checking)
```

Thoughts?

kenguest:
That's fine with me :-) Though I'd be inclined to make it a little bit more robust as it threw an error when I used "markdown" as the generator name (lower case) instead of specifying it in Proper case.

gsherwood:
I've changed the help output. I don't have time to put towards support case insensitivity at the moment.

PR #780, opened 2015-11-21
Add PHPUnit dependency to composer manifest

cmueckedev:
In order to facilitate running tests I recommend adding an explicit dependency to composer.json.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,6 @@ Before you contribute code to PHP\_CodeSniffer, please make sure it conforms to 
 
 Which should give you no output, indicating that there are no coding standard errors. And then:
 
-    phpunit
+    vendor/bin/phpunit
 
 Which should give you no failures or errors. You can ignore any skipped tests as these are for external tools.
--- a/composer.json
+++ b/composer.json
@@ -56,6 +56,9 @@
         "ext-tokenizer": "*",
         "ext-xmlwriter": "*"
     },
+    "require-dev": {
+        "phpunit/PHPUnit": "~4.0"
+    },
     "bin": [
         "scripts/phpcs",
         "scripts/phpcbf"


aik099:
Developers most likely have PHPUnit installed on their machines and Travis CI has it as well. So no need to depend on it.

gsherwood:
Thanks, but I don't use composer to develop PHP_CodeSniffer and I don't need this dependancy. Anyone including PHP_CodeSniffer in this own composer.json would already be including phpunit as well, or not really care about running the PHPCS test suite.

JohnMaguire:
@gsherwood I would like to petition you to reconsider.

While you may not use `composer.json` in any manner, there are 103 users who have contributed code to this project. By adding the dependencies of your project (of which PHPUnit _is_ one, check `tests/`) to `composer.json`, it allows users who are familiar with the typical composer process: `git clone` -> `composer install` to get up and running very quickly.

CONTRIBUTING suggests running `phpunit` prior to submitting a pull request. With that being the case, it should be listed as a dev dependency (i.e. only necessary when developing this project.)

You're correct that people who include PHP_CodeSniffer in their own `composer.json` are likely to be including PHPUnit as well, but again, if you're working on PHP_CodeSniffer, you're doing that in its own clone (thus the only place to add the phpunit dependency is within `composer.json` in this project.)

To be clear, adding a dev dependency will NOT put a hard dependency on PHPUnit when PHP_CodeSniffer is included through an application's `composer.json`.

mrrsm:
I agree with JohnMaguire (and cmueckedev) here. 
By adding phpunit as a dev-dependency it  gives you the added benefit of being able to specify what version of phpunit the user should be using as well.  As of right now it is ambiguous and there is no standard saying what each package manager for every distribution will have so if there is a test that doesn't work in a particular version of phpunit for whatever reason you have a way to control the version people are testing with.

aik099:
At the end developer gets multiple copies of PHPUnit for each project that has it as dev dependency, but essentially it's same PHPUnit every time. Just multiple copies.

JohnMaguire:
@aik099 Actually, having multiple versions installed is a bug not a feature. I do understand where you're coming from, but the fact is the PHPUnit I rely on is not necessarily the PHPUnit you rely on.

This is what composer is meant to solve.

JohnMaguire:
I guess really, I'm having a hard time understanding what the argument _against_ adding this dependency is. I've seen lots of reasons why it's "unnecessary", but why is it unwanted?

aik099:
Faster Travis CI build maybe.

JohnMaguire:
Actually, if you take a look at the Travis CI builds, it doesn't run `composer install`: https://travis-ci.org/squizlabs/PHP_CodeSniffer/jobs/92838873

gsherwood:
I can't use the latest stable PHPUnit version for development because 5 requires PHP5.6 (and 6 next year will require PHP7). I work in an enterprise environment, which means PHP5.3 min for at least a couple more years (I'm considering bumping PHPCSv3 to PHP5.4 min, but might need to leave it at 5.3) and so no new PHPUnit versions for me.

My dependency would therefore install an old unsupported version of PHPUnit for you to use and block newer versions because I can't use them to test on my supported platforms.

I suggest you just use whatever version of PHPUnit you have globally installed for your dev environment because it is unlikely that it is the same as mine. And I'll probably need to use this version past its EOL date, so I shouldn't be forcing that on anyone else.

If you find that the unit tests dont run on newer PHPUnit version, please report bugs and I will try and make them compatible with newer versions as well.

JohnMaguire:
One nice feature of composer is that you can specify acceptable versions of a dependency to install. Generally you would set your minimum version of dependencies to match your own dependencies (so, given that this project says it supports PHP 5.1 and up, you would find a version of PHPUnit which is supported by 5.1).

Unfortunately, composer support wasn't added to PHPUnit prior to it depending on PHP 5.3. PHPUnit 3.7 (when composer support was first added) for example depends on PHP 5.3. All PHPUnit 4.x versions require at least PHP 5.3 as well.

Given that that is the case, a dev dependency of PHPUnit 4.x would probably suffice. I think there might be a fundamental misunderstanding as to how dependencies are resolved as well: You seem to be concerned about PHPUnit conflicting with the application's version of PHPUnit. However, there are two facts that keep this from being an issue.

1. Dev dependencies of application dependencies are NOT resolved by default. This means if your application requires PHPUnit 5.0, it doesn't matter, because the `PHP_CodeSniffer`'s dependency on PHPUnit is *not* resolved.
2. By specifying only a minimum version necessary of PHPUnit, it can co-exist peacefully with any application or library that requires a higher version of PHPUnit.

gsherwood:
But I'd need to specify PHPUnit4 *only*. Although I'd absolutely never use that dependency because I only work off a git clone. So maybe newer versions would work, but you wont be able to test PHPCS properly because you're obviously not testing 5.3+ if you're using PHPUnit5+. So what's the point of specifying a dependancy that I will never use and everyone else will just ignore?

mrrsm:
If you specify phpunit in composer at 4.* that is what I will test against as that is what I will get if I am working on fixes for PHP_CodeSniffer and used require-dev when composer installing.  I am not sure how it would be ignored in this case?

This is a change that will help people who don't have phpunit installed globally and not cause anybody else to notice.  It also makes it a good base standard to verify the tests run against whatever version you have specified phpunit at.

JohnMaguire:
> But I'd need to specify PHPUnit4 _only_.

Hum, why do you say that? Like I said, by specifying a PHPUnit dependency of `>= 4.0` you're allowing any version of 4.x to be installed (I actually thought 5.x too, but I can't prove it, and it wouldn't be a great idea). composer won't try installing a version of PHPUnit incompatible with your version of PHP as far as I can tell.

> So maybe newer versions would work, but you wont be able to test PHPCS properly because you're obviously not testing 5.3+ if you're using PHPUnit5+.

Not quite sure I'm following. PHPUnit 5.x cannot be run on any versions of PHP under 5.6. This means that if you are using PHPUnit 5.x, you can only test this class under PHP 5.6+.

However, PHPUnit 5.x wouldn't actually be installed with the `~4.0` constraint. I believe you might be able to make it installable with `~4`, but it wouldn't be great practice. Theoretically a major version change would indicate backwards incompatibility.

> So what's the point of specifying a dependency I will never use and everyone else will just ignore?

Well that's just it: You won't use it, but again, the 103 (and counting) contributors to this repository can use the dependency to quickly run your tests. Many devs do not like installing *anything* globally, including PHPUnit. This is because of version conflicts. By localizing it within a repository's dependencies, you actually ensure you _don't_ have versioning issues. They wouldn't just be ignoring it.

Today, using the PHPUnit installed globally is not an issue. However, when PHPUnit gets around to implementing PSR-compatible namespacing for example, some versions of PHPUnit would work, and some would not. By specifying a version in `composer.json`, you're also specifying which versions *work*.

JohnMaguire:
I will say this: This really doesn't hurt you in any way, and it doesn't cause the `PHP_CodeSniffer` install process to change. The change will only affect people who *do* use composer, and who *do* want to be able to install their dependencies through it.

gsherwood:
Obviously my fears of conflicts and forcing unsupported versions on developers are unfounded. I've added the require-dev line.

JohnMaguire:
Thank you! :+1:

PR #783, opened 2015-11-23
Add bootstrap file option

JohnMaguire:
This adds a longopt for specifying bootstrap files. An example of why this might be useful:

```php
<?php

PHP_CodeSniffer::$allowedTypes = array_merge(PHP_CodeSniffer::$allowedTypes, ['int', 'bool']);
```

You can set a bootstrap file in `ruleset.xml` using the `<arg>` element. Previously we were modifying `$allowedTypes` in our custom sniff files, but there's a race condition unless you add this to every single sniff file you have (and you end up with lots of dupes unless you run `array_unique` each time.)

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -600,6 +600,17 @@ class PHP_CodeSniffer_CLI
                 }
 
                 $this->values['sniffs'] = $sniffs;
+            } else if (substr($arg, 0, 10) === 'bootstrap=') {
+                $files = explode(',', substr($arg, 10));
+                foreach ($files as $file) {
+                    if (file_exists($file) === false) {
+                        echo 'ERROR: The specified bootstrap file "'.$file.'" does not exist'.PHP_EOL.PHP_EOL;
+                        $this->printUsage();
+                        exit(2);
+                    }
+                }
+
+                $this->values['bootstrap'] = $files;
             } else if (substr($arg, 0, 12) === 'report-file=') {
                 $this->values['reportFile'] = PHP_CodeSniffer::realpath(substr($arg, 12));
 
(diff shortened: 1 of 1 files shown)


gsherwood:
This is an interesting idea, but I feel like it is something that should be included later on in the process. Maybe after all command line arguments have been processed and the run is about to start.

Possibly right here: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/CLI.php#L782

And in the 3.0 branch, probably at the top of the main run() method: https://github.com/squizlabs/PHP_CodeSniffer/blob/3.0/src/Runner.php#L245

Does that still let you do what you need?

JohnMaguire:
@gsherwood I believe so. I figured it probably wasn't the right place for it, but didn't know where was. The main thing is the bootstrap needs to be run before the sniffs are. (In the example I gave, my use case, we want to make sure `$allowedTypes` is updated before any sniffs' checks are run.)

Would you like a separate pull request for each branch?

gsherwood:
@JohnMaguire No need for two PRs. 3.0 isn't stable yet so it may not work for you. It's also completely different, so I'll port it over based on how I've been converting things in there.

If you think the idea of storing the bootstrap location in the `$_cliArgs` member var and then including the file around line 782 is a good one, just modify the PR for that and I'll get it merged in.

JohnMaguire:
@gsherwood I hope I understood correctly! I was able to base my code roughly off the sniffs option above as well.

gsherwood:
> I hope I understood correctly!

Exactly what I was thinking. I can't give it a good test right now, but will test and merge tomorrow. Thanks a lot.

gsherwood:
Thanks a lot for the patch. There 3.0 commit is here: https://github.com/squizlabs/PHP_CodeSniffer/commit/88c587f3ed356dd9cbc3b7dee71ff3ad6fe98d4d

PR #793, opened 2015-11-28
Fix error message for non-existent bootstrap file

JohnMaguire:
`PHP_CodeSniffer::realpath()` either returns false, or a file path. We don't need to check `file_exists()` because `PHP_CodeSniffer::realpath()` does it for us. However, we do want to show a proper filename in the event that the referenced file does not exist.

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -604,14 +604,14 @@ class PHP_CodeSniffer_CLI
             } else if (substr($arg, 0, 10) === 'bootstrap=') {
                 $files = explode(',', substr($arg, 10));
                 foreach ($files as $file) {
-                    $file = PHP_CodeSniffer::realpath($file);
-                    if (file_exists($file) === false) {
+                    $path = PHP_CodeSniffer::realpath($file);
+                    if ($path === false) {
                         echo 'ERROR: The specified bootstrap file "'.$file.'" does not exist'.PHP_EOL.PHP_EOL;
                         $this->printUsage();
                         exit(2);
                     }
 
-                    $this->values['bootstrap'][] = $file;
+                    $this->values['bootstrap'][] = $path;
                 }
             } else if (substr($arg, 0, 12) === 'report-file=') {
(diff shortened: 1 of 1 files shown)


JohnMaguire:
Additionally, I ran into another bug, but I'm less sure how to solve it properly.

One of the ways I was planning to utilize this functionality was by adding the a `<arg>` directive to our `ruleset.xml`. Unfortunately there are a couple issues with this:

1. I can't come up with a good way to resolve the path in relation to the `ruleset.xml`. i.e. I'm imaginging a `bootstrap.php` shipped in the same directory as `ruleset.xml`. `PHP_CodeSniffer::realpath()` currently has no concept of this. I'm not sure if there's an easy way to add it? It would require knowing which `ruleset.xml` is currently being parsed at the time we parse the `bootstrap` argument. This is a bit of a misnomer though because `PHP_CodeSniffer` can process multiple `ruleset.xml`'s, and for all we know `--bootstrap` was passed via a command-line argument.

2. Sort of on the same vein, `<arg name="bootstrap" value="~/path/to/standard/bootstrap.php"/>` currently does not work. I did a little bit of tracing, and I think it has to do with the fact that although command-line arguments are evaluated prior to `process()` being called, the arg from `ruleset.xml` does not appear to be.

For point 2, I had added a log of `$this->values` just after the `$this->setCommandLineValues($args)` call in `getCommandLineValues()` on line 377, and saw that it had parsed the `standard` and other values I had passed in via CLI, but had (I suppose obviously) not interpreted the args out of the `ruleset.xml` yet.

I added a log in `process()` prior to the bootstrap inclusion loop and found the same thing.

However, it does eventually get parsed. I know because I get an error for a non-existent file, for example. Any direction on where I might look to fix this issue would be appreciated. :)

And thank you for this project.

gsherwood:
Thanks for the fix. I also moved the bootstrap file inclusion down to just before the run starts so you can specify it in a ruleset, although you're not going to be able to include it using a relative path just yet. I think that would require a specific bootstrap tag in the ruleset so special processing can take place inside the ruleset processing code. If you need that feature, open a new issue and we can track it there.

PR #796, opened 2015-11-30
New option to avoid indentation reset on T_OPEN and T_CLOSE tags

stronk7:
Right now any T_OPEN_TAG/T_OPEN_TAG_WITH_ECHO/T_CLOSE_TAG do lead,
unconditionally, to current indentation being reset. But it may be
desired for some scripts/standards to keep that current indentation to "persist"
over multiple PHP blocks.

This commit adds a new property (resetIndentationWithOpenClose) to
support both behaviors, defaulting to true, aka keeping BC.

Without this property the only solution is to duplicate the whole
process() method just to switch to the alternative behavior.

Some references:

- The issue in our tracker: https://tracker.moodle.org/browse/CONTRIB-5732
- Test file, only reporting line #6 (property set to true, reset, aka, current behavior in non-strict mode), and reporting both #6 and #18 (property set to false, alternative behavior in non-strict mode) so the current indentation (set to 4 by the T_IF in the first PHP block) remains in the 2nd PHP block: https://github.com/stronk7/moodle-local_codechecker/blob/master/moodle/tests/fixtures/moodle_whitespace_scopeindent.php

FYC, ciao :-)

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/WhiteSpace/ScopeIndentSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/WhiteSpace/ScopeIndentSniff.php
@@ -107,6 +107,18 @@ class Generic_Sniffs_WhiteSpace_ScopeIndentSniff implements PHP_CodeSniffer_Snif
     protected $nonIndentingScopes = array();
 
     /**
+     * Is indentation reset on T_OPEN_TAG and T_CLOSE_TAG?
+     *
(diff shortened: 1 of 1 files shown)


stronk7:
Hi, I'm preventively closing this as far as we found a problem with our indentation customizations. Not sure if the changes in this PR are the cause, but better we close this till we find the culprit.

Ref: https://tracker.moodle.org/browse/CONTRIB-5732?focusedCommentId=384774&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-384774

Issue #797, opened 2015-11-30
Parsing CSS url() value breaks further parsing


SebastianZ:
If `url()` values are parsed, the token array is recreated, which breaks the stack pointer.

Example:

If you have the following CSS the type of the second `opacity` property will be `T_STRING` instead of the expected `T_STYLE`:

```css
div {
  width: 100%;
  height: 50px;
  font-size: 1.2em;
}

.class {
  color: #F00;
  background: linear-gradient(to bottom, #00F, #0F0) repeat scroll 50% 50% #EEE;
  opacity: 0.3;
}

.kachelBorderKachel {
  min-width: 250px;
  max-width: 100%;
  padding-bottom: 50px;
}

.boxAll {
  box-shadow: 2px 3px 3px rgba(100, 100, 100, 0.33);
}

.boxHeader {
  border-left: 1px solid #000;
  border-right: 1px solid #000;
  border-top: 1px solid #000;
  border-bottom: 1px dotted #000;
  background: #1E2A5F linear-gradient(62deg, #1E2A5F 0%, #2B88D9 50%, #5FA1E2 51%, #00AEEA 100%);
}

.metro2Overlay {
  width: 100.1%;
  height: 100.1%;
  background: url(../jquery/cupertino/images/ui-bg_diagonals-thick_90_eeeeee_40x40.png)
      repeat scroll 50% 50% #EEE;
  opacity: 0.3;
  color: #F00;
}
```

Sebastian

gsherwood:
Thanks for the report. More info on PR #798.

SebastianZ:
Thank you for the fast fix! As far as I tested it works fine for me now.

Sebastian

PR #798, opened 2015-11-30
Fix for issue #797

SebastianZ:
Instead of recreating the tokens array skips undefined array items


gsherwood:
I ended up just setting `$stackPtr` to the correct value instead of assuming there will be undefined indexes because I already knew the correct value. I also added a bunch of `-vv` verbose output because I realised that it was a bit empty.

Thanks for the PR though.

SebastianZ:
Good point. Setting the stack pointer correctly seems to be a better approach.

Sebastian

PR #819, opened 2015-12-14
Move coding standards config for PHPCS itself to a dist file

klausi:


gsherwood:
I'm not sure why this change would be needed. The phpcs.xml file is there to set the coding standards for the PHPCS project itself. So it is named the same way as it would be in any other project. It's not an example file.

klausi:
As with phpunit.xml.dist I think the upstream project should come with a *.dist file. If you want to experiment with custom phpunit or phpcs settings then it is nice that you can do that without them showing up in git diff for example.

So I think this is the same case as with phpunit.xml.dist.

klausi:
Oops, I'm seeing just now that PHPCS ships with a phpunit.xml file, not a phpunit.xml.dist file. I would argue that both phpunit.xml and phpcs.xml should ship in their *.dist version to make overriding easier during development.

gsherwood:
Why would you be doing dev inside PHPCS itself and not want to use the standard and test settings that the project requires?

klausi:
Because I might want to customize reporting, example: I always want the "-s" argument when invoking phpcs to always show which actual sniff caused the error. I then just copy the *.dist file and modify it, without causing changes in git and interfering with the patch I'm currently developing.

gsherwood:
Fair enough. I only *just* added support for .dist files in PHPCS, so I can do this now.

gsherwood:
I changed them both.

Issue #822, opened 2015-12-15
Ruleset hard-coded file paths are not used if not running from the same directory as the ruleset


gsouf:
When I run phpcs version  2.5.0 on my project the script never ends and so does phpcbf.
Either I run phpcs with custom arguments or simply with ``./vendor/bin/phpcs`` it will fail (never end)

Now if I downgrade to 2.4.0 it works well.

I'm afraid I can't give more useful information but feel free to ask for details.

UPDATE: It seems that php version does not matter (5.4 5.4 5.6 7.0)

gsouf:
Got it. It's due to file path from xml. It seems the behaviours changed since version 2.4.0. 
Creaitng a patch for that

aik099:
I'm not sure this is needed in 2.x versions because in 3.x versions this behavior is changed to immediately exit when no file contents is provided.

gsouf:
I found where the issue comes from.

This addition in the version 2.5: http://git.io/v0ghw  the ``<file>`` nodes are getting ignored if the xml file is not in the same directory as the cwd.

I use to store my ruleset.xml in an other directory, and then it broke my build. 

That's a compatibility break that should not occur in a minor version update. 

Did my part of the job, no it's up to you to tell if yes or not you want to go back to the previous behaviour. :)

gsouf:
@aik099  even if 3 changed, it's good to fix version 2 ! Isn't it ?

aik099:
> @aik099 even if v3 changed, it's good to fix version 2 ! Isn't it ?

Sure this needs to be fixed in 2.x versions.

gsherwood:
As you've figure out already, the problem is caused by PHPCS not processing hard-coded file paths unless the command is being run from the same directory as the ruleset. This was originally to allow PHPCS to find phpcs.xml files up the tree and use those for processing, but I obviously fixed the file path issue another way and didn't need that extra check in the end because the functionality still works without it.

Until a fix is released, you will need to specify the file paths in your build script. The rest of the ruleset should be processed normally.

gsherwood:
After further investigation, this code was from a fix for the 3.0 version and is not needed in 2.x because of a different way it merges CLI arguments and ruleset values. So looked safe to remove from 2.x after testing.

PR #823, opened 2015-12-15
pipe processing was endless

gsouf:
That's a first issue met in #822 : when no file was provided and phpcs was called without stdin content the script was never ending

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -879,7 +879,11 @@ class PHP_CodeSniffer_CLI
         if (empty($values['files']) === true) {
             // Check if they are passing in the file contents.
             $handle       = fopen('php://stdin', 'r');
-            $fileContents = stream_get_contents($handle);
+            $fileContents = '';
+            stream_set_blocking($handle, 0);
+            while(($content = fgets($handle)) !== false){
+                $fileContents .= $content;
+            }
             fclose($handle);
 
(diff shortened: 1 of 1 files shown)


gsherwood:
I've changed the way STDIN work in 3.0 so that you must pass `-` on the command line to have PHPCS wait for input. But I'm not changing that in 2.x for BC reasons, and I don't want to change it so that it is non-blocking. I use this feature myself for testing the checking and fixing of random code and it is quite handy.

You've only submitted this PR because of an unrelated issue and this wont fix your problem over there (PHPCS will just report that there are no files to check) so I'm going to close this and continue a discussion on the other issue about the root cause.

gsouf:
That would fix travis ci freezing for ten minutes with no error message

gsouf:
Then dont consider I "only submitted this PR because of an unrelated issue"

gsherwood:
> Then dont consider I "only submitted this PR because of an unrelated issue"

I'm not sure what you are trying to say.

gsouf:
I mean your comment was offending and that's not an appropriate response for a contribution to an open source project

gsherwood:
I'm sorry you thought I was being offensive. I think you've completely misunderstood my comment.

I was trying to explain why I didn't want to make this change in the 2.x version and that the change wont actually fix the bug you found. I wanted to continue discussion on the actual bug report and find a solution, which I think I've done.

PR #830, opened 2015-12-23
Generic.PHP.LowerCaseKeyword ignores "array" hint keyword

mathroc:
`array` does not seems to be recognized as a keyword when it's not followed by an opening parenthesis. In the [php documentation](http://php.net/manual/en/reserved.keywords.php) it is shown followed by a parenthesis but I think it should be treated as a keyword. 

otherwise, the following statement from the documentation would not make sense for `array` : 

 > You cannot use any of the following words as constants, class names, function or method names


gsherwood:
Checking the type hint to ensure it is lowercase makes sense to me. Thanks for the patch and the tests.

PR #833, opened 2015-12-28
Composer autoload include statements to make Composer dependencies work

klausi:
Problem: Drupal's Coder defines a PHPCS standard where one sniff uses the Symfony YAML library. That triggers fatal errors because potential Composer dependencies are never included for auto loading.

Let's do it in CLI.php?
PHP-CS-Fixer does something similar: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/php-cs-fixer

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -15,9 +15,9 @@
 error_reporting(E_ALL | E_STRICT);
 
 // Installations via Composer: make sure that we autoload all dependencies.
-if (file_exists($a = dirname(__FILE__).'/../../../autoload.php')) {
+if (file_exists($a = dirname(__FILE__).'/../../../autoload.php') === true) {
     include_once $a;
-} else if (file_exists($a = dirname(__FILE__).'/../vendor/autoload.php')) {
+} else if (file_exists($a = dirname(__FILE__).'/../vendor/autoload.php') === true) {
     include_once $a;
 }
 
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -14,6 +14,13 @@
 
 error_reporting(E_ALL | E_STRICT);
 
+// Installations via Composer: make sure that we autoload all dependencies.
+if (file_exists($a = dirname(__FILE__).'/../../../autoload.php')) {
+    include_once $a;
+} else if (file_exists($a = dirname(__FILE__).'/../vendor/autoload.php')) {
+    include_once $a;
+}
+
 if (is_file(dirname(__FILE__).'/../CodeSniffer.php') === true) {
     include_once dirname(__FILE__).'/../CodeSniffer.php';
 } else {


TomasVotruba:
:+1:

gsherwood:
Thanks for the patch. CLI.php certainly looked like the right place to put this.

danimalweb:
Is it possible to add an ignore for this? For my use case I do not want my autoload file included.

aik099:
@danimalweb , please explain more about the problem you're having, when Composer auto-loader is included.

danimalweb:
I run phpcs via the gulp plugin gulp-phpcs to check my PHP files. I exclude the vendor directory (where I have many php packages used for my application) using the standard exclude pattern in gulp.src e.g:

```return gulp.src(['**/*.php', '!vendor/**/*.*', '!tests/**/*.*'])```

After updating to phpcs 2.5.1 now running my script I recevie lot's of undefined function warnings.

```Fatal error: Call to undefined function apply_filters() in /vendor/valendesigns/option-tree/ot-loader.php on line 91```

FYI, I'm using this in a WordPress theme development setup. These wanrings might be valid issues with the packages I'm autoloading but for the purposes of my checking I don't want to scan this files.

**Comment by aik099** [comment]
The PHP_CodeSniffer isn't calling any code during checking process. That must be something else in Gulp doing that. Probably one of your dependencies has custom autoloader which is expecting some configuration.

I'm not even sure how it's possible that any autoloader is invoked in PHP_CodeSniffer workflow, because it loads checked classes on it's own.

**Comment by danimalweb** [comment]
Thanks for the reply.

Simply rolling back to phpcs 2.5.0 fixed my problems, So maybe its a compatibility issue with this latest version and the gulp-phpcs npm plugin as no other npm plugins or composer libraries have been updated.

**Comment by gsherwood** [comment]
@danimalweb Any chance you could comment out the code that does the autoloading in a 2.5.1 install and see if you are still getting issues? Then we can rule out all other changes between 2.5.0 and 2.5.1. The code is here: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/CLI.php#L17-L22

**Comment by danimalweb** [comment]
@goatherd Installing 2.5.1 and commenting out the code block you suggested fixes my issue.

**Comment by gsherwood** [comment]
@danimalweb Thanks for testing that. At least we know it is the autoloader.

But I'm not sure what to do about it. If you install PHPCS via a composer dependency, it probably should ensure that the composer autoloader is run so custom sniffs will work correctly. Having a command line option to disable this will force the code to run at a later time in the process (it happens before any CLI process begins right now) and seems like a strange option.

It's probably better to figure out why including an autoloader is causing issues with the gulp plugin. It's obviously due to this: https://github.com/valendesigns/option-tree/blob/master/ot-loader.php#L91

Including that option-tree package as a dependency is creating a new object and doing some things in the constructor, including calling a function that doesn't exist yet. I guess it should be fixed there by checking if the function exists before using it. It seems that it is designed to be run in a different environment (presumably WordPress).

**Comment by danimalweb** [comment]
Hi thanks for the feedback.

I thing is some WordPress plugins often get packaged as composer dependencies in order to allow the developer to embed the plugin into a custom WP theme as a dependency which utlizes composer. This is pretty common place even for some of the major premium plugins like ACF Pro.

I removed the option-tree package to see if this issue was isolated to only that dep, but the same happens with another dep I use: https://github.com/TGMPA/TGM-Plugin-Activation/

```Fatal error: Call to undefined function did_action() in /vendor/tgmpa/tgm-plugin-activation/class-tgm-plugin-activation.php on line 1910```

Any further thoughts? Thanks a lot for your insight so far.

**Comment by gsherwood** [comment]
> Any further thoughts? Thanks a lot for your insight so far.

I have almost no knowledge of composer as I don't use it for any projects, so I'm pretty much relying on everyone else for this one.

@klausi any help?

**Comment by aik099** [comment]
According to WordPress plugin development guide (see https://codex.wordpress.org/Writing_a_Plugin) there should be following line on top of plugin file to avoid cases, when file is included prior WordPress code functions:

```php
defined( 'ABSPATH' ) or die( 'No script kiddies please!' );
```

That should solve a problem. Fact, that PHP_CodeSniffer started to include Composer's autoloader just exposed fact, that some of plugins doesn't adhere to plugin writing guidelines.

As @gsherwood said we can't really know what autoloader of each dependency does to disable it (even if we could).

danimalweb:
I can't really count on the coding standards of other developers to get it 100% right before I decide to include a package into my composer deps. The phpcs is awesome and I hope to continue using it as it forces me to write better code.

Is it possible to pass a flag into phpcs to disable this new autoload code? Otherwise, I'll have to stay locked to v2.5.0. I'm sure this will be useful for other developers too.

aik099:
> Is it possible to pass a flag into phpcs to disable this new autoload code? Otherwise, I'll have to stay locked to v2.5.0. I'm sure this will be useful for other developers too.

Not right now. Probably adding a flag won't hurt.

gsherwood:
Does doing a composer global install of PHPCS instead of including it as a dependency of a project stop this problem from occurring?

gsherwood:
> Is it possible to pass a flag into phpcs to disable this new autoload code?

An option like this would mean that PHPCS can never include any composer dependencies as it would need to include the composer autoloader to get them working. Having a CLI option to disable the autoloading would effectively stop PHPCS from working as well. While I have no plans of using any composer packages right now, I don't want to shut the door on that completely.

Hopefully the global install creates and includes a different autoloader. If not, the PHAR file, or a PEAR install, are your other options to use PHPCS right now.

PR #836, opened 2015-12-29
Context sensitive keywords support

kukulich:


gsherwood:
I wonder if it might be easier to think about what tokens are allowed after `const`, `function` and `::` rather than maintain the blacklist. It might be as short as:
- T_STRING
- T_VARIABLE
- T_DOLLAR
- T_OPEN_PARENTHESIS
- A few more I'm sure I have forgotten

kukulich:
@gsherwood Ok, I have just forcepushed new version.

gsherwood:
Thanks a lot for these changes.

PR #838, opened 2016-01-02
CSS indentation fixer changes color codes

klausi:
Here is a test case for demonstration: the line ```background: #072237;``` is changed into ```background: 1072237;```.

Not sure where this should be fixed, so the phpunit tests will fail on this right now.


klausi:
Yay, found the bug in the CSS tokenizer. This should now be ready to be merged.

gsherwood:
Thanks a lot. What a stupid mistake that was :)

PR #844, opened 2016-01-04
Recognize T_SPACESHIP token as operator token

aik099:
This solves 2 problems:

1. the `<=>` operator added in PHP 7 wasn't interpreted as operator token (the `T_SPACSHIP`) by sniffs that handle operators (e.g. `Squiz.WhiteSpace.OperatorSpacing`)
2. if custom standard referred to this `T_SPACESHIP` token and PHP 7 code was checked on PHP 5 version then `T_SPACSHIP` constant wasn't defined

### DIFF
--- a/CodeSniffer/Tokens.php
+++ b/CodeSniffer/Tokens.php
@@ -130,6 +130,11 @@ if (defined('T_POW') === false) {
     define('T_POW', 'PHPCS_T_POW');
 }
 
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for this. I've also replicated the token for older PHP versions.

aik099:
Thanks.

PR #845, opened 2016-01-04
composer: fix dev branch alias

TomasVotruba:


gsherwood:
Actually, wouldnt this be better as an alias like `2.x-dev`, if that is possible?

aik099:
According to https://getcomposer.org/doc/articles/aliases.md what you're suggesting should also work.

PR #848, opened 2016-01-04
Default to PHP_BINARY for syntax tests

deviantintegral:
Without this, you have to manually set php_path in your configuration for this test to work. This also reduces a skipped test in the default testing configuration.

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
@@ -56,31 +56,36 @@ class Generic_Sniffs_PHP_SyntaxSniff implements PHP_CodeSniffer_Sniff
     {
         $phpPath = PHP_CodeSniffer::getConfigData('php_path');
         if ($phpPath === null) {
-            // PHP_BINARY is available in PHP 5.4+
-            if (defined('PHP_BINARY')) {
+            // PHP_BINARY is available in PHP 5.4+.
+            if (defined('PHP_BINARY') === true) {
                 $phpPath = PHP_BINARY;
-            }
-            else {
+            } else {
                 // Fallback for old PHP versions that can be removed once PHP
                 // 5.4 is required.
                 $candidates = array(
-                  PHP_BINDIR . '/php',
-                  PHP_BINDIR . '/php.exe',
-                );
+                               PHP_BINDIR.'/php',
+                               PHP_BINDIR.'/php.exe',
+                              );
 
                 foreach ($candidates as $candidate) {
-                    if (file_exists($candidate) && is_executable($candidate)) {
+                    if (file_exists($candidate) === true
+                        && is_executable($candidate) === true
+                    ) {
                         $phpPath = $candidate;
                         break;
                     }
                 }
 
-                if (empty($phpPath)) {
-                    $phpcsFile->addWarning('Unable to check syntax as php_path is not configured', $stackPtr, 'PHPSyntax');
+                if (empty($phpPath) === true) {
+                    $phpcsFile->addWarning(
+                        'Unable to check syntax as php_path is not configured',
+                        $stackPtr,
+                        'PHPSyntax'
+                    );
                     return;
                 }
-            }
-        }
+            }//end if
+        }//end if
 
         $fileName = $phpcsFile->getFilename();
         $cmd      = "$phpPath -l \"$fileName\" 2>&1";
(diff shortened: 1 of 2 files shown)


CodeSniffer/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
```diff
@@ -56,7 +56,7 @@ public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
     {
         $phpPath = PHP_CodeSniffer::getConfigData('php_path');
         if ($phpPath === null) {
-            return;
+            $phpPath = PHP_BINARY;
```
aik099: The `PHP_BINARY` constant is only available since PHP 5.4 (see http://php.net/manual/en/reserved.constants.php).

deviantintegral: A few options I see:

* PHP 5.3 has been unsupported for over a year, so we just assume PHP_BINARY is available.
* Wrap this in a check for the constant, and fall back to the old behaviour if it's unset.
* Use PHP_BINDIR which is in 5.3 and assume the binary is called 'php'.

aik099: > PHP 5.3 has been unsupported for over a year, so we just assume PHP_BINARY is available.

The PHP version constraint is set in `composer.json` so it must work.

You can use a combination of above suggestions:

* If constant is defined use it.
* If constant doesn't exist guess the binary ( or maybe it's available via `phpinfo();` somehow )
* If guessing failed work as before: return

deviantintegral: I've added fallbacks using PHP_BINDIR, looking for php and php.exe. Any other commonly distributed php executable names? I figure anything unique can still be configured using php_path.

CodeSniffer/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
```diff
@@ -56,7 +56,30 @@ public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
...
+
+                foreach ($candidates as $candidate) {
+                    if (file_exists($candidate) && is_executable($candidate)) {
+                        $phpPath = $candidate;
+                        break;
+                    }
+                }
+
+                if (empty($phpPath)) {
+                    $phpcsFile->addWarning('Unable to check syntax as php_path is not configured', $stackPtr, 'PHPSyntax');
```
aik099: This is a BC break, because under same conditions before no warning was generated.

deviantintegral: As-is, if someone adds the rule to a standard, leaves php_path undefined, and doesn't get any warnings, wouldn't they assume the sniff passed? This burned me on a project recently which is why I added it.

aik099: That is an interesting question really. I'll let @gsherwood answer it, when he'll be back from holidays.

gsherwood:
The 3.0 version does auto-discovery in a different way. The PR is here: https://github.com/squizlabs/PHP_CodeSniffer/pull/571

For compatibility, I would merge a change that simply used PHP_BINARY if available and if the php_path config var is not set, but skip the test and sniff processing otherwise. I don't want to add any auto-discovery code into this branch, or generate a new error if the path is not found, or remove the skip test function.

So pretty much just this for the sniff, and the same again in the test, but with return false:
```php
if ($phpPath === null) {
    // PHP_BINARY is available in PHP 5.4+.
    if (defined('PHP_BINARY') === true) {
        $phpPath = PHP_BINARY;
    } else {
        return;
    }
}
```

If that works for you as well, please amend the PR and I'll merge it in.

deviantintegral:
Sounds good. I've remove the discovery code.

gsherwood:
Just one last thing I think you missed in my last comment (sorry).

I need that shouldSkipTest method in there still because if you are running the tests on <5.4 with no php_path set, it needs to skip the test.

I'm thinking this code:
```php
/**
 * Should this test be skipped for some reason.
 *
 * @return void
 */
protected function shouldSkipTest()
{
    if (defined('PHP_BINARY') === true) {
        return false;
    }

    $phpPath = PHP_CodeSniffer::getConfigData('php_path');
    return (is_null($phpPath));

}//end shouldSkipTest()
```

deviantintegral:
Whoops - thanks, and added that back in.

gsherwood:
Thanks a lot for doing that change, and for this patch in general.

PR #855, opened 2016-01-08
Fix comments capital letter detection for multibyte string.

Daazku:
Accessing string as array actually cut multibyte characters.

That description would pass the test but it shouldn't.
```
/**
 * étude des ...
 */
```

Rewrite the regexes and compare the entire string.

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/Commenting/DocCommentSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Commenting/DocCommentSniff.php
@@ -159,7 +159,7 @@ class Generic_Sniffs_Commenting_DocCommentSniff implements PHP_CodeSniffer_Sniff
             }
         }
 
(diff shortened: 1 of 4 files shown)


aik099:
Why you've closed this?

gsherwood:
> Why you've closed this?

Maybe because it breaks the unit tests.

The original issue is caused by preg_match returning false, although I don't know what the error is.

gsherwood:
> although I don't know what the error is.

Probably because it's only getting the first byte of the multibyte character rather than a regex issue.

gsherwood:
I think I've found the original problem and a possible fix. I'll just turn this into a bug report.

Daazku:
Ya sorry my fixe broke unit test I have have no time to investigate further.

I mostly changed ``preg_match('/\p{Lu}|\P{L}/u', $string[0])`` to ``preg_match('/^(\p{Lu}|\P{L})/u', $string)``

gsherwood:
@Daazku Thanks. Your PR was very helpful in finding the problem.

Issue #582, opened 2015-05-02
Squiz.Arrays.ArrayDeclaration sniff not very configurable


JDGrimes:
The `Squiz.Arrays.ArrayDeclaration` sniff is somewhat generic, but attempting to configure it to disable some errors by placing this in a ruleset will inadvertently disable other errors too:

```xml
	<rule ref="Squiz.Arrays.ArrayDeclaration">
		<exclude name="Squiz.Arrays.ArrayDeclaration.ValueNotAligned" />
		<exclude name="Squiz.Arrays.ArrayDeclaration.KeyNotAligned" />
		<exclude name="Squiz.Arrays.ArrayDeclaration.DoubleArrowNotAligned" />
		<exclude name="Squiz.Arrays.ArrayDeclaration.ValueNotAligned" />
		<exclude name="Squiz.Arrays.ArrayDeclaration.CloseBraceNotAligned" />
		<exclude name="Squiz.Arrays.ArrayDeclaration.ValueNoNewline" />
		<exclude name="Squiz.Arrays.ArrayDeclaration.MultiLineNotAllowed" />
		<exclude name="Squiz.Arrays.ArrayDeclaration.SingleLineNotAllowed" />
	</rule>
```

Basically, I'm not concerned about the alignment rules, but I'd like to still get the other errors, including the `NoComma` error when there is a missing comma, as in the following code:

```php
array(
    'foo' => 1,
    'bar' => 2 // <-- missing comma, should give error.
);
```

However, due to logic of the `processMultiLineArray()` method, this error won't be given if alignment errors are disabled. Actually, it wouldn't be given even if the alignment errors were enabled, because only one error is given per-index within the loop. After a violation is found, `continue` is called. Maybe this is intended behavior, but it makes the sniff less configurable. Could this maybe be made more generic? Maybe custom indentation schemes could be defined?

gsherwood:
It's probably best not to use that sniff unless you want everything it provides. It doesn't split all those checks into various sniffs because the lead-up work to parse the array would take too long over a large code base when run multiple times for each array.

It also stop processing at various places to ensure it doesn't generate incorrect errors.

So yes, it is a complex and non-configurable sniff because it exactly implements a single method of defining arrays for the Squiz standard. I'll leave this open as a feature request, but I don't have any plans on making it configurable. I think it would be better to write a new Generic sniffs (or sniffs) to implement checks that are more commonly used.

JDGrimes:
[...]

PR #863, opened 2016-01-15
Add T_OPEN_SHORT_ARRAY for indentation inGeneric_Sniffs_WhiteSpace_ScopeIndentSniff

gheckenbach:
This would previously report incorrect indentation in the parameters to multiline arrays defined with the short open/close tags:

```
$array = [
    'one',
    'two',
];
```

Saying indentation should be 0 instead of 4, when using `$exact = true`

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/WhiteSpace/ScopeIndentSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/WhiteSpace/ScopeIndentSniff.php
@@ -906,6 +906,34 @@ class Generic_Sniffs_WhiteSpace_ScopeIndentSniff implements PHP_CodeSniffer_Snif
                 continue;
             }//end if
 
+            // Short array openers increase the indent level.
+            if ($tokens[$i]['code'] === T_OPEN_SHORT_ARRAY) {
+                $closer = $tokens[$i]['bracket_closer'];
+                if ($tokens[$i]['line'] === $tokens[$closer]['line']) {
+                    if ($this->_debug === true) {
+                        $line = $tokens[$i]['line'];
+                        $type = $tokens[$i]['type'];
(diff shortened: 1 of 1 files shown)


aik099:
Separate sniffs exist, that deal with array indentation. Unfortunately they are not too generic to be used in other standards.

The `Generic.WhiteSpace.ScopeIndent` is not right sniff to deal with arrays, because arrays doesn't create scope (the `{` and `}`) and the scopes is what this sniff indents.

aik099:
The `$exact` parameter indeed creates a lot of problems for sniff, that isn't naturally indented and that's why it's disabled by default.

gsherwood:
This sniff specifically ignores indentation inside various structures, such as arrays and multi-line function calls, because they tend to have their own rules. I can't merge in this change because it would enforce indentation of at least 4 characters inside arrays, which will instantly conflict with the built-in Squiz sniff that has variable indentation (and any other custom sniffs that might exist).

While I strongly recommend not using the `$exact` option of this sniff (there are so many cases when additional indentation is required for readability and this sniff doesn't understand any of that) I do want to try and find a solution for you.

One thing I might be able to do is just turn off exact indent checking for all arrays, like I do for multi-line function calls. This would still see the arrays as non indenting scopes, and would require 0 spaces for indentation, but would not report any errors with your code. You would still need a sniff to actually enforce the indentation of the array, but PHPCS doesn't have anything like that yet (see #582 for ideas about new sniffs).

So short term, I can probably mute the error. But I can't fix the real problem until some more array sniffs are written that are generically useful (unlike the Squiz sniff).

gheckenbach:
Thanks for the replies. I do agree the array indentation doesn't belong in this sniff, but since it already had the scaffolding for tracking indentation, it was an easy add here.

@gsherwood I would love if you could mute the error, as you suggested. That would allow me to continue using the rest of this sniff, and I can write my own sniff for array indentation.

PR #869, opened 2016-01-19
Recognize __debugInfo as a magic method.

syranez:
PHP 5.6 added the magic method __debugInfo.

http://php.net/manual/en/language.oop5.magic.php#object.debuginfo


gsherwood:
Thanks a lot.

PR #892, opened 2016-02-13
Add Generic.Files.FileEncoding sniff to check the encoding of files

klausi:


CodeSniffer/Standards/Generic/Sniffs/Files/FileEncodingSniff.php
```diff
@@ -0,0 +1,92 @@
...
+
+
+    /**
+     * Processes this sniff, when one of its tokens is encountered.
+     *
+     * @param PHP_CodeSniffer_File $phpcsFile The file being scanned.
+     * @param int                  $stackPtr  The position of the current token in
+     *                                        the stack passed in $tokens.
+     *
+     * @return void
```
aik099: The DocBlock says that `process` method doesn't return anything, but you return a value.

Either the DocBlock is wrong or you shouldn't be returning any specific value.

CodeSniffer/Standards/Generic/Sniffs/Files/FileEncodingSniff.php
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+
+        if ($validEncodingFound === false) {
+            $warning = 'File encoding is invalid, expected %s';
+            $data    = array(implode(' or ', $this->allowedEncodings));
+            $phpcsFile->addWarning($warning, $stackPtr, 'InvalidEncoding', $data);
+        }
+
+        return $phpcsFile->numTokens;
```
aik099: What this return value tells the PHP_CodeSniffer?

CodeSniffer/Standards/Generic/Sniffs/Files/FileEncodingSniff.php
```diff
@@ -0,0 +1,92 @@
...
+     * @return void
+     */
+    public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
+    {
+        // Not all PHP installs have the multi byte extension - nothing we can do.
+        if (function_exists('mb_check_encoding') === false) {
+            return $phpcsFile->numTokens;
+        }
+
+        $fileContent = $phpcsFile->getTokensAsString(0, $phpcsFile->numTokens);
```
aik099: Isn't there any other way to get complete file contents? Building file back (considering that majority of checked files might be already in correct encoding) from tokens might not be the fast process especially for large files.

klausi:
@aik099 good feedback, improved the comments by copying them from the interface.

Not sure what we can do about the file content - I guess we would have to add a method PHP_CodeSniffer_File::getContent() that returns the file content?

CodeSniffer/Standards/Generic/Sniffs/Files/FileEncodingSniff.php
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @return void|int Optionally returns a stack pointer. The sniff will not be
+     *                  called again on the current file until the returned stack
+     *                  pointer is reached. Return (count($tokens) + 1) to skip
+     *                  the rest of the file.
+     */
+    public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
+    {
+        // Not all PHP installs have the multi byte extension - nothing we can do.
+        if (function_exists('mb_check_encoding') === false) {
```
aik099: Maybe instead we can cheat a bit and register 0 tokens in the `register` method?

This way this sniff won't even be called if it can't do anything useful.

CodeSniffer/Standards/Generic/Sniffs/Files/FileEncodingSniff.php
```diff
@@ -0,0 +1,97 @@
...
+     * Returns an array of tokens this test wants to listen for.
+     *
+     * @return array
+     */
+    public function register()
+    {
+        // Not all PHP installs have the multi byte extension - nothing we can do.
+        if (function_exists('mb_check_encoding') === false) {
+            return array();
+        }
```
pfrenssen: Since we are adding an optional dependency on the mbstring extension, it would be nice to add this as a suggestion to `composer.json`, so that users are made aware of it. For example:

```
{
    "suggest": {
        "ext-mbstring": "For checking file encodings"
    }
}
```

This will generate a message on installation:

> squizlabs/php-codesniffer suggests installing ext-mbstring (For checking file encodings)

aik099: Won't it confuse user because the sniff that needs it isn't even enabled by default in any of used standards?

photodude: @aik099 I've seen a lot of packages specify an extra subpackage requirement to use some subpart of a larger package

To that end, I would change the wording to describe when there is a need for `ext-mbstring`

```
{
    "suggest": {
        "ext-mbstring": "Required only if you want to use the  Generic.Files.FileEncoding sniff for checking file encodings"
    }
}
```

klausi:
We forked this sniff into Coder, so I'm not interested to work here any further. Feel free to pick this up and file a new pull request under your own name!

PR #894, opened 2016-02-14
Update to comments for function findPrevious()

ssontakke:
Function description was mentioned as "Returns the position of the next token" for function findPrevious()


gsherwood:
Thanks for fixing that up. It was obviously a copy/paste error from findNext that nobody has noticed.

PR #896, opened 2016-02-15
Add support for a 'rulesetdir' directive in phpcs-ruleset.xml

pfrenssen:
For projects that are using custom standards it would be nice to be able to specify the folder where a custom ruleset is located.

An example use case for this is the Drupal project. The Drupal coding standard is part of a separate library which is not included in the core project. Drupal core supplies an example [phpcs-ruleset.xml.dist](http://cgit.drupalcode.org/drupal/tree/core/phpcs.xml.dist?h=8.1.x) which does not make any assumptions about where the coding standards are installed on a developer's system. Here's an excerpt of the ruleset:

```
<ruleset name="drupal_core">
  <rule ref="Drupal">
    <exclude name="Drupal.Commenting.DocComment"/>
    <exclude name="Drupal.Commenting.FunctionComment"/>
    <exclude name="Drupal.Commenting.HookComment"/>
    ...
  </rule>
</ruleset>
```

This won't work out of the box when a developer runs a coding standards check, since the "Drupal" standard is not found:

```
PHP Fatal error:  Uncaught PHP_CodeSniffer_Exception: Referenced sniff "Drupal" does not exist in ./vendor/squizlabs/php_codesniffer/CodeSniffer.php:1132
```

So a developer needs to either symlink the Drupal coding standard into the PHP CodeSniffer standards directory, or hack the original ruleset and hardcode the path to the standard:

```
<ruleset name="drupal_core">
  <rule ref="/home/pieter/drupal/vendor/drupal/coder/coder_sniffer/Drupal">
    <exclude name="Drupal.Commenting.DocComment"/>
    <exclude name="Drupal.Commenting.FunctionComment"/>
    <exclude name="Drupal.Commenting.HookComment"/>
    ...
  </rule>
</ruleset>
```

With the `<rulesetdir>` directive that is proposed in this PR a developer can declare the location of the custom coding standard in their own ruleset, and include the original ruleset:

```
<ruleset name="drupal_core">
  <rulesetdir dir="/home/pieter/drupal/vendor/drupal/coder/coder_sniffer/Drupal" />
  <rule ref="./phpcs-ruleset.xml.dist" />
</ruleset>
```

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -704,6 +704,11 @@ class PHP_CodeSniffer
         $excludedSniffs = array();
         $cliValues      = $this->cli->getCommandLineValues();
 
+        // Populate the list of ruleset directories.
+        foreach ($ruleset->rulesetdir as $dir) {
+            self::$rulesetDirs[] = (string) $dir['dir'];
+        }
+
         $rulesetDir          = dirname($rulesetPath);
         $rulesetName         = basename($rulesetPath);
         self::$rulesetDirs[] = $rulesetDir;


aik099:
> So a developer needs to either symlink the Drupal coding standard into the PHP CodeSniffer standards directory, or hack the original ruleset and hardcode the path to the standard:

Why developer can't run `phpcs --standard=/absolute/path/to/Drupal/standard` instead?

gsherwood:
I also don't see a need for setting a ruleset dir in a ruleset file. If you have tried some of these other ways and they are not working, please let me know. PHPCS should be smart enough to set the ruleset dirs for you if you are including standards using absolute paths.

If you want to use relative paths, just tell PHPCS where to find your standards:
```
$ phpcs --config-set installed_paths /home/pieter/drupal/vendor/drupal/coder/coder_sniffer/Drupal
$ phpcs --standard=Drupal /path/to/code
```

pfrenssen:
>> So a developer needs to either symlink the Drupal coding standard into the PHP CodeSniffer standards directory, or hack the original ruleset and hardcode the path to the standard:

> Why developer can't run phpcs --standard=/absolute/path/to/Drupal/standard instead?

This will run the a scan with the full coding standard, but that's not what we currently need. I've linked the ruleset that we're using, it is highly customized. We are at the moment only using a handful of sniffs from that standard. There's a reason for this - we recently started an initiative to become fully compliant with our own coding standards. We currently have tens of thousands of violations across thousands of files. We are burning down the violations sniff per sniff, and are maintaining a list of sniffs that have been fixed, so we won't get any regressions.

I'm trying to find a solution so that we can easily let our 3000 contributors scan their code with this custom project-supplied ruleset without having to edit the file, or copy it. People should be free to pull the latest changes from the repo without having to fix conflicts in `phpcs-ruleset.xml.dist`, or having to update their copy so it matches the latest changes in the project-supplied one.

A good solution would be to let the developers save a simple `phpcs-ruleset.xml` that imports `phpcs-ruleset.xml.dist` and overrides a handful of properties while inheriting all the rest.

> &lt;rule ref="home/pieter/drupal/vendor/drupal/coder/coder_sniffer/Drupal" />

This is the example I showed in my second snippet, this requires to either alter the project-supplied ruleset or copy it in its entirety. When I try my approach with a develop-supplied ruleset that includes the project-supplied one it doesn't work, it will include the complete standard, and ignore all the excludes in `phpcs-ruleset.xml.dist`.

```
<ruleset name="drupal_core">
  <rule ref="/home/pieter/drupal/vendor/drupal/coder/coder_sniffer/Drupal" />
  <rule ref="./phpcs-ruleset.xml.dist" />
</ruleset>
```

> I also don't see a need for setting a ruleset dir in a ruleset file. If you have tried some of these other ways and they are not working, please let me know. PHPCS should be smart enough to set the ruleset dirs for you if you are including standards using absolute paths.

Yes I saw that an array of ruleset dirs is populated from rule directives when a ruleset is parsed, this PR exposes this array, so we can add values to it to cover cases where paths cannot be hard coded, like when including a `phpcs-ruleset.xml.dist` file.

> $ phpcs --config-set installed_paths /home/pieter/drupal/vendor/drupal/coder/coder_sniffer/Drupal
> $ phpcs --standard=Drupal /path/to/code

This also does a scan with the full coding standard, triggering the avalanche of tens of thousands of violations.

If we can expose the ruleset dirs using a command line option then this would also be a good solution. It doesn't seem as easy though, since the config options are a key-value combination it would be harder to provide multiple directories.

gsherwood:
> If we can expose the ruleset dirs using a command line option then this would also be a good solution.

You can set config values for a single run by using --runtime-set:
`phpcs --runtime-set installed_paths /path/to/somewhere /path/to/code`

And you can set these in your custom ruleset if you want:
`<config name="installed_paths" value=" /path/to/somewhere"/>`

Does that help?

aik099:
Is Drupal coding standard available through Composer? If yes, then you can take advantage of fact, that recently PHP_CodeSniffer started to include Composer's auto loader.

This way Drupal coding standard would be a dev dependency of a project.

pfrenssen:
@aik099 thanks for the suggestion, this would indeed solve it nicely. I will open a ticket in the Drupal issue tracker to discuss this.

aik099:
Long time ago (when there was PHP_CodeSniffer 1.5 only) I had to copy-paste sniffs from Drupal standard into my own to make it work. This made my standard portable.

klausi:
Yes, we recommend that users set the config for PHPCS when they use the Drupal standard https://www.drupal.org/node/1419988 , example: phpcs --config-set installed_paths ~/.composer/vendor/drupal/coder/coder_sniffer

pfrenssen:
It doesn't work with config-set either. Remember that we use a phpcs.xml.dist file that contains a custom ruleset.

```
$ ../vendor/bin/phpcs --runtime-set installed_paths ../vendor/drupal/coder/coder_sniffer/
PHP Fatal error:  Uncaught PHP_CodeSniffer_Exception: Referenced sniff "Drupal" does not exist in /home/pieter/v/drupal/drupal/vendor/squizlabs/php_codesniffer/CodeSniffer.php:1121
```

```
$ ../vendor/bin/phpcs --runtime-set installed_paths ../vendor/drupal/coder/coder_sniffer/Drupal/
PHP Fatal error:  Uncaught PHP_CodeSniffer_Exception: Referenced sniff "Drupal" does not exist in /home/pieter/v/drupal/drupal/vendor/squizlabs/php_codesniffer/CodeSniffer.php:1121
```

Or when I try to add the `installed_paths` to my custom `phpcs.xml` that includes `phpcs.xml.dist`:

```
<ruleset name="drupal_core">
  <config name="installed_paths" value="../vendor/drupal/coder/coder_sniffer/Drupal/" />
  <rule ref="./phpcs.xml.dist" />
</ruleset>
```

```
$ ../vendor/bin/phpcs 
PHP Fatal error:  Uncaught PHP_CodeSniffer_Exception: Referenced sniff "Drupal" does not exist in /home/pieter/v/drupal/drupal/vendor/squizlabs/php_codesniffer/CodeSniffer.php:1121
```

pfrenssen:
I'm getting somewhere, I need to specify the `installed_paths` relative to the directory where PHP CodeSniffer is installed, not relative to the current directory. This is working:

```
$ ../vendor/bin/phpcs --runtime-set installed_paths ../../drupal/coder/coder_sniffer/ -p
```

But I cannot specify the installed_paths by putting `<config name="installed_paths" value="...">` in my ruleset, that seems to be a bug.

pfrenssen:
The bug is that the config is only parsed _after_ the ruleset has been parsed. But since the ruleset depends on config for the `installed_paths` the config should be processed first.

I have a fix, will push it in an hour or so after lunch

pfrenssen:
I have provided a fix that allows to use the `installed_paths` configuration option inside a custom ruleset in PR #899.

This means that now my problem is fully solved; by supplying the `installed_path` configuration option our contributors can check coding standards without having to hack or copy `phpcs.xml.dist`.

Thanks everybody for the help, much appreciated!

PR #906, opened 2016-03-02
Require SimpleXML

DavidPrevot:
Since the code uses functions from the SimpleXML module
(simplexml_load_string), it should require it. Some distributions split
SimpleXML in a separate package (as Debian recently did), so being
explicit about used modules is helpful.

Bug-Debian: https://bugs.debian.org/816248


gsherwood:
Makes sense. Thanks.

PR #912, opened 2016-03-05
Document needed extensions in package.xml

DavidPrevot:
Mimic existing documentation from composer.json.

Bug-Debian: https://bugs.debian.org/816248


gsherwood:
Thanks

PR #913, opened 2016-03-08
Added Chamilo LMS

ywarnier:
Not sure what the inclusion process for new PHP codebases is (haven't found any), but if you'd like to consider Chamilo, we would love to be included.


gsherwood:
Sorry, added this manually instead of merging.

Reports have been generated and are available here: http://squizlabs.github.io/PHP_CodeSniffer/analysis/chamilo/chamilo-lms/index.html

Thanks for submitting.

Issue #925, opened 2016-03-16
Adding CLI parameter for editor path / command to be able to open invalid file during interactive session


Astinus-Eberhard:
Like --editor-path=<path> or --editor-path="<full cmd to be prefixed by invalid file path>"

And during interactive session:
```
FILE: ....php
----------------------------------------------------------------------
FOUND X ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 ...
----------------------------------------------------------------------

<ENTER> to recheck, [s] to skip, [o] to open in editor or [q] to quit :
```

gsherwood:
See PR for discussion.

PR #926, opened 2016-03-16
Adding CLI parameter to open files on desired editor 

Astinus-Eberhard:
Solution for #925 

One thing: I was unable to run phpunit tests during Composer and Codesinffer autoloader interaction - given the error:
```
PHP Fatal error:  Cannot redeclare class Symfony2_Tests_Formatting_BlankLineBeforeReturnUnitTest in CodeSniffer/Standards/Symfony2DomenyPL/Tests/Formatting/BlankLineBeforeReturnUnitTest.php on line 59
PHP Stack trace:
PHP   1. {main}() vendor/phpunit/phpunit/phpunit:0
PHP   2. PHPUnit_TextUI_Command::main($exit = *uninitialized*) vendor/phpunit/phpunit/phpunit:47
PHP   3. PHPUnit_TextUI_Command->run($argv = *uninitialized*, $exit = *uninitialized*) vendor/phpunit/phpunit/src/TextUI/Command.php:100
PHP   4. PHPUnit_Runner_BaseTestRunner->getTest($suiteClassName = *uninitialized*, $suiteClassFile = *uninitialized*, $suffixes = *uninitialized*) vendor/phpunit/phpunit/src/TextUI/Command.php:123
PHP   5. ReflectionMethod->invoke(*uninitialized*, *uninitialized*) vendor/phpunit/phpunit/src/Runner/BaseTestRunner.php:86
PHP   6. PHP_CodeSniffer_AllTests::suite(*uninitialized*) vendor/phpunit/phpunit/src/Runner/BaseTestRunner.php:86
PHP   7. PHP_CodeSniffer_Standards_AllSniffs::suite() tests/AllTests.php:77
PHP   8. PHPUnit_Framework_TestSuite->addTestSuite($testClass = *uninitialized*) tests/Standards/AllSniffs.php:118
PHP   9. class_exists(*uninitialized*) vendor/phpunit/phpunit/src/Framework/TestSuite.php:264
PHP  10. spl_autoload_call(*uninitialized*) vendor/phpunit/phpunit/src/Framework/TestSuite.php:264
PHP  11. PHP_CodeSniffer::autoload($className = *uninitialized*) vendor/phpunit/phpunit/src/Framework/TestSuite.php:264
```
I believe, that my changes are located in those files not covered by tests and this is tested and used by me in everyday job tasks.

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -269,10 +269,6 @@ class PHP_CodeSniffer
      */
     public static function autoload($className)
     {
-        if (class_exists($className, false) === true) {
-            return;
-        }
-        
         if (substr($className, 0, 4) === 'PHP_') {
             $newClassName = substr($className, 4);
         } else {
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -269,6 +269,10 @@ class PHP_CodeSniffer
      */
     public static function autoload($className)
     {
+        if (class_exists($className, false) === true) {
+            return;
+        }
+        
         if (substr($className, 0, 4) === 'PHP_') {
             $newClassName = substr($className, 4);
         } else {
@@ -1796,11 +1800,12 @@ class PHP_CodeSniffer
             $reportData  = $this->reporting->prepareFileReport($phpcsFile);
             $reportClass->generateFileReport($reportData, $phpcsFile, $cliValues['showSources'], $cliValues['reportWidth']);
 
-            if (empty(PHP_CODESNIFFER_EDITOR_PATH)) {
+            if (empty(PHP_CODESNIFFER_EDITOR_PATH) === true) {
                 echo '<ENTER> to recheck, [s] to skip or [q] to quit : ';
             } else {
                 echo '<ENTER> to recheck, [s] to skip, [o] to open in editor or [q] to quit : ';
             }
+
             $input = fgets(STDIN);
             $input = trim($input);
 
@@ -1811,7 +1816,7 @@ class PHP_CodeSniffer
                 exit(0);
                 break;
             case 'o':
-                if (false == empty(PHP_CODESNIFFER_EDITOR_PATH)) {
+                if (false === empty(PHP_CODESNIFFER_EDITOR_PATH)) {
                     exec(PHP_CODESNIFFER_EDITOR_PATH.' '.$file);
                 }
                 break;
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -135,6 +135,13 @@ class PHP_CodeSniffer
      * @var array
      */
     private $_tokenListeners = array();
+    
+    /**
+     * The editor path / command
+     *
+     * @var string|null
+     */
+    protected $editorPath = null;
 
     /**
      * An array of rules from the ruleset.xml file.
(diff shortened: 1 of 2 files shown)


aik099:
Please rebase. Probably you've created PR based not off the `master` branch.

Astinus-Eberhard:
rebased (sorry ;) )

gsherwood:
I've had a long think about this and decided that I don't think this is the sort of functionality that PHPCS should be providing. Especially given it can't be used anywhere within PHPCS unless you happen to be using interactive mode, which isn't a widely used feature. It feels like adding and maintaining a feature that only a tiny fraction of people would ever use, which I don't like the idea of doing because these have a tendency to be forgotten and break.

I think it's easy enough to just copy the file path and open the file in your editor manually, or via the CLI if you really want to. So it's not like there aren't already other ways to perform the task.

If the interactive runner of PHPCS was a common way of using PHPCS then it would probably make sense to add this (and other) PHPCS options that are only used with interactive mode. But that's just not currently the case.

If you can think of other good uses for an editor path elsewhere in PHPCS, please let me know and we can continue the discussion. Otherwise, thanks for taking the time to put the PR together even though our views differ.

Astinus-Eberhard:
I can agree, that this option will not be commonly used, but still can be very useful in some cases. For example, I am participating some non-trivial projects that were founded over decade ago. There is no build system, and any QA we are doing by hand. Checking & fixing any coding standard violations during interactive session in this case is for now only option, but file paths are too long to fully prints to the screen. I am estimating that I saves 10-30% time spent on QA just because not looking for invalid files and eliminating unnecessary mouse-keyboard actions (especially when a lot of files contains single errors).

There are very small changes introduced by my PR. Beside of that, PHPCS already has similar fancy feature (like desktop notifications, which highly depends on system environment). Please at least check diff out before you through it away. If you are not planning to remove interactive mode nor make it deprecated in some way, I think PHPCS should have this, especially when file paths are too long and are truncated - so you cannot just copy it nor locate easily.

Of course, I do not want to forcing anyone to use on even deploy this. I found this useful, want to share, and thank for appreciate ;)

gsherwood:
> Please at least check diff out before you through it away

Don't worry. I checked the diff and know what the change is. I would have to completely redo the config part in the 3.0 version because it has been completely refactored, but even that's not too hard.

It's not the size of the change that is causing me to question it. It's the fact that once included, I need to keep it supported and documented, and I need to have a good reason why it exists.

Your specific case is an interesting one, but you'd probably get the job done faster if you use a text editor and one of the PHPCS plugins for it. They can scan an entire directory and take you to the results very easily, allowing you to see the errors inside the text editor, fix them, and verify the change. I use Atom with a plugin, but I've also used Sublime in the past and I know PhpStorm is very popular as well. All have plugins for to run PHPCS.

If you have good unit tests, you might also consider changing `phpcs` to `phpcbf` in your command and letting PHPCBF auto-fix everything it can on a first pass. That typically cuts down the number of errors by about 80% but has the disadvantage of changing your code without you actually making the change. But on a massive code base, it comes in very handy.

When I changed PHPCS' own coding standard with 3.0, I used a combination of both. I ran PHPCBF first and then used by normal editor to go through and find/fix the ones I needed to do by hand (mostly docblocks).

Any of those options sound helpful?

Astinus-Eberhard:
Sounds good. I definitely must check it. Nevertheless, for now I am stuck with code that is too big to be refactored at once. We are successively fix files during updates, and we agreed to check only those files that was changed. Any IDE I have been tested before which I know it have plugin for PHPCS can check specified files or whole project structure, but if I have more than 100 files changed strewn around and mixed with other files not changed nor fixed for coding standard in the same places (same directory, very similar filenames, etc.) it is just annoying.

gsherwood:
I'm closing this PR so there isn't confusion about if this will be included or not.

Of course, I change my mind all the time. If anyone else comes across this and feels strongly about it, please comment here.

PR #927, opened 2016-03-17
Adding CLI parameter to open files on desired editor

Astinus-Eberhard:
Solution for #925 for 3.0 branch

Just in case anyone need it.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -428,6 +428,7 @@ class Config
         $this->suffix          = '';
         $this->stdin           = false;
         $this->stdinContent    = null;
+        $this->editorPath      = null;
 
         $standard = self::getConfigData('default_standard');
         if ($standard !== null) {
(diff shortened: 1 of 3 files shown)


src/Config.php
```diff
@@ -482,6 +484,11 @@ public function restoreDefaults()
             $this->showProgress = (bool) $showProgress;
         }
 
+        $editorPath = self::getConfigData('editor_path');
+        if ($editorPath !== null) {
+            $this->editoPath = $editorPath;
```
gsherwood: There is a typo here: `$this->editoPath` should obviously be `$this->editorPath`

gsherwood:
Discussion is on PR #926.

There are a couple of minor changes on here as well that seem to be unrelated. If you think they are required, please submit them separately so we can discuss them.

src/Config.php
Astinus-Eberhard: Fixed ;)

PR #936, opened 2016-03-26
Add Composer standards installer plugin.

schlessera:
This turns the Composer package into a plugin that comes with a custom installer that can install packages of type `phpcs-standard`.

A package that is of type `phpcs-standard` will need to have an `extra` key `standards` that contains an array of entries in the following format:

``` JSON
    "extra"      : {
        "standards": {
            "<standard name 1>" : "<standard path 1>",
            "<standard name 2>" : "<standard path 2>"
        }
    }
```

The `<standard path>` is relative to the package root of the standard package.

For each of the standards defined in this way, the installer will create a symlink in `CodeSniffer/Standards/<standard name>` that points back to the corresponding standard.

### DIFF
--- a/CodeSniffer/ComposerInstaller.php
+++ b/CodeSniffer/ComposerInstaller.php
@@ -1,0 +1,144 @@
+<?php
+/**
+ * Composer installer class to handle installation of additional standards.
+ *
+ * PHP version 5
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @copyright 2006-2016 Squiz Pty Ltd (ABN 77 084 670 600)
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ * @link      http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+use Composer\Installer\LibraryInstaller;
+use Composer\Package\PackageInterface;
+use Composer\Repository\InstalledRepositoryInterface;
+
+/**
+ * Composer installer class to handle installation of additional standards.
+ *
+ * This makes sure that Composer packages of type
+ * `phpcs-standard` are installed in the correct subfolder.
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @author    Alain Schlesser <alain.schlesser@gmail.com>
+ * @copyright 2006-2016 Squiz Pty Ltd (ABN 77 084 670 600)
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ * @version   Release: @package_version@
+ * @link      http://pear.php.net/package/PHP_CodeSniffer
+ */
+class PHP_CodeSniffer_ComposerInstaller extends LibraryInstaller
+{
+
+    const EXTRA_KEY      = 'standards';
+    const STANDARDS_PATH = '../php_codesniffer/CodeSniffer/Standards/';
+    const TYPE           = 'phpcs-standard';
+
+
+    /**
+     * Install the package.
+     *
+     * @param InstalledRepositoryInterface $repo    The repository from where the package was fetched.
+     * @param PackageInterface             $package The package to install.
+     *
+     * @return void
+     */
+    public function install(InstalledRepositoryInterface $repo, PackageInterface $package)
+    {
+        $installPath = $this->getInstallPath($package);
+
+        if ($this->io->isVerbose() === true) {
+            $this->io->write(
+                sprintf(
+                    _('Symlinking PHPCS standards package %1$s'),
+                    $installPath
+                ),
+                true
+            );
+        }
+
+        parent::install($repo, $package);
+
+        foreach ($this->getStandards($package) as $standardName => $standardPath) {
+            $this->linkStandard(
+                $standardName,
+                $installPath.DIRECTORY_SEPARATOR.$standardPath
+            );
+        }
+
+    }//end install()
+
+
+    /**
+     * Whether the installer supports a given package type.
+     *
+     * @param string $packageType Type of the package to check support for.
+     *
+     * @return bool Whether the package is supported.
+     */
+    public function supports($packageType)
+    {
+        return self::TYPE === $packageType;
+
+    }//end supports()
+
+
+    /**
+     * Get the list of standards to add.
+     *
+     * @param PackageInterface $package The package for which to retrieve the standards.
+     *
+     * @return array List of standards.
+     */
+    protected function getStandards(PackageInterface $package)
+    {
+        $extraData = $package->getExtra();
+
+        // Bail early if no "extra" key found.
+        if (empty($extraData) === true) {
+            return array();
+        }
+
+        // Bail early if "extra" key does not contain "standards" key.
+        if (array_key_exists(self::EXTRA_KEY, $extraData) === false) {
+            return array();
+        }
+
+        return (array) $extraData[self::EXTRA_KEY];
+
+    }//end getStandards()
+
+
+    /**
+     * Link a standard to the PHPCS installation.
+     *
+     * @param string $name Name of the standard.
+     * @param string $path Relative path of the standard.
+     *
+     * @return void
+     */
+    protected function linkStandard($name, $path)
+    {
+        if ($this->io->isVeryVerbose() === true) {
+            $this->io->write(
+                sprintf(
+                    _('Symlinking standard "%1$s" to path "%2$s"'),
+                    $name,
+                    $path
+                ),
+                true
+            );
+        }
+
+        $linkPath = getcwd().DIRECTORY_SEPARATOR.self::STANDARDS_PATH.$name;
+
+        $this->filesystem->relativeSymlink($path, $linkPath);
+
+    }//end linkStandard()
+
+
+}//end class
--- a/CodeSniffer/ComposerInstaller.php
+++ b/CodeSniffer/ComposerInstaller.php
@@ -34,7 +34,7 @@ use Composer\Repository\InstalledRepositoryInterface;
 class PHP_CodeSniffer_ComposerInstaller extends LibraryInstaller
 {
 
-    const EXTRA_KEY      = 'standards';
+    const EXTRA_KEY      = 'phpcs-standards';
     const STANDARDS_PATH = '../php_codesniffer/CodeSniffer/Standards/';
     const TYPE           = 'phpcs-standard';
 
--- a/CodeSniffer/ComposerPlugin.php
+++ b/CodeSniffer/ComposerPlugin.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Composer plugin class to register the Composer standards installer.
+ *
+ * PHP version 5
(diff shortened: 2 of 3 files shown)


aik099:
Why?

schlessera:
As cam be seen in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/163 and https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/195, additional standards need a way to enable installation through Composer.

As it stands, the only way to handle this is to add post-install scripts to the additional standards, and this only works when directly installing that package due to a limitation in Composer.

If you want to use such a standard as a (dev)-dependency of another package, the scripts are not run, and you have no way of properly attaching the additional standards to the PHPCS package.

This PR adds an installer because it is the only way to allow such a functionality at a nested level.

Please refer to https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/546 to see how this implementation would look like at the standards package level.

aik099:
> As it stands, the only way to handle this is to add post-install scripts to the additional standards, and this only works when directly installing that package due to a limitation in Composer.

It's not the only way. For example, when standard installed via Composer it can be used like so:

```
phpcs --standard=vendor/standard-vendor/standard-name/StandardName
```

In rare cases (and I guess this is the case you're covering with this PR), when you have bunch of custom made standards one each of them depends on another one this won't work. This however might be solved in PHP_CodeSniffer 3.0 version where namespaced based autoloader is being used.

Unfortunately for all of that to work you need to install PHP_CodeSniffer and needed standards together. For example right now I have PHP_CodeSniffer installed globally via Composer and standard installed only in projects that need it. Then this PR won't help much.

schlessera:
> `phpcs --standard=vendor/standard-vendor/standard-name/StandardName`

This only works if `phpcs` is installed globally and if the Composer packages are installed in the default location.

> In rare cases (and I guess this is the case you're covering with this PR), when you have bunch of custom made standards one each of them depends on another one this won't work. This however might be solved in PHP_CodeSniffer 3.0 version where namespaced based autoloader is being used.

No, the simple fact of having PHPCS as a dependency (instead of installing it globally) makes it very difficult to automate anything as regards to additional standards, without hacking paths in that you should not know about (and which might change depending on the environment).

> Unfortunately for all of that to work you need to install PHP_CodeSniffer and needed standards together.

No, as I see it, you would only need to add the standard you require as a dev-dependency. This should then correctly depend on PHPCS and pull it in.

> For example right now I have PHP_CodeSniffer installed globally via Composer and standard installed only in projects that need it. Then this PR won't help much.

No package should force you to have globally installed stuff only to make the dev-dependencies work. If the package *depends* on something, this should be a * dependency* that is pulled in through Composer.

> Then this PR won't help much.

Correct, in your use case, it won't help much, but it will not hurt either. It will enable other use cases, though.

My specific use case is to have a similar package to [this automated PHPCS precommit check](https://github.com/php-composter/php-composter-phpcs-psr2), but for the [WordPress Standard](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards). So, the package should just pull in the WordPress Standard, and this should in turn depend on PHPCS. With the above PR, the WordPress Standard gets symlinked into the PHPCS folder, so that it *just works*, without needing you to assume anything about the specific Composer settings.

schlessera:
Also, I want to point to the fact that the above PR makes it possible to wire everything together without depending on `vendor` being the specific folder where packages are installed.

gsherwood:
This is very interesting, but will take me a significant amount of time to review given (mostly) my current workload but also my unfamiliarity with composer plugins and the implications of this change. Just a warning - nothing against the code.

schlessera:
@gsherwood Yes, sure.

Here's a basic overview of how Composer plugins work:
1. When dependencies are resolved, Composer will notice that there is a package of type `composer-plugin`: https://github.com/schlessera/PHP_CodeSniffer/blob/schlessera-standards-installer/composer.json#L4
2. For each plugin, it will look for a `"class"` key in the `"extra"` data within `composer.json` to know what class takes care of plugin functionality. Here: https://github.com/schlessera/PHP_CodeSniffer/blob/schlessera-standards-installer/composer.json#L26
3. It will call the `activate()` method of each plugin's class: https://github.com/schlessera/PHP_CodeSniffer/blob/schlessera-standards-installer/CodeSniffer/ComposerPlugin.php#L46-L51
In this case, all the plugin does is let Composer know that we have a custom installer.
4. The custom installer we've told Composer about will be added to the `InstallationManager`.
5. When installing each package, all of the installers are asked whether they `support()` the type of the current package. Our custom installer returns true for the `phpcs-standard` type: https://github.com/schlessera/PHP_CodeSniffer/blob/schlessera-standards-installer/CodeSniffer/ComposerInstaller.php#L83-L87
6. If the installer accepts the type of the package, it can override several methods of dealing with the package installation. Here, we provide a custom `install()` method that first executes the normal installation procedure (downloading and copying into `vendor/<vendor>/<package>`) and then symlinks into the PHPCS `Standards` folder. To know what to symlink, the standard needs to provide information about this in its `"extra"` key, as can be seen here for the WPCS standards: https://github.com/schlessera/WordPress-Coding-Standards/blob/schlessera-standards-installer/composer.json#L27-L33

This works at any nested dependency level (pre/post-install scripts only work for the root package), and can be overridden with to `--no-plugins` option if necessary.

Hope this helps to understand the basic process. Let me know if you need anything else or if you have any other questions, I'm more than happy to help!

schlessera:
After spending some time sorting through the hassle of getting a custom standard working again, I was reminded about this pull request. Any news on this? Anything I can do to help?

aik099:
Now I get how cool this feature actually is. The only problem is that every standard that you was to connect needs to:

* have type set to `phpcs-standard`
* have `extra` key explaining where actual standards are located within package

This is a problem, because there are lots of standards out there on Packagist website and there is no way of knowing if they can/would change their `composer.json` to work.

Maybe we can somehow invert configuration flow and in `composer.json`, where `squizlabs/php_codesniffer` is specified as a dependency in the `extra` key specify mapping of packages, that needs to be symlinked. Not sure if that is possible in Composer plugin because kind of need to first specify mapping and then install package that would use it.

schlessera:
> This is a problem, because there are lots of standards out there on Packagist website and there is no way of knowing if they can/would change their composer.json to work.

Well, worst case is that those that don't intend to change their `composer.json` are stuck with the current behavior, so I don't see how that could be a problem. It's an improvement for those who participate, and it doesn't change a thing for those who don't.

I think the current flow makes more sense. For standards who support it, you just let Composer know about the standard you intend to use. It will pull in the standard, PHPCS itself, and any other dependency that specific standard might need.

schlessera:
BTW, once PHPCS provides a standard, documented way of hooking up additional standards, why shouldn't these standards make the one or two additional changes to their `composer.json` file?

aik099:
> Well, worst case is that those that don't intend to change their composer.json are stuck with the current behavior, so I don't see how that could be a problem.

If standard developer isn't even aware of need of that change, then whoever wants to use his standard needs to ping him to change his `composer.json`. This migration period can be long.

> It's an improvement for those who participate, and it doesn't change a thing for those who don't.

😄 

> I think the current flow makes more sense.

It makes perfect sense if majority of standard providers will adopt it. Probably after merged we need to contact them to adopt as quickly as possible.

> BTW, once PHPCS provides a standard, documented way of hooking up additional standards, why shouldn't these standards make the one or two additional changes to their `composer.json` file?

Due lack of time probably. But if that's not an issue, then I'm perfectly happy with this PR.

gsherwood:
> Any news on this?

I like the idea, but I need to find time to review and understand this properly.

Potherca:
Would this be better suited as an "official" installer in [the Composer Installers repository](https://github.com/composer/installers/)?

There is also [a discussion about a PHPCS Composer Installer in the PHPCompatibility repository](https://github.com/wimg/PHPCompatibility/issues/102).

An implementation has also been created by @frenck: https://github.com/DealerDirect/phpcodesniffer-composer-installer

frenck:
The problem with the code in this PR is within the way composer works.
For example:
you work with stable packages and PHP_CodeSniffer get a new stable. Upon update the new stable ZIP is downloaded en extracted, replacing the full previous install, then the Symlinks would be lost.

This is why I'm against modifying the actual data inside the package in the vendor directory. Symlinks, copy and/or installation directory modifications within the 'vendor' directory is just causing problems on the long run.

The installer I wrote (linked by @Potherca in the comment above), just handles setting up PHP_CodeSniffer using the documented way. It simply tell PHP_CodeSniffer to look into additional directories for coding standards.

schlessera:
@frenck The problem with the symlink is easily fixed by letting Composer know that the symlink should always be done, not only on installation.

But, I honestly don't care what specific method is used to install the standards, as long as there is one method that just works without messing around with paths.

I opted for the current approach so that PHPCS itself is the installer for PHPCS standards. This way, when a PHPCS standard adds PHPCS as its requirements, the installation just works automatically.

schlessera:
@Potherca I think this would not work with the Composer Installers repository, as this is just a "type => fixed path" mapping. There would still be no way for letting PHPCS know about installed standards.

frenck:
@schlessera The early version of my plugin, mentioned by @Potherca indeed did, this was wrong and has been adjusted.

https://github.com/DealerDirect/phpcodesniffer-composer-installer

Actually works by configuring php_codesniffer, without symlinking, without mapping.

Wirone:
I've tested @frenck's installer and it works fine, but I think it would be better if `squizlabs/PHP_CodeSniffer` offered this feature itself, not through external packages.

gsherwood:
There is a bit of debate on this PR about the best way to do things, and I'm not really sure which way to go. But I am getting ready for the 3.0 release, so I don't want to officially support any type of composer installer in 2.x.

To be honest, I'm not sure I want to support a composer plugin at all as I wont use it, and I don't like taking responsibility for features that I know I wont actively maintain.

jrfnl:
@schlessera Composer isn't the only dependency manager out there. As your arguments make *very* clear, this is a Composer problem, not so much a PHPCS problem, so I for one support @Potherca's point of view in this.

schlessera:
I understand that PHPCS prefers not to add something like this. But I am curious to know what other dependency manager you use for PHP code.

Wirone:
@schlessera PECL obviously ;-)

I, personally, think that this feature suits best to this repo. If someone installs CodeSniffer locally/globally also should can install additional standards with minimal overhead (@frenck's installer is great, but it's another meta-package that must be installed in order to handle custom standards).

Is it too much to maintain even if not used by author? Don't think so.

Would be great to see it implemented.

GaryJones:
> If PHPCS needs to sort out dependencies, it should do this through Composer.
> Sorting PHPCS dependencies out through Composer does not work out of the box, because just having both PHPCS and an extra standard as dependencies does not make that standard available to PHPCS.
> To make this work as expected, the relationship between PHPCS and external standards needs to be explicitly stated, either in a separate, third package, or within the "framework" PHPCS itself.

https://github.com/DealerDirect/phpcodesniffer-composer-installer seems to do a pretty good job. Add it along with your Standard and PHPCS dependencies, and it "just works", so that the Standard is available to use with PHPCS immediately.

gsherwood:
Closing as there are other tools to do this. The general issue of installing standards via composer may be addressed in the future, but not in version 3, and probably only after a full refactor and rethink about how standards work.

PR #937, opened 2016-03-27
Shebang can cause PSR1 SideEffects warning

claylo:
I found another spot where the [issue 20481 reported through PEAR](https://pear.php.net/bugs/bug.php?id=20481) occurred during a PSR2 sniff. A PHP script beginning with shebang but containing only comments does not trigger it, but a script defining symbols will trigger the `PSR1.Files.SideEffects.FoundWithSymbols` warning.

```
A file should declare new symbols (classes, functions, constants, etc.) and cause no other side effects, or it should execute logic with side effects, but should not do both. The first symbol is defined on line 5 and the first side effect is on line 1. (PSR1.Files.SideEffects.FoundWithSymbols)
```

This patch fixes that issue in a similar way to the approach found in the commit addressing issue 20481.


gsherwood:
Thanks a lot for the fix and the test.

PR #941, opened 2016-03-31
Remove unused vars

vkhramtsov:
Remove unused vars from code

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -715,7 +715,6 @@ class PHP_CodeSniffer
         $cliValues      = $this->cli->getCommandLineValues();
 
         $rulesetDir          = dirname($rulesetPath);
-        $rulesetName         = basename($rulesetPath);
         self::$rulesetDirs[] = $rulesetDir;
 
         if (is_dir($rulesetDir.DIRECTORY_SEPARATOR.'Sniffs') === true) {
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -1235,8 +1235,6 @@ class PHP_CodeSniffer
                     $value  = (string) $prop['value'];
                     $values = array();
                     foreach (explode(',', $value) as $val) {
-                        $v = '';
-
                         list($k,$v) = explode('=>', $val.'=>');
                         if ($v !== '') {
                             $values[$k] = $v;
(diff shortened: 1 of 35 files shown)


aik099:
How you've detected that code is unused? I'm asking because https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Generic/Sniffs/PHP/NoSilencedErrorsSniff.php#L67 page also qualifies (the variable is replaced immediately after declaration), but wasn't included in PR.

vkhramtsov:
I'm using phpmd (https://packagist.org/packages/phpmd/phpmd) for checking code. I can also check for cases when variable replaced immediately after definition if it required in this PR

aik099:
> I can also check for cases when variable replaced immediately after definition if it required in this PR

Yes, please do so.

vkhramtsov:
Done

gsherwood:
I'm not going to merge this in. A quick glance at just the first file shows a removal of `$firstContent`, which which is used just two lines down. And I see that the call to `getDefaults()` is blindly removed from CLI.php just because the return value is not used. But this would stop the CLI from actually working.

I don't have the time to go through each and every one of these places right now and they aren't going to be causing any problems, so i'm going to close this.

If anyone else has time to go through them and submit a PR with just the obvious ones, I'd be happy to review that. Thanks for submitting this, but even a single false positive is going to cause real issues.

vkhramtsov:
`A quick glance at just the first file shows a removal of $firstContent, which which is used just two lines down.`
As I see two lines down, you overwrite `$firstContent` without taking care about it's value. So this definition is obsolete.
Code passed all tests after modification. Do you keep all, even unused code after introducing?

Thank you.

gsherwood:
> As I see two lines down, you overwrite $firstContent without taking care about it's value. So this definition is obsolete.

Yeah sorry, I stuffed that up. There are other examples in there that will break though. I'm never going to merge in a changeset like this without reviewing every single change one by one.

vkhramtsov:
No problem, I will wait for your feedback. Actually I need to know your position on this change set, so after review I can change it, return some removed code and add some comments for people who will look into the code later.
Thank you.

PR #946, opened 2016-04-06
Throw tag sniff supporting re-thrown exceptions

samlev:
This update lets PHPCS detect when an exception is caught and re-thrown as a variable. It requires the re-thrown exception to be explicitly listed in the function comment, and avoids problems where re-thrown exceptions are listed correctly, but show PHPCS errors as excessive `@throws` tags.

For example, this would now be valid:

``` php
/**
 * Calls a function that may raise an exception, or may return false. In either instance, raise an exception
 *
 * @throws NewException When a brand new exception is thrown.
 * @throws OldException When an exception is caught and re-thrown.
 */
function thisThrowsExceptions() {
    try {
        $val = mightThrowExceptions();
    } catch (OldException $e) {
        throw $e;
    }

    if ($val === false) {
        throw new NewException('$val is false');
    }

    // The rest of the code.
}
```

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
@@ -97,8 +97,15 @@ class Squiz_Sniffs_Commenting_FunctionCommentThrowTagSniff extends PHP_CodeSniff
             while ($currPos < $currScopeEnd && $currPos !== false) {
                 /*
                     If we can't find a NEW, we are probably throwing
-                    a variable, so we ignore it, but they still need to
-                    provide at least one @throws tag, even through we
+                    a variable.
+
+                    If we're throwing the same variable as the exception container
+                    from the nearest 'catch' block, we take that exception, as it is
+                    likely to be a re-throw.
+
+                    If we can't find a matching catch block, or the variable name
+                    is different, it's probably a different variable, so we ignore it,
+                    but they still need to provide at least one @throws tag, even through we
                     don't know the exception class.
                 */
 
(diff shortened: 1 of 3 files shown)


aik099:
What is rethrown exception is:

* rethrown in the closure defined in a that method
* rethrown conditionally

?

samlev:
I'm not sure if I understand your question.

A re-thrown exception is an exception that you catch, and then throw again (because PHP doesn't support exception bubbling - an exception has to be caught in the same scope or the scope above where it's thrown). It's commonly done when you're working with a class or a function which may throw an exception, but you just want it to bubble up, instead of dealing with it directly yourself.

Take the code in this [codebap](http://codepad.org/CZxPeWQs).

It's good practice to catch and re-throw exceptions that we're aware are possible (even if they don't happen). Sometimes you may want to add extra context to the exception, or run a logging function, but if you're not going to deal with the exception at that point, it's good to catch it and re-throw it, rather than letting it bubble up silently.

No-one using your code/functions/classes/library should be *surprised* by an exception - it should be explicitly documented, and this change to the PHPCS `@throws` commenting sniff allows you to properly document the expected/possible behaviour of your program, instead of being unable to document exceptions which you are aware of, catch, and re-throw.

aik099:
What I mean is following code:

```php
<?php

class A {

	public function throwFromClosure()
	{
		return function() {
			throw new Exception('message');
		}
	}

	public function conditionalRethrow()
	{
		try {
			callSomethingThatThows();
		}
		catch ( Exception $e ) {
			if ( $e instanceof InvalidArgumentException ) {
				throw $e;
			}
		}
	}

}
```

In both cases fact, that `throw` statement is present in the method doesn't guarantee that exception will be thrown by that method.

aik099:
With an uncoditional rethrow like in your original example you don't know actual exception class to report in error message. You can use what was used in `catch (...)` statement, but that not necessarily best way.

For example when an `InvalidArgumentException` is thrown but you catch the `LogicException` (the `InvalidArgumentException` is sub-class from it), then in DocBlock you want PHP_CodeSniffer to expect `InvalidArgumentException` and not `LogicException`.

samlev:
OK, now I think I understand.

1. In the instance of the closure, `throwFromClosure` _wouldn't_ throw an exception, because it _returns_ the closure without calling it. It's an anonymous function, which is inherently difficult to document.

2. Correct, in so much that it would expect you to document `Exception` as something thrown, instead of `InvalidArgumentException`. I would argue that in that case, you would be better to have a `catch` case for `InvalidArgumentException`, and deal with it explicitly, rather than catching the broadest possible type of Exception, then only dealing with expected cases.

This change is not perfect - for example, it can't account for situations where the variable is replaced with a different type of Exception, and then thrown.

It is, however, a vast improvement on the existing rule which _stops_ you from explicitly documenting exceptions that you are aware of, and deliberately re-throw.

I don't think that static code analysis will ever catch every possible edge case for documentation, but the existing rule forces you to write worse documentation, and this pull request allows you to write more accurate documentation.

aik099:
> This change is not perfect - for example, it can't account for situations where the variable is replaced with a different type of Exception, and then thrown.

For that same reason we haven't implemented it before. Too much edge cases and we don't want false positives to happen.

We'll wait for @gsherwood opinion on this one.

samlev:
To put it another way, the examples that you pointed out are, in and of themselves, "code smell", albeit code smell which is very difficult to detect or work around. This pull request aims to reduce the codes smell of undocumented `throw` calls when the exception being thrown is an exception that is already explicitly caught.

Another sane modification to make would be to analyze a throw exception to see if it is already caught within the same scope. For example, with this code, it would force you to document a `@throws` despite the fact that nothing is actually thrown out of the function.

```php
function selfContainedException() {
    try {
        if (someFunctionFails()) {
            throw new Exception();
        }
    } catch (Exception $e) {
        // This isn't important, so it can be ignored.
    }
}
```

samlev:
bump @gsherwood

gsherwood:
Sorry it took so long to get around to this. Last year was pretty crazy.

Thanks a lot for this PR. It works nicely.

PR #950, opened 2016-04-10
Allow multiple default_standard values in CodeSniffer.conf

schnittstabil:
This PR fixes the following weird behavior:
```sh
$ phpcs --standard=PSR1,PSR2 src/
$ phpcs --config-set default_standard PSR1,PSR2
Config value "default_standard" added successfully
$ phpcs src/
ERROR: the "PSR1,PSR2" coding standard is not installed. The installed coding standards are PSR2, PHPCS, PEAR, PSR1, MySource, Squiz and Zen
```

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -1079,7 +1079,7 @@ class PHP_CodeSniffer_CLI
                 $standard = 'PEAR';
             }
 
-            return array($standard);
+            return explode(',', $standard);
         }//end if
 
         $cleaned   = array();


gsherwood:
I'd like to merge this change in, but why are your tests creating temp dirs?

schnittstabil:
The test simulates an environment without `phpcs.xml` and `phpcs.xml.dist` in the cwd, because these  [always take priority](https://github.com/squizlabs/PHP_CodeSniffer/blob/2.6.0/CodeSniffer/CLI.php#L1058-L1076) over values set by `PHP_CodeSniffer::setConfigData`.

Otherwise `$sut->validateStandard(null)` would return something like `…/PHP_CodeSniffer/phpcs.xml.dist`…

gsherwood:
> Otherwise $sut->validateStandard(null) would return something like …/PHP_CodeSniffer/phpcs.xml.dist…

Assuming you are running them from a git clone I suppose. I really don't want unit tests creating their own environments or writing files. If you want to submit a PR without the unit test, I'll merge the one-liner in. I'd rather not have a test than have them creating temp dirs. I run them hundreds of times a day.

schnittstabil:
@gsherwood Done; Thanks a lot.

gsherwood:
Thanks a lot for doing that, and for the contribution.

Issue #810, opened 2015-12-07
Zend Naming Convention Sniff bug


ARH-Digital:
Using Zend rules. 
Zend/Sniffs/NamingConventions/ValidVariableNameSniff.php

Incorrectly ignores T_OBJECT_OPERATOR,
example, 

`$Index->short_name = $Data;`
`135 | ERROR | Variable "short_name" is not in valid camel caps format`

`short_name` should not be considered a variable, or it should be excluded due to the operator. This is a snake_cased value for a database column. It is not proceeded by a $ either.

gsherwood:
There is no way for PHPCS to know if that member var has been set with an invalid name, or if it has been set to match a DB column name, or if it is being access using __get(). Either way, the member var name is now in the correct format and so it throws an error.

The alternative is to never check any accessed member vars. I'm not sure that it is a good change.

aik099:
In my standard I've modified the `ValidVariableNameSniff` to only check class property declaration, not usage. This way if the property name is wrong I get 1 error instead lots of them on every usage place.

arlaneenalra:
I just ran into this.. Looking at the code I can see a few possible fixes ...

- Remove this block: 
 https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Zend/Sniffs/NamingConventions/ValidVariableNameSniff.php#L83-L112
  Which would get rid of all reference time checks of member vars ..
 - Change the reported sniff here https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Zend/Sniffs/NamingConventions/ValidVariableNameSniff.php#L104 to match the sniff here https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Zend/Sniffs/NamingConventions/ValidVariableNameSniff.php#L185  This should probably also be done on the contains Numbers sniffs and would make this a maskable sniff while maintaining the disallowance of snake_case for variables in the local scope.  I'll probably be throwing a PR out there in a bit ...

gsherwood:
Closing this as this isn't a change I intend to make to the sniff.

PR #954, opened 2016-04-12
Distinguish MemberVar reference sniffs from local variable sniffs. (issue 810)

arlaneenalra:
See #810

Update the reported sniff for member variable references to actually claim to be member variable references so that they can be appropriately filtered as such.

The second of the options suggested in https://github.com/squizlabs/PHP_CodeSniffer/issues/810#issuecomment-209101675

### DIFF
--- a/CodeSniffer/Standards/Zend/Sniffs/NamingConventions/ValidVariableNameSniff.php
+++ b/CodeSniffer/Standards/Zend/Sniffs/NamingConventions/ValidVariableNameSniff.php
@@ -101,11 +101,11 @@ class Zend_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_CodeSniff
                     if (PHP_CodeSniffer::isCamelCaps($objVarName, false, true, false) === false) {
                         $error = 'Variable "%s" is not in valid camel caps format';
                         $data  = array($originalVarName);
-                        $phpcsFile->addError($error, $var, 'NotCamelCaps', $data);
+                        $phpcsFile->addError($error, $var, 'MemberVarNotCamelCaps', $data);
                     } else if (preg_match('|\d|', $objVarName) === 1) {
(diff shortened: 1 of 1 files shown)


aik099:
This is a BC break.

arlaneenalra:
That is true ... I could try breaking this out into a new sniff in under the generic sniffs.  Mainly I'm looking for a way allow member variables to be snake_case while forcing local variables to be camelCase so the individual sniff itself doesn't matter to me.  Would that be more palatable?

aik099:
@arlaneenalra , I also had such need plus I needed to check only variable declaration and not usages. I had to create my custom sniff for that: https://github.com/aik099/CodingStandard/blob/master/CodingStandard/Sniffs/NamingConventions/ValidVariableNameSniff.php

It's exactly forcing class members (properties) to be in camel case and all other variables in snake case.

arlaneenalra:
@aik099 Nice implementation, it feels cleaner than the base Zend one (albeit that may be because it strips the contains number test.)  I can see enough combinations in this that I was considering taking the Zend base and adding a configuration option to pick the allowed case for member and local vars.  Do you have any objection to me taking your version and updating it accordingly? (Unless of course you want to ;)  Either way I can see something like that covering #810 and maybe a few other lurking issues or frustrations and, assuming it get's accepted, avoiding the need for custom code in the future.

aik099:
> Nice implementation, it feels cleaner than the base Zend one

I've used Squiz sniff (if I'm not mistaken) as base and removed not needed parts.

> Do you have any objection to me taking your version and updating it accordingly? (Unless of course you want to ;) 

You mean using modified version of my sniff in your project? Sure, use it as you need, because it's open source.

arlaneenalra:
I'm going to close this out for the moment and take a different tack.

PR #957, opened 2016-04-17
Operator spacing sniff errors when plus is used as part of a number

klausi:


gsherwood:
Thanks for the fix. I confess that I've never even tried to use a number like this in PHP code.

Issue #959, opened 2016-04-19
Call-time pass-by-reference false positive if there is a square bracket before the ampersand


realmfoo:
It's a valid bitwise operator:
```
myfunc($var[0] & $var);
```

gsherwood:
Thanks for the report and the fix.

PR #960, opened 2016-04-19
Fixed bug #959 : Call-time pass-by-reference false positive if there is a square bracket before the ampersand

realmfoo:
Fixes issue #959


gsherwood:
Thanks a lot for the patch and the test.

Issue #962, opened 2016-04-20
Null coalescing operator (??) not detected as a token


joelposti:
I’m extending Squiz.WhiteSpace.OperatorSpacing to check for whitespace around elvis operator (?:) and null coalescing operator (??). I already managed to get the sniff to work correctly with elvis operator. The sniff is registered to listen to tokens `T_INLINE_THEN` and `T_INLINE_ELSE` which match the elvis operator correctly. I would have guessed that null coalescing operator would be equal to two `T_INLINE_THEN` tokens, but instead the operator is not detected at all. I could not find a token for the operator in `CodeSniffer/Tokens.php`.

aik099:
I think token isn't supported currently and isn't considered as operator token. I've sent #844 PR before to add support for spaceship operator. You can send new PR for `??` operator using my PR as base.

aik099:
> elvis operator 

Elvis operator, really? Maybe you wanted to say `elseif` operator?

aik099:
Another thing is that PHP < 7.0 needs to recognize this new PHP 7 token as well to ease test writing.

joelposti:
Thanks for the pull request tip! However, I do not believe we can add support for the null coalescing operator as easily because PHP itself does not provide a token for the operator (https://secure.php.net/manual/en/tokens.php). I believe it is required to code support for the operator into `CodeSniffer/Tokenizers/PHP.php`.

I really did mean to say elvis operator. Elvis operator is a compression of the ternary operator. `$oof = $foo ?: $bar` is equal to `$oof = $foo ? $foo : $bar`. https://en.wikipedia.org/wiki/Elvis_operator

aik099:
> However, I do not believe we can add support for the null coalescing operator as easily because PHP itself does not provide a token for the operator (https://secure.php.net/manual/en/tokens.php)

That's not a problem. PHP_CodeSniffer has many useful tokens, that PHP tokenizer doesn't provide.

joelposti:
I found out the token name while running phpcs in extra verbose mode (-vv). The token is `T_COALESCE`. If there is a space between the question marks (syntax error), they are recognized as two `T_INLINE_THEN` tokens. I’m in the process of creating a pull request.

gsherwood:
Thanks for the report and the fix.

PR #963, opened 2016-04-20
T_COALESCE as operator token

joelposti:
This adds support for T_COALESCE token that corresponds to null coalescing operator (??) added in PHP 7.0.

Fixes issue #962.

### DIFF
--- a/CodeSniffer/Tokens.php
+++ b/CodeSniffer/Tokens.php
@@ -141,6 +141,10 @@ if (defined('T_SPACESHIP') === false) {
     define('T_SPACESHIP', 'PHPCS_T_SPACESHIP');
 }
 
+if (defined('T_COALESCE') === false) {
+    define('T_COALESCE', 'PHPCS_T_COALESCE');
+}
+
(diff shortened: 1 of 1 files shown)


aik099:
What about fallback code for PHP < 7?

joelposti:
Doesn't the addition on lines 144-147 serve as fallback code for PHP < 7?

joelposti:
Or do you mean additions similar to the changes made in commit c5fda3e? Unfortunately I do not have PHP 5.x environment readily available for developing that.

aik099:
> Or do you mean additions similar to the changes made in commit https://github.com/squizlabs/PHP_CodeSniffer/commit/c5fda3efd8b54d00483927b566af30ab2e416373? Unfortunately I do not have PHP 5.x environment readily available for developing that.

Yes, I mean exactly them.

gsherwood:
Thanks for the fix. I've also added support for T_COALESCE to the tokenizer for older PHP versions.

joelposti:
Thank you for the merge and maintaining this excellent piece of software!

PR #964, opened 2016-04-20
callable and self tokens should be converted to T_RETURN_TYPE too

kukulich:


aik099:
Please also add tests covering code changes.

kukulich:
@aik099 Unfortunately I didn't find any test where I could add new checks.

aik099:
You can look at commit, where T_RETURN_TYPE token support was added initially. I bet that commit or commits right after that are adding these tests.

kukulich:
@aik099 No, there are not tests for this. There are only tests for ArrayDeclarationSniff: https://github.com/squizlabs/PHP_CodeSniffer/commit/71d0292dbfd17cf68fbfacf08cc9b4c330a79a2c

aik099:
Then you can add your tests to same sniff to be consistent.

aik099:
There are no dedicated tokenizer tests, because it's tested indirectly through sniffs that provide PHP code for tokenizing.

kukulich:
@aik099 "self" and "callable" in ArrayDeclarationSniff really don't make any sense. And as I said earlier I didn't find any suitable sniff tests (for "self" or "callable") where I could add new checks.

aik099:
Doesn't https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Sniffs/Functions/FunctionDeclarationSniff.php sniff look suitable? It checks function declaration.

gsherwood:
Thanks for the PR. I also include T_PARENT as this is valid as well, according to the RFC.

Issue #920, opened 2016-03-14
Request for first member var exception on Squiz.WhiteSpace.MemberVarSpacing


photodude:
Squiz.WhiteSpace.MemberVarSpacing currently enforces a new line before all the member Vars including the first member var. It would be helpful (and would help avoid unnecessary custom sniffs), if there was an option to allow for an exception to the newline requirement on the first member var.

glen-84:
#725 ?

gsherwood:
Closing as a duplicate of #725

Issue #725, opened 2015-09-30
MemberVarSpacingSniff – allow var on first line of class


glen-84:
This is not allowed:

``` php
class X
{
    public $x;

    public $y;
}
```

`Expected 1 blank line before member var; 0 found`

glen-84:
The same is true for `Squiz.WhiteSpace.FunctionSpacing` (before/after). I'd like to specify 1 line between methods, but not above the first method or below the last.

gsherwood:
This will now be possible from version 3.1.0. To have a single blank line between member vars, but no blank line before the first member var, put this in your ruleset.xml file:
```xml
<rule ref="Squiz.WhiteSpace.MemberVarSpacing">
    <properties>
        <property name="spacingBeforeFirst" value="0"/>
    </properties>
</rule>
```

PR #972, opened 2016-04-25
Exception to the newline requirement on the first member var after class opener.

photodude:
Based on the draft PHPCS2 Joomla Custom member var sniff https://github.com/photodude/coding-standards/blob/phpcs-2/Joomla/Sniffs/WhiteSpace/MemberVarSpacingSniff.php#L102-L137

Addresses issues #920 and #725

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/MemberVarSpacingSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/MemberVarSpacingSniff.php
@@ -42,6 +42,7 @@ class Squiz_Sniffs_WhiteSpace_MemberVarSpacingSniff extends PHP_CodeSniffer_Stan
      */
     public $firstVarException = false;
 
+
     /**
      * Processes the function tokens within the class.
      *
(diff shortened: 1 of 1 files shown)


CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/MemberVarSpacingSniff.php
```diff
@@ -34,6 +34,16 @@ class Squiz_Sniffs_WhiteSpace_MemberVarSpacingSniff extends PHP_CodeSniffer_Stan
 
 
     /**
+     * The exception to the newline requirement on the first member var after class opener.
+     * false for newline requirement on the first member var after class opener
+     * true for no newline requirement on the first member var after class opener
+     *
+     * @var bool
+     */
+    public $firstVarException = false;
```
aik099: This is a BC break.

photodude: Maybe I'm not seeing what you're seeing, How is this a BC break?

What I see is a new API end point adding additional functionality which is controlled by a new member var parameter. 

The existing behavior for this sniff is unchanged... unless you set the new member var parameter to true thus turning on the exception.

aik099: Ah, I thought that you've wrapped existing code with an IF, which was disabled by default. Now I see that it's completely new code (maybe fixing part is copy-pasted from somewhere).

photodude: Yes, it's listed in the PR description.

photodude:
@gsherwood can this be considered for merge?

gsherwood:
Sorry, just getting around to look at this PR.

If options are going to be added to this sniff, I think it is better to have 2 new options; 1 to specify how many blank lines should be between member vars, and a second that specifies how many blank lines before the first member var.

The same set of options could probably be added to Squiz.WhiteSpace.FunctionSpacing. This sniff actually has a `spacing` property already to define how many blank lines should be between functions, but it doesn't have an option to say how many blank lines should be before the first function found in a file or class.

If you don't have time to look into those two new options, I'll get around to this eventually.

photodude:
Sounds like a plan. I'm not sure when, or if, I'll get around to looking into making the modifications you suggested.

wilsonge:
See https://github.com/squizlabs/PHP_CodeSniffer/pull/1355 for an attempted implementation

photodude:
Closing in favor of #1355 (thanks for making the requested modifications @wilsonge )

wilsonge:
No stress :)

PR #976, opened 2016-04-25
Added ignoreNewlines parameter to Squiz.WhiteSpace.ObjectOperatorSpacing

AlexHowansky:
The Squiz.WhiteSpace.OperatorSpacing sniff has a parameter named ignoreNewlines. Thus, I can disallow:

`$a = $b+$c+$d;`

While allowing:

```
$a =
    $b +
    $c +
    $d;
```

I would like to do the same for object operators, and disallow:

`$foo -> bar() -> baz();`

While allowing:

```
$foo
    ->bar()
    ->baz();
```

However, the Squiz.WhiteSpace.ObjectOperatorSpacing sniff does not have an ignoreNewlines parameter, so the latter is always flagged.

This is an enhancement request to add an ignoreNewlines parameter to the Squiz.WhiteSpace.ObjectOperatorSpacing sniff. Thanks.

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/ObjectOperatorSpacingSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/ObjectOperatorSpacingSniff.php
@@ -30,6 +30,13 @@
 class Squiz_Sniffs_WhiteSpace_ObjectOperatorSpacingSniff implements PHP_CodeSniffer_Sniff
 {
 
+    /**
+     * Allow newlines instead of spaces.
+     *
+     * @var boolean
+     */
+    public $ignoreNewlines = false;
+
 
     /**
      * Returns an array of tokens this test wants to listen for.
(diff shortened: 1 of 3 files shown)


photodude:
@AlexHowansky in your first example would the operator being first be allowed, or optionally be allowed?
As in:

```
$a =
        $b
     + $c
     + $d;
```

or 

```
$a = $b
     + $c
     + $d;
```

AlexHowansky:
Note the ignoreNewlines parameter in the Squiz.WhiteSpace.OperatorSpacing sniff was already there -- I did not add it. If set to true, it already allows both cases in your example. I added the same parameter to the Squiz.WhiteSpace.ObjectOperatorSpacing sniff, with the intent of having it work the same way. So, all the following would be accepted if you have it set to true:

    $foo->bar()->baz();

    $foo
        ->bar()
        ->baz();

    $foo->
        bar()->
        baz();

gsherwood:
Sorry it took so long for me to review this. The change worked really well, so I've merged it in. Thanks a lot.

PR #978, opened 2016-04-25
Added Generic.Formatting.AlphabeticalMethodNames sniff.

AlexHowansky:


gsherwood:
This feels a bit too specific for PHPCS. I've never seen a coding standard that requires methods in alphabetical order (is there a common one?). It also feels like developers would require options like how to treat uppercase characters and underscores.

Unless this is a common requirement for coding standards, I think it is best that it stays in a custom standard.

AlexHowansky:
Yes, you're right -- I do not know of a common standard that requires this, it's just something we do here in my local dev group. I do not intend for it to be on by default for any of the included standards, so I did not add it to any existing ruleset.xml files. I simply figured it would be nice to have available for anyone if desired in their own custom ruleset.xml file. This is why I put it in Generic -- from what I see, none of the standards "include all" from Generic, they just pull in singular specific sniffs as needed. Thus, adding new sniffs to Generic won't affect any existing standards but will still allow people to add it to their own custom rulesets. It's no different than the Generic.PHP.SAPIUsage or Generic.VersionControl.SubversionProperties sniffs -- none of the standards use them but they're available for custom rulesets.

aik099:
> I see, none of the standards "include all" from Generic,

I did previously, but @gsherwood recommended not to do so. But maybe somebody still does.

AlexHowansky:
> But maybe somebody still does.

Newly added and unwanted sniffs can be easily excluded. If anybody has an existing custom ruleset that includes all in Generic, then they *must* already be using excludes to skip all the sniffs that are mutually exclusive. (Like UpperCaseConstants vs LowerCaseConstants.)

If you're going to use the logic "somebody somewhere might be including all in Generic" as a rationale to not add any new sniffs, then we can never add any new sniffs...

aik099:
Which sniff comes in and which doesn't is decided purely on general usefulness of the sniff.

IMO the sniff that checks method sorting in a class is something too specific to be useful by general public. It indeed is very useful as custom sniff for whoever developed it originally, but not that useful for majority of phpcs users. This is not an official statement, just my personal review of the sniff.

gsherwood:
Before I could accept this, I think it would probably need:
- an option to put non-alphanumeric chars first or last
- an option to put numbers first or last
- an option to put uppercase first or last
(or maybe just some option so that order can be specified: special,numbers,lower,upper)
- code to handle non-English chars for all the above

I still don't feel like this would be widely used, but if you want to take the time to improve it to this level, I'll merge it in. Once I do, all bug reports with it come directly to me, which is why I need to be careful about what sniffs go in because I'm creating more work for myself. Nothing against you personally, or the sniff.

gsherwood:
Closing as I believe this is too specific and would need the options and tests specified before I could  take it on for future support.

PR #984, opened 2016-04-28
Ignore lines to indent in Generic.WhiteSpace.ScopeIndentSniff

joelposti:
This adds a functionality for ignoring specified lines when using Generic.WhiteSpace.ScopeIndentSniff. It complements the existing `ignoreIndentationTokens` and `nonIndentingScopes` functionalities. The use case is similar: sniffs that extend Generic.WhiteSpace.ScopeIndentSniff can make the parent class ignore some lines.

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/WhiteSpace/ScopeIndentSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/WhiteSpace/ScopeIndentSniff.php
@@ -100,6 +100,16 @@ class Generic_Sniffs_WhiteSpace_ScopeIndentSniff implements PHP_CodeSniffer_Snif
     private $_ignoreIndentationTokens = array();
 
     /**
+     * List of lines not needing to be checked for indentation.
+     *
+     * Useful to allow Sniffs based on this to easily ignore/skip some
+     * lines from verification.
+     *
+     * @var int[]
+     */
+    public $ignoreLines = array();
+
+    /**
      * Any scope openers that should not cause an indent.
      *
      * @var int[]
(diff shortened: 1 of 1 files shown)


aik099:
If you want to ignore lines during PHPCS checks just specify `@phpcsIgnore` (I don't remember exact name) annotation to mark such lines. Not sure, but maybe there is a way to ignore specific sniff for specific lines as well.

joelposti:
Thanks for the hint! However, I do not have just a few lines to ignore which could be annotated like you said. I have certain structures which I want to indent according to my rules, not according to those imposed by Generic.WhiteSpace.ScopeIndentSniff. However, I do not wish to completely disable the sniff because it is useful. Also, I do not want to copy the whole code of `process()` to my own sniff because that breaks update-proofyness in a major way; updates to Generic.WhiteSpace.ScopeIndentSniff would not be reflected on my sniff.

aik099:
That's why I never enable exact scope indent parameter.

gsherwood:
I don't think sniffs should have a setting that specifies a list of lines to ignore. The sniff would then be custom to a specific file. Ignoring lines is something done via the core recording and reporting code.

There are some annotations to remove error for lines (https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file) but you can't limit this to a specific sniff yet, and you probably don't want to go putting these all over your files anyway.

joelposti:
It doesn’t have to be public. It could be protected which would not expose the property to ruleset.xml. That would also mean that ignoring lines could only be used per file basis: in a sniff that extends Generic.WhiteSpace.ScopeIndentSniff. Would you accept the change if the property was protected?

gsherwood:
> Would you accept the change if the property was protected?

I wouldn't, for the same reasons.

> Does CodeSniffer create a new instance of each sniff for each file it processes?

No, just a single instance. The lines would carry over.

PR #990, opened 2016-05-03
Allow input paths / files to be passed in a file

andrei-propertyguru:
This PR allows running the Code Sniffer on multiple PHP files given as input and output a single report file. This is more convenient than passing the files on the command line(command line limits, simpler handling of input, etc). 

Unix throws an error in case many files(>hundreds) are passed on the command line:
`-bash: <process>: Argument list too long`

This solves the above issue.

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -1254,7 +1254,7 @@ class PHP_CodeSniffer_CLI
         echo '    [--severity=<severity>] [--error-severity=<severity>] [--warning-severity=<severity>]'.PHP_EOL;
         echo '    [--runtime-set key value] [--config-set key value] [--config-delete key] [--config-show]'.PHP_EOL;
         echo '    [--standard=<standard>] [--sniffs=<sniffs>] [--encoding=<encoding>]'.PHP_EOL;
-        echo '    [--extensions=<extensions>] [--ignore=<patterns>] [--bootstrap=<bootstrap>] <file> ...'.PHP_EOL;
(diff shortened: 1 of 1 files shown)


aik099:
Why you're passing bunch of files instead of passing a folder and then excluding not needed files via ruleset?

andrei-propertyguru:
@aik099 I am using it in a script that checks only the files changed between 2 git commits. These files can be in multiple folders, and in some cases ( large refactors, white space / line ending changes ) the file list can be huge ( over 10k )
Also running this on the whole app will take 2-3 hours.

aik099:
I have similar task but what I do is use build server to run checks on all project and then compare to reports to associate issues found with changes made in particular commit.

andrei-propertyguru:
Large refactors like moving entire sections of code, or changing the line endings from windows to UNIX, should be done in a single commit though. This can include thousands of files.

Having a file list argument is a common pattern in unix tools.

aik099:
> Having a file list argument is a common pattern in unix tools.

Any examples of such tools?

andrei-propertyguru:
rsync, tar, wget, ctags to name a few

aik099:
Nice.

gsherwood:
Thanks for the contribution. This is a nice feature to have.

I did make some slight changes because it was implemented as `phpcs --file-list /path/to/file` but documented as `phpcs --file-list=/path/to/file`. I used the `--file-list=` format as it is consistent with the other options.

@andrei-propertyguru I'd like to thank you in the changelog. If you'd be happy for me to do that, just let me know what name to use.

andrei-propertyguru:
@gsherwood thank you. You can add me as Blotzu.

gsherwood:
> You can add me as Blotzu.

Thanks.

Issue #993, opened 2016-05-05
phpcs reads from stdin which requires a terminal, and hangs if there are none.


masterzen:
With version 2.6.0, `phpcs` now tries to read content from `stdin`. 

This of course either requires a terminal or an incoming stream from another process.
When running `phpcs` as part of a git pre-commit hook in `git gui` or any other GUI around git which doesn't allocate a terminal, this completely blocks the pre-commit process and and the GUI itself, the process is stuck in T state on linux or osx.

gsherwood:
As discussed on the linked PR, I've tried to replicate this issue using a pre-commit hook and `git gui`, but I'm not having any problems on 2.6.0.

Here is a quick recording of what I was doing on OS X:
![git gui](https://cloud.githubusercontent.com/assets/185976/15059008/2be736f8-1363-11e6-9ef1-decad541b293.gif)

So presumably, STDIN is available for my GUI but possibly not for yours.

Are you able to debug the code and see if there is an easy way to detect the difference between your command line and the GUI? Maybe something as simple as `fopen()` returning false?

masterzen:
Many thanks @gsherwood for your efforts trying to reproduce the issue.

The [bash script](https://gist.github.com/codemis/8225337) I'm using is doing:
~~~bash
...
OUTPUT=$($PHPCS_BIN -s --standard=$PHPCS_CODING_STANDARD $EXTENSIONS $ENCODING $IGNORE $STAGED_FILES)
RETVAL=$?
...
if [ $RETVAL -ne 0 ]; then
    echo "$OUTPUT" | less
fi
exit $RETVAL
~~~

This might be the reason. 
I'm going to change the script to not run in a subshell.

masterzen:
To make this script working one need to use:
~~~bash
OUTPUT=$(echo '' | $PHPCS_BIN -s --standard=$PHPCS_CODING_STANDARD $EXTENSIONS $ENCODING $IGNORE $STAGED_FILES)
...
~~~

To force the existence of stdin (at least that we can open it).

gsherwood:
Strange. I was using the exact same script you just linked to (you linked it in the PR as well) and was getting the result I posted. I also didn't think STDIN disappears when using subshells.

nnnikolay:
Just want to share my case

I recently update phpcs to 2.6 and I've started to have the same problem with STDIN. I've never had it before.

I'm using phing to automate some build steps, and I'm running it from pre-push hook like this
```
[...]

PR #994, opened 2016-05-05
Fix #996 - do not read from STDIN if there is no terminal

### LINKED ISSUE #996: phpcbf could fix better
The PSR2 rule that states : "The closing parenthesis and the opening brace of a multi-line function declaration must be on the same line".

While this is automatically fixable it does that while leaving an empty new line. Would it be possible to remove that extraneous line?

```
    public static function create(
        $addressee = null,
        $street = null,
        string $streetOrdinality = null,
        string $streetOrdinalitySuffix = null,
        string $postalCode = null,
        string $populatedPlace = null,
        Country $country = null,
        CountryRegion $region = null,
        Coordinate $geoLocation = null
    ): Address
    {
        return new static(...omited
    }
```
Will be fixed as
```
    public static function create(
        $addressee = null,
        $street = null,
        string $streetOrdinality = null,
        string $streetOrdinalitySuffix = null,
        string $postalCode = null,
        string $populatedPlace = null,
        Country $country = null,
        CountryRegion $region = null,
        Coordinate $geoLocation = null
    ): Address {
    
        return new static(...omited
    }
```
While the following would be more appropriate
```
    public static function create(
        $addressee = null,
        $street = null,
        string $streetOrdinality = null,
        string $streetOrdinalitySuffix = null,
        string $postalCode = null,
        string $populatedPlace = null,
        Country $country = null,
        CountryRegion $region = null,
        Coordinate $geoLocation = null
    ): Address {
        return new static(...omited
    }
```

masterzen:
In pre-commit hooks, it can happen that there are no terminal or
contents to read (if the hook runs in a GUI), in which case attempting
to read from STDIN will hang the whole process.
This makes sure there is something to read from STDIN.

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -389,20 +389,21 @@ class PHP_CodeSniffer_CLI
 
         $this->setCommandLineValues($args);
 
-        // Check for content on STDIN.
-        $handle = fopen('php://stdin', 'r');
-        if (stream_set_blocking($handle, false) === true) {
-            $fileContents = '';
-            while (($line = fgets(STDIN)) !== false) {
+        // Check for content on STDIN
+        $handle       = fopen('php://stdin', 'r');
+        $read         = array($handle);
+        $write        = null;
+        $except       = null;
+        $fileContents = '';
+        if (stream_select($read, $write, $except, 0) === 1) {
+            while ($line = fgets($handle)) {
                 $fileContents .= $line;
-                usleep(10);
             }
+        }
 
-            stream_set_blocking($handle, true);
-            fclose($handle);
-            if (trim($fileContents) !== '') {
-                $this->values['stdin'] = $fileContents;
-            }
+        fclose($handle);
+        if (trim($fileContents) !== '') {
+            $this->values['stdin'] = $fileContents;
         }
 
         return $this->values;


aik099:
If you supply files/folders to check in command arguments it still looks in STDIN?

gsherwood:
Unfortunately, this looks a lot like the first code I wrote for this feature, which has problems: https://github.com/squizlabs/PHP_CodeSniffer/commit/bd571ae9206ab671121e3f77d0fb29a3bc6cb669

I went through a range of versions before getting to the current code. The way it reads STDIN, including the `usleep()` and non-blocking reads, was essential for getting all IDE plugins to work correctly. Changing that is going to take a lot of manual testing again.

Otherwise, I'll see if I can replicate the problem you are having and work an additional check into the code. But I can't merge this in as-is.

Other related commits in this STDIN saga: https://github.com/squizlabs/PHP_CodeSniffer/commit/80b156de8d440b6245e6b1dd7f8df777d99064fa https://github.com/squizlabs/PHP_CodeSniffer/commit/8d669c50e4122c592a634580d9ea8b30c4edd0e1 https://github.com/squizlabs/PHP_CodeSniffer/commit/63fafe07d6cb8588b787475a3cc041f52b05a4ac

masterzen:
@gsherwood, thanks for the explanation. 

Maybe we could just add a command line arguments for preventing to read from `STDIN`?

To elaborate a bit on my setup, the pre-commit hook is a bash script that looks like [this one](https://gist.github.com/codemis/8225337).

Do you want me close this PR and continue the discussion in the issue #993 ?

gsherwood:
> Do you want me close this PR and continue the discussion in the issue #993 ?

Good idea. I'll keep commenting over there.

PR #995, opened 2016-05-06
Refactoring: Remove ELSE block and have just IF condition

ssontakke:
Either if/else will always get executed, so we can move default values outside conditions. This will reduce the Cyclomatic Complexity.

If you are fine with this change, then I can take care of finding and replacing other such instances where we can remove else part. We would skip this type of refactoring where if-else block has heavy processing.

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -813,10 +813,12 @@ class PHP_CodeSniffer_File
         $severity=0,
         $fixable=false
     ) {
-        if ($stackPtr === null) {
-            $line   = 1;
-            $column = 1;
-        } else {
+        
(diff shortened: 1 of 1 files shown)


aik099:
Not sure if build failure is related.

gsherwood:
I prefer the code as it is because it is explicit to me why I have chosen the default values. I am not part of the "no else" code movement. I really like ELSE conditions.

I'm closing this because it's not fixing a specific problem and this just isn't my personal coding style in this specific case. I don't think it is a good use of your time to go looking for else statements that can be removed. But if you happen to find some really obvious ones that are causing performance or readability issues, I'd be happy to review them.

PR #998, opened 2016-05-09
set 'orig_column' when changing tabs to spaces

### LINKED ISSUE #982: Wrong column reported when using tabs
The wrong column is reported when the `--tab-width` parameter is set and the error line uses tabs

UziTech:
Set 'orig_column' on tokens when tabs are changed to spaces and use 'orig_column' on errors and warnings for the correct column

fixes #982

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -1490,7 +1490,7 @@ class PHP_CodeSniffer_File
         $tokenizerType = get_class($tokenizer);
         $ignoring      = false;
         $inTests       = defined('PHP_CODESNIFFER_IN_TESTS');
-        $tabLines      = [];
+        $tabLines      = array();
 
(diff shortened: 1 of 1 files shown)


CodeSniffer/File.php
```diff
@@ -1482,6 +1490,7 @@ private static function _createPositionMap(&$tokens, $tokenizer, $eolChar, $enco
         $tokenizerType = get_class($tokenizer);
         $ignoring      = false;
         $inTests       = defined('PHP_CODESNIFFER_IN_TESTS');
+        $tabLines      = [];
```
gsherwood: You can't use short array syntax in version 2.x of PHPCS. I can fix this after merging, but you might be able to get to it before I do.

gsherwood:
Closed #982 , so this is no longer needed.

PR #1006, opened 2016-05-17
Use php's internal defined function list for sniff

michaelbutler:
We can no longer rely on calling `function_exists` to determine whether the function token in question is a "PHP internal" function. Mainly because PHP_CodeSniffer now uses the vendor/autoload.php file (if it exists) to load classes. This is a shared autoload system, that could cause many user defined functions to exist in the same memory space under which PHPCS is running.

Since the whole point of the sniff is to check only PHP built in (inbuilt) functions, I've made the change in this PR to check the function using `get_defined_functions`.

I didn't change or add any unit tests, since I couldn't think of any additional use cases needed. I suppose the only risk is if a given function returns `true` for function_exists() but does not exist in get_defined_functions()['internal'].

- phpcs passes
- phpunit passes

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/PHP/LowercasePHPFunctionsSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/PHP/LowercasePHPFunctionsSniff.php
@@ -95,11 +95,10 @@ class Squiz_Sniffs_PHP_LowercasePHPFunctionsSniff implements PHP_CodeSniffer_Sni
         }
 
         // Make sure it is an inbuilt PHP function.
-        // PHP_CodeSniffer doesn't include/require any files, so no
-        // user defined global functions can exist, except for
-        // PHP_CodeSniffer ones.
+        // PHP_CodeSniffer can possibly include user defined functions
+        // through the use of vendor/autoload.php.
         $content = $tokens[$stackPtr]['content'];
-        if (function_exists($content) === false) {
+        if ($this->isBuiltInFunction($content) === false) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


CodeSniffer/Standards/Squiz/Sniffs/PHP/LowercasePHPFunctionsSniff.php
```diff
@@ -119,4 +118,22 @@ public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
...
 
+    /**
+     * Determine whether the functionName in question is a built-in
+     * (or extension-provided) function.
+     *
+     * @param string $functionName Name of the function to check
+     *
+     * @return bool
+     */
+    protected function isBuiltInFunction($functionName)
```
aik099: Performance-wise this is very bad, because `isBuiltInFunction` is called for every checked function.

Since function list won't change anyway I'm proposing to:

1. create list of functions in sniff constructor and use it here
2. in created function list set function name as key in array and `true` as value
3. check using `isset($this->definedFunctions) === true`

Checking by key is faster, then doing `in_array`.

gsherwood: I like the above ideas. I'll add one more. Don't make the call an additional function. It's cleaner code, but the additional function call for every T_STRING token that appears to be a function name actually adds significant overhead when checking large codebases. I know, it's horrible. But if it shaves a few seconds off a run, it's good.

michaelbutler: Thanks for the tips, will do! Think the array hash map should be a static
variable, or not necessary because there will only ever be one instance?

aik099: There is only 1 instance of the sniff created, so no need for static variable.

michaelbutler:
added the hash map for O(1) lookups. Note that I used array_flip instead of setting every value to 'true'. I usually find it's faster to take a flat array and call array_flip() on it to convert to a hashmap, with the value being the index it was. `isset()` still works as expected.

gsherwood:
I finally got around to merging. Thanks for the PR.

PR #1017, opened 2016-05-21
Makes "Squiz.Commenting.FunctionComment" sniff understand description in "@return" tag

### LINKED ISSUE #1015: The Squiz.Commenting.FunctionComment sniff doesn't allow description in @return tag
Code example:

```php

class ClassName {

	/**
	 * (PHP 5 &gt;= 5.0.0)<br/>
	 * Move forward to next element
	 *
	 * @return void Any returned value is ignored.
	 * @link   http://php.net/manual/en/iterator.next.php
	 */
	public function next()
	{
		$this->offset++;
	}

}
```

I'm getting `Squiz.Commenting.FunctionComment.InvalidNoReturn` error (message `Function return type is not void, but function has no return statement`) because `@return` statement has a description after it. This is allowed in DocBlocks - https://www.phpdoc.org/docs/latest/references/phpdoc/tags/return.html .

I think, that we should explode `@return` tag content into 2 parts by space, since spaces aren't allowed in returned type itself and only check 1st part as returned type.

Related Issue: #1018

aik099:
Closes #1015

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -97,9 +97,12 @@ class Squiz_Sniffs_Commenting_FunctionCommentSniff extends PEAR_Sniffs_Commentin
                     }
                 }
 
+                // Ignores description, that might come after return type.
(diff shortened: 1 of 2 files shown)


gsherwood:
Thanks for the PR.

PR #1019, opened 2016-05-21
Update File.php - possibly unused variables

ssontakke:
# Removed unused statement : $sniff     = $code;
# Removed unused variable $isWin and logic to set that value

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -993,7 +993,6 @@ class PHP_CodeSniffer_File
         // Work out which sniff generated the error.
         if (substr($code, 0, 9) === 'Internal.') {
             // Any internal message.
-            $sniff     = $code;
             $sniffCode = $code;
         } else {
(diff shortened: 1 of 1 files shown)


CodeSniffer/File.php
```diff
@@ -1321,7 +1319,7 @@ public function getWarningCount()
 
 
     /**
-     * Returns the number of successes recorded.
+     * Returns the number of successes recorded. Currently not in use.
```
gsherwood: This comment is irrelevant. It's part of the API that can be used by both custom sniffs and custom reports.

gsherwood:
The $sniff leftover args looks to be left there after a big refactor of processing in commit https://github.com/squizlabs/PHP_CodeSniffer/commit/55c5962f09b0a8af0f3275a997859e19a2e66740#diff-acee2709ae4492fd666b32bf4ad8ef93

I've merged the changes and removed the docblock comment change I commented on. Thanks for the other changes.

PR #1020, opened 2016-05-21
Update CLI.php - Update return type to match the function defination.

ssontakke:
Update return type to match the function defination.


gsherwood:
Thanks for the PR.

PR #1021, opened 2016-05-21
Update CLI.php - Update return type to match the function definition.

ssontakke:
Update return type to match the function defination.


gsherwood:
This is old code leftover from an old method of processing CLI arguments. It hasn't been needed since commit https://github.com/squizlabs/PHP_CodeSniffer/commit/6e210d7924ea7eb34a1d7bde5f9abf5285e1d1b3

Thanks for pointing it out. I've removed the line.

PR #1032, opened 2016-06-11
Fix PHP notices when file extensions defaults are not present

klausi:
Sometimes the internal values of CLI.php are not initialized. This happens on test runs in Drupal's coder. I'm trying to add

```xml
<arg name="extensions" value="php,module,inc,install,test,profile,theme,css,info,txt,md,yml"/>
```
to ruleset.xml to have some extension default values.

The commit is my best guess at fixing the notice, or is there another place where $this->values should be initialized?

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -722,6 +722,9 @@ class PHP_CodeSniffer_CLI
                     $this->values['standard'] = explode(',', $standards);
                 }
             } else if (substr($arg, 0, 11) === 'extensions=') {
+                if (isset($this->values['extensions']) === false) {
+                    $this->values['extensions'] = array();
+                }
                 $this->values['extensions'] = array_merge($this->values['extensions'], explode(',', substr($arg, 11)));
             } else if (substr($arg, 0, 9) === 'severity=') {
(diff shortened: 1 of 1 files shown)


gsherwood:
I can't figure out why you are having problems just yet, but I can replicate your issues with coder and it does appear to just be a unit testing thing given how a normal run works:

PHPCS sets default values for all settings here: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/CLI.php#L266

When CLI values are set, the first thing that happens is that these defaults are loading into the `values` member var here: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/CLI.php#L423

I'm not sure how command line values would be processed without the default values being set first, given that the `processLongArgument` method is only ever called from `setCommandLineValues` and that's the method that loads in the defaults. So I'll need to look more into what is happening during the test run.

gsherwood:
Gosh I'm an idiot sometimes. During unit testing, the defaults are intentionally not loaded. The code is right there in my second link: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/CLI.php#L420

gsherwood:
I think your fix is in the right spot. Thanks for the PR.

Issue #652, opened 2015-07-27
Squiz.WhiteSpace.FunctionSpacing last method no space


ofbeaton:
It would be nice if `Squiz.WhiteSpace.FunctionSpacing.After` would not require a blank line after a method if it is the last method in a class.

``` php
class A 
{
  public function method1()
  {
  }

  public function method2()
  {
  }

  public function method2()
  {
  }
}
```

This would be much nicer than enforcing a space after the last method. You could switch it to enforce a space before a method, but then you'd run into the same problem but at the top (assuming no properties).

aik099:
In fact I think it should also be checking for empty line above function declaration. Currently other sniff checks that, but only if class has any properties.

techi602:
Sadly FunctionSpacing is useless without ability to skip first and last method because it violates PSR-2 
(at least the example in PSR-2 example 1.1)

So this rule


```
    <rule ref="Squiz.WhiteSpace.FunctionSpacing" >
        <properties>
            <property name="spacing" value="1" />
        </properties>
    </rule>
```

violates  http://www.php-fig.org/psr/psr-2/#1-1-example

```
<?php
namespace Vendor\Package;

use FooInterface;
use BarClass as Bar;
use OtherVendor\OtherPackage\BazClass;

class Foo extends Bar implements FooInterface
{
    public function sampleFunction($a, $b = null)
    {
        if ($a === $b) {
            bar();
        } elseif ($a > $b) {
            $foo->bar($arg1);
        } else {
            BazClass::bar($arg2, $arg3);
        }
    }

    final public static function bar()
    {
        // method body
    }
}
```

@TomasVotruba @OndraM 

Do you guys have some sort of control for function line spacing in your projects? 

Thanks

This PR is open for more than a year :cry:

TomasVotruba:
I am not sure, but it think there is some hidden dependency on other sniff, that takes care of this issue.
Because I use whole "psr2" standard and I've never came across this.

techi602:
@TomasVotruba the problem with default PSR-2 ruleset is that it does not cover the spaces between methods at all
https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/PSR2/ruleset.xml

So you can have 5 spaces between methods without warning :/

I guess there are not many people having this issue but my junior developers are generating the random amount of spaces between methods in almost every PR :smile:

aik099:
[...]

PR #1041, opened 2016-06-21
Squiz.WhiteSpace.FunctionSpacing - don't check spacing after last block

smatyas:
This PR tries to solve the issue reported in #652 by skipping the check after the last block. I'm not sure if that's what everybody wants, but I would be totally happy with this solution. Or it could be a sniff parameter like the spacing itself.


gsherwood:
I missed this PR when it came in, but I'm closing now as the sniff will need properties to decide if it checks the first and last methods in a class rather than making a change for all users.

PR #1044, opened 2016-06-24
Travis: coverage added, old PHP versions dropped

TomasVotruba:
Also HHVM dropped, since PHP 7 is blazing fast.


TomasVotruba:
Wrong project.

PR #1045, opened 2016-06-24
gitignore: composer.lock that is already missing added

TomasVotruba:


gsherwood:
Why would there be a composer.lock file in the root dir of PHPCS? You don't composer install PHPCS.

TomasVotruba:
When you do `composer update`, there is.

gsherwood:
> Where is located on you machine then?

I don't use composer to work with PHPCS. Hence why I was asking.

So if you composer update a project that includes PHPCS, a composer.lock file gets dumped into the root of the PHPCS dir?

gsherwood:
Forgot to say: I've got no problem merging this. I just want to understand it a bit more before I do.

TomasVotruba:
When I install this package, e.g. to make PR, it has `comopser.lock` root in it.

Every time I commit sth, I have to skip `composer.lock` not to commit it. Best practice is to have it `.gitignore`, if not app type project.

aik099:
There is an alternative approach to commit `composer.lock` file instead.

TomasVotruba:
There is already nice explanation [why and when commit or not to commit composer.lock](http://stackoverflow.com/a/24247443/1348344).

aik099:
If you're working on a library development, then committing `composer.lock` is also beneficial for same reasons, why it is for projects.

For this particular library (phpcs) the `composer.lock` doesn't pose any value, because there will never be any dependencies listed in `composer.json` in the first place.

phansys:
>there will never be any dependencies listed in composer.json in the first place

What do you mean @aik099? [`composer.json`](https://github.com/squizlabs/PHP_CodeSniffer/blob/eb79bdec8c2238dc91c5008251e1365cba06c9f4/composer.json#L54-L62) has already some requirements defined, even dev ones:
```json
"require": {
    "php": ">=5.1.2",
    "ext-tokenizer": "*",
    "ext-xmlwriter": "*",
    "ext-simplexml": "*"
},
"require-dev": {
    "phpunit/phpunit": "~4.0"
}
```

aik099:
Since PHP_CodeSniffer can be installed through PEAR (where `composer.json` has no effect) nobody will specifying any dependencies in `require` part of `composer.json`.

TomasVotruba:
We are getting somewhere else. PEAR is not an issue, nor composer require dependencies.

> When I install this package, e.g. to make PR, it has `comopser.lock` in the root.

aik099:
I get that. For this library it makes sense to ignore `composer.lock` are you did in this PR.

TomasVotruba:
@aik099 I didn't understand your comments. This is much more clear, thank you.

TomasVotruba:
Thank you @gsherwood  & @aik099

PR #1047, opened 2016-06-27
Use `bool` and `int` instead of `boolean` and `integer`

oligriffiths:
Per https://wiki.php.net/rfc/scalar_type_hints_v5 PHP7 uses `bool` and `int` as aliases are removed. However https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer.php#L2079 causes `bool` to be `boolean` and `int` to be `integer`

Can this be changed to reflect the actual correct values rather than the aliases?

Thanks

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -2024,13 +2024,13 @@ class PHP_CodeSniffer
         } else {
             $lowerVarType = strtolower($varType);
             switch ($lowerVarType) {
-            case 'bool':
-                return 'boolean';
+            case 'boolean':
+                return 'bool';
             case 'double':
             case 'real':
                 return 'float';
-            case 'int':
-                return 'integer';
+            case 'integer':
+                return 'int';
             case 'array()':
                 return 'array';
             }//end switch
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -184,9 +184,9 @@ class PHP_CodeSniffer
      */
     public static $allowedTypes = array(
                                    'array',
-                                   'boolean',
+                                   'bool',
                                    'float',
-                                   'integer',
+                                   'int',
                                    'mixed',
                                    'object',
                                    'string',


aik099:
This is a BC break.

oligriffiths:
@aik099 Ok, how can we make this not BC, it seems like a bug tbh. Boolean and Integer have always been aliases...

oligriffiths:
Cool, yeah a config flag would be best. I'll re-work this implementation.

aik099:
@oligriffiths , Just search in code for `integer` and `boolean` to also fix relevant sniffs, that do the fixing.

@gsherwood , I'm wondering why type-related code is located inside `CodeSniffer` class and not in actual sniffs, that use it. That currently prevents from adding sniff-option instead of cli-option to change that behavior.

michaelbutler:
I also found it weird that the types list is a property on the CodeSniffer class... as a temporary fix in our own project we have overridden the sniff to accept bool and int explicitly. 
+1 for making it a configuration flag.

cyle:
config flag would be dope af

gsherwood:
> I'm wondering why type-related code is located inside CodeSniffer class and not in actual sniffs, that use it.

Because it was a hack nearly 10 years ago. No other real reason.

Doc comment sniffs are a hard thing to write because they need to share a lot of code and config, but few people agree on a complete standard for them. So developers tend to want a larger number of smaller checks so they can mix and match, which means sharing code and config between the sniffs is even more important. Without a central location to store this, you get a lot of repeated code. But I'm not sure I have the energy to write a completely new system for doc comment sniffing.

aik099:
Then I guess new cli-option is the right approach then. That will leave code in `CodeSniffer` class.

ofbeaton:
I just ran into this myself.
```
/**
 * @param integer
 */
public function myMethod(int $param1)
{
```

was very odd, since `integer` was reported as invalid for the type hint and `int` was reported as invalid for the docblock.

Since we're [forced](http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration) to use `int` in the type hint, perhaps the docblock could accept `int` or `integer`?

This is different than what is proposed above because it would not be a BC break. It would however be less strict and not enforce consistent use. 

I'm not against some kind of flag but I thought I'd offer this suggestion.

If nobody has cycles to work on this I volunteer.

aik099:
@ofbeaton , actually this is different problem, where long type name (e.g. `intereger`) is enforced in DocBlock, but PHP 7 scalar type hint uses short type name (`int`). Probably needs to be reported separately.

ofbeaton:
@aik099 Are you sure? From reading your response I believe we're talking about the same thing. I proposed being able to use `int` or `integer` in the DocBlock to represent the same type hint `int`. This could be instead of or in addition to a flag. I can already use `integer` in the DocBlock and `int` in the type hint, that works just fine from what I can tell. It just looks/feels odd.

aik099:
> @aik099 Are you sure? From reading your response I believe we're talking about the same thing.

Ah, you are right.

> I proposed being able to use int or integer in the DocBlock to represent the same type hint int. This could be instead of or in addition to a flag.

The https://github.com/squizlabs/PHP_CodeSniffer/pull/1047#issuecomment-228774998 seems the only viable option for now.

ctrl-f5:
Also ran into this issue, even when disbling the sniff, the phpcbf will still convert them to the long version, which we definitely do not want. especially since php7 will use the short notation.

If anything the short should become the default and the long version should be enabled with a config flag

ctrl-f5:
I have implemented this as a command line argument in PR #1195.
please review and comment.

virgofx:
Any update on this PR?

gsherwood:

This change will modify the behaviour of PHPCS too much for inclusion in the 2.x branch, so I will close this PR. Making these changes in a future 3.x release is still something I want to do, so feel free to open an issue or PR on the 3.x branch if this is still something you want to work on. If not, it is something I will eventually get around to looking at myself.

PR #1049, opened 2016-06-29
Version control reports can show notices when the report width is very small

aminin:
The command
```
phpcs --standard=PSR2 --encoding=utf-8 -n -p -s --report=gitblame src/
```
produces warning
```
PHP Warning:  str_repeat(): Second argument has to be greater than or equal to 0 in /home/anton/.composer/vendor/squizlabs/php_codesniffer/CodeSniffer/Reports/VersionControl.php on line 244
```
if the sniffed code contains a Squiz.WhiteSpace.ControlStructureSpacing.SpacingBeforeClose error


gsherwood:
There was a bigger problem with report width - the sources were not being truncated like they are in other reports. So I've fixed that, and a minor width issue, and added an indicator for which sources are fixable. Should be good now.

Thanks for reporting the problem.

Issue #682, opened 2015-08-15
Property data type in "@var" comments isn't auto-fixed


aik099:
I know, that for function arguments and return value we're automatically fixing data types (e.g. "bool" to "boolean", etc.).

However I've noticed, that same doesn't happen for class property data types:
```php
<?php

class Sample {

	/**
	 * Don't execute hooks, before event processing
	 *
	 * @var bool
	 */
	public $SkipBeforeHooks = false;

}
```

aik099:
Might be fixed in #1061.

gsherwood:
> Might be fixed in #1061.

Confirmed fixed.

PR #1061, opened 2016-07-07
fix for capitalization of var types

### LINKED ISSUE #1060: Add fixable error IncorrectVarType "Array vs array"
Request for a fixable error for `Expected "array" but found "Array" for @var tag in member variable comment from Squiz.Commenting.VariableComment.IncorrectVarType`

Possibly could be a minor enhancement to the existing error that just checks if the expected type is there but was incorrectly capitalized and allows for a fix. similar to the check/fix for int vs integer for function return types.

photodude:
need to include the lower cased var types in the switch to ensure a fix to capitalized var types
This should fix #1060


aik099:
Does this fix all places, where type can be seen (property, method param, method return value) or just property type?

aik099:
Also sniff test needs to be added, that would demonstrate how capitalized types are detected and fixed.

photodude:
@aik099 The unit test should be updated now, just waiting (and was waiting) for the tests to complete to make sure everything checks out.

photodude:
@aik099 I've added a "fixed" file but It doesn't seem to be checked. I know my first round didn't fix all of the errors so something should error if the fixed file was verified. I'm not sure what needed to be added/changed to use this new fixed file in the unit tests

aik099:
> ... I've added a "fixed" file but It doesn't seem to be checked. ...

You can easily check that. Just remove all contents from that file and try running unit tests. You should get large diff showing how things should be fixed.

photodude:
> Does this fix all places, where type can be seen (property, method param, method return value) or just property type?

@aik099 This should fix the capitalization of var types everywhere that the `suggestType()` function is used in the sniffs.

aik099:
> https://github.com/squizlabs/PHP_CodeSniffer/pull/1061#issuecomment-231184744

Then it's better to add unit tests to each sniff, that uses `suggestType` method, to guarantee that it works in all places.

photodude:
@aik099 removing all contents from the fixer file didn't cause errors, I don't think it's being checked.

photodude:
@aik099 I believe the only other sniff that uses `suggestType()` already [has in it's unit test an item that covers most of the changes here](https://github.com/photodude/PHP_CodeSniffer/blob/patch-3/CodeSniffer/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc#L219-L235).

Although doesn't have a fixed file in the unit tests.

gsherwood:
> removing all contents from the fixer file didn't cause errors nor did it produce a large diff, I don't think it's being checked.

The reason you aren't getting failures for the .fixed file is because you haven't changed the sniff to use `addFixableError` instead of `addError` for the specific error message you are wanting to fix. If you do that, you'll get a bunch of unit test failures in that file.

But also, there is no code in this PR to actually fix the error. The sniff would need to be modified to check the return value of `addFixableError` and then use the Fixer class to modify the file and replace the invalid var type with the valid one.

If you want to give this a go, please modify the PR and I'll review the changes again. If you aren't sure how to do it, just let me know and I'll do that bit for you. But the change would look something like this (note I haven't tested this):
```php
$varType       = $tokens[($foundVar + 2)]['content'];
$suggestedType = PHP_CodeSniffer::suggestType($varType);
if ($varType !== $suggestedType) {
    $error = 'Expected "%s" but found "%s" for @var tag in member variable comment';
    $data  = array(
              $suggestedType,
              $varType,
             );

    $fix = $phpcsFile->addFixableError($error, ($foundVar + 2), 'IncorrectVarType', $data);
    if ($fix === true) {
        $phpcsFile->fixer->replaceToken(($foundVar + 2), $suggestedType);
    }
}
```

photodude:
@aik099 I'm working on a system that I don't have set up to run tests locally which is why I temporarily made that change.

aik099:
> https://github.com/squizlabs/PHP_CodeSniffer/pull/1061#issuecomment-231359935

No much setup needed really. Usual one:

1. clone
2. `composer install`
3. `phpunit `

photodude:
This specific system doesn't have php or anything else setup to test locally.

photodude:
@gsherwood this should be ready for consideration

aik099:
@photodude , it seems, that you've implemented what I've reported in #682.

gsherwood:
Sorry it took a while to get back to this. It works well now. Thanks a lot.

photodude:
You are welcome, thank you.

PR #1064, opened 2016-07-11
Read stdin input from the dup()’ed non-blocking handle

addaleax:
Read stdin from the handle opened via `fopen('php://stdin')`, not the original `STDIN` handle which may not have been affected by the `stream_set_blocking()` call.

I’d look for confirmation from others but I’m pretty sure this fixes https://github.com/squizlabs/PHP_CodeSniffer/issues/993.

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -399,7 +399,7 @@ class PHP_CodeSniffer_CLI
         $handle = fopen('php://stdin', 'r');
         if (stream_set_blocking($handle, false) === true) {
             $fileContents = '';
-            while (($line = fgets(STDIN)) !== false) {
+            while (($line = fgets($handle)) !== false) {
                 $fileContents .= $line;
                 usleep(10);
             }


gsherwood:
I've put out a call for testers on the linked issue because I can't replicate any of the problems myself. Still, it looks like an obvious correction to the code, but I'm just not sure if I can call issue #993 fixed with it or not.

addaleax:
That’s perfectly understandable; if you know your way around Node.js a bit, you might be able to use the test case provided in nodejs/node#7652 to reproduce it (I *think* it’s because Node.js spawns child processes with `socketpair()`s instead of `pipe()`s which then behave differently when `dup()`ing them… but that’s just a guess).

masterzen:
Unfortunately that's something I already tested and this doesn't fix the issue for me.

gsherwood:
So it sounds like this fix doesn't fix the other bug report, but it obviously fixes your problem and it's an obvious correction, so thanks for the PR.

PR #1065, opened 2016-07-12
Added a possibility to skip the file validation by cleaning up listeners.

ruslanbaydan:
Added the additional condition to make possible skipping the validation if in a sniff would be cleaned up the listeners and tokens lists.

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -532,6 +532,7 @@ class PHP_CodeSniffer_File
                 if (isset($this->_ignoredListeners[$listenerData['class']]) === true
                     || (isset($listenerIgnoreTo[$listenerData['class']]) === true
                     && $listenerIgnoreTo[$listenerData['class']] > $stackPtr)
+                    || empty($this->_listeners) === true
                 ) {
                     // This sniff is ignoring past this token, or the whole file.
                     continue;


gsherwood:
I'm not sure why the $this->_listeners array would ever be empty when the foreach loop is using it. What exactly is the use case that you need this change for?

ruslanbaydan:
We call PHP_CodeSniffer_File::cleanUp method in our custom Sniff class to prevent the following validation for mixed html and php files, because we would like to skip the syntax validation for our view classes which have .php extension.

gsherwood:
Thanks for providing more info. I'm still not 100% sure what you are trying to do, but calling cleanUp probably isn't the best way to do it.

The only reason to call cleanUp is so that one sniff can stop all others sniffs from processing the current file. But the problem there is that you can't be sure which sniff is going to be run first, so you can't be sure your custom sniff will get to call cleanUp before other sniffs are executed. While each run of PHPCS might be producing consistent results now, that might not always be the case. The order of sniffs shouldn't be important during a run.

Maybe you'd be better off just ignoring your view classes using the ruleset.xml file. You can add ignore rules for specific sniffs, as long as you are able to exclude your view files using a regular expression.

Here is some sample config from the annotated ruleset: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml
```xml
 <!--
    You can hard-code ignore patterns for specific sniffs,
    a feature not available on the command line. Please note that
    all sniff-specific ignore patterns are checked using absolute paths.

    The code here will hide all messages from the Squiz DoubleQuoteUsage
    sniff for files that match either of the two exclude patterns.
 -->
    <exclude-pattern>*/tests/*</exclude-pattern>
    <exclude-pattern>*/data/*</exclude-pattern>
 </rule>
```

If the problem is that you can't tell if a .php file is a view file without first looking at the content, then that seems to be a fairly big problem. Hopefully that's not the case or else I imagine finding the right file to edit can be difficult.

ruslanbaydan:
Thank you for the detailed answer.

We solved the priority problem with use the Sniff folder prefix starts from Aaa.
Unfortunately we can't define the particular directories list as a exclude-pattern, because they are vary from project to project and we are developing pre-commit validation tool to use from the projects.
So we define if the content of a file is an our view and skip the following validation for it.

gsherwood:
> We solved the priority problem with use the Sniff folder prefix starts from Aaa.

You shouldn't really rely on this if possible. The order of sniffs is not documented and could change in the future, breaking your entire system.

> Unfortunately we can't define the particular directories list as a exclude-pattern, because they are vary from project to project

If projects have different settings, can they just use different standards, or include a phpcs.xml file in the root and use that? They can all share the same sniffs, but they can define a different `--ignore` command line argument.

If none of that works, you're probably going to need to a proper custom filtering system, which isn't coming until version 3.0. In the 3.0 branch right now, you can create a custom filter class so you get to decide if a file is processed or not. The feature description is:
```
  - Added support for custom filtering of the file list
    -- use --filter=/path/to/filter.php to filter the list
    -- extend \PHP_CodeSniffer\Filters\Filter to also get the basic PHPCS ext/path filtering
    -- extend \PHP_CodeSniffer\Filters\ExactMatch to get the basic filtering + ability to use blacklists and whitelists
    -- included filter \PHP_CodeSniffer\Filters\GitModified is a good example of an ExactMatch filter
    -- use --filter=gitmodified to only check modified and added files in a git repo
```

A good example is one of the built-in filters that limits checking to files added/modified in git: https://github.com/squizlabs/PHP_CodeSniffer/blob/3.0/src/Filters/GitModified.php

You would probably do something similar, but use a blacklist instead of a whitelist, or do something totally custom. So while PHPCS is figuring out which files to process, you could be finding all the view files in a project and telling PHPCS to ignore them. Then PHPCS wont spend time tokenzing them at all, or running any sniffs.

But there is nothing like this in the 2.x version, and it's too big of a feature to try and add there.

If I merge this in, or provide some sort of official way of ignoring a file from within a sniff, I'm basically saying that this is now a supported way of doing custom filtering, which I don't really want to do given I'm adding the filtering feature to 3.0. So I have to think about this a bit more. Maybe I'll add it and then immediately document it's removal for 3.0.

ruslanbaydan:
> You shouldn't really rely on this if possible. The order of sniffs is not documented and could change in the future, breaking your entire system.
You are right it is not a best solution to rely on it. Just a small hack to temporary avoid the validation for them until we implement Sniffs to perform the custom validation of our views. We can use it from a custom tag with the one line of extra validation and then return to the original repository.

> If projects have different settings, can they just use different standards, or include a phpcs.xml file in the root and use that?
Our goal is to use the same standards for all of the projects. We are going to adapt the tool to support all of them.

Regarding filters mechanism, it is very good improvement and should be useful for us at this point. But we would not like to ignore the files at all, as I mentioned, it's temporary, so we can go with the quick solution.

Thank you for your recommendation.

gsherwood:
> We can use it from a custom tag with the one line of extra validation and then return to the original repository.

If it is just temporarily, then that sounds like the best idea. Thanks for suggesting that solution. I'll close this now, but please let me know if you have more problems and need it revisited.

PR #1067, opened 2016-07-14
Add fixer to ForbiddenFunctionsSniff.

nkovacs:


aik099:
Not sure if that is possible considering, that alternative function might have different arguments or their order.

nkovacs:
You're right, though with the defaults in Squiz_Sniffs_PHP_ForbiddenFunctionsSniff and this one that's not an issue, and it's a shame phpcbf can't fix those automatically.

gsherwood:
A couple of people have asked for this, but it's not something I'm going to add.

This sniff is designed to be extended, or used with a custom list of functions in a ruleset, so you can't rely on it only ever fixing what is shown in the core code. I consider changing function calls as too risky for a core sniff. Perfectly fine for custom sniffs, but not this specific one.

Thanks for taking the time to submit this though. You might want to consider using this sniff in a custom standard so it does the auto-fixing for you.

Issue #1063, opened 2016-07-10
[Feature request] Sniff for alternative opening tags ?


jrfnl:
The `Generic.PHP.DisallowShortOpenTag` sniff checks for PHP short open tags, however there are also some alternative syntaxes which have been removed in PHP7:
```php
<% echo 'content'; %>
<%= 'content'; %>
<script language="php">
echo 'content';
</script>
```
Ref: http://php.net/manual/en/language.basic-syntax.phptags.php

Would there be interest in a sniff for this ?

And if so - should this be added to the `Generic.PHP.DisallowShortOpenTag` sniff or should a separate sniff be created ?

PR #1069, opened 2016-07-15
Make the `Generic.PHP.DisallowShortOpenTag` sniff fixable.

jrfnl:
This PR makes the `Generic.PHP.DisallowShortOpenTag` sniff fixable.

The `myconfig.ini` file and the adjustment to the `travis` script are needed for the fixer unit tests to pass as when `short_open_tag` is set to `off`, the fixer won't fix it correctly and the tests will fail ;-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ php:
   - 7.0
   - hhvm
 
+before_script:
+  - if [ $TRAVIS_PHP_VERSION != "hhvm" ]; then phpenv config-add myconfig.ini; fi
+
 script:
   - if [ $TRAVIS_PHP_VERSION != "hhvm" ]; then php scripts/phpcs --config-set php_path php; fi
   - phpunit -d date.timezone=Australia/Sydney tests/AllTests.php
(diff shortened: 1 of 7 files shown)


jrfnl:
I've got a similar sniff to this - fixable as well - ready for alternative PHP open tags as I suggested in #1063.

That one will also need a `myconfig.ini` file.

Both branches pass the travis unit tests if run individually with the `myconfig.ini` file included, but when I combine them I keep getting a failure and haven't figured out how to fix that one.
https://travis-ci.org/jrfnl/PHP_CodeSniffer/builds/144907266

In an ideal situation I should have two different `.fixer` files for this PR - one for when `short_open_tag = On` and one for when it is `Off`.
And three for the Alternative PHP tags sniff to test different combinations of `short_open_tag` and `asp_tags` set to `On/Off`.

The `config.ini` file is the only way I could think of getting the fixer test to pass at this moment, but for some strange reason that's not having the desired effect when I combine the fixes.

Any suggestions on how to deal with this ?

jrfnl:
Update: Found a working solution & sending in the sister-PR now.

gsherwood:
The only problem with this change is that unless I enable short_open_tags on my local installs of PHP, I'm going to always have failing unit tests because the `.fixed` file assumes the setting is enabled.

I haven't merged because I haven't found a way around this yet. I could skip the test if the setting is disabled, but then the actual checking features are only being checked by travis, so I might need to break this up into multiple test files and find a way to selectively skip some of them.

gsherwood:
I've reorganised those test files and added a small feature to let sniff unit tests decide what files they want tested.

Thanks for submitting this patch, especially with that myconfig.ini file.

jrfnl:
Sounds like a good solution for the unit tests. I'll have a look in more detail at your commit and rebase & try to apply the same to the sister-pr #1084.

PR #1070, opened 2016-07-15
Add sniff for class declaration to `generic`

### LINKED ISSUE #1068: [Feature request] Sniff for class opening braces on same line ?
Would there be any interest in having a sniff for class opening braces on the same line as the class statement ?
I imagine it could be added to `Generic` ?

All currently existing sniffs for class opening structures I could find check for the brace on a new line AFAICS.

I'd be happy to send in a PR.

jrfnl:
In contrast to similar existing sniffs in other rulesets, this sniff checks that the brace of the class declaration is on the same line as the class statement.

Fixes #1068


gsherwood:
I think the name of this new sniff isn't quite right. Because this lives in the Generic standard, and is designed to just check the opening brace position instead of several declaration rules, I'd suggest calling it OpeningBraceSameLine.

It would eventually be worth adding an OpeningBraceNewLine into the same directory and then using that in the standards that require this, allowing them to remove some of their checks. But that's a different job.

What do you think about the sniff name?

jrfnl:
Makes sense to me. Adjusted the file names (+ added the new files to the package.xml file).

jrfnl:
Also added a new unit test and minor code fix to allow for multi-line class declarations.

gsherwood:
Another great PR. Thanks a lot.

jrfnl:
You're welcome.

PR #1073, opened 2016-07-16
Add a sniff to detect the usage of the backtick operator.

jrfnl:
Working on a project now which wants to add a sniff like this as part of their security-sniffs and figured it would be re-usable for others as well.

### DIFF
--- a/CodeSniffer/Standards/Generic/Tests/PHP/BacktickOperatorUnitTest.php
+++ b/CodeSniffer/Standards/Generic/Tests/PHP/BacktickOperatorUnitTest.php
@@ -1,0 +1,85 @@
+<?php
+/**
+ * Unit test class for the BacktickOperator sniff.
+ *
+ * PHP version 5
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @author    Marc McIntyre <mmcintyre@squiz.net>
+ * @copyright 2006-2014 Squiz Pty Ltd (ABN 77 084 670 600)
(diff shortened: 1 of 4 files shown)


CodeSniffer/Standards/Generic/Tests/PHP/BacktickOperatorUnitTest.php
```diff
@@ -0,0 +1,89 @@
...
+     * The key of the array should represent the line number and the value
+     * should represent the number of errors that should occur on that line.
+     *
+     * @return array<int, int>
+     */
+    public function getErrorList()
+    {
+        $errors = array();
+
+        if ($this->isSafeModeEnabled() === false) {
```
aik099: Instead you can:

1. also expect errors
2. add `shouldSkipTest` function that would call `$this->isSafeModeEnabled() === false`
3. in fact you can inline the `isSafeModeEnabled` function into `shouldSkipTest`

jrfnl: Good point - adjusting now.

jrfnl: Originally I intended to check for the backtick character in other tokens when safe mode was enabled and throw a warning, but decided against it. This code was left behind from that.

gsherwood:
I ended up making the error message say "forbidden" instead of "strongly discouraged" because it was an error instead of a warning, and it also matches other error messages. But the message and type (error/warning) can be overridden in a ruleset.xml file, so I don't think that should be a problem.

Thanks for the new sniff.

jrfnl:
@gsherwood That's absolutely fine by me and shouldn't be an issue for the depending projects (in my case).

PR #1074, opened 2016-07-16
Make the LongConditionClosingCommentSniff more flexible.

jrfnl:
* Allow for setting a different lineLimit from the config file.
* Allow for a different comment format to be set from the config file.

Also: minor spelling corrections.

Includes unit tests.
Original test files have been renamed to prevent unit tests failing because of property value bleed through.

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/Commenting/LongConditionClosingCommentSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/Commenting/LongConditionClosingCommentSniff.php
@@ -59,7 +59,16 @@ class Squiz_Sniffs_Commenting_LongConditionClosingCommentSniff implements PHP_Co
      *
      * @var int
      */
-    protected $lineLimit = 20;
+    public $lineLimit = 20;
+
+    /**
+     * The format the end comment should be in.
+     *
+     * The placeholder %s will be replaced with the type of condition opener.
+     *
+     * @var string
+     */
+    public $commentFormat = '//end %s';
 
 
     /**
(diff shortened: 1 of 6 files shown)


gsherwood:
I like the feature, but I don't know why you needed different test files. I'd prefer a single one.

You have already figured out how to change sniff settings, so I'm not sure what bleed through you would have been getting. You would presumably put a couple of test cases in after the current tests to check that the sniff settings were working, and change the settings using comments.

Can you give me more information as to why that doesn't work? Maybe something else needs fixing.

jrfnl:
> I don't know why you needed different test files. I'd prefer a single one.

There is no problem having only one test file. It's just that the original test files don't have a `fixed` version and to have everything in one file, that would need to be added. To me, that was outside the scope of this PR. So having a separate test file which *does* have a fixed version for the tests contained in it made sense.

When someone gets round to creating a `fixed` version for the original test files, the files can be merged together.

> You have already figured out how to change sniff settings, so I'm not sure what bleed through you would have been getting.

The bleed through only happens when the original test files run after the new one - which won't happen now I've renamed them -.

The issue is that I can pass string variables using a `@codingStandardsChangeSetting` comment, but the string variable gets cut off at the first space it encounters. So resetting the `commentFormat` to `//end %s` using a `@codingStandardsChangeSetting` comment doesn't work as it will result in `//end`.

gsherwood:
> The issue is that I can pass string variables using a @codingStandardsChangeSetting comment, but the string variable gets cut off at the first space it encounters. So resetting the commentFormat to //end %s using a @codingStandardsChangeSetting comment doesn't work as it will result in //end.

Thanks, exactly what I needed to know. I'll need to find a way to supporting this then.

jrfnl:
@gsherwood I did look into fixing that and found the `explode(' ', ... )` which causes it, but was afraid I'd affect other things if I changed it. (I'm not familiar enough with the underlying code yet.)

gsherwood:
I've made a change to that code to allow @codingStandardsChangeSetting comments to contain spaces in the final value part. Any chance you could try that out and see if you can use a single test file now?

I can merge this in and just make the changes myself, but figured it would be best to ask if you wanted to change the PR first.

jrfnl:
@gsherwood All done. This did mean I had to create the fixed version for the other unit tests as well.

gsherwood:
> This did mean I had to create the fixed version for the other unit tests as well.

I didn't think of that. Thanks for doing it though. I appreciate it.

gsherwood:
Thanks again for a good feature and for all that unit test work.

jrfnl:
You're welcome! Thanks for merging this.

PR #1076, opened 2016-07-17
Fix bug where functions in mixed/uppercase would not be recognized.

jrfnl:
The `Generic.PHP.ForbiddenFunctions` sniff can easily be extended to also sniff for other functions.

However, there is a bug which prevents functions in mixed- or uppercase from being recognized.

As functionnames in PHP are case-insensitive both for userland functions as well as build-in functions, the function should always do a case-insensitive function name check.

In the current code the comparison was not consistently done in a case-insensitive manner. This PR fixes that.

No unit tests are provided with the PR as I can't seem to influence the `$forbiddenFunctions` array just and only for the unit test without actually extending the class.

A real world example of the bug can be found here:
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/WPCS/feature/fix-bug-in-discouraged-functions/WordPress/Sniffs/PHP/DiscouragedFunctionsSniff.php#L64
With the associated unit test failing: https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/145281345

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/PHP/ForbiddenFunctionsSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/PHP/ForbiddenFunctionsSniff.php
@@ -191,7 +191,7 @@ class Generic_Sniffs_PHP_ForbiddenFunctionsSniff implements PHP_CodeSniffer_Snif
             }
         }//end if
 
-        $this->addError($phpcsFile, $stackPtr, $function, $pattern);
+        $this->addError($phpcsFile, $stackPtr, $tokens[$stackPtr]['content'], $pattern);
 
     }//end process()
 
@@ -220,7 +220,7 @@ class Generic_Sniffs_PHP_ForbiddenFunctionsSniff implements PHP_CodeSniffer_Snif
         }
 
         if ($pattern === null) {
-            $pattern = $function;
+            $pattern = strtolower($function);
         }
 
         if ($this->forbiddenFunctions[$pattern] !== null
--- a/CodeSniffer/Standards/Generic/Sniffs/PHP/ForbiddenFunctionsSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/PHP/ForbiddenFunctionsSniff.php
@@ -102,6 +102,9 @@ class Generic_Sniffs_PHP_ForbiddenFunctionsSniff implements PHP_CodeSniffer_Snif
             }
         }
 
+        $this->forbiddenFunctionNames = array_map('strtolower', $this->forbiddenFunctionNames);
+        $this->forbiddenFunctions     = array_combine($this->forbiddenFunctionNames, $this->forbiddenFunctions);
+
         return array_unique($register);
 
     }//end register()


gsherwood:
From what I can tell, the problem you are having is:

A sniff has registered a forbidden function using mixed case. So something like getHTML(). But the sniff assumes you've put everything in as lowercase because it strtolower()'s all the function names it finds before comparing them to the array.

So what this PR does is change all the registered function names to lowercase so the sniff developer doesn't have to.

Is that correct?

If so, why not just change the function name in your sniff so it is lowercase? This PR doesn't make the error message any more correct (the found function name is always lowercase there) so it seems like this just adds unnecessary processing time, allbeit a tiny amount.

jrfnl:
> But the sniff assumes you've put everything in as lowercase because it strtolower()'s all the function names it finds before comparing them to the array.

Yes - basically the sniff just `strtolower()`s one side of the equation, not the other side. That seems like a clear bug to me.

It doesn't say anywhere that function names for the ForbiddenFunctions sniff have to be provided in lowercase, so providing them as they are defined is not a strange 'mistake' to make and it in actual fact makes the extended sniff more self-descriptive.

> If so, why not just change the function name in your sniff so it is lowercase? 

If I was the only one creating the sniffs, that would be an option as a work-around for the bug, but I'm not. And basically telling every sniff developer out there to add function names in lowercase even though it's counter-intuitive to how the function is defined seems quite roundabout for something which - to me - is clearly a bug.

gsherwood:
I don't think this is a clear bug at all. The sniff works just fine, but has a requirement that you list your forbidden functions in lowercase. It's not documented, and it should be.

> And basically telling every sniff developer out there to add function names in lowercase even though it's counter-intuitive to how the function is defined seems quite roundabout way for something which - to me - is clearly a bug.

If you are so strongly concerned with how the function is defined in code, then I think the real solution here would have to also address the error message. That message does not show the function name with either the case the developer used, or the case you defined in the forbidden list. It's always lowercase. Nobody except the sniff developer ever gets to see that mixed case you put into the array.

Without a change to the error message, this PR is just adding processing time to every run of PHPCS so that you don't have to lowercase 4 letters in a sniff file. That doesn't feel like a good trade-off to me because the end user isn't getting anything. If the error message used the correct case, then the end user would get a better error message and the change would provide benefit.

What do you think?

jrfnl:
> If you are so strongly concerned with how the function is defined in code, then I think the real solution here would have to also address the error message. 

I agree. However depending on the implementation - what is leading ? how the sniff developer defined it or the function name as was found in the code ? - the changes would be far more extensive which is why I held off on that.

If 'function name as was found in the code' would be ok to use, that change is trivial, so making it now.

gsherwood:
> If 'function name as was found in the code' would be ok to use, that change is trivial, so making it now.

I actually think that's the best one to use in this case. Then the developer sees what they wrote. Thanks for making the change.

jrfnl:
Excellent! Thanks for merging.

PR #1079, opened 2016-07-19
Allow lower case namespace in constant definitions

dschniepp:
The current version in combination with hhvm raises issues like https://github.com/zircote/swagger-php/pull/319. hhvm interprets the namespaces more strict than php.

This pull request will allow constant definitions with a lower case namespace like `define('bar\foo\baz\VALID_NAME_WITH_NAMESPACE', true)`. Currently it's not allowed and you need to change the code to `define('BAR\FOO\BAZ\VALID_NAME_WITH_NAMESPACE', true);` though the namespace is in lower case like `namespace bar\foo\baz;`.


gsherwood:
Nice change. Thanks.

PR #1084, opened 2016-07-22
Add sniff to disallow alternative PHP open tags.

### LINKED ISSUE #1063: [Feature request] Sniff for alternative opening tags ?

jrfnl:
This sniff checks  for the alternative PHP syntaxes - ASP tags and script tags - which have been removed in PHP7:
```php
<% echo 'content'; %>
<%= 'content'; %>
<script language="php">
echo 'content';
</script>
```
Ref: http://php.net/manual/en/language.basic-syntax.phptags.php

If PHPCS is run on PHP < 7, the `script` tags will be fixable.
If PHPCS is run on PHP < 7 + `asp_tags` is on, the ASP tags will be fixable.

Fixes #1063

Similar to #1069, the `myconfig.ini` file and the adjustment to the `travis` script are needed for the fixer unit tests to pass as when `asp_tags` is set to off, the fixer won't fix it correctly and the tests will fail ;-)

The `travis` and `myconfig.ini` changes will conflict with #1069, but depending on whichever PR will get merged first - if any get merged at all -, I will adjust the other one to remove the conflict.


CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowAlternativePHPTagsSniff.php
```diff
@@ -0,0 +1,253 @@
...
+class Generic_Sniffs_PHP_DisallowAlternativePHPTagsSniff implements PHP_CodeSniffer_Sniff
+{
+
+
+    /**
+     * Whether ASP tags are enabled or not.
+     *
+     * @var bool
+     */
+    private $_asp_tags = false;
```
aik099: Not sure if property name matches coding standard used in sniffs. Probably it should be `aspTags` instead.

CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowAlternativePHPTagsSniff.php
```diff
@@ -0,0 +1,253 @@
...
+
+
+    /**
+     * Returns an array of tokens this test wants to listen for.
+     *
+     * @return array
+     */
+    public function register()
+    {
+        if (version_compare(PHP_VERSION, '7.0.0alpha1', '<') === true) {
```
aik099: Why checking against `7.0.0alpha1` version and not just `7.0.0`. It's unlikely that somebody would use alpha version.

CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowAlternativePHPTagsSniff.php
jrfnl: Good point (though the phpcs code style check does not trigger on it - maybe a rule should be added to the project's own phpcs.xml file ?)
Anyway, fixed it. Let me know if you'd like me to squash the commits.

(am working with three different sniff projects at the moment using different code styles, so this was a left over from one of the others :sunglasses: )

CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowAlternativePHPTagsSniff.php
jrfnl: There is at least [one other place](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Generic/Tests/PHP/DisallowShortOpenTagUnitTest.php#L51) where versions like this are checked against pre-release version nrs, so I figured I'd better follow suit.

All the same, IMHO, it would be better to use the `PHP_VERSION_ID` constant as that way pre-release versions would be automatically included.
I looked into using that, but that constant was only [introduced in PHP 5.2.7](http://php.net/manual/en/reserved.constants.php#reserved.constants.core), while PHPCS has a minimum PHP requirement of 5.1.2.

It would be easy enough to define the constant in the PHPCS loader in case it's not available and I'd be happy to send in a PR to that end, but I figured that ought to be a separate issue/PR.

jrfnl: I've addressed the version check now in a separate PR #1088

jrfnl:
Rebased this branch & squashed the commits. I've also applied the same logic for the splitting of the test files as @gsherwood did in https://github.com/squizlabs/PHP_CodeSniffer/commit/3ce573638c061c7b9d7d126a33a5858522875056 for the ShortOpenTags sniff so the tests can still pass on a local environment with `asp_tags` set to `Off`.

gsherwood:
Finally got around to merging this in. Thanks a lot for the contribution.

jrfnl:
You're welcome & thanks for merging this. Hope others will find this useful too.

PR #1086, opened 2016-07-25
Additional development functions

grappler:
Add a few more functions that are normally used in development but should not be used in production.

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/PHP/DiscouragedFunctionsSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/PHP/DiscouragedFunctionsSniff.php
@@ -41,9 +41,12 @@ class Squiz_Sniffs_PHP_DiscouragedFunctionsSniff extends Generic_Sniffs_PHP_Forb
      * @var array(string => string|null)
      */
     public $forbiddenFunctions = array(
-                                  'error_log' => null,
-                                  'print_r'   => null,
-                                  'var_dump'  => null,
+                                  'error_log'             => null,
+                                  'print_r'               => null,
+                                  'var_dump'              => null,
+                                  'var_export'            => null,
+                                  'debug_backtrace'       => null,
+                                  'debug_print_backtrace' => null,
                                  );
 
     /**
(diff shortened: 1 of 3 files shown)


CodeSniffer/Standards/Squiz/Sniffs/PHP/DiscouragedFunctionsSniff.php
```diff
@@ -41,9 +41,12 @@ class Squiz_Sniffs_PHP_DiscouragedFunctionsSniff extends Generic_Sniffs_PHP_Forb
      * @var array(string => string|null)
      */
     public $forbiddenFunctions = array(
-                                  'error_log' => null,
-                                  'print_r'   => null,
-                                  'var_dump'  => null,
+                                  'error_log'             => null,
+                                  'print_r'               => null,
+                                  'var_dump'              => null,
+                                  'var_export'            => null,
```
aik099: These are used if you want to create custom error handler. So probably would end in error, when looking at Symfony's Components (VarDumper and Debug).

grappler: @aik099 The first three were already here before I created the PR. It should help prevent having debug code in production. I don't know much about Symfony but are you going to use those components in a production environment?

aik099: > The first three were already here before I created the PR.

I've added comment about functions you've added (last 3), not first 3 ones.

> I don't know much about Symfony but are you going to use those components in a production environment?

It is possible.

gsherwood:
I don't want to add any more functions to this list because it is in the Squiz standard and we do use those functions you've listed for custom errors handlers in pretty much all our products.

But you also don't need to use that sniff if you want to create your own forbidden function list. To do what you are after, add this to your ruleset.xml file:
```xml
<!-- Forbid some functions -->
  <properties>
    <property name="forbiddenFunctions" type="array" value="error_log=>null,print_r=>null,var_dump=>null,var_export=>null,debug_backtrace=>null,debug_print_backtrace=>null,trigger_error=>null,set_error_handler=>null"/>
  </properties>
</rule>
```
Is that good enough for what you want to do?

grappler:
Ok, thats understandable. Will add them to our own list.

gsherwood:
Let mw know if you have any problems.

PR #1088, opened 2016-07-26
Improve version comparisons by using PHP_VERSION_ID.

jrfnl:
Inspired by: https://github.com/squizlabs/PHP_CodeSniffer/pull/1084#discussion_r71835858

A PHP version number is build up as `{major}.{minor}.{release}{extra}`.
In 99.9% of cases where a version comparison is needed, the `{extra}` (`alpha/beta/RC`) information can/should be disregarded.
`PHP_VERSION` is a string which contains the complete version number, including `{extra}`.
`PHP_VERSION_ID` is an integer which contains the release version number *without* the `{extra}`

So using `PHP_VERSION_ID` allows to do accurate version comparisons without taking the `{extra}` into account *and* without an extra function call to `version_compare()`.

As `PHP_VERSION_ID` only became available in PHP 5.2.7, it needs to be defined in the application for older PHP versions.

Ref: http://php.net/manual/en/reserved.constants.php#reserved.constants.core

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -14,8 +14,15 @@
 
 error_reporting(E_ALL | E_STRICT);
 
+// Make sure version id constant is available.
+if (defined('PHP_VERSION_ID') === false) {
+    $version = explode('.', PHP_VERSION);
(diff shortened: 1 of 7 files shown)


jrfnl:
Rebased for merge conflict.

gsherwood:
Very good idea. Thanks.

PR #1090, opened 2016-07-26
Allow return type hints without space before colon in OpeningFunctionBraceKernighanRitchie sniff

VasekPurchart:
Currently if you you use return type hints with a space before colon ` : `, f.e.:

```php
<?php

function myFunction() : Something {
    return null;
}
```

everything works fine, but if you prefer not including the space before colon `: `, like this:

```php
<?php

function myFunction(): Something {
    return null;
}
```
then a `SpaceAfterBracket` error is reported.

It is also a problem when defining closures.

This also influences other sniffs, such as `PEAR.Functions.FunctionDeclaration` or inherently `Squiz.Functions.MultiLineFunctionDeclaration` (where I stumbled upon this).

This pull request should allow both syntaxes, see tests, I hope I didn't forget about something.

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/Functions/OpeningFunctionBraceKernighanRitchieSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Functions/OpeningFunctionBraceKernighanRitchieSniff.php
@@ -153,7 +153,7 @@ class Generic_Sniffs_Functions_OpeningFunctionBraceKernighanRitchieSniff impleme
             return;
         }
 
-        if ($tokens[($closeBracket + 1)]['code'] !== T_WHITESPACE) {
+        if ($tokens[($closeBracket + 1)]['code'] !== T_WHITESPACE && $tokens[($closeBracket + 1)]['code'] !== T_COLON) {
             $length = 0;
         } else if ($tokens[($closeBracket + 1)]['content'] === "\t") {
             $length = '\t';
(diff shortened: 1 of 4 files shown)


aik099:
This is a BC break.

gsherwood:
Yes, this is a BC break, but I think the change makes sense. The role of this sniff is not to enforce spacing for the return type syntax - just for the brace. The next release is 2.7.0, so it's already a minor breaking release for PHPCS, so this is a good time to make the change.

The reason it does this is because return types never existed, so it could just check for spacing after the closing parenthesis because that was the same as checking before the brace. But it can no longer make that assumption.

For this reason, I didn't merge the PR. I actually changed the error message and code to be specific that a single space is required before the brace. I changed the logic to check that location as well. With the PR, the colon was seen as a single whitespace, allowing the developer to put any number of spaces before the opening brace, which isn't what this sniff wants.

Thanks a lot for the PR, even though I went with a more breaking change.

VasekPurchart:
> The role of this sniff is not to enforce spacing for the return type syntax

Yes, that was exactly my way of thinking.

Yes, your change makes more sense, because it is more generic, thanks for looking into this :+1:

jgedarovich:
@gsherwood will there be a 2.7.0 release before 3.x? I see that the latest 2.x release is 2.6.2. Thanks.

gsherwood:
> will there be a 2.7.0 release before 3.x?

Yes. I'm sure there will be several new releases before 3.0 is stable.

PR #1093, opened 2016-07-29
Fix FunctionCallSignatureSniff

grongor:
The `findNext` method doesn't return token at the `$end` position. In that case we want the `$end` token to be returned (the close bracket) so there should be `$closeBracket + 1`. At least I think so - it solved my problem. I debugged it couple times and this seems to be the valid and intended `$end`.

The problem was with method `isMultiLineCall` - it didn't detect multi-line call correctly.
```
return [
    'tags'  => array_merge($tags, [
        'service'  => $this->validateService($service),
        'language' => $this->language->getCode(),
    ]),
    'level' => $level,
    'extra' => $extra,
];
```

This is clearly a multi-line call but the method failed to detect that because:

1. `$tags` is on the same line
2. `[ ... ]` doesn't end with comma so the while loop ends
3. Closing brace `)` is not returned as the result of the `findNext` method because the `$end` is not included (and it was set to the closing brace position)

To fix that I simply extended the `$end` by one more token (to include it in the search results).

### DIFF
--- a/CodeSniffer/Standards/PSR2/Sniffs/Methods/FunctionCallSignatureSniff.php
+++ b/CodeSniffer/Standards/PSR2/Sniffs/Methods/FunctionCallSignatureSniff.php
@@ -76,7 +76,7 @@ class PSR2_Sniffs_Methods_FunctionCallSignatureSniff extends PEAR_Sniffs_Functio
 
         // We've reached the last argument, so see if the next content
         // (should be the close bracket) is also on the same line.
-        $next = $phpcsFile->findNext(PHP_CodeSniffer_Tokens::$emptyTokens, ($end + 1), $closeBracket, true);
+        $next = $phpcsFile->findNext(PHP_CodeSniffer_Tokens::$emptyTokens, ($end + 1), $closeBracket + 1, true);
         if ($next !== false && $tokens[$next]['line'] !== $tokens[$end]['line']) {
             return true;
         }


aik099:
Please:

1. explain what problem were you having (code sample, standard used), not the way how you fixed it
2. add a test case that demonstrates what is fixed

Also please

grongor:
Sorry, I was yet without my morning coffee :) I added an explanation of the problem/fix. I wanted to add some test case but there are no tests for the snifffs - or am I wrong? Can you point me in the right direction? Thanks.

gsherwood:
By the PSR2 definition, that `array_merge()` call is *not* a multi-line function call.

Yes, I realise it goes over multiple lines. But you have to ignore any newlines that are within array definitions, closures, strings, inner function calls etc when calculating if it is multi-line or not.

PHP-FIG actually added errata just for this because people complained about the confusion so much (including myself): https://github.com/php-fig/fig-standards/pull/178

Phil actually blogged about it here: https://philsturgeon.uk/php/2013/10/15/psr2-v-codesniffer-psr2/

So this specific code sample is 100% valid PSR2 as far as I can tell. Does it looks right to you once you've considered the PSR2 errata?

grongor:
Ohh ... now I understand. Thank you for the explanation and I'm sorry for taking your time. I will try to go around that somehow in my standard :)

Have a nice day.

gsherwood:
> Thank you for the explanation

No problem. Glad it made sense.

Issue #1034, opened 2016-06-14
Squiz FunctionDeclarationArgumentSpacing gives incorrect error when first arg is a variadic


firegate666:
I have the following function

    function o(...$args) {
        $args_len = sizeof($args);
        ...
    }

which should be valid, but the code sniffers complains about it

     51 | ERROR | [x] Expected 1 space between type hint and argument
     |       |     "$args"; 0 found
     |       |     (Squiz.Functions.FunctionDeclarationArgumentSpacing.SpacingAfterHint)

Within the same file, there is a similar method that uses the variadic parameter as second parameter

    function p($text, ...$arguments) {
        $text = call_user_func_array('vsprintf', [$text, $arguments]);
        print $text;
        print PHP_EOL;
    }

no complains here

gsherwood:
Fixed now. Thanks for reporting.

firegate666:
Awesome, thx for fixing it.

PR #1096, opened 2016-08-01
Squiz FunctionDeclarationArgumentSpacing gives incorrect error/fix when variadic operator is followed by a space

grongor:
Before this patch code like:
```function myFunc(... $args) {}```
would become:
```function myFunc($args) {}```
... effectively breaking the code.

I wasn't sure if I also should handle the space removal (between T_ELLIPSIS and the parameter) so I left it like that but if you want I can adjust the code to handle that as well.


gsherwood:
This is a further fix for bug #1034

gsherwood:
Thanks for the bug report and fix, but I made a slightly different change to solve this problem. I changed where the sniff was looking for whitespace a bit further up because it was already trying to do that but failing when the variadic operator was followed by whitespace.

PR #1098, opened 2016-08-02
Add a new utility method to find implemented interface names.

jrfnl:
Based on `findExtendedClassName()`. Allows to retrieve an array with the names of the interfaces a class implements.

Credits go to @westonruter (with a minor fix by me).

This is a method used within the WPCS project, but could possibly be useful for others too.

[Edit] Example use case: check if a method is in an extended class or implements an interface and if so, exempt it from method name convention checks as when overloading a method from the parent/interface complying with the method name conventions may not be possible.

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -3715,4 +3715,56 @@ class PHP_CodeSniffer_File
     }//end findExtendedClassName()
 
 
+    /**
+     * Returns the name(s) of the interface(s) that the specified class implements.
+     *
+     * Returns FALSE on error or if there are no implemented interface names.
+     *
+     * @param int $stackPtr The stack position of the class.
+     *
+     * @return array|false
+     */
+    public function findImplementedInterfaceNames($stackPtr)
+    {
+        // Check for the existence of the token.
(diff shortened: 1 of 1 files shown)


gsherwood:
Seems like a helpful method. Thanks for the contribution.

I also added a couple of very basic tests for the method given it isn't being used by any included sniffs and so isn't being tested in any way yet.

jrfnl:
Thanks for that! I wasn't sure where or how to add the unit tests in this case, though I'd confirmed that it was working as intended through unit tests in the sniff in WPCS.

gsherwood:
> I wasn't sure where or how to add the unit tests in this case

Not a problem at all. I rarely write tests for those core methods directly because they are almost considered private to PHPCS and tested through sample code used by individual sniff tests. But I did for this one just because I wouldn't have a sniff to use if anyone ever did find a problem with it.

jrfnl:
FYI: The PHPCompatibility project now also uses this method for two sniffs (though they'd have to up their minimum PHPCS version to defer to the method in PHPCS, so using a copy of the method for now).

See https://github.com/wimg/PHPCompatibility/pull/144 for some examples using the method.

Issue #878, opened 2016-01-26
Support for group use declaration (PHP7)


PierreLvx:
With PHP 7:

> Classes, functions and constants being imported from the same namespace can now be grouped together in a single use statement.

[Source + example code](http://php.net/manual/fr/migration70.new-features.php#migration70.new-features.group-use-declarations).

Using the PSR2 Standard, the following happens (sample output):

```
----------------------------------------------------------------------
FOUND 2 ERRORS AND 1 WARNING AFFECTING 2 LINES
----------------------------------------------------------------------
1 | WARNING | [ ] A file should declare new symbols (classes,
|         |     functions, constants, etc.) and cause no other
|         |     side effects, or it should execute logic with
|         |     side effects, but should not do both. The first
|         |     symbol is defined on line 15 and the first side
|         |     effect is on line 13.
13 | ERROR   | [x] There must be one USE keyword per declaration
13 | ERROR   | [x] Closing brace must be on a line by itself
----------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

I can get rid of the `one USE keyword` error excluding `PSR2.Namespaces.UseDeclaration`. I can also work around the side effects warning excluding `PSR1.Files.SideEffects`, though not ideal.

However, I have no idea how work around the closing brace error, and I don't want to exclude that rule since I still want it to apply on the rest of the code.

Is there a clean way to work around this issue ?

gsherwood:
The side effects and brace errors are caused by PHPCS assigning the curly braces to the function keyword in `use function ...`, which is incorrect. Fixing that should fix both of those.

For the use declaration error, I don't think I can fix that. PSR2 is pretty clear about having one declaration per line, so they would need some errata to change that rule and allow for PHP7 syntax, if they want that to be allowed. But as you've said, you can ignore it in your ruleset.

gsherwood:
The issues are actually caused by the curly braces being assigned to the `use` keyword (or a `namespace` keyword if you have one up top. Fixing that issue will fix the side effects and brace placement errors.

gsherwood:
I've assigned the curly braces new tokens so there is no confusion with sniffs and the tokenizer. It also lets sniff detect this syntax and apply coding standards to it more easily.

PierreLvx:
Excellent, thanks :pray:

luispabon:
[...]

PR #1099, opened 2016-08-03
 Group use declarations are incorrectly fixed by the PSR2 standard

jasonmccreary:


jasonmccreary:
Per our discussion here https://github.com/squizlabs/PHP_CodeSniffer/issues/878, this is a first pass at a patch to *ungroup* grouped `use` statements when *fixing* for the PSR-2 coding standard.

While this passes a basic test, it would not work for *mixed* grouped `use` statements. For example:

```php
use some\ns {
    ClassA,
    const CON,
    function a_func
};
```

I'm not even sure such code is allowed in PHP 7. But I'm glad to adjust if so.

Also, the code could probably stand for some optimizations as I'm not as familiar with the API.

gsherwood:
I think it looks pretty good, but I've got a couple of suggestions for you to consider.

1. It appears that mixed use group syntax is valid as per the RFC (https://wiki.php.net/rfc/group_use_declarations) so the sniff should handle fixing that. It doesn't look too hard given your code though.

2. It would be good to have tests for grouped `use function` and `use const` statements as well as the mixed group. Ripping them straight from the RFC would give this sample code:
```php
use function foo\math\{ sin, cos, cosh };
use const foo\math\{ PI, E, GAMMA, GOLDEN_RATIO };
use foo\math\{ Math, const PI, function sin, function cos, function cosh };
```
Which should be converted to:
```php
use function foo\math\sin;
use function foo\math\cos;
use function foo\math\cosh;
use const foo\math\PI
use const foo\math\E;
use const foo\math\GAMMA;
use const foo\math\GOLDEN_RATIO;
use foo\math\Math
use const foo\math\PI;
use function foo\math\sin;
use function foo\math\cos;
use function foo\math\cosh;
```

3. During fixing, it might be easier to just construct the new use statement lines, add that content before the `use` keyword, then wipe everything from the old use statement. This also ensures every token inside the grouped use statement is modified, so other sniffs wont get a go to change things until the next fixing loop - making conflicts impossible.

My general comment is: you've done a really good job on one of the most complex fixers. Thanks for the PR.

CodeSniffer/Standards/PSR2/Sniffs/Namespaces/UseDeclarationSniff.php
```diff
@@ -68,14 +68,53 @@ public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
         }
 
         // Only one USE declaration allowed per statement.
-        $next = $phpcsFile->findNext(array(T_COMMA, T_SEMICOLON), ($stackPtr + 1));
-        if ($tokens[$next]['code'] === T_COMMA) {
+        $next = $phpcsFile->findNext(array(T_COMMA, T_SEMICOLON, T_OPEN_USE_GROUP), ($stackPtr + 1));
```
aik099: Is `T_OPEN_USE_GROUP` created also, when PHP 5 code is parsed OR PHP 5 is used to parse PHP 7 code?

gsherwood: Yes. It's a PHPCS-specific thing. PHP doesn't tokenise that code any differently, so PHPCS does it for all PHP versions.

jasonmccreary:
Thanks. I will add tests for the mixed group `use` statements.

I thought about regenerating the `use` statements. What concerned me was something like the following:

```php
use some\ns\{
    // use whatever from here
    whatever,
};
```

Thoughts?

jasonmccreary:
I have updated this to handle *mixed groups*.

I didn't rewrite it to rebuild the use statements as I felt that could get more complicated when attempting to account for things like comments. Nonetheless, open to suggestions as I'm sure this could be optimized.

gsherwood:
> I opted not to rebuild the use statements as I felt that may be more complicated when trying to account for things like comments.

I didn't consider that. You've absolutely picked the right way.

gsherwood:
This was a great PR. Thanks a lot.

jasonmccreary:
Anytime! I use PHPCS on many projects. Glad I could give back.

Issue #1103, opened 2016-08-05
Squiz.Commenting.FunctionComment type hint breaks on multi-version php support


ofbeaton:
I support PHP 5.6 and PHP 7.0 for my application, and I'm trying to use `Squiz.Commenting.FunctionComment` in my ruleset.

Unfortunately many of the type hints are not available for PHP 5.6. I use travis to make sure my application works, including phpcs, on php 5.6 and php 7.0.

The PHP 5.6 build works great but when it runs against php 7.0, I get type hint errors because my docblocks say a variable is `string` but I didn't provide the type hint in the function.

I looked at the code and I see:

```
} else if (version_compare(PHP_VERSION, '7.0.0') >= 0) {
```

this is clever but means that I can either make my code have syntax errors in 5.6, or style errors in 7.0.

I propose that that a new configuration setting be added to specify the version that phpcs should use when evaluating the code. If it is not provided, it should default to PHP_VERSION.

In this way I could lock my sniffs to php 5.6 enforcement until such a time I can drop php 5.6 and only support php 7.0, no matter what version of php is currently running phpcs.

gsherwood:
Can you not just run your build using the path to php 5.6? Like `/path/to/php5.6/php phpcs ...`

ofbeaton:
This would only work for the development environment, and require all developers to keep php 5.6 installed at all times. It would not work for having phpcs as part of continuous integration testing. 

I suppose it could be argued it does not belong there, but I feel that it's a great place to put it and make sure that all commits have indeed followed the ruleset. This turns travis into a similar service as https://styleci.io/

Here are some examples of my own public projects where I have historically used phpcs with travis-ci.
https://github.com/ofbeaton/db-ping/blob/master/.travis.yml (phpcs only)
https://github.com/ofbeaton/command-running/blob/master/.travis.yml (phpcs + phpunit)
https://github.com/ofbeaton/php-timeago/blob/master/.travis.yml (phpcs + phpunit)
https://github.com/ofbeaton/console-tester/blob/master/.travis.yml (phpcs + phpunit)

I don't want to suppress all type hints errors, after all I use the PHP 5.6 ones.

ofbeaton:
I would be remiss in not mentioning that another solution exists, to disable `phpcs` only in my php 7.0 using something like:
```
- bash -c 'if [ "$TRAVIS_PHP_VERSION" != "7.0" ]; then bin/phpcs; fi;'
```

[...]

PR #1104, opened 2016-08-05
add config `php_version` to override PHP_VERSION_ID in sniffs

ofbeaton:
Provides enhancement requested in #1103, allows your phpcs to target a specific php version since some sniffs (three currently) provide different checks based on php version.

Allows user to specify `--platform=PHP_VERSION_ID` on command line, or `<platform>` in `phpcs.xml`. Can be used for both phpcs and phpcbf. Modified the Squiz Function Comment to use it if present.

Defaults to current PHP_VERSION_ID.

Documentation updates cannot be made in a PR.

This is similar to the composer config setting `platform` https://getcomposer.org/doc/06-config.md#platform but for phpcs.

I almost exclusively live on an intranet with SVN, so this is my first PR, sorry if I did anything wrong.

Ran codestyle and phpunit before submitting. Not sure if I need to also provide a test.

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -238,6 +238,11 @@ class Squiz_Sniffs_Commenting_FunctionCommentSniff extends PEAR_Sniffs_Commentin
      */
     protected function processParams(PHP_CodeSniffer_File $phpcsFile, $stackPtr, $commentStart)
     {
+        $phpVersion = PHP_CodeSniffer::getConfigData('php_version');
+        if ($phpVersion === null) {
+            $phpVersion = PHP_VERSION_ID;
+        }
+
         $tokens = $phpcsFile->getTokens();
 
         $params  = array();
@@ -382,7 +387,7 @@ class Squiz_Sniffs_Commenting_FunctionCommentSniff extends PEAR_Sniffs_Commentin
                         $suggestedTypeHint = 'callable';
                     } else if (in_array($typeName, PHP_CodeSniffer::$allowedTypes) === false) {
                         $suggestedTypeHint = $suggestedName;
-                    } else if (PHP_VERSION_ID >= 70000) {
+                    } else if ($phpVersion >= 70000) {
                         if ($typeName === 'string') {
                             $suggestedTypeHint = 'string';
                         } else if ($typeName === 'int' || $typeName === 'integer') {
--- a/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -32,6 +32,13 @@ if (class_exists('PEAR_Sniffs_Commenting_FunctionCommentSniff', true) === false)
 class Squiz_Sniffs_Commenting_FunctionCommentSniff extends PEAR_Sniffs_Commenting_FunctionCommentSniff
 {
 
+    /**
+     * The current PHP version.
+     *
+     * @var integer
+     */
+    private $_phpVersion = null;
+
 
     /**
      * Process the return comment of this function comment.
@@ -238,9 +245,11 @@ class Squiz_Sniffs_Commenting_FunctionCommentSniff extends PEAR_Sniffs_Commentin
      */
     protected function processParams(PHP_CodeSniffer_File $phpcsFile, $stackPtr, $commentStart)
     {
-        $phpVersion = PHP_CodeSniffer::getConfigData('php_version');
-        if ($phpVersion === null) {
-            $phpVersion = PHP_VERSION_ID;
+        if ($this->_phpVersion === null) {
+            $this->_phpVersion = PHP_CodeSniffer::getConfigData('php_version');
+            if ($this->_phpVersion === null) {
+                $this->_phpVersion = PHP_VERSION_ID;
+            }
         }
 
         $tokens = $phpcsFile->getTokens();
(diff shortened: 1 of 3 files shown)


CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -238,6 +246,22 @@ protected function processThrows(PHP_CodeSniffer_File $phpcsFile, $stackPtr, $co
      */
     protected function processParams(PHP_CodeSniffer_File $phpcsFile, $stackPtr, $commentStart)
     {
+        if ($this->_platform === null) {
```
aik099: All of this can be moved to sniff constructor to avoid checking each time if platform was set, because neither PHP version nor CLI values change during single run.

CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -382,7 +406,7 @@ protected function processParams(PHP_CodeSniffer_File $phpcsFile, $stackPtr, $co
                         $suggestedTypeHint = 'callable';
                     } else if (in_array($typeName, PHP_CodeSniffer::$allowedTypes) === false) {
                         $suggestedTypeHint = $suggestedName;
-                    } else if (PHP_VERSION_ID >= 70000) {
+                    } else if ($this->_platform >= 70000) {
```
aik099: Similar checks exist in other sniffs. Maybe they needs to be addressed as well.

CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
ofbeaton: other sniffs do this at the top of `process()` but this sniff does not have such a method. Do you still recommend I move it into the constructor, or should I somehow add a `process()` method to this sniff?

CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
ofbeaton: You're right, it is used elsewhere! I'm updating the code for those as well now. Thank you for catching this.

CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
ofbeaton: upon more careful review, $phpcsFile is not available in a constructor. This is why similar features like `tabWidth` do it like I have done here. If there is indeed a way, I'd be happy to use it.

CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
ofbeaton: I've updated the pull request with the only other place I can add the check, PSR1 ClassDeclarationSniff

ofbeaton:
I've updated my PR with another sniff.

I wasn't able to use the platform in certain sniffs. For example `Generic_Sniffs_PHP_DisallowAlternativePHPTagsSniff` uses `PHP_VERSION_ID` in the `register()` method where `$phpcsFile` is unavailable.

Similarly, it is not available in unit tests, which makes sense since I'm not providing the platform on the command line to phpunit.

I'm also not convinced using platform in `PSR1_Sniffs_Classes_ClassDeclarationSniff` is beneficial, given the lowest version phpcs itself will run outclasses the PHP_VERSION_ID requirement here. 

However maybe this is an example of how `platform` is useful, to target a version prior to what phpcs itself requires, which could be important in v3.* and the version bumb in the future.

I consider this PR ready to merge if there are no further comments.

gsherwood:
Unless the PHPCS core is using this fake PHP version ID (it isn't, and I can't see it ever needing to) then I don't think this should be a dedicated command line value. Having something like this for a single sniff (or even a couple) is not how CLI arguments are used.

A better way to do this would be to use a config var. Here is an example sniff that checks for syntax errors: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Generic/Sniffs/PHP/SyntaxSniff.php#L57

By default, it uses the PHP binary that PHPCS was run with. But if you want to use a different one, you specify the path in the config var and the sniff will use that instead.

Config value can either be set permanently, or can be set for a single run using `--runtime-set`: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#setting-configuration-options

You can also put them in ruleset.xml files:
```xml
<!--
    You can hard-code config values used by sniffs directly
    into your custom standard. Note that this does not work
    for config values that override command line arguments,
    such as show_warnings and report_format.

    The following tag is equivalent to the command line argument:
    --runtime-set zend_ca_path /path/to/ZendCodeAnalyzer
 -->
 <config name="zend_ca_path" value="/path/to/ZendCodeAnalyzer"/>
```
I would suggest a config var called `php_version` that any sniff could use. Then I'd get these sniffs using that value. If it is empty, I'd get them to use the built-in constant value.

Sound like an ok solution?

ofbeaton:
Made the suggested changes, struggled with git for an hour to undo everything I had previously done, finally got it, redid the suggested changes, retested, and now resubmitted!

The config method also allowed me to add `Generic_Sniffs_PHP_DisallowAlternativePHPTagsSniff`, so this was a better solution overall. Thank you for the input and guidance @gsherwood.

Should be ready to be merged.

ofbeaton:
fyi, I didn't make a $this->phpVersion to set like you did with tabWidth, since you didn't in the example you gave. If you'd like me to do that, I'd be happy to modify the code and check it in.

aik099:
Looks compact and easily readable. Maybe test suite needs to be updated to specify that config option.

gsherwood:
First, thanks for taking the time to change this PR. I'm glad my input helped find a solution we're both happy with. Nice work getting all the stuff backed out from git as well :)

> fyi, I didn't make a $this->_phpVersion to set like you did with tabWidth, since you didn't in the example you gave. If you'd like me to do that, I'd be happy to modify the code and check it in.

I think storing the result in a private member var would be a good idea because of how often they get called.

It would probably be a good idea for the syntax sniff I linked as well, given it would be called for every file being checked. All those function calls do add up. I'd better go do that :)

ofbeaton:
Done. If you think I should also modify the tests to respect the phpVersion config, let me know, but I have a feeling they should remain independent.

gsherwood:
That was quick :) And yes, I think the tests should remain as they are as well.

Thanks a lot. I've got a meeting soon, but I'll merge this in later today.

gsherwood:
Merged now. Thanks a lot for the contribution.

ofbeaton:
You're welcome, and thanks for merging my PR!

Issue #811, opened 2015-12-08
Add option to force all errors to display, even within @codingStandardsIgnore sections


davidscotson:
If you're running CodeSniffer against code written by other people, you might want to get a full report of potential issues without giving them the ability to suppress any errors by using // @codingStandardsIgnore comments to skip files, or lines.

Reading through the docs there doesn't seem to be an override setting that allow this level of reporting.

gsherwood:
I've added an `--ignore-annotation` CLI argument to ignore all `@codingStandard..`. code comments. You can also set this value in a ruleset.xml file if you don't want to allow these annotations in your coding standard.

PR #1105, opened 2016-08-06
Add --force-check to display all errors without ignoring of any files.

tumerov:
Provides enhancement requested in #811. Allows your phpcs to prevent of ignoring any @codingStandardsIgnore sections and to ignoring of any modifications of current coding standard via @codingStandardsChangeSetting section.

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -1728,11 +1728,14 @@ class PHP_CodeSniffer
             $filePath = $file;
         }
 
-        // Before we go and spend time tokenizing this file, just check
+        $cliValues = $this->cli->getCommandLineValues();
+
+        // If force check not set,
+        // before we go and spend time tokenizing this file, just check
         // to see if there is a tag up top to indicate that the whole
         // file should be ignored. It must be on one of the first two lines.
         $firstContent = $contents;
-        if ($contents === null && is_readable($filePath) === true) {
+        if (isset($cliValues['forceCheck']) === false && $contents === null && is_readable($filePath) === true) {
             $handle = fopen($filePath, 'r');
             stream_set_blocking($handle, true);
             if ($handle !== false) {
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks for the PR, but I'm not going to merge it in because I'm not convinced this is actually a good feature to implement. Developers rarely use those annotations to hide coding standard errors, so there is typically a good reason when they do. So just showing them feels like the wrong thing to do.

I think that the only way I'd consider implementing this feature is if the errors were displayed with some sort of annotation to indicate that they were ignored but are now being shown. I'm not convinced that this is even a required or useful feature, but it would at least give you an error report where you could see which errors are valid and which ones you've turned on temporarily.

I've left the feature request open to see if any more support comes for it, so you might want to post your use case over there. There current one doesn't make a lot of sense to me given you're not typically imposing your own coding standard on code you don't contribute to. If developers are worried about other developers muting errors, a sniff to detect the usage of the coding standard ignore comments is easy to build and provides a talking point in code reviews.

tumerov:
Thanks for the response.

Unfortunately, sniff for detecting the usage of the coding standard ignore comments will not work because phpcs ignores all files which have `@codeStandardsIgnoreFile` and prevent them to be tokenized and scanned by any sniff. Of course, the functionality to detect these special comments can be easy implemented by another code before running the phpcs. But it would be nice if everything would be in one tool.

I understand the reason for using ignoring - and it is a really good feature.
But also it can add some vulnerabilities - I'm not convinced, that all people, who use phpcs for basic check the code written by other people (for marking the places for deep code review) knows about special comments for skipping parts of the code from the check.
Besides that, this is just an option, and a person, who will use it, will know what he wants and what expects to get.

Sincerely

gsherwood:
> Besides that, this is just an option, and a person, who will use it, will know what he wants and what expects to get.

Every option that gets added makes the code a bit more complex, and gives me something else I need to support. That's why I prefer to add options that I think provide the most benefit. As I've already said, I can't see the use cases for this, so I could never explain to someone why they'd ever want to use this option. That makes it hard to support.

Thanks again for the PR, even though we have a clear difference of opinion. I'll close it for now, but I will come back to it if issue #811 gets more support and clearer use cases. Then I can be sure I'm adding and supporting a feature people will use.

PR #1107, opened 2016-08-09
contributing.md remove args so it uses phpcs.xml.dist

ofbeaton:
The current `CONTRIBUTING.md` says to run `php scripts/phpcs . --standard=PHPCS -np` which makes it ignore the default `phpcs.xml.dist` and also scans the whole `vendor` directory. 

By removing the arguments it will use the `phpcs.xml.dist` file and not scan the vendor directory, plus keep up to date with whatever defaults you change in the future.

Alternatively it if you want to keep it listing everything, it could be changed to `php scripts/phpcs CodeSniffer.php scripts CodeSniffer --standard=PHPCS -np` so that vendor and tests are ignored.


gsherwood:
Thanks. Completely forgot about that.

PR #1117, opened 2016-08-14
Recognize `self` as type hint in `getMethodParameters()`.

jrfnl:
Improve type hint recognition:
- `self` is a valid type hint and was so far not being recognized as such. Ref: http://php.net/functions.arguments#functions.arguments.type-declaration
- `parent` and `static` are _not_ valid type hints, but if they are there, they were probably intended as such.
- For both the above, made sure not to confuse the type hint with the default value.

New array value `raw`:
- Add raw content of the parameter parts to the return array for use in further processing.

Includes adjusted unit tests + added unit test for `self`

P.S.: the alignment of the setting of the `raw` array value is a bit out of sync, but that's due to a bug in another sniff. The PHPCS code style was borking on proper alignment.

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -2769,13 +2769,14 @@ class PHP_CodeSniffer_File
 
         $vars            = array();
         $currVar         = null;
+        $paramStart      = ($opener + 1);
         $defaultStart    = null;
         $paramCount      = 0;
         $passByReference = false;
         $variableLength  = false;
         $typeHint        = '';
 
-        for ($i = ($opener + 1); $i <= $closer; $i++) {
+        for ($i = $paramStart; $i <= $closer; $i++) {
             // Check to see if this token has a parenthesis or bracket opener. If it does
             // it's likely to be an array which might have arguments in it. This
             // could cause problems in our parsing below, so lets just skip to the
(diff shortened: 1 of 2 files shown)


jrfnl:
Any feedback on this ?

gsherwood:
> Any feedback on this ?

Sorry, will get to merging this soon. My only comment is: looks like a great change. Thanks.

jrfnl:
Anything I can do to move this PR and it's sister #1193 forward ?

gsherwood:
> Anything I can do to move this PR and it's sister #1193 forward ?

Take my kids for a few weeks :)

More seriously, I'm back working on PHPCS and will get through the backlog.

jrfnl:
> Take my kids for a few weeks :)

Uh... for the good of your kids, I'd strongly advise against that ;-)

> More seriously, I'm back working on PHPCS and will get through the backlog.

Awesome. Thanks for all you do, just thought I'd check in to see if I could do anything to make it easier. Would combining this PR and #1193 (which both touch the same method) help any ?

gsherwood:
> P.S.: the alignment of the setting of the raw array value is a bit out of sync, but that's due to a bug in another sniff. The PHPCS code style was borking on proper alignment.

That's actually a sniff setting in the PHPCS standard that enforces a max of 12 spaces for padding. So if it takes more than 12 spaces for alignment, it enforces no alignment. The setting is this:
```xml
 <!-- Have 12 chars padding maximum and always show as errors -->
 <rule ref="Generic.Formatting.MultipleStatementAlignment">
  <properties>
   <property name="maxPadding" value="12"/>
   <property name="error" value="true"/>
  </properties>
 </rule>
```
But I actually made a change to the index name - "raw" became "content" just to match the index of the main tokens array. As it turns out, it's long enough to force alignment of the equal signs. Hopefully that doesn't cause you much of a problem.

Thanks a lot for this PR, and for adding tests as well.

jrfnl:
> ...  a change to the index name - "raw" became "content" ... Hopefully that doesn't cause you much of a problem.

No problem at all & happy to help.

PR #1122, opened 2016-08-19
Makes Squiz MethodScopeSniff sniff Traits

donatj:
It currently sniffs classes and interfaces, this just adds traits to the list of things sniffed.

Scrutinizer was complaining about one of our traits having a method without visibility defined and I was like "phpcs should have caught this!" huzzah, it was not checking traits.

Easy fix.


gsherwood:
Thanks for the PR, and the tests.

PR #1125, opened 2016-08-21
Bug in Squiz.Operators.ComparisonOperatorUsage

rhorber:
The sniff which enforces identity operators does not work correctly on inline ifs.
The check for the parenthesis opener uses $stackPtr instead of $end (which is the closer).


aik099:
Please also add a test case that would cover this behavior.

rhorber:
I created a new pull request #1127, with tests.
So I close this.

aik099:
FYI: You could have updated this PR as well.

PR #1127, opened 2016-08-23
Fixed test in ComparisonOperatorUsageSniff and added tests for it.

rhorber:


gsherwood:
Thanks a lot for the fix, and the tests.

rhorber:
You're welcome.

PR #1131, opened 2016-08-25
Fix in Squiz.CSS.DuplicateClassDefinitionSniff

rhorber:
It now handles a media block as a scope.
This prevents false positives caused by the same class in multiple media queries/blocks.


gsherwood:
Thanks a lot for doing this. I'm not sure when I would have got time. I really appreciate it.

rhorber:
No problem, you're welcome :)
Let me know, if I could help with something other.

Issue #1136, opened 2016-08-31
Provide XSL linking


BorislavSabev:
Hello.

It would be a nice feature to be able to pass an .xsl file s path so it is embedded in the XML when generated.
The required element is: `<?xml-stylesheet type="text/xsl" href="$PATH_SET$"?>`
The path would be relative to the target file. 

Something like:
`./phpcs --report=xml --report-file=target.xml --report-xsl=phpcs.xsl`

This would result in the following element being included in the XML:
`<?xml-stylesheet type="text/xsl" href="phpcs.xsl"?>`

I have some XSL that I can provide as a default one. It is better to be able to see the styled XML in a browser as a nice default. 

What do you think on this issue?

gsherwood:
I don't like putting top-level CLI options in for such specific features, but the idea in general is a good one. I'm just not sure how to achieve it.

BorislavSabev:
@gsherwood I don't know how phpcs builds the XML internally but the general place to put this in should be the XML generation phase.

Do you know if there is access to the CLI params in `PHP_CodeSniffer_Reports_Xml->generate()`?
How do you generally send CLI params down trough the layers of phpcs?

I am willing to send a PR if you lend a head in getting to know phpcs's internals so that I can do it fast.

BorislavSabev:
@gsherwood  I can see you don't directly pass the $cliValues to the actual Report generatior class.
They're passed only to `Reporting->printReport()` which is one level above.
I've started implementing it. I will see how far I go.

BorislavSabev:
@gsherwood and here is the PR:
https://github.com/squizlabs/PHP_CodeSniffer/pull/1137

gsherwood:
As I said in my previous comment, I wouldn't put in a top-level CLI option for something like this, which is what the PR does. This switch is only useful to a single report type, so it should not be something that the core even knows about.

The way I was thinking about trying to achieve this some time in the future (only for 3.x, not for 2.x) would be to perhaps support templates for reports. Although I don't really believe that would be useful.

I'm not going to merge in that PR for the above reason, so the best way to achieve what you are after right now is just to create your own custom report class. Copy the XML one, hack it up to put in the XSL line you want, then use it with `--report=/path/to/your/report.php`.

gsherwood:
Closing as I've had no additional interest or PRs for this.

BorislavSabev:
Hey @gsherwood is there a way to solve this in the latest version?

[...]

PR #1137, opened 2016-08-31
 #1136 - Provide XSL linking first proposition

BorislavSabev:
My changes are working. I don't think it is fully-done as it is almost certain I missed something that the project adheres to. However here is the PR. The XSL is included at the top of the XML file if --with-xsl-file=<xslFilePath> is passed.

The changes Include:
 - new CLI param --with-xsl-path=<xslFilePath>
 - input parsing of the new param
 - docs for the new param
 - Defined report types' constants to match against
 - Reporting modification in printReport()
 - Modification in PHP_CodeSniffer_Reports_Xml

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -1810,7 +1810,7 @@ class PHP_CodeSniffer
                 continue;
             }
 
-            $reportClass = $this->reporting->factory('full');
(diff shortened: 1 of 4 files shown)


gsherwood:
Closing. See #1136 for the discussion of this issue.

PR #1144, opened 2016-09-05
Fix indentation in PEAR multiline function declaration fix

astehlik:


astehlik:
Currently this pull request only contains the tests for the problem.

If someone can give me a hint how to solve it, I'll try to provide a fix.

gsherwood:
You get this result because you are running the sniff in isolation. Some fixes, like this one, don't bother trying to get indents correct as they leave that up to the main scope indent sniff to figure out. This saves them having to do things like check for comments in the indent area, and checking for inline HTML inside template files.

So this particular fix just throws the bracket onto a new line and let the scope indent sniff position it, and comes back around a second time to bring the opening brace onto the same line as well.

If you want to change the sniff to put the indent in itself, you have to first figure out how far indented the function is by looking at the tokens between it and the previous line. You might be able to get away with just looking at the column position of the first token non-whitespace/non-comment token and then putting in spaces before the `(` to align it with that token. But it needs more unit tests than the one you've added to be sure it is working.

astehlik:
Thank you for this extensive feedback and the great explanations :)

If there is a sniff that fixes the indentation in that special case, I will gladly use it. Then there is no point in changing this one.

I just figured out how stupid I was. I added `<rule ref="Generic.WhiteSpace.ScopeIndent">` to my ruleset and now the indentation if fine.

Thanks again for your help!

gsherwood:
> I just figured out how stupid I was.

You're not stupid. I don't think it is unreasonable to expect that a sniff will fix your code exactly the way you want. It's more a design decision to keep rules like indenting off in their own sniff due to their complexity.

I'm going to re-open this and leave it as an enhancement because I still do like to get sniffs to look at the existing indents and positions fixed code correctly, if possible.

PR #1146, opened 2016-09-06
Play nice with sniff libraries with alternative unit test setups.

jrfnl:
If there are `installed_paths` set up in the config and one of the standards in the `installed_paths` uses an alternative way to unit test - this applies in particular to the [PHPCompatibility](https://github.com/wimg/PHPCompatibility) sniff library -, running the unit tests for any of the other standards using the normal PHPCS unit test set up will fail as fatal errors will be thrown for (sniff) files which cannot be found.

This minor adjustment prevents these fatal errors while leaving the responsibility for running the tests which are set up differently with the sniff library in question.

### DIFF
--- a/tests/Standards/AllSniffs.php
+++ b/tests/Standards/AllSniffs.php
@@ -36,7 +36,16 @@ require_once dirname(__FILE__).'/AbstractSniffUnitTest.php';
  */
 class PHP_CodeSniffer_Standards_AllSniffs
 {
-
+    /**
+     * Error storage.
+     *
+     * Stores the test file path for each test file which either could not be
+     * translated to a corresponding sniff file or where the corresponding
+     * sniff file could not be found.
+     *
+     * @var array
+     */
+    public static $includeFailures = array();
 
     /**
      * Prepare the test runner.
@@ -64,6 +73,14 @@ class PHP_CodeSniffer_Standards_AllSniffs
 
         $isInstalled = !is_file(dirname(__FILE__).'/../../CodeSniffer.php');
 
+        // Optionally allow for ignoring the tests for one or more standards.
+        $ignoreTestsForStandards = getenv('PHPCS_IGNORE_TESTS');
+        if ($ignoreTestsForStandards === false) {
+            $ignoreTestsForStandards = array();
+        } else {
+            $ignoreTestsForStandards = explode(',', $ignoreTestsForStandards);
+        }
+
         $installedPaths = PHP_CodeSniffer::getInstalledStandardPaths();
         foreach ($installedPaths as $path) {
             $path      = realpath($path);
@@ -80,6 +97,10 @@ class PHP_CodeSniffer_Standards_AllSniffs
             }
 
             foreach ($standards as $standard) {
+                if (in_array($standard, $ignoreTestsForStandards, true)) {
+                    continue;
+                }
+
                 $testsDir = $path.DIRECTORY_SEPARATOR.$standard.DIRECTORY_SEPARATOR.'Tests'.DIRECTORY_SEPARATOR;
 
                 if (is_dir($testsDir) === false) {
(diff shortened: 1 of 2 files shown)


gsherwood:
I can't merge this in as this is going to silently cause invalid sniff tests in PHPCS itself, or in custom standards that follow the same testing format, to be skipped. So you will think you've committed a passing test, but it's actually not being run at all.

jrfnl:
@gsherwood Any suggestions on how else to handle this ? Would throwing an exception instead of silently failing be considered for merge ? Any pointers appreciated and I'll gladly rework the PR to something acceptable.

jrfnl:
Oh and to clarify: the main issue I'm trying to solve is the unit test suite loading mechanism looking for `sniff` files for test files which don't relate to a specific sniff, but test underlying methods in an abstract class, largely similar to the tests contained in https://github.com/squizlabs/PHP_CodeSniffer/tree/master/tests/Core/File

Is there another way you can think of in which I can prevent the `suite()` method from throwing fatal errors for those situations ? If there is, I'll gladly send in a PR to the sniff library in question to change things there.

gsherwood:
> Would throwing an exception instead of silently failing be considered for merge ?

I think throwing an exception is fine because it will alert me (and others) to bad tests, but will it help with this particular external standard?

> Is there another way you can think of in which I can prevent the suite() method from throwing fatal errors for those situations ?

Nothing comes immediately to mind, and my plate is so full at the moment that I can't see any time that I'll be able to sink into looking for one.

The goal is good, but I've been saved by that fatal error a few times so I need to keep that alert in there. So if we can find a solution that keeps it, I'll be happy to merge it.

jrfnl:
@gsherwood Ok, I've come up with an alternative solution.

This solution consists of two parts based on various iterations while trying to solve this.

Part 1 is checking an environment variable which can be passed in through the `phpunit.xml` file which allows for indicating one or more standards to ignore when setting up the test suite.
This is the simplest way to prevent the issue.
As there won't be that many people working on several standards which have their own test suite, for those few people to add the environment variable to their `phpunit.xml` file for the projects which do use the PHPCS native test suite, should not be much of a problem.

Part 2 is not necessary, but might be a nice to have (=result of an earlier iteration).
This involves the catching of the file path names of the files which are causing issues and reporting them back as part of the phpunit results in a neat list without errors being thrown before starting to run the unit tests.

screenshot

If you could let me know if this solution direction is acceptable and what you'd prefer: just part 1 (minimal solution) or include both parts, I'll adjust the PR to match.

For part 1, I can imagine it might also be useful to add something about the availability of this option to the wiki.

jrfnl:
Thanks for getting back to me this quickly! I've adjusted the PR to what I outlined above (both parts).

> Have you had a chance to see in the PHPCompatiblity project would be happy with it, or whatever other standard you are testing?

The issue only really crops up for devs who develop sniffs - and therefore run tests - for several standards where at least one of the other standards has their own unit test set-up - like PHPCompatibility - and to be honest, I'm not even sure if there are more of us, let alone other projects but PHPCompatibility which have their own unit test setup ;-)

I have prepared a section for the `README.md` of the PHPCompatibility project - PR https://github.com/wimg/PHPCompatibility/pull/217 - with information about running the test suites for other sniff libraries without conflicts, so we might get response from other devs about it there.


---

:fast_forward: For anyone who comes across this thread at a later point: the most elegant solution would have been to be able to use the `filter` argument from the PHPUnit command line to bypass loading of standards, combined with the environment variable for when no `filter` argument would be passed in, but I can't seem to get access to the parsed value at the point where I'd need it and parsing the PHPUnit command line command again within the PHPCS test suite seems over the top.

jrfnl:
@gsherwood Both related PRs for PHPCompatibility have been merged and can therefore be considered accepted.

gsherwood:
@jrfnl Sorry, I got swamped with work and only just got around to merging and testing this. Thanks a lot for the PR, and thanks for sorting it all out with PHPCompatibility as well.

jrfnl:
@gsherwood Not to worry. Life happens to all of us ;-) Thanks for merging this.

Issue #1008, opened 2016-05-18
PSR-2 alternative condition syntax demanding a space before colon


amcsi:
`if ($something):` <- PHPCS is complaining here that there's a space missing before the colon. I couldn't find it in the PSR-2 documentation that this is needed in the alternative syntax. Also this is really unusual for me; I've mostly seen the alternative syntax's colon without a space in other code I've seen before.

gsherwood:
PSR-2 doesn't specify any specific rules for alternate syntax, so PHPCS applies what rules it can.

PSR-2 says `There MUST be one space between the closing parenthesis and the opening brace`, so PHPCS requests that you place a single space after the closing parenthesis.

If you think alternate syntax should follow different rules, I encourage you to ask on the mailing list. You've got almost no chance of getting PSR-2 changed, but maybe you'll be able to get something put into PSR-12.

amcsi:
There are more things PSR-2 doesn't specify rules for than does. I don't see how you come to the conclusion that if a rule is not defined, then "apply this rule" rather than not apply any rule at all. I could argue that I think the switch-case rules should apply, because we're dealing with a colon, and thus there should never be spaces before colons.

I'm not saying that I want that to be the case, I'm just a bit bothered by the fact that I would need to follow a part of a coding style that not only does not exist, but is not popular. I looked at the first 10 results in Google when searching for "alternate php syntax", and nowhere in the 10 results did I see anyone put spaces before the colon.

So this doesn't make sense to me.
In any case do you happen to know where that google group is? Ever since the the PHP-FIG website changed, I can't find the PSR-2 group link anymore.

gsherwood:
> In any case do you happen to know where that google group is?

It's just one big group for everything: https://groups.google.com/forum/?fromgroups#!forum/php-fig

PR #1148, opened 2016-09-07
Allow configuring the number of spaces before colons in alternative control structures

nkovacs:
This is a follow-up to #1008. I don't think the PSR2 standard should be checking colons at all, but I'm not going to argue about that.

The problem is that if you want to create a standard that expects 0 spaces for colons (since that is how most people write code), you have to copy the entire sniff and modify it, and that sniff needs to handle the normal syntax too, since you have to disable the original sniff.

I made the sniff configurable, so all you need to do is change one setting in your standard.

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/ControlStructures/ControlSignatureSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/ControlStructures/ControlSignatureSniff.php
@@ -27,6 +27,13 @@ class Squiz_Sniffs_ControlStructures_ControlSignatureSniff implements PHP_CodeSn
 {
 
     /**
+     * How many spaces should precede the colon if using alternative syntax.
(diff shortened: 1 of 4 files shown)


ramsey:
I'd love to see this merged in. Is this something you're still considering for inclusion?

jkrehm:
Any chance of this getting merged in? I would love to have this option to placate my IDE.

devantoine:
This PR is almost one year old but we need it. It would be nice to be able to set the number of spaces before colons.

gsherwood:
I've committed a change to allow this option. I didn't merge only due to the number of conflicts.

New option is documented here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#squizcontrolstructurescontrolsignature

Thanks for this PR.

PR #1157, opened 2016-09-12
Fix DisallowMultipleStatements not working with <?=

### LINKED ISSUE #1156: Generic.Formatting.DisallowMultipleStatements errors when mutliple short echo tags are used on the same line
In #423 you made an exception to Generic.Formatting.DisallowMultipleStatements for multiple php open/close tags in the same line.

This works inconsistently. The following are all okay:

```php
<div class="<?php echo $class; ?>" id="<?php echo $id; ?>"></div>
<div class="<?php echo $class ?>" id="<?php echo $id ?>"></div>
<div class="<?= $class ?>" id="<?= $id ?>"></div>
```

But this isn't:

```php
<div class="<?= $class; ?>" id="<?= $id; ?>"></div>
```

I've added a failing test: https://github.com/nkovacs/PHP_CodeSniffer/tree/disallow-multiple-statements-bug

nkovacs:
Fixes #1156


gsherwood:
Thanks a lot.

PR #1162, opened 2016-09-18
Test on Current HHVM

photodude:
The travis precise container HHVM version is really outdated being 3.6. Current HHVM is 3.18 STS. We now have the [option to specify LTS versions](https://docs.travis-ci.com/user/languages/php#HHVM-versions) or to just use latest on trusty. (most repositories are still opting for latest)
#### Summary
- Test on Current HHVM
- Use hhv\* for checks in case we decide to also use LTS builds like hhvm-3.12 and hhvm-3.15


photodude:
Bump for merge consideration @gsherwood

photodude:
Bump for merge consideration @gsherwood, the conflicts here have been solved

.travis.yml
```diff
@@ -9,24 +8,28 @@ php:
...
 
 script:
-  - if [ $TRAVIS_PHP_VERSION != "hhvm" ]; then php scripts/phpcs --config-set php_path php; fi
+  - if [[ $TRAVIS_PHP_VERSION != hhv* ]]; then php scripts/phpcs --config-set php_path php; fi
   - phpunit tests/AllTests.php
-  - php scripts/phpcs CodeSniffer.php CodeSniffer --standard=PHPCS --report=full -np
-  - if [[ $TRAVIS_PHP_VERSION != "hhvm" && ${TRAVIS_PHP_VERSION:0:1} != "7" ]]; then pear package-validate package.xml; fi
-  - if [ $TRAVIS_PHP_VERSION != "hhvm" ]; then php scripts/build-phar.php; fi
-  - if [ $TRAVIS_PHP_VERSION != "hhvm" ]; then php phpcs.phar CodeSniffer.php CodeSniffer --standard=PHPCS --report=full -np; fi
+  - php scripts/phpcs CodeSniffer.php CodeSniffer --standard=PHPCS --report=summary -np
```
jrfnl: Why are you also changing the report type ? (here and three lines down)

photodude: I didn't notice the change, that is all.  Is there a reason you changed the report to full when you did the php 5.2 stuff? I didn't see an explanation in the commit history for the necessity of the change when I just checked for when this changed.

jrfnl: I didn't change that and AFAIK that change didn't have anything to do with the PHP 5.2 stuff.

photodude: My misunderstanding, I was looking at the wrong commit information on my phone
I found it in 36052ebb85a9ecaaac29fe95cab3cae6ece2dbb7

Will make the change

gsherwood:
I don't really want to maintain a custom HHVM runner for Travis. I test on HHVM because it's a one-liner in the travis config file, but I can't even get the thing compiled for local testing, so I find supporting it nearly impossible. I've thought about dropping support entirely several times.

Also, The travis build with this change looks to have not worked. PHPUnit wont even run, so I imagine a different version of PHPUnit also needs to be installed here to make this change worthwhile.

photodude:
One of the problems is the "one liner" is an extremely old eol version of hhvm and has an exceptional number of php5 compatibility issues.

I can try rewriting this for the travis beta trusty container to shorten the lines a little since this project would not be affected by the mysql socket bug they have

It looks like phpunit 4.8.x should be installing via composer, both 4.8.x and 5.x I have working on other repos. there should be no issue there. Not sure why `The command "phpunit tests/AllTests.php" exited with 1.`

photodude:
@gsherwood Running on the beta trusty container fixes the HHVM build and keeps the setup simpler.

gsherwood:
> Running on the beta trusty container fixes the HHVM build and keeps the setup simpler.

That looks a bit easier to understand. Thanks.

PR #1168, opened 2016-09-21
Updated FunctionCommentSniff to allow `@inheritDoc`

Zizaco:


CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -74,6 +74,19 @@ protected function processReturn(PHP_CodeSniffer_File $phpcsFile, $stackPtr, $co
...
+        
+        // Skip if function comment contains inheritDoc
+        for ($i=0; $i < 10; $i++) {
+            if (! isset($tokens[$commentStart+$i])) {
+                break;
+            }
+
+            $tokenContent = strtolower($tokens[$commentStart+$i]['content']);
+
+            if (strstr($tokenContent, '{@inheritdoc')) {
```
aik099: According to docs (see https://www.phpdoc.org/docs/latest/guides/inheritance.html#the-inheritdoc-tag) this tag only inherits description.

In Symfony however it's used to inherit following:

* short description
* long description
* parameter list (the `@param` tag)
* return type (the `@return` tag)

gsherwood:
This is not the correct way to do this. Proper support for @inheritDoc doesn't skip checking the function comment just because someone puts this tag in. The tag is just used as a placeholder for importing the description from a parent. All other rules must still apply.

I can't merge this in, and I doubt the official solution is what you are looking for. It is more likely that you just use @inheritDoc to skip documenting an entire block, which might be what specific frameworks do, but wont be going into the Squiz coding standard.

PR #1176, opened 2016-09-26
Added @codingStandardsChangeRule annotation to change/disable rulesets per file

DaFox:


gsherwood:
I don't quite understand what this feature does and why you need it.

The codingStandardsChangeSetting annotation already allows you to change sniff settings as files are processed. The code that processes that annotation directly modifies the public member vars inside the sniff classes while they are running so that they work correctly. While this code would modify those property values in the ruleset array, they wont be used by the sniffs, so nothing would happen.

It looks like this code would only work if you limit its use to changing things related to error messages, such as the type and severity. These are sourced directly from the ruleset array during error message creation.

If this is what you are using it for, I would suggest that annotations are not the best way to do this. I don't like the codingStandardsChangeSetting annotation either, but I needed to build it to support unit testing. In fact, I hate the idea of annotations all together. I would prefer a syntax that can be used in a ruleset.xml file to define specific sniff settings per file in the same way you can include or exclude entire sniffs per file.

But a change like this would only make sense in the 3.x branch. I don't want to put features like this into the 2.x branch as it is a lot of effort to forward-port changes like this to 3.x, and I want to try and keep 2.x as stable as possible while it is winding down.

gsherwood:
Closing due to no feedback and previous comment.

PR #1183, opened 2016-10-01
Fix DocBlock of `PHP_CodeSniffer_Tokenizers_JS::tokenizeString()`

ocean90:
Seems to be copy/pasted from `PHP_CodeSniffer_Tokenizers_PHP::tokenizeString()`.


gsherwood:
Thanks for fixing that.

PR #1187, opened 2016-10-05
Remove duplicate code from the Squiz `ArrayDeclaration` sniff.

jrfnl:
The three code blocks for `T_ARRAY`, `T_OPEN_SHORT_ARRAY` and `T_CLOSURE` were nearly the same.
This PR simplifies the code by folding the three blocks into one and only having a small if/else block for the actual difference between the three.


gsherwood:
Thanks for the cleanup.

PR #1193, opened 2016-10-19
Allow for nullable type hints and add a `nullable_type` indicator.

jrfnl:
PHP 7.1 introduces nullable types.

This PR allows for recognizing these in the `getMethodParameters()` method.
The starting `?` will be added to the type hint and a new parameter array key `nullable_type` will be set as well.

Includes additional unit test for this functionality, however the unit test breaks the test suite as the nullable type hint is seen as a parse error in PHP < 7.1. This is due to the test code being contained in the same file as the actual tests.

@gsherwood Any suggestions how I should work around this ? Remove the unit tests ? Move the test code to a separate file ? (but how will the PHPCS test suite handle that ?)

Ref:
- http://php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types
- https://wiki.php.net/rfc/nullable_types

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -2774,6 +2774,7 @@ class PHP_CodeSniffer_File
         $passByReference = false;
         $variableLength  = false;
         $typeHint        = '';
+        $nullableType    = false;
 
         for ($i = ($opener + 1); $i <= $closer; $i++) {
             // Check to see if this token has a parenthesis or bracket opener. If it does
@@ -2806,7 +2807,7 @@ class PHP_CodeSniffer_File
                 break;
             case T_ARRAY_HINT:
             case T_CALLABLE:
-                $typeHint = $this->_tokens[$i]['content'];
+                $typeHint .= $this->_tokens[$i]['content'];
                 break;
             case T_STRING:
                 // This is a string, so it may be a type hint, but it could
@@ -2843,6 +2844,12 @@ class PHP_CodeSniffer_File
                     $typeHint .= $this->_tokens[$i]['content'];
                 }
                 break;
+            case T_INLINE_THEN:
+                if ($defaultStart === null) {
+                    $nullableType = true;
+                    $typeHint    .= $this->_tokens[$i]['content'];
+                }
+                break;
             case T_CLOSE_PARENTHESIS:
             case T_COMMA:
                 // If it's null, then there must be no parameters for this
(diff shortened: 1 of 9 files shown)


aik099:
> Includes additional unit test for this functionality, however the unit test breaks the test suite as the nullable type hint is seen as a parse error in PHP < 7.1. This is due to the test code being contained in the same file as the actual tests.

You need to add code, that would allow parser running on PHP 5 to parse this nullable types as well.

jrfnl:
@aik099 No amount of code will allow a PHP5 parser to understand syntax which will only be introduced in PHP 7.1 ....

aik099:
Do PHP_CodeSniffer scalar type tests work currently on PHP5? I mean tests that check that `string` typehint is correctly recognized when used as parameter typehint in a method.

If they exists (I'm not sure if they exists) and work, then I see no big problems to emulate support for `?` parsing prior to typehint, do you?

gsherwood:
I typically try and emulate new tokens in older PHP versions. When I get a chance to look at this PR, I'll also take a look at emulating this. If I can, great. If I can't, I'll have to figure out how to make the tests pass. Either way, I'll figure something out. PHP hasn't beaten me yet :)

jrfnl:
@aik099 @gsherwood The actual functionality is no problem at all - the tokenizer can handle it fine in any PHP version.

The problem is with the unit tests. And even then it doesn't *have* to be a problem.
Currently for unit testing utility methods, the unit tests themselves and the "example code to parse" live in the same file. That's what the issue is. As that file has to be parsed for the unit tests and the parser will encounter an unknown syntax in the example code and fail with a parse error.
If the "example code to parse" could live in a different file for utility methods, there would be no problem whatsoever as then the file containing the example code would not need to be parsed, only tokenized.

I already created a sniff using a (copy of) the adjusted method for the PHPCompatibility library (https://github.com/wimg/PHPCompatibility/pull/282) and as you can see the unit tests for that sniff pass fine on all supported PHP versions: https://travis-ci.org/wimg/PHPCompatibility/builds/168821994

gsherwood:
> Currently for unit testing utility methods, the unit tests themselves and the "example code to parse" live in the same file. That's what the issue is.

Yep, I see that now. I confess that I didn't actually look before.

I don't remember why those tests were done like that. I don't think I actually wrote them, so I suspect there isn't a technical reason for it at all.

jrfnl:
> I don't remember why those tests were done like that. I don't think I actually wrote them, so I suspect there isn't a technical reason for it at all.

So, you'd be ok with me moving the test code out to a separate file ? I'll need to run some tests to see how the unit test framework would handle that, but I think it should be fine.
If I get it working - should adjust this for all the utility method unit tests ? To keep them consistent and to prevent similar issues in the future ?

gsherwood:
> So, you'd be ok with me moving the test code out to a separate file ?

I wouldn't just be ok. I'd be very grateful.

> If I get it working - should I adjust this for all the utility method unit tests ?

Yes please.

jrfnl:
Okidoki. All done. For the three test files for utility methods which contained test case code, I've created separate test case files.
The other two utility method test files contained the code inline - not as a separate block - so I've left those as is.

I have one other PR open (#1117) which touches the same method `getMethodParameters()` as this PR. Would you like me to add that commit to this PR to prevent merge conflicts ?

jrfnl:
Anything I can do to move this PR and it's sister #1117 forward ?

gsherwood:
Finally merged it in. Thanks a lot for the change, and especially for fixing those tests files for me. I really really appreciate that.

jrfnl:
You're very welcome ;-)

PR #1195, opened 2016-10-20
add flag to use short scalar types

ctrl-f5:
added flag `--short-scalar-types` or `-t` which switches out `boolean` and `integer` for `bool` and `int`

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -204,6 +204,13 @@ class PHP_CodeSniffer
                                    'callable',
                                   );
 
+    /**
+     * Flag to configure the use of short scalar types
+     *
+     * @var bool
+     */
+    public static $useShortScalarTypes = false;
+
 
     /**
      * Constructs a PHP_CodeSniffer object.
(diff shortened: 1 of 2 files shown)


aik099:
What about auto-fixing. Right now sniffs expect short types to be reported as error for them to fix automatically.

It would be logical instead of allowing short types alongside with long types just:

1. remove long types before adding short ones
2. make auto-fixing code aware of this changes and autofix long types into short types, when command option is used

Have you checked that if specified in `ruleset.xml` new argument works as expected?

gsherwood:
Thanks for the PR, but I don't want a CLI option for this. I have the associated issue on my todo list and I'll get to it when I finally get some time. The resulting change is likely to just be a core from a specific version (probably 3.0 due to BC breaks).

ctrl-f5:
I did test it with the argument in the xml, works fine.

also automatic fixer works as expected. same as before without the option, and long types are converted to short when the options is set.

This makes it not break BC, so it can easily be merge into 2.x branch, and you can implement it as you like in 3.x

virgofx:
Any update on short scalar support? Most of the PSR-5 docs and PHP7 docs use this.

ctrl-f5:
I've noticed there's a RC for 3.0, was this feature added?

gsherwood:
> I've noticed there's a RC for 3.0, was this feature added?

It has not been added. It's still on my high priority list, but it's tracked here: https://github.com/squizlabs/PHP_CodeSniffer/pull/1047

virgofx:
@gsherwood Thanks for the update. Adding the pull to my watch list :)

PR #1196, opened 2016-10-21
Option to pass an array as a list of possible values to search for

lenaorobei:
Added option to pass an array as a list of possible values to search for in following methods:
- findPrevious
- findNext
- findFirstOnLine


aik099:
Any tests, that you can add to ensure this code works?

lenaorobei:
@aik099 I added tests for each of these methods.

gsherwood:
Closing as this was forgotten but has not been needed since. Also wont merge + 3.5.0 changes will make the problem worse.

PR #1200, opened 2016-10-26
Svn-pre-commit hook: use env in shebang

sserbin:
Same way as scripts/phpcs and scripts/phpcbf do.

### DIFF
--- a/scripts/phpcs-svn-pre-commit
+++ b/scripts/phpcs-svn-pre-commit
@@ -1,4 +1,4 @@
-#!@php_bin@
+#!/usr/bin/env php
 <?php
 /**
  * A commit hook for SVN.


aik099:
You're suggesting that this hook doesn't work by default due invalid shebang?

sserbin:
That's right. Kind of makes it unsable when installed through composer. 
This is also in line with e0beeefdf90c4b8c693af8bc7000e15cf0c24ca6

aik099:
Ah, I see now. The `@php_bin@` was probably replaced with actual PHP path only during PEAR install.

gsherwood:
I'll merge this in and fix the package.xml (this change makes the script unusable with PEAR) but be aware that this script has been removed from the 3.0 version.

gsherwood:
Thanks for the PR.

PR #1201, opened 2016-10-27
Update ConstructorNameSniff to fix OldStyle methods

PatrickRose:
We're about to switch to PHP7 and having this sniff automatically fix things for us would make it a lot easier!

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/NamingConventions/ConstructorNameSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/NamingConventions/ConstructorNameSniff.php
@@ -84,8 +84,11 @@ class Generic_Sniffs_NamingConventions_ConstructorNameSniff extends PHP_CodeSnif
 
         if (strcasecmp($methodName, $className) === 0) {
             if (in_array('__construct', $this->_functionList) === false) {
-                $error = 'PHP4 style constructors are not allowed; use "__construct()" instead';
-                $phpcsFile->addError($error, $stackPtr, 'OldStyle');
+                $error     = 'PHP4 style constructors are not allowed; use "__construct()" instead';
+                $toReplace = $phpcsFile->findNext(T_STRING, $stackPtr);
+                if ($phpcsFile->addFixableError($error, $stackPtr, 'OldStyle') === true) {
+                    $phpcsFile->fixer->replaceToken($toReplace, '__construct');
+                }
             }
         } else if (strcasecmp($methodName, '__construct') !== 0) {
             // Not a constructor.
--- a/CodeSniffer/Standards/Generic/Tests/NamingConventions/ConstructorNameUnitTest.inc.fixed
+++ b/CodeSniffer/Standards/Generic/Tests/NamingConventions/ConstructorNameUnitTest.inc.fixed
@@ -1,0 +1,52 @@
+<?php
+class TestClass extends MyClass
+{
+
+    function TestClass() {
+        parent::MyClass();
+        parent::__construct();
+    }
+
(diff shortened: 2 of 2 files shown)


aik099:
What about `parent::SomeParentClass()` calls? Are they currently reported as an error and should they be fixed?

gsherwood:
I don't like including fixers that do things like change method names because it's possible that they'll produce code that causes a project to break. So PHPCS core fixers are limited to formatting changes only.

As this isn't a formatting fix, it's not the sort of PR I would merge into the core.

I understand how it is useful, and you can certainly put the sniff into your own custom standard to use it, but I would have to figure out a way of classifying fixers as "safe" and "unsafe" before a feature like this could be included in PHPCS.

PatrickRose:
This is certainly safe though isn't it? I've already checked that there
isn't a method called __construct already

gsherwood:
> This is certainly safe though isn't it?

In my opinion, it isn't. Changing the name of a function, or class, or other structure is something that should be considered as refactoring, and done with knowledge of the code base through experience or analysis. PHPCS only changes formatting.

gsherwood:
Forgot to add: PHPCS *can* be used for refactoring as well, but this is a question about if a change is going into the core, and so being supported by myself. I'm not comfortable supporting PHPCS as a refactoring tool because PHPCS only processes one file at a time.

jrfnl:
I concur with @gsherwood on this and would like to add this additional argument for this specific case:

Renaming the PHP4 style constructor can easily break interoperability with other code.

Consider the following:
```php
// Your code
class MyClass {
	function MyClass() {
		// do constructor tasks
	}
}

// Someone else's code which extents your code
class OtherClass extends MyClass {
	function __construct() {
		parent::MyClass(); // The rename will break this call causing a fatal error 'function not found'.
		// do other constructor tasks
	}
}
```

Blindly renaming the constructor will break any code which extends or otherwise builds onto your code.
While in your specific use case you may know for sure that there are no dependents, you cannot assume the same for other people using this sniff.

> This is certainly safe though isn't it? I've already checked that there isn't a method called __construct already

Additionally, checking that there isn't a method called `__construct()` definitely isn't enough. The class should also not be within a namespace. A method shadowing the class name in a namespaced class is never considered a constructor (ok, except in 5.3.0-5.3.1) and therefore should not be renamed.

CodeSniffer/Standards/Generic/Sniffs/NamingConventions/ConstructorNameSniff.php
```diff
@@ -84,8 +84,11 @@ protected function processTokenWithinScope(
 
         if (strcasecmp($methodName, $className) === 0) {
             if (in_array('__construct', $this->_functionList) === false) {
-                $error = 'PHP4 style constructors are not allowed; use "__construct()" instead';
-                $phpcsFile->addError($error, $stackPtr, 'OldStyle');
+                $error     = 'PHP4 style constructors are not allowed; use "__construct()" instead';
+                $toReplace = $phpcsFile->findNext(T_STRING, $stackPtr);
```
dereuromark: This can be moved inside the if.

gmponos:
I would like also to point out that this sniffs considers `old` constructors static methods.

So in case you have the following:

class MyClass{

    public static function myClass(){ echo 'test'; }
}

MyClass::myClass();

Then this rule will also change the function above and most likely your code will break.

gsherwood:
Closing as this isn't the sort of fixer I want to support in the core. It would still be worthwhile as a custom coding standard (maybe a migration standard) that people could download and use.

Issue #604, opened 2015-05-28
Selectively disable rules with codingStandardsIgnoreStart and codingStandardsIgnoreLine


westonruter:
PHPCS supports comment directives like:

``` php
// @codingStandardsIgnoreFile
// @codingStandardsIgnoreStart
// @codingStandardsIgnoreLine
```

However, these are all or nothing. What would be even more useful is if we can selectively ignore certain sniffs or errors within sniffs. For instance:

``` php
$q = new WP_Query( array(
    'posts_per_page' => 200, // @codingStandardsIgnoreLine WordPress.VIP.SlowDBQuery.slow_db_query 
) );
```

or

``` php
// @codingStandardsIgnoreStart WordPress.VIP.SlowDBQuery.slow_db_query
$q = new WP_Query( array(
    'posts_per_page' => 200,
) );
// @codingStandardsIgnoreEnd WordPress.VIP.SlowDBQuery.slow_db_query
```

This draws inspiration from JSCS and its directives like:

``` js
//jscs:disable requireCamelCaseOrUpperCaseIdentifiers
```

We have this in the [WordPress-Coding-Standards](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards) project in a sniff-specific ad hoc system of end-of-statement comments:

``` php
echo $foo; // xss ok
```

Which suppresses the `WordPress.XSS.EscapeOutput` sniff. But it would be very useful if this was standardized in PHPCS itself and available to all sniffs

JDGrimes:
:+1:

GaryJones:
What format would the ignoring of multiple disabled rules take? Comma-separation?

westonruter:
Comma-separated rules is what JSCS uses: http://jscs.info/overview.html#disabling-specific-rules

```js
// jscs:disable requireCurlyBraces, requireDotNotation
```

jjeaton:
:+1:

thiemowmde:
:+1:

DanielSiepmann:
:thumbsup:

michaelhogg:
:+1:

pdelre:
+1

andrewhowdencom:
scss-lint disables lint on a given "scope". See: https://github.com/brigade/scss-lint#disabling-linters-via-source
``
// scss-lint:disable ExampleSniff
``
[...]

PR #1211, opened 2016-11-11
Add rule suppresion for a line or scope

mrubiosan:
Hey there. How do you feel about adding the functionality to ignore some rules for a given line of code, or block? Right now, you can use the // @codingStandardsIgnoreStart notation, but it has the downside of ignoring all of them. With this feature you can specify exactly which violation you are aware of and should be ignored.

I'm not familiar with the tokenizing process and unaware of the myriad of possibilities, so most likely there's something wrong with how I'm traversing them.

So let this serve as a proof of concept at least. The way it works:
```php
<?php
// @codingStandardsIgnoreRule(Generic.Commenting.Todo)
function foo()
{
    //TODO: blah
    echo 'foo!';
}

/**
 * You can also use multiline or doc comments, and the class name instead of the rule name
 * @codingStandarsIgnoreRule(Generic_Sniffs_Commenting_TodoSniff)
 */
function bar()
{
    //TODO: bleh
    echo 'bar!';
}
```

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -233,6 +233,13 @@ class PHP_CodeSniffer_File
     private $_ignoredCodes = array();
 
     /**
+     * An array of rules to ignore for a given token range
+     *
+     * @var array()
+     */
+    private $_ignoredRulesForLines = array();
+
+    /**
      * The total number of errors raised.
      *
      * @var int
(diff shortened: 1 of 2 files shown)


gsherwood:
There is an open feature request for this here: https://github.com/squizlabs/PHP_CodeSniffer/issues/604

So yes, I would consider adding it, and I probably will add it sometime in the future. But I have to first figure out what the syntax should be, and I think I'd like to change the way those ignore comments actually look as well. They are too long and not consistent with other tools that have since added their own syntax for it.

I'd also only put this into the 3.x branch - to be released as 3.0.0 soonish - because the way everything works is completely different and I wouldn't want to write this feature twice.

So unfortunately, I wont merge this PR into PHPCS due to the target version and the fact I want to change the syntax of this feature (a BC break). Thanks for the time you've taken to put it together though. It might be worth popping over to the issue I linked earlier and commenting there about the syntax.

PR #1213, opened 2016-11-14
Include composer.lock

olvlvl:
Hi,

Because PHP CodeSniffer is an application it should include the `composer.lock` file, so that everybody checking a branch will use the exact same dependencies. It's a Composer recommendation: https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file

This is Composer's lock file for instance: https://github.com/composer/composer/blob/master/composer.lock

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 /phpunit.xml
 .idea/*
 /vendor/
-composer.lock
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,1128 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "This file is @generated automatically"
+    ],
+    "hash": "050f971076701f366d7cbc7413a49217",
+    "content-hash": "f097f359dc7041ad5aab6e9964fcf21a",
+    "packages": [],
+    "packages-dev": [
+        {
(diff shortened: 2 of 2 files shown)


aik099:
This particular application has no production dependencies, because it also can be installed through PEAR. Does adding the `composer.lock` give any advantage in that case?

olvlvl:
It only matters when you install the application using Composer.

PatrickRose:
Common misconception about composer.lock is it has an effect for package
consumers - it doesn't. Composer uses composer.json in that case.

It *does* have an effect for package contributions since it forces people
to work with the same dev/production packages.

aik099:
> Common misconception about composer.lock is it has an effect for package
> consumers - it doesn't. Composer uses composer.json in that case.

@PatrickRose , thanks for clarifying that. That was my point exactly.

olvlvl:
Yes, composer.lock has no effect on package consumers, it only affect the application itself, that's the point of my PR. Once you make a release and the dependencies are safely locked inside the `composer.lock` file, anyone cloning the repository and installing the dependencies will have the exact same ones as when the release was made, guaranteeing that the application will execute in the exact same fashion.

gsherwood:
The only dep that will be included in the lock file is the dev dependency on PHPUnit, which means nothing to just about every user of PHPCS. There is no need to commit the lock file for a stand-alone product like PHPCS.

olvlvl:
You know what, I forgot that composer install dev dependencies by default, I didn't realize they were phpunit's, not phpcs'. My bad.

PR #1231, opened 2016-12-08
Tokenise << and >> in JS as bitwise shifts

### LINKED ISSUE #1230: JS tokeniser incorrectly tokenises bitwise shifts as comparison
With code like `4 << 2`, the JS tokeniser incorrectly makes this into `T_LESS_THAN, T_LESS_THAN`. This should instead be `T_SL`. (Ditto for `T_SR` instead of `T_GREATER_THAN, T_GREATER_THAN.)

rmccue:
Previously, these were treated as two comparison operators.

Fixes #1230.

### DIFF
--- a/CodeSniffer/Tokenizers/JS.php
+++ b/CodeSniffer/Tokenizers/JS.php
@@ -202,6 +202,8 @@ class PHP_CodeSniffer_Tokenizers_JS
                               ':'         => 'T_COLON',
                               '<'         => 'T_LESS_THAN',
                               '>'         => 'T_GREATER_THAN',
+                              '<<'        => 'T_SL',
+                              '>>'        => 'T_SR',
                               '<='        => 'T_IS_SMALLER_OR_EQUAL',
                               '>='        => 'T_IS_GREATER_OR_EQUAL',
                               '!'         => 'T_BOOLEAN_NOT',


aik099:
Please also add test case to confirm, that it will work in future.

rmccue:
@aik099 I'd be happy to, but couldn't actually find any existing JS tokeniser tests. Can you point me in the right direction? :)

aik099:
The tokenizers aren't tested directly, but indirectly through sniffs, that are using them. So I guess you need to:

1. find sniff that does something with the `<<` and `>>`
2. in its fixture JS file add sample code with `<<` and `>>` that would cause warning/error from that sniff to be raised
3. add/update fixed fixture file with a fixed version

Since `<<` was treated as 2 `<` before, then there should be 2x errors reported in lines in JS files, where they are present. Just a guess.

gsherwood:
I'll write the tests - it's not a problem.

There is the `>>>`, `>>>=`, `>>=', and `<<=` operators that also needs to be supported, so I'll add those in as well.

gsherwood:
Thanks for the PR. All those operators are now working.

rmccue:
Thanks, apologies for not getting to the tests :)

PR #1238, opened 2016-12-13
Fix "Uninitialized string offset" notice on PHP 5.2

### LINKED ISSUE #1237: Uninitialized string offset in PHP Tokenizer on PHP 5.2
I'm receiving a PHP notice on PHP5.2 with PHPCS 2.7.1/master since commit 531a73143a17098b422d319787473c5279d352f1 which is breaking our builds on PHP5.2

`Notice: Uninitialized string offset: 0 in /tmp/phpcs/CodeSniffer/Tokenizers/PHP.php on line 688`

jrfnl:
Quick & dirty fix for the issue reported in #1237. A better fix might still be needed.

Fixes #1237

### DIFF
--- a/CodeSniffer/Tokenizers/PHP.php
+++ b/CodeSniffer/Tokenizers/PHP.php
@@ -685,7 +685,7 @@ class PHP_CodeSniffer_Tokenizers_PHP
             */
 
             if ($stackPtr > 1
-                && $tokens[($stackPtr - 1)][0] === T_PAAMAYIM_NEKUDOTAYIM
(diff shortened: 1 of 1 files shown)


gsherwood:
I use is_array to check if the tokens a simple or complex, so I used that again in this instance. It's what I should have been doing all along - checking if the previous token was a complex token, and then if it is actually a double colon. But it obviously slipped my mind that, at this point, some tokens are still simple non-array tokens.

Thanks for the report.

jrfnl:
No worries & thanks for adding the fix. I've restarted the builds for open PRs and they are now passing.

PR #1240, opened 2016-12-14
False positive for function names starting with triple underscore

jrfnl:
Function and method names starting with a double underscore are reserved by PHP itself for magic functions/methods.

A triple underscore function name prefix, however, should not trigger this error. (I'm not saying it's a good idea, but it's out there in userland code)

Fixed in the two non-proprietary standards which have checks for double underscored function/method names.
Includes additional unit tests for all four sniffs.

This PR does not cover similar checks in the proprietary PEAR and Squiz standards.

Fixes https://github.com/wimg/PHPCompatibility/issues/297

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/NamingConventions/CamelCapsFunctionNameSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/NamingConventions/CamelCapsFunctionNameSniff.php
@@ -126,7 +126,7 @@ class Generic_Sniffs_NamingConventions_CamelCapsFunctionNameSniff extends PHP_Co
         $errorData = array($className.'::'.$methodName);
 
         // Is this a magic method. i.e., is prefixed with "__" ?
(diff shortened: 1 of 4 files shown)


aik099:
Your code also does cover cases when 3+ (not just exactly 3) underscores are used.

jrfnl:
> Your code also does cover cases when 3+ (not just exactly 3) underscores are used.

Correct, it does.

gsherwood:
Thanks for the fix. I also ended up fixing it in the PEAR and Squiz standards because they were trying to do the same thing.

jrfnl:
@gsherwood You're welcome ;-)

jrfnl:
P.S.: I did originally also add it for the PEAR and Squiz standards (locally), but ended up removing it as the unit tests I added for it were triggering the CamelCase rules for those sniffs, so no matter what, the function name would end up not being acceptable.

PR #1243, opened 2016-12-14
Update composer.json

gmponos:
Not sure but I think this the branch alias should be dev-3.x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "3.x-dev"
+            "dev-3.x": "3.x-dev"
         }
     },
     "autoload": {


gsherwood:
You're not sure. I'm not sure. Nobody has complained. Probably best not to change anything.

gmponos:
What I mostly meant is that I haven't tested it. I thought that you might have some more experience in composer.

I will test it and let you know.

gsherwood:
> I thought that you might have some more experience in composer.

Unfortunately, I have none. The composer.json changes are typically provided by other people, so I have no idea if there is a problem. But nobody has reported one.

Given that 3.0 will be merged into master pretty soon, it may not matter anyway.

gmponos:
> Given that 3.0 will be merged into master pretty soon, it may not matter anyway.

if that's the case then you are correct. It wont matter.

PR #1247, opened 2016-12-17
Allow message type changing from the ruleset to work for complete sniffs

jrfnl:
Currently a `<type>` can be set to change the message type of a message from the ruleset.
If the sniff to which the `type` change is applied passes error codes, this only works on error code level, i.e. `Standard.Subset.Sniff.ErrorCode`, however, if the sniff doesn't pass any error codes, it works on sniff level, i.e. `Standard.Subset.Sniff`.

This seemed inconsistent to me and makes the usage unpredictable as that means that the code needed to set this, is dependent on whether error codes are passed or not.

This PR fixes this inconsistency and makes it so the `<type>` parameter can be set both for individual error codes as well as for complete sniffs.

Ref: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml
> ```xml
> 
>  <rule ref="Generic.Commenting.Todo.CommentFound">
>   <type>error</type>
>  </rule>
>  <rule ref="Squiz.Strings.DoubleQuoteUsage.ContainsVar">
>   <type>warning</type>
>  </rule>

In other words, if this PR would be merged, the following would also become valid:
```xml
 <type>warning</type>
</rule>
```

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -1020,8 +1020,10 @@ class PHP_CodeSniffer_File
         }
 
         // Make sure this message type has not been set to "warning".
-        if (isset($this->ruleset[$sniffCode]['type']) === true
-            && $this->ruleset[$sniffCode]['type'] === 'warning'
+        if ((isset($this->ruleset[$sniffCode]['type']) === true
+            && $this->ruleset[$sniffCode]['type'] === 'warning')
+            || (isset($sniff, $this->ruleset[$sniff]['type']) === true
+            && $this->ruleset[$sniff]['type'] === 'warning')
         ) {
             // Pass this off to the warning handler.
             return $this->_addWarning($error, $line, $column, $code, $data, $severity, $fixable);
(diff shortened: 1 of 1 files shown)


aik099:
> This seemed inconsistent to me and makes the usage unpredictable as that means that the code needed to set this, is dependent on whether error codes are passed or not.

The code needs to be set all the time, because it allows to change specific error properties (e.g. type, message).

gsherwood:
This is something I added to the 3.0 version, but didn't want to backport. From the 3.0.0a1 changelog:
> * You can now set the severity, message type, and exclude patterns for an entire sniff, category, or standard
>    * Previously, this was only available for a single message

There was also this related change:
> * You can now include a single sniff code in a ruleset instead of having to include an entire sniff
>    * Including a sniff code will automatically exclude all other messages from that sniff
>    * If the sniff is already included by an imported standard, set the sniff severity to 0 and include the specific message you want

Also note that sniff codes are no longer optional in 3.0 either. They never should have been, but I left it that way for backwards compatibility, but never threw any warnings to indicate that the code was not set.
> * Sniff codes are no longer optional
>    * If a sniff throws and error or a warning, it must specify an internal code for that message

So the inconsistency here is not actually in PHPCS itself - it's in the sniffs. They should all be using error codes and so there should be no feature that exists to set the error type for an entire sniff at once.

So in 3.0, the inconsistency of sniffs will be removed as the code is enforced, and the feature to set various settings sniff-wide has been properly introduced.

I'm not going to merge this PR in because those sniff-wide settings are something I've been keeping for 3.0 and I want them to be introduced all together. So in this case, I'd suggest setting error codes for all messages that a sniff produces and then changing the message type based on that code. That will also ensure the sniffs work with 3.0.

Does that explain the behaviour a bit better?

jrfnl:
Hi @gsherwood Thanks for the reply.

It's a shame this feature won't be in the 2.x branch too as it's a PITA to add it for a sniff with a lot of different error codes at the moment.

All the same, I'm very glad to hear it will be in PHPCS 3.0, though for some standards it may well be quite a while before those will be converted to work with 3.0 as the changes will completely break compatibility for users of the standard with PHPCS 1.x and 2.x.

> Also note that sniff codes are no longer optional in 3.0 either.

Good thing I've been adding them to the standards I'm involved with in that case ;-)

gsherwood:
Forgot to close this.

PR #1254, opened 2016-12-28
Consider CodeSniffer.conf packaged inside of PHAR

### LINKED ISSUE #949: CodeSniffer.conf packaged in PHAR is ignored
Starting release 2.3.1 (a90531b), CodeSniffer.conf placed inside PHAR is not considered when reading configuration data. The file instead is looked for in the same directory as the *.phar file itself.

The original behavior makes sense in cases when PHP_CodeSniffer is packaged as part of another application inside its own PHAR, and the package contains with the default settings (say I want the default standard to be PSR2 instead of PEAR).

A possible solution is to consider two files: the external file overrides the settings defined in internal file. All changes are saved to the external file.

morozov:
Fixes #949.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -1602,23 +1602,34 @@ class Config
             return self::$configData;
         }
 
-        $path = '';
-        if (is_callable('\Phar::running') === true) {
-            $path = \Phar::running(false);
-        }
+        $paths  = array();
+        $isPhar = false;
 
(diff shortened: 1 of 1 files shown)


gsherwood:
Closing: see https://github.com/squizlabs/PHP_CodeSniffer/issues/949#issuecomment-354935820

PR #1263, opened 2017-01-08
Added new token T_NULLABLE

kukulich:
`?` for nullable type hints is tokenized as `T_INLINE_THEN` now and `:` for return type hints is therefore very often tokenized as `T_INLINE_ELSE`. I've added new token `T_NULLABLE` for `?` before type hints to fix this.


gsherwood:
Thanks a lot for doing this change. I added a few failing unit tests before merging and this fixed them all up.

PR #1264, opened 2017-01-08
Array return type hint is sometimes detected as T_ARRAY_HINT instead of T_RETURN_TYPE

kukulich:


aik099:
Can you please also update tests to cover this behavior?

gsherwood:
I can't replicate the error this is trying to fix. While this change wont break anything, I would like to add a unit test to expose the problem. Can you provide some sample code to expose the issue please.

kukulich:
@gsherwood I've just prepared best fix. Here is the code to reproduce the error:

```php
<?php

function foo() : array
{
	$foo;
}

```

gsherwood:
> Here is the code to reproduce the error

Thanks. I thought I tried that already, but I must have stuff up the syntax or something. Easy to replicate it with that code.

gsherwood:
Your first fix was much simpler, and it worked fine. Why did you change it?

kukulich:
@gsherwood The previous fix was fixing only the symptoms, now I've tried to fix the place where the bug really begin. However if you are ok with the simpler fix, I can change the pull request again :)

gsherwood:
> However if you are ok with the simpler fix, I can change the pull request again :)

Yep, I'm ok with the simple fix. It just keeps the return type detection logic in the one place.

I know what you were trying to do with the second patch (I do the same thing sometimes) but I think you were right the first time on this one.

kukulich:
Ok, changed again :)

gsherwood:
Thanks for changing it back, and thanks for the fix.

Issue #1267, opened 2017-01-11
Fixer incorrectly handles filepaths with repeated dir names


orx0r:
Fixer incorrectly handles filepath, e.g.:

For example, if filePath is `/app/path/app/file.php` and cwd is `/app/`, expected filepath should be `path/app/file.php`
But in fact, filepath became wrong - `pathfile.php`

Because `str_replace` replaces all occurrences in target filepath:
https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Fixer.php#L246
```php
$cwd      = getcwd().DIRECTORY_SEPARATOR;
$filename = str_replace($cwd, '', $filePath);
```

orx0r:
#1269

PR #1269, opened 2017-01-11
Fixed bug #1267 : Fixer incorrectly handles filepath

orx0r:
Fixer incorrectly handles filepath, e.g.:

For example, if filePath is `/app/path/app/file.php` and cwd is `/app/`, expected filepath should be `path/app/file.php`
But in fact, filepath became wrong - `pathfile.php`

Because `str_replace` replaces all occurrences in target filepath:
https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Fixer.php#L246
```php
$cwd      = getcwd().DIRECTORY_SEPARATOR;
$filename = str_replace($cwd, '', $filePath);
```

### DIFF
--- a/CodeSniffer/Fixer.php
+++ b/CodeSniffer/Fixer.php
@@ -243,8 +243,13 @@ class PHP_CodeSniffer_Fixer
             $filePath = $this->_currentFile->getFilename();
         }
 
-        $cwd      = getcwd().DIRECTORY_SEPARATOR;
-        $filename = str_replace($cwd, '', $filePath);
+        $cwd = getcwd().DIRECTORY_SEPARATOR;
+        if (strpos($filePath, $cwd) === 0) {
+            $filename = substr($filePath, strlen($cwd));
+        } else {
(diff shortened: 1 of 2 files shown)


gsherwood:
The fix looks good, but I can't merge it in unless you remove the use of the `mb_substr` function. This would add a requirement to PHPCS for the MB extension, but it's also not needed in this case, and the mixing of mb_substr and strlen would cause issues as well.

orx0r:
@gsherwood now use simply substr and it seems well

aik099:
I understand what path handling was fixed, but can you please post here:

1. path combinations that caused issue to happen
2. what exactly was the issue more specifically

?

aik099:
My bad. Haven't noticed connected issue with all the details. It's a good practice to specify issue in PR description, not other way around 😉

orx0r:
@aik099 Ok, thanks!

gsherwood:
Thanks for doing that for me. It's working well.

PR #1274, opened 2017-01-14
Sniff: fix File param name

TomasVotruba:


gsherwood:
Given the file's namespace, the comment would need the fully qualified class name instead of `File`, so I couldn't merge this in.

But this is just one of more than 200 places where the old class name was used. I didn't get around to replacing them all before, but I've committed that change now. I had to change some fixers, and add a new one, to get it automated because I didn't want to align all the comments by hand.

Thanks for prompting me to fix these comments.

TomasVotruba:
Thank you!

Issue #1280, opened 2017-01-18
Uninitialized string offset in PHP Tokenizer on PHP 5.2


jrfnl:
I'm receiving a PHP notice on PHP5.2 with PHPCS master since commit https://github.com/squizlabs/PHP_CodeSniffer/commit/b2b4b2861972280d3217800f03fb033415bdd7a6 which is breaking our builds on PHP5.2

Notice: Uninitialized string offset: 0 in /tmp/phpcs/CodeSniffer/Tokenizers/PHP.php on line 690
Notice: Uninitialized string offset: 0 in /tmp/phpcs/CodeSniffer/Tokenizers/PHP.php on line 694

gsherwood:
This sort of thing is typically caused by not checking if the token is an array or not. I've just added that check to try it out, but there are so many 5.2 failures that I can't really tell if it did anything. Can you try that code out for me please?

jrfnl:
Thanks for that! Your fix looks a lot simpler than mine ;-)

I've restarted the test build I created & linked to in #1283 which at the time was failing against master & passing against the PR branch.

https://travis-ci.org/jrfnl/PHPCompatibility/builds/193080359

Looks like it's passing against both now, so as far as I can see, that part is fixed now.

gsherwood:
Thanks for getting back to me.

PR #1281, opened 2017-01-18
Let's start testing the build on PHP 5.2 to avoid further breaks by PHP notices

jrfnl:
And as expected and reported in #1280, the build for PHP 5.2 is currently breaking.

[Edit] Just looked through the details of the breaking build and most of the issues seem to be to do with the `\` `T_NS_SEPARATOR` not being recognized in PHP 5.2, haven't found one demonstrating the current issue yet....

How would you like me to proceed with this ? Exclude namespace related tests for PHP 5.2 ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 language: php
 
 php:
+  - 5.2
   - 5.3
   - 5.4
   - 5.5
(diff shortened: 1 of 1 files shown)


gsherwood:
I could have sworn 5.2 wasn't available on Travis. Thanks for putting this in.

jrfnl:
They took it out briefly, but as WordPress still has a minimum requirement of PHP 5.2 and is such a big player, they brought it back very quickly....

gsherwood:
Looks like a lot more than those notices are breaking on 5.2. Seems I haven't emulated the NS separator for that version and that's causing all sorts of breaks. How painful :(

jrfnl:
Correct, the NS separator is the main issue for PHPCS builds breaking on 5.2 and that will hide any other issue for the moment.

gsherwood:
I'm going to have to revert this because PHPCS is never going to work properly on PHP 5.2 unless you are only using PHP 5.2 syntax.

When a NS separator appears in content, token_get_all emits a warning and completely removes that content from the token array, so there is no way for PHPCS to get it back unless it manually tokenizes PHP code itself (like it does for JS), which I'm not going to do.

Example here: https://3v4l.org/0lp1i

Code being tokenized is:
```php
<?php
class testFECNNamespacedClass extends \testFECNClass {}
```
It should look like this:
```
0 : T_OPEN_TAG => <?php\n
1 : T_CLASS => class
2 : T_WHITESPACE =>  
3 : T_STRING => testFECNNamespacedClass
4 : T_WHITESPACE =>  
5 : T_EXTENDS => extends
6 : T_WHITESPACE =>  
7 : T_NS_SEPARATOR => \
8 : T_STRING => testFECNClass
9 : T_WHITESPACE =>  
10 : {
11 : }
```
But it looks like this:
```
Warning: Unexpected character in input:  '\' (ASCII=92) state=1 in /in/0lp1i on line 6
0 : T_OPEN_TAG => <?php\n
1 : T_CLASS => class
2 : T_WHITESPACE =>  
3 : T_STRING => testFECNNamespacedClass
4 : T_WHITESPACE =>  
5 : T_EXTENDS => extends
6 : T_WHITESPACE =>  
7 : T_STRING => testFECNClass
8 : T_WHITESPACE =>  
9 : {
10 : }
```
I'd don't really expect the NS sep to appear as a token, but I'd at least expect it to be tacked on the front of the class name (token 7). But as that would be invalid, PHP appears to throw it out instead.

So I'm going to remove the 5.2 test again because I just don't have the time to go through all the tests and add exceptions for 5.2. Plus, PHPCS 3 wont support 5.2, so that's also not something that needs to be taken forward.

jrfnl:
I understand. I have to admit, I was actually a bit surprised this was merged while the build was failing with it :-)

Anyway, I'll just keep an eye on any changes in the Tokenizer class and will let you know (again) if we see breakage on PHP 5.2.

gsherwood:
> I have to admit, I was actually a bit surprised this was merged while the build was failing with it :-)

I can't get a 5.2 build running locally, so that's basically the only way for me to figure out what is broken. But the answer just happened to be "too much".

> Anyway, I'll just keep an eye on any changes in the Tokenizer class and will let you know (again) if we see breakage on PHP 5.2.

Thanks.

jrfnl:
FYI: Just came across another token which is does not seem to be backported to PHP 5.2: T_NOWDOC and the start/end variants.

PR #1282, opened 2017-01-18
Fix phpcs: env: php\r: No such file or directory

qRoC:


gsherwood:
What exactly is this change doing?

scripts/phpcbf
```diff
@@ -1,23 +1,46 @@
 #!/usr/bin/env php
+
```
qRoC: fixed, sorry :)

aik099:
Not sure why empty line between shebang and `<?php` does fix it.

Maybe you can post exact error message you're getting? I suppose you've posted only a fragment of it as PR title.

aik099:
So empty line at the end of a file solves that problem? Still not getting how this could fix anything. Maybe that file had mixed line endings in it and by adding/removing line you fixed that?

qRoC:
Just replacing "\r" to "\n" on first line fixes this error.

gsherwood:
I still don't know what error you are trying to solve. Please take some time to write a comment about what this PR is trying to fix and how it is fixing it.

You also mention a `\r` line ending, but these files only have `\n` line endings, which I've just reconfirmed. Plus, this PR doesn't currently change the line ending on the first line anyway.

gsherwood:
Closing as I still don't know what this PR is for.

Issue #1278, opened 2017-01-17
Improve discoverability of configurable sniffs / public member vars


jrfnl:
Currently to find out if a sniff has any configuration options, one has to look at the sniff files themselves.
In my experience, this results in a lot of people not knowing about the fact that some sniffs can be configured, let alone, using these features.
And for people who have PHPCS installed as a phar file, looking at the source code is not even an option (unless they look at it here).

I believe a wiki page could help improve the discoverability.

Would you interested in adding a wiki page with a list of the current properties per standard ?

I'd be willing to put in the work to try and gather the information for the page.

Related to #1140

gsherwood:
I've wanted to do this for a while, ideally using a script, but I was also planning that the Doc Generator stuff could be used as well. Manually putting it all together and making it useful with examples would be a big task, but it would be a great resource and I'd certainly add it to the wiki.

But I think it's something that really requires a full doc of every included sniff, what they do, how to include them in a ruleset, how the configuration options change that behaviour, and examples of how to set those options in a ruleset. It's a big job.

jrfnl:
> It's a big job.

I'm fully aware of this, but am willing to put in the time. I might not be able to get everything done in one go, but will keep you informed of progress.

What about if I put work in progress in a gist for feedback (and easy copy/paste to the wiki) ?

I've just finished preparing the same for the WP coding standards (not yet online as I've been cleaning things up which I came across while doing it, waiting for the merges before publishing).
If you like, I'll put a link up to the end result of the WP version once it goes online.

jrfnl:
Update: That was surprisingly less painful than expected. I've finish creating the list, will tidy up later today or tomorrow and put a draft online.

jrfnl:
Ok, tidying up took a bit more time, but you can find the page I've prepared for the wiki here:
https://gist.github.com/jrfnl/ff1555af507a03f3df00462b94d029ab

It's completely up to date up to the latest commit https://github.com/squizlabs/PHP_CodeSniffer/commit/8120537a8967c3135a2f01db663d11fc2890be2c and already includes the proposed new properties as per PR #1318

gsherwood:
Oh wow! You're amazing.

I got inundated with PHPCS issues over the last few days so I haven't looked at this properly yet, but I promise I'll get this committed as soon as I can.

Thanks so much.

gsherwood:
[...]

PR #1287, opened 2017-01-19
Squiz: property visibility

jrfnl:
While working on the list of public properties which can be changed from rulesets (#1278), I came across two properties in the Squiz sniffs for which it seemed unlikely that these were intended to be public.


gsherwood:
I think this makes sense.

PR #1288, opened 2017-01-19
Generic/DuplicateClassName: fix property visibility

jrfnl:
While working on the list of public properties which can be changed from rulesets (#1278), I came across this property for which it seemed unlikely that it was intended to be public.

This is a complex property which expects a multi-level array to be set based on the classes found by the sniff.


jrfnl:
FYI: build failure is an issue with travis & should be fixed by restarting the build. It's not an issue with the PR.

gsherwood:
Yep, I'm pretty sure that's just a mistake.

PR #1289, opened 2017-01-19
Generic / ByteOrderMark: property visibility

jrfnl:
While working on the list of public properties which can be changed from rulesets (#1278), I came across this property in the `Generic_Sniffs_Files_ByteOrderMarkSniff` for which it seemed unlikely that it was intended to be public.


gsherwood:
Makes sense to me. I can't see a reason to ever need this as a setting.

PR #1291, opened 2017-01-19
AbstractVariableSniff: detect variables in anonymous classes.

jrfnl:


aik099:
I think you should also check other sniffs listening for `T_CLASS` token to see if they also need to listen for `T_ANON_CLASS` token.

jrfnl:
> I think you should also check other sniffs listening for T_CLASS token to see if they also need to listen for T_ANON_CLASS token.

That's like saying that for the spelling PRs I just send in, I should have checked every line of documentation in every file to be sure I'd fixed all spelling errors....

I PRed what I came across and as it's in an abstract, it will fix the issue for a number of sniffs. I'm not claiming completeness.

aik099:
Not that bad 😄 . Just look for `T_CLASS` mentions within `register` methods of the sniffs.

gsherwood:
To test this, I added anon class support to a couple of sniffs that only check for classes, and uncovered a few problems in the PHP_CodeSniffer_File class, which I also fixed. So hopefully anon class support is a lot better now.

Thanks for the PR.

jrfnl:
Awesome! Thanks ;-)

PR #1293, opened 2017-01-20
getMethodParameters(): Allow for closures.

jrfnl:
The method itself works perfectly for closures, it just didn't allow itself to be used for them.

Includes allowing for closures in two sniffs which use the `getMethodParameters()` method.
Includes additional unit test(s) for these sniffs.

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -2742,7 +2742,7 @@ class PHP_CodeSniffer_File
 
 
     /**
-     * Returns the method parameters for the specified T_FUNCTION token.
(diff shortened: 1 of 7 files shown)


gsherwood:
Works great. Thanks a lot.

jrfnl:
You're very welcome.

N.B.: saw the changelog - for sniff devs it might be interesting to know that the underlying method `getMethodParameters()` has changed as well.

gsherwood:
> for sniff devs it might be interesting to know that the underlying method getMethodParameters() has changed as well.

Good idea. I've updated it.

PR #1295, opened 2017-01-20
Replace < and > with &t; and &gt; on output

dingo-d:
When used in a plugin (https://github.com/ernilambar/ns-theme-check/) if characters are not written as character entities, they will get rendered which breaks the error report looks. This prevents it.

### DIFF
--- a/CodeSniffer/Reports/Full.php
+++ b/CodeSniffer/Reports/Full.php
@@ -178,8 +178,8 @@ class PHP_CodeSniffer_Reports_Full implements PHP_CodeSniffer_Report
 
                         echo '] ';
                     }
-
-                    echo $errorMsg.PHP_EOL;
+                    
+                    echo str_replace( '<', '&lt;', str_replace( '>', '&gt;', $errorMsg ) ).PHP_EOL;
                 }//end foreach
             }//end foreach
         }//end foreach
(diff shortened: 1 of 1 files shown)


CodeSniffer/Reports/Full.php
```diff
@@ -178,8 +178,8 @@ public function generateFileReport(
 
                         echo '] ';
                     }
-
-                    echo $errorMsg.PHP_EOL;
+
```
aik099: Please remove trailing spaces.

CodeSniffer/Reports/Full.php
```diff
@@ -178,8 +178,8 @@ public function generateFileReport(
 
                         echo '] ';
                     }
-
-                    echo $errorMsg.PHP_EOL;
+                    
+                    echo str_replace( '<', '&lt;', str_replace( '>', '&gt;', $errorMsg ) ).PHP_EOL;
```
aik099: Please use `htmlspecialchars($errorMsg, ENT_QUOTES, 'UTF-8', false)` function instead.

gsherwood:
The full report is a human readable report, and also the default. I'm not going to replace human readable characters with HTML entities in this report. If you want to programmatically read PHPCS output, use something like the JSON, XML or CSV report types and then display the output however you want.

dingo-d:
Ok, I'll notify the plugin author. Thanks for the explanation.

PR #1297, opened 2017-01-20
NestingLevel: minor documentation fix.

jrfnl:
Add two missing words.


gsherwood:
Thanks for fixing those docs

PR #1299, opened 2017-01-23
Package.xml: Remove reference to non-existent file.

jrfnl:
Noticed this error in the Travis build.

`Error: File "./CodeSniffer/Standards/PEAR/Tests/Commenting/FunctionCommentUnitTest.inc.fixed" in package.xml does not exist`


gsherwood:
Actually, that file should exist - but seems to have been blown away before I committed it or something. Was supposed to be part of this commit: https://github.com/squizlabs/PHP_CodeSniffer/commit/7e92909a0e3e43cbda69a6b0f76c02c597a9f696

So I'll close this, and recreate the file instead. Thanks for pointing it out.

PR #1301, opened 2017-01-23
Remove two superfluous checks for fixer->enabled.

jrfnl:
These are no longer needed since commit 5def2acbe3911e2aea08ac8b8eb4e4d64330021f / `addFixableError() and addFixableWarning() now only return true if the fixer is enabled`


gsherwood:
Nice catch. Must have been some of the very first fixer code there. Thanks.

PR #1303, opened 2017-01-24
allows for closures

uniquexor:
Checks for calling closures and other methods that have their own
tokens, like unset()


uniquexor:
It seems that there is a configuration error in CI, thus php 7 test errored :/

gsherwood:
Thanks a lot for the PR

PR #1305, opened 2017-01-25
Fixed: comments are no longer removed

uniquexor:
Used to remove everything between a semicolon and a nonempty token. Now it jumps over the comments leaving them as is only removing whitespaces between a comment and a semicolon. For exmple, if we have code like this:
`$test = $a /* + $b */  ;`
Before the comment would be removed:
`$test = $a;`
Now the comment remains:
`$test = $a; /* + $b */`

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/Functions/FunctionCallArgumentSpacingSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Functions/FunctionCallArgumentSpacingSniff.php
@@ -40,7 +40,7 @@ class Generic_Sniffs_Functions_FunctionCallArgumentSpacingSniff implements PHP_C
     {
         $tokens = PHP_CodeSniffer_Tokens::$functionNameTokens;
 
-        // for calling closures
+        // For calling closures.
         $tokens[] = T_VARIABLE;
 
         return $tokens;


aik099:
Can you maybe rephrase so that PR has:

- short title that explains what is done
- description that explains in detail if needed

Maybe add some examples describing what is changed.

uniquexor:
Sorry. Have updated the name and description.

aik099:
It obviously changes the placement of the comment itself (was before `;` but now it's after `;`). Is that acceptable?

uniquexor:
I think it is more acceptable than just removing it entirely, no?

aik099:
Yes, but people who were expecting old fixing behavior might be surprised to see how it will work after a fix. I personally don't place `// ...` comments on same line with a code. Just a guessing here really.

uniquexor:
Neither do I, but I have colleagues, who do :) I can imagine a situation where some code has been commented out and after running a beautifier it would be lost. For example, imagine a situation like this:
`$c = $b /* + $c */ ;`
If someone, for some reason temporarily commented out the code, it would get lost after running this sniff.
In my opinion running a code beautifier, should not result in a loss of code, even if it's commented out.

aik099:
That's more close to life example. You'd better also mention `/* ... */` in your unit tests to confirm, that it isn't removed.

uniquexor:
What do you mean? There is a case for `/* ... */` comment in both JS and PHP unit tests.

aik099:
Ah, then I must have missed that. Sorry. Then all is fine.

aik099:
I've meant exactly what you've shown, in tests:

```php
$c = $b /* + $c */ ;
```

There are only examples with comment coming after `;` before fixing.

uniquexor:
Well, there is a line in PHP test:
`$test = $this->testThis()  /* comment here */  ;`
and in JS test:
```
alert('hi')   /* Comment between */
;
```
Both, in my opinion, cover your case

uniquexor:
Sorry, I'm new to git... The last commit should have probably been a new pull request. Is there a way for me to remove it from here?

jrfnl:
@uniquexor You can hard reset your branch to the previous commit. This will cause you to lose the change though. If you're working on several issues, it might be useful to start working in separate branches for each.

PR #1310, opened 2017-01-25
Squiz / FunctionComment: fix detection of incorrect types when @return tag has description

### LINKED ISSUE #1018: The Squiz.Commenting.FunctionComment sniff doesn't detect incorrect types when @return tag has description
Example code:

``` php
/**
 * Return description function + mixed return.
 *
 * @return string|int This is a description.
 */
public function myFunction() {
    return 5;
}
```

In above example the replacement of `string|int` into `string|integer` should be suggested, but this doesn't happen because instead of `int` the `int This is a description.` is parsed as data type.

Fixing this can be problematic, because on of supported data type format includes spaces: `array(int => string)` for example.

Most likely `PHP_CodeSniffer::suggestType` method needs to be changed to allow understanding that case OR a code added to separate description from return type as done for `@param` tag.

Related issue: #1015

jrfnl:
Also incidentally fixes a _"Function return type is not void, but function has no return statement"_ error being thrown when it shouldn't be when using multiple types including void, i.e. `array|void`.

Includes unit tests specific to this issue in a separate file + fixed file.

Fixes #1018

Note: the unit tests are in a separate file as the "main" unit test file for this sniff does not have a `.fixed` version and adding that is currently not an option as there are a number of other bugs in the sniff.
Once all the bugs are solved & a `.fixed` file for the main unit test file can be added, the tests for this issue can be added to the main file.

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -81,8 +81,16 @@ class Squiz_Sniffs_Commenting_FunctionCommentSniff extends PEAR_Sniffs_Commentin
                 $error = 'Return type missing for @return tag in function comment';
                 $phpcsFile->addError($error, $return, 'MissingReturnType');
             } else {
+                // Support both a return type and a description.
+                $split = preg_match('`^((?:\|?(?:array\([^\)]*\)|[\\\\a-z0-9\[\]]+))*)( .*)?`i', $content, $returnParts);
+                if (isset($returnParts[1]) === false) {
+                    return;
+                }
+
+                $returnType = $returnParts[1];
+
                 // Check return type (can be multiple, separated by '|').
-                $typeNames      = explode('|', $content);
+                $typeNames      = explode('|', $returnType);
                 $suggestedNames = array();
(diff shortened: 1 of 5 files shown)


jrfnl:
FYI: I'm preparing follow up PRs for several of the other fixer bugs in this sniff, but will pull them one by one to prevent continuous merge conflicts.

fabacino:
Looking forward to this fix :)

gsherwood:
Thanks for this PR. It worked great.

I ended up merging the unit tests into the main file and adding a `.fixed` file for it. There were only a couple of bugs exposed, which I fixed in this commit: https://github.com/squizlabs/PHP_CodeSniffer/commit/4079d5fdbc8917861680fc99fc26584bcf5e80d8

jrfnl:
@gsherwood thanks for merging & fixing this. I did already have the fixes ready & waiting though, which could have saved you some work (see [comment above](https://github.com/squizlabs/PHP_CodeSniffer/pull/1310#issuecomment-275259284)) ;-)
I'll run a check to see if all the things I found have been addressed and if anything is still left over from my original fixes, I'll send you a separate PR.

gsherwood:
> I'll run a check to see if all the things I found have been addressed and if anything is still left over from my original fixes, I'll send you a separate PR.

:) thanks

PR #1315, opened 2017-01-27
Add some configuration options to Squiz.Commenting.FunctionComment

fabacino:


CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -33,6 +33,34 @@ class Squiz_Sniffs_Commenting_FunctionCommentSniff extends PEAR_Sniffs_Commentin
...
+     * @var boolean
+     */
+    public $paramCommentCapitalStart = true;
+
+    /**
+     * Check that param comments end with the full stop.
+     *
+     * @var boolean
+     */
+    public $paramCommentFullStop = true;
```
aik099: Isn't this a requirement by PhpDoc?

CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -33,6 +33,34 @@ class Squiz_Sniffs_Commenting_FunctionCommentSniff extends PEAR_Sniffs_Commentin
...
+     * @var int
+     */
+    public $spacingAfterParamName = 1;
+
+    /**
+     * Check that param comments start with a capital letter.
+     *
+     * @var boolean
+     */
+    public $paramCommentCapitalStart = true;
```
aik099: Isn't this a requirement by PhpDoc?

CodeSniffer/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
fabacino: AFAIK a full stop is not a mandatory requirement, but I think it is useful to make this setting customizable in either way, as there are many projects who do not use them (e.g. Symfony, CodeIgniter, etc.)

gsherwood:
A couple of notes:

1. The two new boolean properties shouldn't be needed. Sniffs don't need settings to hide errors because you can just exclude the error message code itself inside a ruleset.xml file. Given you're already in there setting this config option, you're better off letting the sniff generate the error, but reducing its severity (to 0 if you want to hide it completely) or changing it to a warning.

2. I'm not sure why someone would want to configure the number of spaces after param types and names. Doing so would render a few other related sniffs broken because they assume you are trying to align values and have a min of 1 space to shorten lines. Is there a good reason to have multiple spaces? Is there an existing published coding standard that requires this?

fabacino:
Thanks for your feedback.

1. Makes perfect sense, I will remove the boolean properties.
2. I think it aids readability, especially if multi line descriptions are involved.

```php
/**
 * @param integer $index Some Index as the first argument
 * @param integer $value The value as second argument
 * @param integer $error Error as third argument with an even longer

 * @param integer  $index  Some Index as the first argument
 * @param integer  $value  The value as second argument
 * @param integer  $error  Error as third argument with an even longer
 **/
```

```php
/**
 * @param integer $index Index as first argument
 * @param integer $value Value as second argument with some other
 *      properties, needing multiple lines to properly describe it.
 * @param integer $error Error as third argument with an even longer
 *      description, so again, we must break it onto multiple lines.

 * @param integer  $index  Index as first argument
 * @param integer  $value  Value as second argument with some other
 *      properties, needing multiple lines to properly describe it.
 * @param integer  $error  Error as third argument with an even longer
 *      description, so again, we must break it onto multiple lines.
 **/
```

This is, of course, a completely subjective matter. Currently I have to extend `Squiz_Sniffs_Commenting_FunctionCommentSniff` and overwrite the whole `processParams` function to get this behavior. Since copying some 350 loc just to change two integers does not seem like a very good idea, I changed them into properties. I didn't notice any problems when checking against the Squiz standard, but if this breaks other sniffs, I will try to refactor the checking into a separate function with an optional parameter for the number of spaces which defaults to 1.

gsherwood:
@fabacino Sorry, I wasn't looking for you to change everything around - just asking for an opinion :)

But if this change makes life easier for you, I'm very happy to merge it in. If you'd prefer to have config options for the number of spaces, it's not going to break anything and so I'd merge that in as well. Either way, breaking out the code into these new methods looks like a good change.

Have a think about you want merged and let me know.

fabacino:
@gsherwood No problem at all. For me, either way works fine, but I'd go with the separate methods. Using multiple spaces seems uncommon at best and probably does not deserve its own config options. If it turns out that other people find this useful as well, we can still add them at a later time.

gsherwood:
> No problem at all. For me, either way works fine, but I'd go with the separate methods.

Too easy then. I've merged it in and it will be released in the next 2.x and 3.x versions. Thanks a lot for the PR.

PR #1320, opened 2017-01-28
Issue #1319: findnext inconsistency, option 1

### LINKED ISSUE #1319: Inconsistent behaviour findPrevious() vs findNext()
I've been tripped up by this a number of times already, so putting this out here for consideration.

If the `$start` and `$end` token positions you pass to `findPrevious()` are the same, the function will work as expected. However, the same can not be said about the `findNext()` method as in that case it will always return `false`.

This is caused by the `findPrevious()` method treating the end position as *inclusive* `for ($i = $start; $i >= $end; $i--)`, while the `findNext()` method treats it as **exclusive** `for ($i = $start; $i < $end; $i++)`.

There are two possible solutions for this:
* Either check for `$start` and `$end` being the same and adding 1 to the `$end` in that case before entering the loop. This will lead to the least difference with current behaviour. - PR #1320
* Or make the loop in `findNext()` inclusive as well. This might lead to more breakage, but would make the functions more consistent in use. - PR #1321

jrfnl:
Fixes #1319 with the smallest possible BC break.

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -3405,6 +3405,10 @@ class PHP_CodeSniffer_File
             $end = $this->numTokens;
         }
 
+        if ($start === $end) {
+            ++$end;
+        }
+
         for ($i = $start; $i < $end; $i++) {
             $found = (bool) $exclude;
             foreach ($types as $type) {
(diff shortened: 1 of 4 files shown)


gsherwood:
I'm going to close the two PRs that provide options for #1319 because I only want to make a change like this in 3.x, and I'm not entirely sure which change yet. Discussion can continue over on the main issue.l

jrfnl:
@gsherwood Would you like me to pull these PRs against the v3 branch ?

gsherwood:
> Would you like me to pull these PRs against the v3 branch ?

No thanks. I don't see any reason for a PR until there is a discussion about what change would be made. And I'm not going to make any changes before the 3.0 release anyway.

jrfnl:
So what would you need to help you decide ? And what can I do to help ?

As this - as you correctly pointed out before - would be a BC break, this feels like something which should be in the 3.0.0 version and not in a later 3.x.

gsherwood:
> this feels like something which should be in the 3.0.0 version and not in a later 3.x

It's just not going to be in 3.0. I'm not making a change that would require even more upgrades and another RC. I need to get it released, so this change is basically on hold for the time being.

jrfnl:
On hold until version 4 in that case ?

gsherwood:
> On hold until version 4 in that case ?

Maybe

PR #1321, opened 2017-01-28
Issue #1319: findnext inconsistency, option 2

### LINKED ISSUE #1319: Inconsistent behaviour findPrevious() vs findNext()

jrfnl:
Fixes #1319 with full consistency between behaviour of the `findNext()` and `findPrevious()` methods, but with a larger BC break.

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -3401,11 +3401,11 @@ class PHP_CodeSniffer_File
     ) {
         $types = (array) $types;
 
-        if ($end === null || $end > $this->numTokens) {
-            $end = $this->numTokens;
+        if ($end === null || $end >= $this->numTokens) {
+            $end = ($this->numTokens - 1);
         }
 
-        for ($i = $start; $i < $end; $i++) {
+        for ($i = $start; $i <= $end; $i++) {
             $found = (bool) $exclude;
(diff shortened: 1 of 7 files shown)


gsherwood:
I'm going to close the two PRs that provide options for #1319 because I only want to make a change like this in 3.x, and I'm not entirely sure which change yet. Discussion can continue over on the main issue.l

jrfnl:
@gsherwood Would you like me to pull these PRs against the v3 branch ?

PR #1322, opened 2017-01-29
Avoid a fatal error when calling a method on bool

antograssiot:
If the file was already included, `include_once` will return `true` and the function call on  the boolean will throw a fatal error
`PHP Fatal error:  Uncaught Error: Call to a member function unregister() on boolean``

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -62,7 +62,11 @@ if (class_exists('PHP_CodeSniffer\Autoload', false) === false) {
                     && file_exists(__DIR__.'/../../autoload.php') === true
                 ) {
                     self::$composerAutoloader = include_once __DIR__.'/../../autoload.php';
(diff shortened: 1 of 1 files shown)


gsherwood:
This is something I'm actively working on and trying out some test cases for. If the autoloader has already been included, I need to try and get it another way or I can't use it to provide proper custom sniff autoloading.

So I wont merge this PR in because I'm trying to find a fix for the problem, but thank you very much for submitting it and reporting the issue. It's another test case for me to try.

gsherwood:
Just an FYI that I ended up using `instanceof` in the fix for this, but I needed to use `\Composer\Autoload\ClassLoader` instead of `Composer\Autoload\ClassLoader` to get it working. Otherwise it was always `false`.

antograssiot:
Great thx for the feedback.

PR #1323, opened 2017-01-30
Create WebCompat.php report file

dingo-d:
This report will return Full report but suitable for web based output (e.g. in plugins or web apps).

### DIFF
--- a/CodeSniffer/Reports/WebCompat.php
+++ b/CodeSniffer/Reports/WebCompat.php
@@ -32,7 +32,7 @@
  * @version   Release: @package_version@
  * @link      http://pear.php.net/package/PHP_CodeSniffer
  */
-class PHP_CodeSniffer_Reports_Full implements PHP_CodeSniffer_Report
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for the submission, but I don't think this report belongs in PHPCS. It's not in a format that I believe will be useful to a wide range of people given it's just the full report with htmlspecialchars.

If you need this report for a project, you can specify the path to the report like this: `phpcs --report=/path/to/WebCompat.php /path/to/code`

dingo-d:
Well I commented in the code, that it could be useful for any plugins - WordPress is currently working with phpcs for checking the coding standards in the themes, and the plan is to have a plugin that will check the theme from within WordPress. Also one could build web apps for checking code standards online as well. Thought that it could be useful to cover that part.

jrfnl:
@dingo-d You can just as easily use the existing checkstyle or xml report formats for that and combine those with an XSL stylesheet.

dingo-d:
I'll have that in mind if I'll do something like that in the future :)

PR #1324, opened 2017-01-30
Allow failures for PHP nightly.

jrfnl:
Currently PHPUnit is failing on PHP nightly and therefore the build is failing.

This change will prevent complete builds failing for failures which occur only in PHP nightly.

FYI: the underlying issue in PHPUnit / Comparator has already been addressed but is not contained in the version used by Travis yet.
Ref: https://github.com/sebastianbergmann/phpunit/issues/2453
Ref: https://github.com/sebastianbergmann/comparator/pull/30

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,13 @@ php:
   - hhvm
   - nightly
 
+matrix:
+  fast_finish: true
+
+  allow_failures:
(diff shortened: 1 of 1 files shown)


gsherwood:
Good idea. Thanks.

PR #1326, opened 2017-01-31
Fixed: comments are no longer removed

uniquexor:
Used to remove everything between a semicolon and a nonempty token. Now it jumps over the comments leaving them as is only removing whitespaces between a comment and a semicolon. For exmple, if we have code like this:
`$test = $a /* + $b */ ;`
Before: the comment would be removed:
`$test = $a;`
Now: the comment remains:
`$test = $a; /* + $b */`

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/SemicolonSpacingSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/SemicolonSpacingSniff.php
@@ -88,9 +88,14 @@ class Squiz_Sniffs_WhiteSpace_SemicolonSpacingSniff implements PHP_CodeSniffer_S
         $fix = $phpcsFile->addFixableError($error, $stackPtr, 'Incorrect', $data);
         if ($fix === true) {
             $phpcsFile->fixer->beginChangeset();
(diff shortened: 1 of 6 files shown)


aik099:
Don't send PR from `master` branch, because sending another PR would mix commits between them.

And this is probably variation of #1305.

uniquexor:
Yup, that was the reason, why I closed that request. I've unintentionally added a new commit to it. I've since recreated everything with the right commits and in the future I will not use master branch. Sorry for the confusion I've caused... New to git and just wanted to contribute to this wonderful project. :/

aik099:
But now you're made same mistake again: sent PR from `master` branch. You should:

1. create topic branch (usually named to indicate what changes as included)
2. push that branch to your repo's fork
3. send PR from that pushed branch

uniquexor:
I understand it now and as I've said, I won't do the same mistake the third time (I hope :D). 
Or should I close this request and do another after creating a new branch?

aik099:
If you promise not to open another PR from same `master` branch, then of course you can keep this PR as-is.

uniquexor:
I promise :D

gsherwood:
@uniquexor Thanks for this. Really good work.

Issue #1312, opened 2017-01-26
Squiz.WhiteSpace.ScopeKeywordSpacing not respect `final` keyword


evilebottnawi:
Not error if i use:
```php
class MyClass
{
    final  public function test() // Here two spaces after final keywork
    {
        return 'foo-bar';
    }
}
```
But `final` not scope, it is keyword. Maybe related to another sniff.

aik099:
That sniff only checks space between of scope and `function` word.

Using that logic there should different sniff that check spacing for `final` and `abstract` keywords in function declaration. I have no if there are.

aik099:
There is `PSR2.Methods.MethodDeclaration` sniff that does something with `final` word, but it does a lot of other stuff as well.

evilebottnawi:
@aik099 maybe new sniff?

aik099:
Sure. Something that check keyword spacing before function keyword. And their order maybe.

gsherwood:
Closing as I've had no additional interest or PRs for this.

jrfnl:
I'm actually interested in a sniff like this - which would check all keywords - and am willing to create it.
Should be easy enough as most of the logic is already available in the aforementioned sniff.

PR #1329, opened 2017-02-01
Add new sniff MethodPrefixSpacing to Squiz coding standard

fabacino:
The sniff ensures that there is only a single space after the keywords `static`, `abstract`, `final` and the visibility keywords. I made a new sniff instead of enhancing the existing `ScopeKeywordSpacingSniff`, because the name would be misleading.

This, however, raises a couple of questions:
1. The existing `ScopeKeywordSpacingSniff` is now obsolete and could be removed but I don't know what the procedure is for removing a sniff (users will get an error if they reference this sniff in their ruleset).
2. If it does not get removed, both sniffs will complain, if there is more than one space after a visibility keyword or `static`. I am not sure if this would be considered a problem.

If this is all too much trouble and the misleading name is of no concern, I can also put these changes into the existing sniff.

This PR addresses issue #1312

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/MethodPrefixSpacingSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/MethodPrefixSpacingSniff.php
@@ -1,0 +1,97 @@
+<?php
+/**
+ * Squiz_Sniffs_WhiteSpace_MethodPrefixSpacingSniff.
+ *
+ * PHP version 5
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @author    Marc McIntyre <mmcintyre@squiz.net>
+ * @copyright 2006-2014 Squiz Pty Ltd (ABN 77 084 670 600)
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ * @link      http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+/**
+ * Squiz_Sniffs_WhiteSpace_MethodPrefixSpacingSniff.
+ *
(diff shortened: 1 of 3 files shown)


jrfnl:
Just my two pennies: I would consider `MethodPrefixSpacing` a confusing name as I'd associate it with prefixing the method *name*. What about `MethodKeywordSpacing` ?

You could prevent a BC break for the removal of the `ScopeKeywordSpacingSniff`, by instead of removing it, making it extend the new sniff and just fall through to the parent class methods in the new sniff. Optionally, a _"deprecated sniff, use x instead"_ warning message could be thrown the first time the sniff is triggered.

aik099:
Cool. You can also add another sniff, that would do the same for properties.

jrfnl:
Just thinking: wouldn't this sniff be better placed in the `Generic` library and then added to the `Squiz` standard via the ruleset ?

aik099:
Of course it should be located in `Generic` ruleset.

jrfnl:
Hmm.. just reviewed the actual code and still very much struggling with the sniff name as it's not about **_methods_** at all. It also - quite rightly - checks the spacing for visibility keywords for properties and the `abstract` keyword for a class,

fabacino:
Thank you both for your input.

I named the sniff `MethodPrefixSpacing` because the property in the token class containing all the relevant keywords used in this sniff is also named `methodPrefixes`. I see @jrfnl's point and agree that `MethodKeywordSpacing` would be a better name than the current one but I think there is still room for improvement.

The sniff actually covers the `static`, `abstract`, `final` and the visibility keywords for both method and variable definitions. If anybody has a good idea for an appropriate name, please share. The best one I can come up with atm is just `KeywordSpacing` but this might be too fuzzy.

fabacino:
How about `ModifierKeywordSpacing`? All the keywords are some kind of modifiers, either for the visibility, the extensibility (`final`, `abstract`) or the "staticness".

PR #1335, opened 2017-02-04
Generic.CodeAnalysis.UselessOverridingMethodSniff should exclude constructors with protected/private scope

gmponos:
Let's say that we have the following example:

```php

class ClassA {
    public function __construct(){}
}

class ClassB extends ClassA {
    private function __construct(){
        parent::__construct();
    }

   public function create(){
       return new self();
   }
}
```

The above is detected by the `Generic.CodeAnalysis.UselessOverridingMethodSniff` as usesless override but it is not since it makes the constructor method private and you can't use it anymore. The only way for you to initialize a `ClassB` is using `create` functon.

I made this optional using a parameter in order to be in the choice of the developer if he will use this feature.

There might be still cases where you have:

```php

class ClassA {
    protected function __construct(){}
}

class ClassB extends ClassA {
    protected function __construct(){
        parent::__construct();
    }

   public function create(){
       return new self();
   }
}
```

The obove case is still a uselless override. So that's why I added the parameter. So that the developer can choose if he really needs to detect these cases.


gsherwood:
Because PHPCS doesn't analyse an entire project at once, it can never determine if your private or protected constructor is actually useful or not. This is why this sniff generates warnings instead of errors. The message itself even says "**Possible** useless method overriding detected" because it is never sure.

I don't think adding an option, or making it ignore private/protected constructors, makes the sniff better. For those that take the time to set the option, it just means that the sniff may not catch a possible error.

Warnings are there for manual code review, so you don't need to strive to have them completely removed. But if you prefer to have no warnings, I suggest not using this sniff (or any of the code analysis ones) and using a different tool that analyses the entire project to generate accurate error messages about these sort of architectural issues.

Thanks for the PR regardless.

PR #1336, opened 2017-02-04
Remove an unused param inside UselessOverridingMethodSniff

gmponos:
Remove an unused variable.
Simple thing.


gsherwood:
Thanks for finding and fixing that.

PR #1337, opened 2017-02-04
Created a new Generic.WhiteSpace.LanguageConstructSpacingSniff

### LINKED ISSUE #1317: Add more in LanguageConstructSpacingSniff
I was wondering if

`yield`, `throw`, `namespace` and `use` should also be inside [here](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php#L43)

gmponos:
Closes #1317


CodeSniffer/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
```diff
@@ -0,0 +1,106 @@
...
+     *
+     * @return array
+     */
+    public function register()
+    {
+        return array(
+                T_ECHO,
+                T_PRINT,
+                T_RETURN,
+                T_INCLUDE,
```
aik099: Not sure about `include`, `require`, `include_once` and `require_once` because they can be written in 2 ways:

* as keyword: `include_once 'file.php';`
* as function call: `include_once('file.php');`

In later case it will be auto-fixed as `include_once ('file.php');` which will trigger warning in another sniff checking function calls.

gmponos: These TOKENS already existed in the deprecated one. I just copied and added more tokens. The sniff already used to handle such cases.

gmponos:
Now that I think about it I must add also tests. I forgot about it.

CodeSniffer/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
aik099: Then probably I reported that issue for deprecated sniff as well. Then don't change anything here.

gmponos:
Not sure how to make the tests pass for 5.3 and 5.4 version.

CodeSniffer/Standards/Generic/Tests/WhiteSpace/LanguageConstructSpacingUnitTest.inc
```diff
@@ -0,0 +1,55 @@
...
+$obj = new MyClass();
+$obj = new    MyClass();
+
+return;
+return $blah;
+return   $blah;
+return($blah);
+
+yield $blah;
+yield   $blah;
```
gmponos: Not sure how to make this test pass for version 5.3 and 5.

gmponos:
Hello... I moved this PR to use the new version of codesniffer but still I don't know how to pass PHP 5.4

Same as [here](https://github.com/squizlabs/PHP_CodeSniffer/pull/1458)

jrfnl:
> I don't know how to pass PHP 5.4

For most versions it is a case of cleaning up the added code (docblock at the top) - see: https://travis-ci.org/squizlabs/PHP_CodeSniffer/jobs/230469601#L224

As for PHP 5.4, see my remark [in the other issue](https://github.com/squizlabs/PHP_CodeSniffer/pull/1458#issuecomment-300259102)

gmponos:
Since this https://github.com/squizlabs/PHP_CodeSniffer/issues/1513 PR was merged the tests of the current one are passing and it can now be merged I think.

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
```diff
@@ -0,0 +1,88 @@
...
+        return array(
+                T_ECHO,
+                T_PRINT,
+                T_RETURN,
+                T_INCLUDE,
+                T_INCLUDE_ONCE,
+                T_REQUIRE,
+                T_REQUIRE_ONCE,
+                T_NEW,
+                T_YIELD,
```
jrfnl: What about also adding `T_YIELD_FROM` ?

jrfnl:
Just wondering: should the "old" sniff be removed/excluded from PHPCS native rulesets which contain it and the "new" sniff be added to them instead ?

gmponos:
1. I have checked all the other rulesets (hope I didn't miss anything) and it does not seem like any other ruleset besided Squiz was using it.
2. I could also exclude it from Squiz ruleset but I was not sure if this is considered backwards compatible.

Let's say that someone has a custom ruleset and includes all the Squiz ruleset.

If he had in his custom ruleset `Squiz.WhiteSpace.LanguageConstructSpacingSnif` silenced or converted to warning instead of error or maybe someother case (that does not come to my mind at the moment) he will have to adjust his custom ruleset.

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
gmponos: I saw the conversation on the other PR and I also thought about it but I need to understand something here is the T_YIELD_FROM equals to just `from` or `yield from`?

gmponos: OK... I will need to check it out if this sniff has the same behaviour with two word tokens.. I am curious about it.

gmponos:
Seems it still can't be merged because of this blocking issue https://github.com/squizlabs/PHP_CodeSniffer/issues/1572

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
gmponos: hello @jrfnl

As I suspected this sniff is not working as expected with two word tokens (check the [travis](https://travis-ci.org/squizlabs/PHP_CodeSniffer/jobs/256926241))

Do you have any suggestion about this?

My suggestion would be:

- Remove these test cases but add comments on the sniff and inside test and fix this on a different PR.
- Remove the T_YIELD_FROM add comments on the sniff that T_YIELD_FROM needs to be added and solve this on a different PR.

@gsherwood  what do you think?

webimpress: @gmponos I would go with something like that:
```php
if ($tokens[$stackPtr]['code'] === T_YIELD_FROM) {
    $space = substr($tokens[$stackPtr]['content'], -5, 4);

    if ($space !== ' ') {
        // error and fix
    }
}
```

I know, it is only specific for that one token, but in future if we have another 'two words' token we can prepare something more generic.

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
```diff
@@ -0,0 +1,89 @@
+<?php
+/**
+ * Ensures all language constructs contain a single space between themselves and their content.
+ *
+ * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @copyright 2006-2015 Squiz Pty Ltd (ABN 77 084 670 600)
```
webimpress: I think, as it is new file year here should be 2017, not 2015.

webimpress: I think, as it is new file year here should be 2017, not 2015.

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
```diff
@@ -0,0 +1,89 @@
...
+     * @param int                         $stackPtr  The position of the current token in
+     *                                               the stack passed in $tokens.
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+
+        if ($tokens[($stackPtr + 1)]['code'] === T_SEMICOLON) {
```
webimpress: We need check first if `$tokens[($stackPtr + 1)]` is set.

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
```diff
@@ -0,0 +1,89 @@
...
+            // No content for this language construct.
+            return;
+        }
+
+        if ($tokens[($stackPtr + 1)]['code'] === T_WHITESPACE) {
+            $content       = $tokens[($stackPtr + 1)]['content'];
+            $contentLength = strlen($content);
+            if ($contentLength !== 1) {
```
webimpress: Not sure about this check. What if the character after keyword is single tab (`\t`) or just new line (`\n`)?

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
gmponos: Please have in mind that these are copy pasted from the deprecated one.

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
gmponos: > Please have in mind that these are copy pasted from the deprecated one.

same here.

webimpress: Then I would improve it, as it is a new sniff. Seems to be wrong for me, in cases I've mentioned.

gmponos: First of all there is an issue risen here. If one of these two sniffs has changes/fixes then the sync between them can be easily forgotten.

I wonder if it would be better instead of copy-pasting the code from the Squiz sniff (as I did) the new Generic sniff to extend the Squiz sniff and only override the `register` function with more tokens.

About the corrections that you are saying I would suggest you to create a PR for the Squiz sniff and have the current PR pending until these changes are merged. Once your PR is merged we can proceed here either with the way I suggest above or by copy-pasting again the code along with the fixes. The reason is that 

 - I like to keep my PR focused on the main target and the goal of this PR was adding more tokens. It is just a way of how I prefer to work.
- Also these fixes need to be made also on the Squiz sniff.

If for those two suggestion above @gsherwood has another opinion I don't mind to go by that.

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
gmponos: done.

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
jrfnl: > I wonder if it would be better instead of copy-pasting the code from the Squiz sniff (as I did) the new Generic sniff to extend the Squiz sniff and only override the register function with more tokens.

I personally would do it the other around: have the `Generic` sniff contain the code and let the `Squiz` version extend and overload the `register()` method if so desired.

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
jrfnl: I second @webimpress' suggestion here, though I would change the second line as that would still not get the right info:
```php
$space = stri_replace(array('yield', 'from'), '', $tokens[$stackPtr]['content']);

// Or maybe even better (simpler), just do:
if (strtolower($tokens[$stackPtr]['content']) !== 'yield from') {
    // error and replace content
}
```

webimpress: > ```php
> // Or maybe even better (simpler), just do:
> if (strtolower($tokens[$stackPtr]['content']) !== 'yield from') {
>     // error and replace content
> }
> ```

Oh, yeah, it even better and much clearer! Thanks @jrfnl

src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
webimpress: Or maybe we should just update list of keywords in `Squiz`... ? The same I've proposed in https://github.com/squizlabs/PHP_CodeSniffer/pull/1492 and it looks like it will be merged for 3.1.0 release.

gmponos: > Or maybe we should just update list of keywords in Squiz... ?

As I said I like to keep my PR focused and that was my main goal but if you read the related issue at the top I had to go by what @gsherwood said.

> I personally would do it the other around: have the Generic sniff contain the code and let the Squiz version extend and overload the register() method if so desired.

I find it reasonable.

gmponos: This is resolved in another PR and I will follow the suggestion of @jrfnl about Squiz extending the Generic one.

gmponos:
I believe it's fine now. Hope you can check it.

gsherwood:
Thanks a lot for the PR. I've merged it in now, with a couple of small changes:
1. I removed the old Squiz sniff so this becomes a "move" of a sniff from the Squiz standard to the Generic standard. I updated the Squiz ruleset and put a note in the changelog about the BC break.
2. I added support for multi-line `yield from` statements because if someone ever uses one, the sniff was breaking fairly badly.
3. When `yield from` is not formatted correctly, only a single error is raised rather than trying to check after the `from` as well. This doesn't affect auto-fixing because a second loop fixes it, but it stop multi-line `yield from` statements from making a mess of the error report.

The sniff will be available from version 3.3.0.

gsherwood:
...and I forgot to fix the tests to support older PHP version. Will do that now.

jrfnl:
@gsherwood As you say, removing the sniff is a BC-break. More than that, it is a problematic BC break for external standards which support more PHPCS versions than just the latest & greatest.

Would you please reconsider deprecating the old sniff (and letting it extend the new sniff for now) and only removing the old sniff in v 4.0.0 ?

gsherwood:
> Would you please reconsider deprecating the old sniff (and letting it extend the new sniff for now) and only removing the old sniff in v 4.0.0 ?

Fair point. I've restored the Squiz sniff to its old behaviour. I don't think it should implement any of the new checks, so it does not extend the new sniff.

I've also fixed the multi-line T_YIELD_FROM issue that was causing the unit tests to fail on old PHP versions.

gmponos:
Thanks for merging and sorry for any inconvenience caused.

jrfnl:
@gsherwood Much appreciated!

PR #1341, opened 2017-02-07
Saving STDIN input to `stdin` CLI value before passing it to `$phpcs-…

### LINKED ISSUE #1340: STDIN file contents not being populated in some cases
In case the PHP Code Sniffer's CLI falls back to re-reading the STDIN in https://github.com/squizlabs/PHP_CodeSniffer/blob/ee3ce4b8ab6b5c815a06fea52ab480a69343134e/CodeSniffer/CLI.php#L1002,L1007 the value is not saved to the CLI values as `stdin` and thus the `$fileContents` is not reachable from within the Sniffs as if it would be in case the initial reading succeeded: https://github.com/squizlabs/PHP_CodeSniffer/blob/ee3ce4b8ab6b5c815a06fea52ab480a69343134e/CodeSniffer/CLI.php#L405,L418

I'm hitting the issue when passing an output of the `svn can` command to the PHPCS using pipes. When using the `cat` for outputting the file and piping it to PHPCS, everything seems to be working fine. This is, imho, related to the fact the initial stream reading is not blocking (which [can't be changed to blocking](https://github.com/squizlabs/PHP_CodeSniffer/issues/953#issuecomment-214611797)).

Looks like saving the value to CLI values before it's passed to `$phpcs->processFile()` method might resolve the issue.

david-binda:
…>processFile()` method in case the input is obtained during the re-read in `PHP_CodeSniffer_CLI::process` method

fixes #1340

### DIFF
--- a/CodeSniffer/CLI.php
+++ b/CodeSniffer/CLI.php
@@ -1013,6 +1013,7 @@ class PHP_CodeSniffer_CLI
                 $this->printUsage();
                 exit(2);
             } else {
+                $this->values['stdin'] = $fileContents;
                 $phpcs->processFile('STDIN', $fileContents);
             }
         }


gsherwood:
I can't replicate the issue directly, but the fix makes perfect sense. Thanks for sending it in.

PR #1343, opened 2017-02-09
Refactor the spacing around the object operator into separate functions

fabacino:
The separate functions make life easier if you want to extend the class and do additional checks.


gsherwood:
Thanks for the PR. Nice easy merge.

PR #1347, opened 2017-02-14
PSR2.Methods.FunctionCallSignature strips some comments during fixing

uniquexor:
This is a somewhat hard to come by bug, but still... The problem was, when using PSR2.Methods.FunctionCallSignatureSniff with requiredSpacesBeforeClose set to 1 and you had a situation like this:
```
foo('Testing
    multiline text'
 );
```
Notice a space before the closing parenthesis. The sniff would try to fix it but would get stuck, since it was removing a single token before the parenthesis. What I've done is I try to move the closing parenthesis to the end of the function call parameter, while jumping over any inline T_COMMENT and moving it before any /* */ comments or any other token, so for example, if we had a situation like this:
```
foo('Testing
    multiline text: ' // . $text
 );
```
The fixed code would look like this:
```
foo('Testing
    multiline text: ' ) // . $text
;
```

But if we had a situation like this:
```
foo('Testing
    multiline text: ' /* . $text */
 );
```
The fixed code would look like this:
```
foo('Testing
    multiline text: ' /* . $text */ );
```

### DIFF
--- a/CodeSniffer/Standards/PEAR/Sniffs/Functions/FunctionCallSignatureSniff.php
+++ b/CodeSniffer/Standards/PEAR/Sniffs/Functions/FunctionCallSignatureSniff.php
@@ -242,7 +242,7 @@ class PEAR_Sniffs_Functions_FunctionCallSignatureSniff implements PHP_CodeSniffe
 
         // Checking this: $value = my_function(...[*]).
         $spaceBeforeClose = 0;
-        $prev = $phpcsFile->findPrevious(PHP_CodeSniffer_Tokens::$emptyTokens, ($closer - 1), $openBracket, true);
+        $prev = $phpcsFile->findPrevious(T_WHITESPACE, ($closer - 1), $openBracket, true);
         if ($tokens[$prev]['code'] === T_END_HEREDOC || $tokens[$prev]['code'] === T_END_NOWDOC) {
             // Need a newline after these tokens, so ignore this rule.
             return;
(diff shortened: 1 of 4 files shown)


gsherwood:
The sniff was able to fix that space correctly because the fixer loops until there are no more errors. So it fixed it, but it took two loops.

The real problem was fixing the placement of the closing parenthesis, which was stripping comments at the end of the argument list. This PR fixes that, as well as allowing the fixer to fix the space in a single loop. So thank you very much for this.

I made some slight modifications to not do all this work in the simple case of incorrect padding before the closing parenthesis, and I also kept the semicolon with the closing parenthesis where possible.

uniquexor:
Thank you. In that case there might have been some other sniff making things worse in my code, since leaving a space did hang up the whole beautifier. It would make 50 passes and stop with an error. Maybe that could be the feature of the v3 - detecting which sniffs cause the infinite loop :)
Either way, I'm glad we fixed it.

gsherwood:
> Maybe that could be the feature of the v3 - detecting which sniffs cause the infinite loop :)

If you use PHPCS with the `--report=diff` option instead of PHPCBF, you'll get a diff of the changes PHPCBF would make. If you also add `-vv` you'll see what fixes are being made and what fixes and being rejected due to conflicts. This is the best way to see which sniffs are conflicting during fixing.

If you use `-vvv` instead of `-vv` you'll also see the output of the file after each pass in the fixing process. I'd only use this when checking small sample files.

PR #1349, opened 2017-02-14
Squiz.Strings.DoubleQuoteUsage.NotRequired message is badly formatted when string contains a CR newline char

uniquexor:
On files with CRLF new lines the message would get distorted, since \r chars would not get converted.


gsherwood:
Thanks a lot for the fix. It worked great.

Issue #1216, opened 2016-11-18
Function/Variable: short/long phpDoc


f3l1x:
Hi guys.

I have a little problem, perhaps I don't know how to define it in ruleset.

I would prefer this:
```
/**
 * @return int
 */
protected function foo()
```

Instead of this:
```
/** @return int */
protected function foo()
```

-----

And also I would like to allow both of following:

```php
/** @var int */
private $a;
```

```php
/** 
 *@var int 
 */
private $b;
```

-----

Can you help me please? Thank you very much.

gsherwood:
For the first problem, you can include the two sniffs `Squiz.Commenting.DocCommentAlignment` and `Generic.Commenting.DocComment` in a ruleset.xml file along with your other checks. This will throw errors for your first code block and allow `phpcbf` to automatically reformat your code blocks as well.

For the second problem, there are no sniffs that allow those formats. Including the two sniffs that I mentioned above will also throw errors for your last two code blocks and reformat them like this:
```php
/**
 * @var int
 */
private $a;

/**
 * @var int
 */
private $b;
```

PR #1351, opened 2017-02-15
Add config options to allow short data type aliases.

fabacino:
There are two new options `useIntDataType` and `useBoolDataType` in Squiz/FunctionCommentSniff which
will allow the usage of `int` instead of `integer` and `bool` instead of `boolean` respectively within doc comments.

Some examples:

```php
// OK with both options set to true, otherwise NOK
/**
 * @param int                $offset The offset.
 * @param int|bool           $value  The mixed value.
 * @param array(int => bool) $map    The fancy mapping.
 */

// OK with both options set to false, otherwise NOK (default & current behaviour)
/**
 * @param integer                   $offset The offset.
 * @param integer|boolean           $value  The mixed value.
 * @param array(integer => boolean) $map    The fancy mapping.
 */
```

This PR is an alternative to #1047 and fixes/answers #1216.

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -83,6 +83,21 @@ class PHP_CodeSniffer
     const STABILITY = 'stable';
 
     /**
+     * The data types to be used for variable type checks.
+     *
+     * @var string
+     */
+    const TYPE_ARRAY    = 'array';
+    const TYPE_BOOL     = 'bool';
+    const TYPE_FLOAT    = 'float';
+    const TYPE_INT      = 'int';
+    const TYPE_MIXED    = 'mixed';
+    const TYPE_OBJECT   = 'object';
+    const TYPE_STRING   = 'string';
+    const TYPE_RESOURCE = 'resource';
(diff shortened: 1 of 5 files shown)


gsherwood:
Thanks for the PR, but I have plans to change the sniff to use the short data types by default so I don't think there is any need for an option. This will only be in the 3.0 version - I don't want to change this functionality in 2.x.

If you want to do up a PR for 3.x with the change to short types, I'd be happy to review that. Otherwise, I will get around to it when I get time.

Thanks again.

fabacino:
Thanks for your feedback. I am not sure I understand your point of view though.

While I agree that it makes sense to change the default values to the short versions for the sake of consistency (starting from 3.0), I fail to see why there should not be an option for the user to override these defaults. phpDocumentor does not specify whether to use long or short data types, therefore I think it makes sense to leave that decision up to the user. The combination of `int` and `boolean` for example is afaik not uncommon at all, but would not be supported out of the box if there were no option to allow `boolean`.

PR #1355, opened 2017-02-17
Exception to the newline requirement on the first member var after class opener.

wilsonge:
Based on the draft PHPCS2 Joomla Custom member var sniff https://github.com/photodude/coding-standards/blob/phpcs-2/Joomla/Sniffs/WhiteSpace/MemberVarSpacingSniff.php#L102-L137

Addresses issues #920 and #725

This is a successor PR to #972 based on the feedback by @gsherwood which @photodude hadn't had time to work on yet

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/MemberVarSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/MemberVarSpacingSniff.php
@@ -97,34 +97,27 @@ class MemberVarSpacingSniff extends AbstractVariableSniff
         $foundLines = ($tokens[$first]['line'] - $tokens[$prev]['line'] - 1);
 
 		// Exception to the newline requirement on the first member var after class opener.
-		if($this->firstVarException)
-		{
-    		if ($foundLines > 0 && $tokens[$prev]['code'] === T_OPEN_CURLY_BRACKET)
-    		{
-    			$error = 'Expected 0 blank lines before first member var; %s found';
-    			$data  = array($foundLines);
-    			$fix   = $phpcsFile->addFixableError($error, $stackPtr, 'FirstMember', $data);
-    			if ($fix === true)
-    			{
-    				$phpcsFile->fixer->beginChangeset();
-    				for ($i = ($prev + 1); $i < $first; $i++)
-    				{
-    					if ($tokens[$i]['line'] === $tokens[$prev]['line'])
-    					{
-    						continue;
-    					}
-    					if ($tokens[$i]['line'] === $tokens[$first]['line'])
-    					{
-    						break;
-    					}
-    					$phpcsFile->fixer->replaceToken($i, '');
-    				}
-    				$phpcsFile->fixer->endChangeset();
-    			}
-    		}
+        if($this->firstVarException) {
+            if ($foundLines > 0 && $tokens[$prev]['code'] === T_OPEN_CURLY_BRACKET) {
+                $error = 'Expected 0 blank lines before first member var; %s found';
+                $data  = array($foundLines);
+                $fix   = $phpcsFile->addFixableError($error, $stackPtr, 'FirstMember', $data);
+                if ($fix === true) {
+                    $phpcsFile->fixer->beginChangeset();
+                    for ($i = ($prev + 1); $i < $first; $i++) {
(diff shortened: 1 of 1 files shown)


photodude:
@gsherwood any thoughts on merging this?

CodeSniffer/Standards/Squiz/Sniffs/WhiteSpace/MemberVarSpacingSniff.php
```diff
@@ -102,22 +118,34 @@ protected function processMemberVar(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
...
+        $expected   = $this->spacing;
+        $isFirst    = $tokens[$prev]['code'] === T_OPEN_CURLY_BRACKET;
+
+        // We allow specifying a different number of spaces to the first operator.
+        if ($isFirst === true) {
+            $expected = $this->firstMemberSpacing;
+            $error    = 'Expected %s blank lines before first member var; %s found';
+        }
+
+        if ($foundLines === $this->spacing || ($isFirst === true && $foundLines === $this->firstMemberSpacing)) {
```
ramsey: I found a bug with this line...

If I have a case where `$this->spacing` is set to `1` and `$this->firstMemberSpacing` is set to `0` and `$isFirst` is `true`, then if I happen to have exactly one space above the first member, I would expect to receive an error saying:

    Expected 0 blank lines before first member var; 1 found

However, since `$foundLines === $this->spacing` is evaluated first, and it is `true` (since `$foundLines` is `1`), then no error occurs.

I think the proper approach would be to change this line to:

``` php
if ($foundLines === $expected) {
```

wilsonge: Ahh good spot. Fixed :)

apaarmann:
As many popular projects  (some references in https://github.com/squizlabs/PHP_CodeSniffer/pull/1464) use no spacing between the class opening and first var declaration, shouldn't `$firstMemberSpacing` have default value as 0 ?
I know this change won't be backward compatible but maybe its the right way to go.

gsherwood:
> shouldn't $firstMemberSpacing have default value as 0 ?

It must be defaulted to 1 to maintain backwards compatibility. Projects that require no spaces are probably not even using the current sniff as it is not configurable, so there is no BC issue with them.

wilsonge:
Rebased to fix conflicts

photodude:
@wilsonge would you fix the CS items
```php
FILE: src/Standards/Squiz/Sniffs/WhiteSpace/MemberVarSpacingSniff.php
 23 | ERROR | [x] Expected "integer" but found "int" for @var tag in member variable comment
 31 | ERROR | [x] Expected "integer" but found "int" for @var tag in member variable comment
```

wilsonge:
Done :) Thanks for the ping :)

photodude:
You are welcome, thanks for fixing those items.

src/Standards/Squiz/Sniffs/WhiteSpace/MemberVarSpacingSniff.php
```diff
@@ -86,22 +102,34 @@ protected function processMemberVar(File $phpcsFile, $stackPtr)
...
+        $expected   = $this->spacing;
+        $isFirst    = $tokens[$prev]['code'] === T_OPEN_CURLY_BRACKET;
+
+        // We allow specifying a different number of spaces to the first operator.
+        if ($isFirst === true) {
+            $expected = $this->firstMemberSpacing;
+            $error    = 'Expected %s blank lines before first member var; %s found';
+        }
+
+        if ($foundLines === $this->spacing || ($isFirst === true && $foundLines === $this->firstMemberSpacing)) {
```
astehlik: I got an error when I pass in zero as a parameter:

```xml
<property name="firstMemberSpacing" value="0" />
```
It seems to me a cast to int is missing:

```php
$foundLines === (int)$this->firstMemberSpacing
```

gsherwood:
Thanks for this PR, although I didn't merge it in as I did things a little differently once I added a bunch of test cases. See the linked commit for the diff if you are interested.

photodude:
thanks @gsherwood glad to see this feature added.

wilsonge:
Thanks!

PR #1362, opened 2017-02-19
Allow using pear sniffs with tabs

wilsonge:
In @Joomla we want to use these sniffs, but we use tabs instead of spaces. This has left us with a handful of (not so nice options). Copy and keep our own copies of the sniffs with these changes in, changing the message - except the message shows the number of spaces instead of tabs required (even through the autofixer works as expected), so basically this is less about the autofixing and more about displaying a user friendly message.

This is basically the same code taken from the `Generic.Whitespace.ScopeIndentSniff`


photodude:
@gsherwood any thoughts on merging this?

wilsonge:
Rebased to fix conflicts

gsherwood:
I've decided to remove this from the 3.2.0 milestone because I'd like to find a better way of allowing any sniff to support tab indents without copy/pasting code and tests.

The fact that PHPCS will replace tabs with spaces while checking (assuming you use `--tab-width`) already allows for consistent checks, but there isn't any easy way to know if this happened and change error messages accordingly.

In addition, auto fixing should be able to apply whitespace indents and have the fixer automatically insert the correct type of indent based on the tab width setting and if the indent was already using tabs or not.

Obviously, if you've included a sniff like `Generic.WhiteSpace.DisallowSpaceIndent` then your indents are going to be replaced by tabs no matter what the other sniffs do during auto-fixing. So this really only applies when you are running a single fixer over your code base. Still, it's not that much effort to also include the indent sniff that you want at the same time to ensure the fixes use the correct indent, so I feel like there is an easy workaround for this already. The message is where no workaround currently exists.

wilsonge:
@gsherwood it's been a year since the last comment here - have you had any thoughts about how you'd like to progress with this. I'm willing to implement something else if you have an idea on how to achieve it. Of course your comment is totally on the money.

gsherwood:
> It's been a year since the last comment here - have you had any thoughts about how you'd like to progress with this.

I have not had time to look into this and I'm not sure when I will.

wilsonge:
In which case can I persuade you to reconsider this. I know it's not the most optimal thing ever - but it saves us having to create 4 sniffs just to work around this in Joomla (because too many end users got confused 🤷‍♂️ )

gsherwood:
> In which case can I persuade you to reconsider this.

I will, when I get time to do some PHPCS dev again.

Issue #834, opened 2015-12-28
PSR2.ControlStructures.SwitchDeclaration does not handle if branches with returns


Majkl578:
Consider the following code as an example:

``` php
<?php

switch ($foo) {
    case 1:
        if ($bar === 1) {
            return 1;
        } elseif ($bar === 2) {
            return 2;
        } else {
            return 3;
        }
    case 2:
        return 4;
}
```

Although the `case 1` body always returns and thus is not fall-through, the error is generated: `There must be a comment when fall-through is intentional in a non-empty case body.`

gsherwood:
I'm not sure how hard this will be to fix. I might be able to just look for ELSE statements directly inside the case and search for returns in there, but I would also have to recursively check other nested ELSE statements inside as well.

In the meantime, the most obvious suggestion would be to make the return explicit, which also adheres to the common no-else-return rule (or even the no-else and return-early rules) because the ELSE is technically unnecessary:

```php
switch ($foo) {
    case 1:
        if ($bar === 1) {
            return 1;
        } elseif ($bar === 2) {
            return 2;
        }

        return 3;
    case 2:
        return 4;
}
```

Majkl578:
I understand this is not easy to fix, one has to check all branches recursively. Explicit return is of course a solution, and also more clean. :)

By the way, this problem also applies to `throw`.

dereuromark:
@Majkl578 Or continue ;)
But yeah, one should probably avoid this kind of nesting, better to refactor into a sub-method here.

gsherwood:
Fixed by PR #1363

PR #1363, opened 2017-02-19
PSR2.ControlStructSwitchDeclaration does not handle if branches with returns

fabacino:
PSR-2 requires a comment when intentionally falling through a `case` within a `switch`.  Before reporting this error, we now check if there is an if/else statement at the end of the `case` block which always terminates. This reduces the number of false positives.

This PR fixes bug #834 .

### DIFF
--- a/src/Standards/PSR2/Sniffs/ControlStructures/SwitchDeclarationSniff.php
+++ b/src/Standards/PSR2/Sniffs/ControlStructures/SwitchDeclarationSniff.php
@@ -179,21 +179,19 @@ class SwitchDeclarationSniff implements Sniff
                 continue;
             }
 
-            $nextCode = $phpcsFile->findNext(
-                T_WHITESPACE,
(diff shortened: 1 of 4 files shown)


gsherwood:

In its effort to reduce false positives, this change has the potential to cause some instability due to false negatives, so I would prefer it to go into the 3.x branch only so the 2.x behaviour is not affected.

it's a very worthwhile change that I would like to see in PHPCS, and you've taken the time to produce a lot of tests cases, so I hope you're able to re-target it for 3.x. If not, please let me know and I'll make the changes myself.

fabacino:
Thanks for your feedback. I'll update the PR for a merge into 3.x.

fabacino:
Should be good to go now.

gsherwood:
Thanks a lot for doing that. The PR looked really good and the change appears to work really well too.

PR #1368, opened 2017-02-21
getcwd returns false in phpstorm

hboomsma:
When using phpcs in phpstorm, the current path becomes unset when a sniff detects an error or warning. This results in `getcwd()` returning false. `chdir`ing to false won't work and emits a warning which triggers phpstorm to fail. (I run phpstorm EAP).

### DIFF
--- a/CodeSniffer.php
+++ b/CodeSniffer.php
@@ -880,9 +880,15 @@ class PHP_CodeSniffer
             // Change the directory so all relative paths are worked
             // out based on the location of the ruleset instead of
             // the location of the user.
-            $inPhar = self::isPharFile($rulesetDir);
+            $inPhar     = self::isPharFile($rulesetDir);
+            $currentDir = getcwd();
+
+            // If started from a removed directory, there will be no current directory.
+            if ($currentDir === false) {
+                throw new PHP_CodeSniffer_Exception('Current directory does not exist.');
+            }
+
             if ($inPhar === false) {
-                $currentDir = getcwd();
                 chdir($rulesetDir);
             }
 


gsherwood:
Do you know why the current path become unset when PHPStorm runs PHPCS? It seems like a potentially serious problem that PHPCS can't rely on some file system functions. Patching it away doesn't feel like the right solution without knowing more about what is going on.

hboomsma:
Hi @gsherwood , I'm still investigating as it is not easily reproducible. My hunch is that is you remove the directory you were in when you started phpstorm you get into trouble. I'll find time next week to test this hypothesis.

hboomsma:
Happens when I remove active projects in PHPStorm while opend. Not a good thing to do anyway. The current code produces a notice in that case, when the `error_level` does not allow those to show, everything works fine, but if it doesn't the notices pollute the output.

gsherwood:
I'm not sure I should be making changes to cover this case. It sounds like PHPCS shouldn't even run given it's being executed inside a directory that no longer exists, and this change would mask potentially critical errors caused by something other than phpstorm.

hboomsma:
Ik agree, would you like a pull request to throw an Exception? At the moment you only get a notice which is not helping much in the case of a insufficient error level.

gsherwood:
> would you like a pull request to throw an Exception?

Yeah, I think that's a good idea. If you want to submit that as a change, I'll merge it in.

hboomsma:
Still found no time, to dig into this, so I can not realistically assume that I will in the near future. Thank you for your time and effort regarding this pull request.

PR #1374, opened 2017-03-01
Add ESLint sniff

### LINKED ISSUE #1210: Configuration options - eslint?
I used JSHint for my linting so far, but the more I read I can see that ESLint is becoming more and more popular. In the [config wiki](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Configuration-Options) I only see that the mentioned several linters (CSSLint, JSHint, JSLint...), but no mention of ESLint.

Does this means that you do not support ESLint, or can I simply do 
```
phpcs --config-set eslint_path ... path to linter ...
```

and it will work, or should I just stick with JSHint when sniffing php files?

rmccue:
Fixes #1210.

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
@@ -6,7 +6,7 @@
  *
  * @category  PHP
  * @package   PHP_CodeSniffer
- * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @author    Ryan McCue <ryan+gh@hmn.md>
  * @copyright 2006-2017 Squiz Pty Ltd (ABN 77 084 670 600)
  * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
  * @link      http://pear.php.net/package/PHP_CodeSniffer
@@ -19,7 +19,7 @@
  *
  * @category  PHP
  * @package   PHP_CodeSniffer
- * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @author    Ryan McCue <ryan+gh@hmn.md>
  * @copyright 2006-2017 Squiz Pty Ltd (ABN 77 084 670 600)
  * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
  * @version   Release: @package_version@
--- a/CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
@@ -1,0 +1,147 @@
+<?php
+/**
+ * Generic_Sniffs_Debug_ESLintSniff.
+ *
+ * PHP version 5
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @copyright 2006-2017 Squiz Pty Ltd (ABN 77 084 670 600)
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ * @link      http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+/**
+ * Generic_Sniffs_Debug_ESLintSniff.
+ *
+ * Runs eslint on the file.
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @copyright 2006-2017 Squiz Pty Ltd (ABN 77 084 670 600)
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ * @version   Release: @package_version@
+ * @link      http://pear.php.net/package/PHP_CodeSniffer
+ */
+class Generic_Sniffs_Debug_ESLintSniff implements PHP_CodeSniffer_Sniff
+{
+    /**
+     * A list of tokenizers this sniff supports.
+     *
+     * @var array
+     */
+    public $supportedTokenizers = array('JS');
+
+
+    /**
+     * ESLint configuration file path.
+     *
+     * @var string|null Path to eslintrc. Null to autodetect.
+     */
+    public $configFile = null;
+
+
+    /**
+     * Returns the token types that this sniff is interested in.
+     *
+     * @return int[]
+     */
+    public function register()
+    {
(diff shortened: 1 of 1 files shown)


CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
```diff
@@ -0,0 +1,147 @@
...
+                $config_file = $candidates[0];
+            }
+        }
+
+        $eslint_options = array( '--format json' );
+        if (empty($config_file) === false) {
+            $eslint_options[] = sprintf('--config %s', $config_file);
+        }
+
+        $cmd  = sprintf(
```
gsherwood: - Why not a simple string concat?
- PHPCS doesn't use `-` in variable names; please conform to the existing coding standards.
- The eslint path and the filename need to be escaped before being used please

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
```diff
@@ -0,0 +1,147 @@
...
+                 1 => array(
+                       'pipe',
+                       'w',
+                      ),
+                 2 => array(
+                       'pipe',
+                       'w',
+                      ),
+                );
+        $proc = proc_open($cmd, $desc, $pipes);
```
gsherwood: All other external tool sniff get away with a simple shell_exec(). Why does this one need to use steams?

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
```diff
@@ -0,0 +1,147 @@
...
+        fclose($pipes[0]);
+        $stdout = stream_get_contents($pipes[1]);
+        $stderr = stream_get_contents($pipes[2]);
+        fclose($pipes[1]);
+        fclose($pipes[2]);
+
+        // Close, and start working!
+        $code = proc_close($proc);
+
+        if ($code > 0) {
```
gsherwood: Would be better to return early here rather than indent the rest of the sniff.

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
```diff
@@ -0,0 +1,147 @@
+<?php
+/**
+ * Generic_Sniffs_Debug_ESLintSniff.
+ *
+ * PHP version 5
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Greg Sherwood <gsherwood@squiz.net>
```
gsherwood: Author should be set to your  name, or the names of whoever wrote the sniff.

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
```diff
@@ -0,0 +1,147 @@
...
+ */
+
+/**
+ * Generic_Sniffs_Debug_ESLintSniff.
+ *
+ * Runs eslint on the file.
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Greg Sherwood <gsherwood@squiz.net>
```
gsherwood: Author should be set to your  name, or the names of whoever wrote the sniff.

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Processes the tokens that this sniff is interested in.
+     *
+     * @param PHP_CodeSniffer_File $phpcsFile The file where the token was found.
+     * @param int                  $stackPtr  The position in the stack where
+     *                                        the token was found.
+     *
+     * @return void
+     * @throws PHP_CodeSniffer_Exception If jslint.js could not be run
```
gsherwood: This sniff doesn't actually throw an exception. This is just a copy/paste error.

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
```diff
@@ -0,0 +1,147 @@
...
+        fclose($pipes[2]);
+
+        // Close, and start working!
+        $code = proc_close($proc);
+
+        if ($code > 0) {
+            $data = json_decode($stdout);
+            // Detect errors.
+            if (json_last_error() !== JSON_ERROR_NONE) {
+                $error = 'Unable to run eslint: %s';
```
gsherwood: Sniffs don't add error messages when the external tool fails. Most don't do anything, but one does throw an exception. It's probably best to do nothing so that all the tool sniffs are consistent. In the future, a change could be made to all of them to either throw exceptions or add a warning about a tool failing to run.

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
rmccue: It checks the response code from ESLint, which isn't available via `shell_exec()`. Could potentially use `exec()` instead (I was actually using stderr previously, but stopped)

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
rmccue: > PHPCS doesn't use `-` in variable names; please conform to the existing coding standards.

Ah, this wasn't picked up by phpcs running on the code, so missed it. Might be another good sniff to add ;)

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
rmccue: Interestingly, JSLintSniff doesn't throw either.

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
rmccue: I personally found it super useful to have the error reported. Happy to remove here, but IMO changing them all to report errors would be better.

rmccue:
@gsherwood Fixed all of these up, but still using `proc_open` as noted above. Can change to `exec()` if you'd prefer.

CodeSniffer/Standards/Generic/Sniffs/Debug/ESLintSniff.php
gsherwood: If you just need the return value, exec() would cut out a lot of code here and make the intention clearer.

rmccue:
Switched out for `exec()`, should be good to go 👍

gsherwood:
Thanks for making those changes. Finally merged in.

PR #1379, opened 2017-03-08
'bool' and 'int' instead of 'boolean' and 'integer'

marius-balteanu:
- Replace boolean and integer with bool and int 
- Update tests and doc blocks 

This PR is related to:
- #1047
- #1195

We're very interested to have this change in the 3.0 version of CodeSniffer because currently we modified in our local CI server the library in order to allows bool and int types. 

@gsherwood, any feedback is appreciated.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -136,7 +136,7 @@ class Config
      * If FALSE, arguments that are not command line options or file/directory paths
      * will be ignored and execution will continue.
      *
-     * @var boolean
+     * @var bool
      */
     public $dieOnUnknownArg;
 
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -81,7 +81,7 @@ class File
      * If TRUE, the file was loaded from a local cache.
      * If FALSE, the file was tokenized and processed fully.
      *
-     * @var boolean
+     * @var bool
      */
     public $fromCache = false;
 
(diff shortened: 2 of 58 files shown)


gsherwood:
I'm not going to merge this change in because it just changes comments. PHPCS itself uses the long form of int and bool in comments and I'm only going to change that if I decide to change the sniffs that enforce the commenting standards as well (I probably will eventually).

If I do that, I'll use PHPCBF to make these modifications for me to ensure the fixer is working.

marius-balteanu:
Thanks for your response, @gsherwood.

As I said, I'm very interested to help you implement this change in 3.0 (of course, if you agree with this change and you need my help, if not, I'll stop here).

My changes from the PR were made in 3 steps:
1. [manually] change from integer/boolean to int/bool in src/Util/Common.php
2. [**using PHPCBF**] replace integer/boolean with int/bool in comments
3. [manually] update tests

From your comment, I understand that you think that I made manually the changes in the comments, but I didn't or I miss something else.

Please let me know if you want to work on this (and which are the next steps from your point of view) or I should wait for your implementation.

gsherwood:
Thanks for getting back to me. I don't want to update any of the code for this change yet. It's something I'm considering doing in the 3.0 branch, but I need to look at it more closely before I go and change the included coding standards.

PR #1384, opened 2017-03-10
Make the unit test suite compatible with PHPUnit v6

### LINKED ISSUE #1383: Unit tests failing under PHPUnit 6.x
Travis has upgraded the PHPUnit version used for the PHP 7.x builds to PHPUnit 6.x.

The current testsuite is not compatible with this. This impacts both PHPCS as well as any external standards which unit test their standard using the PHPCS build in unit test mechanism.

I'm working on a fix for this.

jrfnl:
Since today Travis uses PHPUnit 6.x for PHP 7+. The unit tests were not compatible with this and failing because of it.

These relatively minor changes fix that and make the testsuite compatible with both PHPUnit < 6 as well as 6+.

Ref: https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-6.0.0

I've also run a test with an external standard using the PHPCS test suite to confirm that this will fix it for them too without any effort on their side and have confirmed it will.
See: https://travis-ci.org/jrfnl/WordPress-Coding-Standards/builds/209592330 - the PHP 7+ builds fail when using PHPCS `master`, but pass when using the branch contained in this PR.

Fixes #1383

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,14 +19,20 @@ matrix:
     - php: nightly
     - php: hhvm
 
+before_install:
+  # Circumvent a bug in the travis HHVM image - ships with incompatible PHPUnit.
+  - if [[ $TRAVIS_PHP_VERSION == hhv* ]]; then composer self-update; fi
+  - if [[ $TRAVIS_PHP_VERSION == hhv* ]]; then composer require phpunit/phpunit:~4.0; fi
+  - if [[ $TRAVIS_PHP_VERSION == hhv* ]]; then composer install; fi
+
 before_script:
   - if [[ $CUSTOM_INI == "1" && ${TRAVIS_PHP_VERSION:0:1} == "5" ]]; then phpenv config-add php5-testingConfig.ini; fi
   - if [[ $CUSTOM_INI == "1" ]] && [[ ${TRAVIS_PHP_VERSION:0:1} == "7" || $TRAVIS_PHP_VERSION == "nightly" ]]; then phpenv config-add php7-testingConfig.ini; fi
 
 script:
-  - composer install
   - if [[ $TRAVIS_PHP_VERSION != hhv* ]]; then php bin/phpcs --config-set php_path php; fi
-  - vendor/bin/phpunit tests/AllTests.php
+  - if [[ $TRAVIS_PHP_VERSION != hhv* ]]; then phpunit tests/AllTests.php; fi
+  - if [[ $TRAVIS_PHP_VERSION == hhv* ]]; then vendor/bin/phpunit tests/AllTests.php; fi
   - if [[ $CUSTOM_INI != "1" ]]; then php bin/phpcs --no-cache --parallel=1; fi
   - if [[ $CUSTOM_INI != "1" &&  $TRAVIS_PHP_VERSION != hhv* && $TRAVIS_PHP_VERSION != "nightly" ]]; then pear package-validate package.xml; fi
   - if [[ $CUSTOM_INI != "1" &&  $TRAVIS_PHP_VERSION != hhv* ]]; then php scripts/build-phar.php; fi
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "ext-simplexml": "*"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.0"
+        "phpunit/phpunit": "^4.0 || ^5.0 || ^6.0"
     },
     "bin": [
         "bin/phpcs",
--- a/package.xml
+++ b/package.xml
@@ -82,6 +82,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
      <file baseinstalldir="" name="AbstractSniffUnitTest.php" role="test" />
      <file baseinstalldir="" name="AllSniffs.php" role="test" />
     </dir>
+    <file baseinstalldir="" name="bootstrap.php" role="test" />
     <file baseinstalldir="" name="AllTests.php" role="test" />
     <file baseinstalldir="" name="TestSuite.php" role="test" />
    </dir>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit>
+<phpunit backupGlobals="true" bootstrap="tests/bootstrap.php">
     <testsuites>
         <testsuite name="PHP_CodeSniffer Test Suite">
             <file>tests/AllTests.php</file>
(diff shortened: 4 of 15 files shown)


orlangur:
Basically, here is a much cleaner solution from linked ticket: 
> what WordPress core is doing: https://core.trac.wordpress.org/ticket/40086

It seems to be workable for any external standard as well.

Does it really double the build time?

jrfnl:
@orlangur You don't seem to grasp the fact that the solution in WP is a plaster on the wound, not a cure.

As I said in #1383:
> There are a number of reasons I went with the chosen solution rather than Composer:
>    * Every single external standard which uses the PHPCS test suite would have to adjust their build script for that to work. My solution is a "silent" one and their builds will just pass again.
>    * The Travis build for PHPCS currently does not use composer
>    * It would double the build time
>    * The PHPUnit requirements for composer are different for builds on PHP 5.2-5.3, PHP 5.4-5.6 and PHP7+.
>    * You can't just limit it to PHPUnit 5.x as then we'll have the same problem as soon as PHP 7.2 hits.




> Does it really double the build time?

Test it & see for yourself.

gsherwood:
I think this is a clever solution to get around having to wait for composer to install PHPUnit and all it's dependencies. Personally, I don't really mind how long the builds take, so I assume this is actually a problem for other projects not wanting to install their composer deps to run tests?

Have you thought about wget'ing a PHPUnit phar and running that? I wonder if this would allow the tests to run without the extra composer time, but still keep the testing framework clean.

I know this would require projects that run their own PHPCS tests to change their build config, but the number of projects that do this is very small, and their builds should already be breaking.

Just thinking out loud here. I was already planning on trying to make the test suite forwards compatible, so I would have probably gone the same way as this PR, so I'm not opposed to just merging it in.

jrfnl:
> Have you thought about wget'ing a PHPUnit phar and running that?

I did, but that would have the same issue as a composer install, in the sense that you'd need different versions of the PHPUnit phar for testing on different PHP versions, as well as the question of whether testing on PHP 7.2/nightly will still be possible as it's not officially supported for the PHPUnit 5.x series.
Testing on PHP 7.2/nightly might not break immediately, but chances are it will over the next couple of months.

As you said: this PR makes the test suite forward compatible or rather: it makes the test suite back- ànd forward compatible with the least amount of changes while maintaining the ease of use for external standards.

> the number of projects that do this is very small,

Is it really ? What with the test suite being so easy to extend, I would expect every external standard to be unit tested :sunglasses:  .

> and their builds should already be breaking

Only if they pushed/pulled anything in the last few days as the change in the Travis images is recent, so not everyone will have noticed yet.

jrfnl:
P.S.: the only thing I can think of which may need some tweaking at some point is the regex to get the version nr: https://github.com/squizlabs/PHP_CodeSniffer/pull/1384/files#diff-cecbf887c605916a3e2ddf91d93eb341R27

The regex currently does not allow for alpha/beta/RC builds of PHPUnit, but for the Travis builds those aren't used anyway, so this would only become an issue if a sniff dev running the test suite locally would use those kind of build of PHPUnit.

jrfnl:
As PHPCS 3.x has a minimum requirement of PHP 5.4, this PR can be simplified (a lot) by using `class_alias()` in a PHPUnit bootstrap file instead of extending upstream classes.

@gsherwood Would you like me to update the PR to that effect or is this PR effectively dead ?

orlangur:
> this PR can be simplified (a lot) by using `class_alias()` in a PHPUnit bootstrap file

Yeah, and it can be improved even more (in terms of simplicity & reliability) by installing PHPUnit via Composer 😉 Those wrappers are a pure evil for further maintenance...

orlangur:
Oh, it is already implemented properly by the way: https://github.com/squizlabs/PHP_CodeSniffer/commit/1280f49a291714944dbb450dff1ed70975c7dc90

So, this PR can be closed.

jrfnl:
@orlangur You've made your point of view quite clear before. Your current remarks do not add any value to the discussion as they just repeat what you said before.

orlangur:
True, just summarized what was said before as it was not refuted.

gsherwood:
> Would you like me to update the PR to that effect or is this PR effectively dead ?

I would really like to see a simplified version if you want to take the time to write one. This PR isn't dead, although it's not high on the priority list right now. But if the change can be made to look cleaner, I'll put it onto a release milestone so I don't forget about it again.

jrfnl:
> I would really like to see a simplified version if you want to take the time to write one.

@gsherwood Thanks for your response. I've rebased & updated the PR with the simplified version. All builds passing ;-)

gsherwood:
> I've rebased & updated the PR with the simplified version

I'm loving this. Thanks.

jrfnl:
One side-note for external classes which use the PHPCS unit test suite as a basis for their unit tests:
* They would either need to run their tests from the PHPCS installation directory, i.e.:
`php_codesniffer>phpunit tests/AllTests.php --filter=StandardName`
* Or they would need to run the bootstrap themselves (or add it to their own bootstrap):
`some_standard>phpunit /path/to/phpcs/tests/AllTests.php --filter=StandardName --bootstrap=/path/to/phpcs/tests/bootstrap.php`

wmathes:
Asking out of curiosity :) This PR covers the last issue for the 3.1 milestones and appears to be solved, though not yet reformatted. Now it stalled for a good 6 weeks. Is this still being worked on? Should people wait for 3.1 because it's "around the corner" or is dev-master the way to go for some time? ^_^

gsherwood:
> Asking out of curiosity :) This PR covers the last issue for the 3.1 milestones and appears to be solved, though not yet reformatted. Now it stalled for a good 6 weeks. Is this still being worked on? Should people wait for 3.1 because it's "around the corner" or is dev-master the way to go for some time? ^_^

Don't worry, nothing is stalled. PRs come in and I get to them when I have time. There are things in 3.1 that have been waiting a lot longer than 6 weeks to get into a release.

3.1 is just around the corner now that I've managed to clear the feature backlog. This is the last planned changed before release, although I have at least one bug fix to review before I set a date. I'd expect it to be out late this week or sometime next week.

wmathes:
Great! Best of luck then. :)

gsherwood:
I've now merged this in, but needed to make a few changes to the test runner to get everything working when PEAR installed. I've tried this with an external standard, but can you please also check this with the WordPress standards before I release?

jrfnl:
@gsherwood Just tested it. As WPCS still supports both PHPCS 2.x as well as 3.x, it uses its own bootstrap for the unit tests, so it did need a small change to load the PHPCS bootstrap from within the WPCS bootstrap, but other than that, looking good.

The downstream PR with the minor change needed for WPCS is here: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1130

gsherwood:
> but other than that, looking good.

Thanks for testing that for me.

PR #1388, opened 2017-03-13
fixes missing variable declaration in test coverage

LC43:
My code had this error, sorry for this. it also removes extra variable declaration.


gsherwood:
It doesn't matter if the code is valid. This test file actually has a syntax error in it to make sure the sniff (and tokenizer) can cope with that. The file itself is not parsed - it is just tokenized during static analysis, so it doesn't actually matter if the code does anything, or if vars are declared properly or not.

The important thing is that the code structure you provided for testing replicates the problem, which it did.

But I'll merge this in anyway so you don't look at your test code and feel bad about it :)

LC43:
eheh i see, thanks!

PR #1390, opened 2017-03-16
Made able to check anonymous function as standard function

andras-tim:
I wanted to write some rules for anonymous function, but I couldn't get definition of function.
(e.g. I couldn't check for an anonymous function is static or not)

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -2709,6 +2709,10 @@ class PHP_CodeSniffer_File
     public function isAnonymousFunction($stackPtr)
     {
         $tokenCode = $this->_tokens[$stackPtr]['code'];
+        if ($tokenCode === T_CLOSURE) {
+            return true;
+        }
+
         if ($tokenCode !== T_FUNCTION) {
             throw new PHP_CodeSniffer_Exception('Token type is not T_FUNCTION');
         }


devdrops:
@andras-tim hi! I'm just a PHPCS user and subscriber on this repo (that's how I've noticed you PR), so I'd like to add a small suggestion to your contribution. Here, on `CodeSniffer/File.php`: 

```php
if ($tokenCode !== T_FUNCTION 
    && $tokenCode !== T_CLOSURE
) {
    throw new PHP_CodeSniffer_Exception('Token type is not T_FUNCTION or T_CLOSURE');
}
```
IMHO, as you're using `&&` which will only be satisfied if the both conditions are matched, it would be best to describe the message as below:

```php
if ($tokenCode !== T_FUNCTION 
    && $tokenCode !== T_CLOSURE
) {
    throw new PHP_CodeSniffer_Exception('Token type is neither T_FUNCTION nor T_CLOSURE');
}
```

What do you think? Hope I'm not being annoying here :wink:

andras-tim:
Dear @devdrops, I copied from `getMethodParameters()`, and I didn't want to change the appearance.

devdrops:
@andras-tim I see. Well, in that case, thank you for your feedback! :wink:

CodeSniffer/File.php
```diff
@@ -2709,8 +2709,10 @@ public function getDeclarationName($stackPtr)
     public function isAnonymousFunction($stackPtr)
     {
         $tokenCode = $this->_tokens[$stackPtr]['code'];
-        if ($tokenCode !== T_FUNCTION) {
```
jrfnl: AFAICS you could simplify this a bit by leaving the original code as is and just adding the following snippet above it:
```php
if ($tokenCode === T_CLOSURE) {
    return true;
}
```

jrfnl:
As for using the `getMethodProperties()` to determine whether a anonymous function is static: this comes across as quite inefficient as that method is checking for so much more - none of which can be used in combination with closures.

Below is a simpler way to check whether a closure is static.

```php
    protected function isClosureStatic(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
    {
        $tokens    = $phpcsFile->getTokens();
        $prevToken = $phpcsFile->findPrevious(PHP_CodeSniffer_Tokens::$emptyTokens, ($stackPtr - 1), null, true, null, true);

        return ($prevToken !== false && $tokens[$prevToken]['code'] === T_STATIC);
    }
```

CodeSniffer/File.php
andras-tim: Oh, sorry, I was a bit careless.

andras-tim:
@jrfnl, don't know, how can be a `T_FUNCTION` anonymous, because as far as I know, the anonymous function is `T_CLOSURE`. Or I missed something?

So, the `$isClosure` property in `getMethodProperties()` is valid?

jrfnl:
> l, don't know, how can be a T_FUNCTION anonymous, because as far as I know, the anonymous function is T_CLOSURE. Or I missed something?

@andras-tim The `T_CLOSURE` token is not a native PHP token, but one coming from PHPCS. I'm not sure without testing this, but I imagine that this token might not always be correctly backfilled for older PHP versions (PHP 5.1/5.2 which are still supported by PHPCS) ?

It might also be a code artifact left over from when PHPCS did not replace the token while it does do so now.

> What do you think, should I copy your isClosureStatic() function into File.php?

I'd use that logic in your sniff. I wouldn't add it to the `File.php` file, but that's just me.

andras-tim:
> > l, don't know, how can be a T_FUNCTION anonymous, because as far as I know, the anonymous function is T_CLOSURE. Or I missed something?
> 
> @andras-tim The T_CLOSURE token is not a native PHP token, but one coming from PHPCS. I'm not sure without testing this, but I imagine that this token might not always be correctly backfilled for older PHP versions (PHP 5.1/5.2 which are still supported by PHPCS) ?

@jrfnl, I think, in this case I should add `T_CLOSURE` support to `getMethodProperties()`. What do you think?

gsherwood:
I think this is a case of old code that pre-dates the T_CLOSURE token: https://github.com/squizlabs/PHP_CodeSniffer/commit/0d78e10b206c797c9fbde9aaeaf785c1afecaee4

I think the isAnonymousFunction() method can be removed (I'd do this in 3.x only) and I think other functions that reference it in the core (there are 2 of them) can be changed to not call it and still work fine.

In sniffs that check anon functions, they already have to listen for T_CLOSURE and so there is no need to check if that is anon or not; the tokenizer has already decided it is and assigned a different token type.

So if you want to check if that closure is static, I'd just use the code @jrfnl has posted above. I wouldn't add that into the core though - it's not a complex enough piece of code to require support in there.

Adding T_CLOSURE support to getMethodProperties() does seem like a bit of a waste due to all the extra processing that happens in there. But it could just be modified to check if the token is a T_CLOSURE right at the top, check for static, then return a value with just that information in it (everything else in the array would use the default values). Still, I'd question if that is really useful, and I'd probably remove the `isClosure` index from the return value because the token type gives you that information now.

gsherwood:
I ended up removing `isAnonymousFunction()`, removing the `is_closure` index from `getMethodProperties()` and adding T_CLOSURE support to `getMethodProperties()`.

The reason I decided to add T_CLOSURE support is because `getMethodParameters()` already supports both T_FUNCTION and T_CLOSURE, so this is better for consistency.

I won't be making these changes in 2.x though as they are breaking changes.

PR #1393, opened 2017-03-20
DisallowMultipleAssignments sniff: fix false positive for PHP4 compatible code

jrfnl:
While PHP4 code shouldn't be used anymore, there are other sniffs which deal with that.
This sniff should in that regard be version agnostic.


gsherwood:
Thanks for this PR.

jrfnl:
You're welcome ;-)

PR #1395, opened 2017-03-20
Added long parameter `phpVersion`

boesing:
Parse long parameter `phpVersion` and store into `configData` array which is used by other logic within PHP Codesniffer.


jrfnl:
@boesing The PHP Version can already be overruled.... `$ phpcs --config-set php_version 50403`

See: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Configuration-Options#setting-the-php-version
Was added in: https://github.com/squizlabs/PHP_CodeSniffer/pull/1104

boesing:
As @jrfnl mentioned, the value can already be overridden.

PR #1399, opened 2017-03-22
Travis tweak: Run the test with & without the custom ini files.

jrfnl:
In the current setup, the tests were only run using the custom ini files.
This meant that the test-cases for which the default ini is expected, were never being run.

I've now added a custom environment variable which toggles whether the default ini values or the custom values should be used.

I've selectively chosen two PHP versions on which to run with the custom ini file based on the sniffs involved.
The other PHP versions will now use the default ini values, making the build testing via travis more comprehensive as both situations will be tested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,15 @@ matrix:
   include:
     - php: 5.3
       dist: precise
+    - php: 5.3
+      dist: precise
+      env: CUSTOM_INI=1
     - php: 5.4
     - php: 5.5
     - php: 5.6
     - php: 7.0
+    - php: 7.0
+      env: CUSTOM_INI=1
     - php: 7.1
     - php: nightly
     - php: hhvm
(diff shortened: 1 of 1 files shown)


jrfnl:
Rebased after the merge of #1400 to show the build still passes.

gsherwood:
Do you think there is any benefit is running both the default and custom ini on each PHP version rather than just picking a couple?

jrfnl:
> Do you think there is any benefit is running both the default and custom ini on each PHP version rather than just picking a couple?

@gsherwood As the custom ini values only apply to a handful of sniffs, the benefits would be very limited. The PHP versions I selected are the "breaking points" where the ini values start making a difference one way or another for the relevant sniffs.
It might be more relevant to do so in the future if there would be more sniffs added which need to be tested with custom ini values.

For now, I could adjust the matrix to run against both the default and custom ini values for the selected PHP versions if you like.

Just let me know and I'll make the necessary adjustments.

gsherwood:
> For now, I could adjust the matrix to run against both the default and custom ini values for the selected PHP versions if you like.

I think this might be best. Or, at least run both the default and custom ini's on PHP 5.3 and 7.0 as well, just to make sure they function with the values on/off.

jrfnl:
> Or, at least run both the default and custom ini's on PHP 5.3 and 7.0 as well, just to make sure they function with the values on/off.

@gsherwood I believe that is what I proposed or did you mean something different ?

Anyways, I've made this change. For the two extra builds, I've set it up so only the unit tests are run for efficiency. The other build tests are skipped as they are already covered by the builds with the default ini values.

gsherwood:
> I believe that is what I proposed or did you mean something different ?

Yep, that's what I was thinking. Thanks a lot.

> For the two extra builds, I've set it up so only the unit tests are run for efficiency. The other build tests are skipped as they are already covered by the builds with the default ini values.

Good idea!

PR #1400, opened 2017-03-22
DisallowShortOpenTag: throw warning for potential short open tags in inline HTML

### LINKED ISSUE #1398: DisallowShortOpenTagSniff does not always recognize short open tags
When the `short_open_tags` ini setting is turned off on the system PHPCS is being run on, short open tags used within `T_INLINE_HTML` are not recognized by the sniff.

If you have no control over the servers where code is being deployed (open source projects), this can be dangerous as there might well be short open tags remaining in the code in that case and the server on which the code is deployed might have the setting turned on.

While these can't reliably be fixed, they can be warned about when found.

Test case:

```php
Some content <? echo $var; ?> Some more content

Some content <?
echo $var;
?> Some more content
```

Related #1234

jrfnl:
When the ini setting `short_open_tag` is `off` on the machine running PHPCS, short open tags would not be recognized in `T_INLINE_HTML` tokens.
This could lead to undesirable consequences if the code would later be deployed on a machine where `short_open_tag` is set to `on`.

This PR fixes that and will throw a warning if a use of short open tags is detected in inline HTML.

Fixes #1398

Note: the `getSnippet()` method is a 1-on-1 copy of the one used in the `DisallowAlternativePHPTags` sniff.
It would be better if that method would be placed in another file and just called by both sniffs, but I wasn't sure which class would be most suited to place the method in.

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowShortOpenTagSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowShortOpenTagSniff.php
@@ -38,10 +38,17 @@ class Generic_Sniffs_PHP_DisallowShortOpenTagSniff implements PHP_CodeSniffer_Sn
      */
     public function register()
     {
-        return array(
-                T_OPEN_TAG,
-                T_OPEN_TAG_WITH_ECHO,
-               );
+        $targets = array(
+                    T_OPEN_TAG,
+                    T_OPEN_TAG_WITH_ECHO,
+                   );
+
+        $shortOpenTags = (boolean) ini_get('short_open_tag');
+        if ($shortOpenTags === false) {
+            $targets[] = T_INLINE_HTML;
+        }
+
+        return $targets;
 
     }//end register()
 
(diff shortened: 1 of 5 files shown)


gsherwood:
The tests are failing on the .3 file because the error lists shows lines 13, 16, 21 with warnings and there aren't even 21 lines in the file. I'm going to change that to what I get, which is the `Internal.NoCodeFound` warning on line 1, then the sniff warnings on lines 3, 6, and 11.

I'm also changing the message just a little, and changing the DisallowAlternativeTags messages to match that change. I hadn't noticed it before - obviously very old code :)

jrfnl:
> The tests are failing on the .3 file
>... the .1 file is also failing when short_open_tag is off

Oh darn, sorry about that, I've moved the unit tests around a couple of times and clearly forgot to change the line numbers the last time.
Once #1399 has been merged, things like that will actually be caught by Travis, so that should prevent this kind of mishap in the future.

> I'm also changing the message just a little ... I hadn't noticed it before - obviously very old code :)

*grin* please change it to your liking, the other came from me as well. Just happy to be able to contribute so these are available to the wider public :heart_eyes:

gsherwood:
> Oh darn, sorry about that

Oh it's no problem. Makes me understand the change better as well. Just wanted to document why I changed all this stuff so it's not confusing. 

I've committed the test changes now and they appear to be working ok.

Thanks for another great contribution.

jrfnl:
:+1:

PR #1402, opened 2017-03-30
FEATURE: Migrate docs from wiki to sphinx

DanielSiepmann:
* to have everything in a single repository.
* Allowing to check missing docs on Pull Requests with new or changed
  features.
* Also allows others to fork docs and adjust them via Pull Requests.
* Also allows us to render them via Sphinx to different formats like
  HTML, PDF, ...
* Allows hosting on read the docs.

The result is available at http://phpcs-fork.readthedocs.io/en/feature-docs/

I will also adjust to your wishes and of course remove the domain if you don't like the idea or will host it on your own.
The service is free and setup is done in 5 Minutes.


gsherwood:
I appreciate you doing this, but as I said on Gitter, I am happy with the docs being on the Github wiki and would only move them elsewhere if I built a proper site for PHPCS, which I have no plans (or time) to do right now.

PR #1408, opened 2017-04-04
Tokens: Move `T_COALESCE_EQUAL` from `$operators` to `$assignmentTokens`

jrfnl:
The `T_COALESCE_EQUAL` operator compares and assigns to a variable:
```php
$a ??= $b; // When $a is set, use $a, otherwise assign $b to $a.
```

Looking at the token arrays, I believe was incorrectly placed with `$operators` and should be in `$assignmentTokens` instead.

Having it in `$operators` could cause issues with sniffs like `Squiz.Formatting.OperatorBracket`.

### DIFF
--- a/CodeSniffer/Tokens.php
+++ b/CodeSniffer/Tokens.php
@@ -422,20 +422,21 @@ final class PHP_CodeSniffer_Tokens
      * @var array(int)
      */
     public static $assignmentTokens = array(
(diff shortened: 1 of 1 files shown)


gsherwood:
You are 100% correct. Was just a mistake. Thanks for fixing it.

PR #1411, opened 2017-04-05
Squiz/OperatorSpacing: Ignore operators in declare statements.

### LINKED ISSUE #1163: Add option for Squiz.WhiteSpace.OperatorSpacing to invert for declare
It would be really nice to use the following configuration to indicate that declare statements should match `declare(strict_types=1);`

``` xml
<rule ref="Squiz.WhiteSpace.OperatorSpacing">
    <properties>
        <property name="invertDeclares" value="true"/>
    </properties>
</rule>
```

jrfnl:
Fixes #1163


gsherwood:
Thanks. That was nice andeasy :)

PR #1415, opened 2017-04-08
Tokens: Add new `$ooScopeTokens` property.

jrfnl:
Add new `$ooScopeTokens` property representing all tokens which typically open class-like constructs.

This is particularly useful if a sniff needs to behave differently when it encounters a token within one of those scopes, like in the case of NamingConvention sniffs.

Applied the use of the property in a few places in existing sniffs.


jrfnl:
Rebased for merge conflicts / v3.0 release.

gsherwood:
Works nicely. Thanks.

jrfnl:
You're welcome Thanks for merging.

PR #1420, opened 2017-04-11
PEAR/ClassComment: Apply the class comment rules to traits as well.

jrfnl:
Under the "PEAR rules were created before traits existed" principle, I figured traits should comply with the class comment rules as well.

I've not added anonymous classes as to add that, some token walking would be needed allow for the anonymous class to be assigned to a variable. Let me know if you'd like me to change/add that anyway.


gsherwood:
Thanks

jrfnl:
You're welcome ;-)

PR #1423, opened 2017-04-14
Squiz/BlockComment: Fix missing error code for one message.

jrfnl:


gsherwood:
I added a type for this error in 3.x when I made error codes mandatory. There were only a couple of places where a code was missing. Commit is here: https://github.com/squizlabs/PHP_CodeSniffer/commit/c587b2744535b195d3249c8c74ace2c2d18cad0f

I've gone ahead and backported those codes as I clearly forgot to do it at the time. Thanks.

jrfnl:
Great. Thanks!

PR #1424, opened 2017-04-14
Squiz/ClassComment: Fix incorrect sniff description in class docblock.

jrfnl:
The description did not reflect what the sniff was really checking for.

It might be that the other items were intended to be added at some point, but they are currently not in the sniff.


gsherwood:
Thanks

PR #1425, opened 2017-04-14
Squiz/ClosingDeclarationComment: use `$data` rather than composing the error message in parts

jrfnl:


gsherwood:
Good change. Thanks.

PR #1431, opened 2017-04-25
PHPCBF can't fix short open tags when they are not followed by a space

ghunti:
I was using `phpcbf` to fix some legacy files that contained short opening tags without an empty space between them and the content: `<?echo` (this is valid to the PHP engine)
The problem is that `phpcbf` was replacing the short opening tag with a huge string like `<?phpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpphpecho`

I've added a test that shows the failure + the fix that I found for the problem.
Keep up the good work.
Cheers

### DIFF
--- a/CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowShortOpenTagSniff.php
+++ b/CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowShortOpenTagSniff.php
@@ -73,10 +73,11 @@ class Generic_Sniffs_PHP_DisallowShortOpenTagSniff implements PHP_CodeSniffer_Sn
             $fix   = $phpcsFile->addFixableError($error, $stackPtr, 'Found', $data);
             if ($fix === true) {
                 $correctOpening = '<?php';
-                if (isset($tokens[$stackPtr + 1]) === true && $tokens[$stackPtr + 1]['code'] !== T_WHITESPACE) {
-                    // Avoid creation of invalid open tags like <?phpecho if the original was <?echo
+                if (isset($tokens[($stackPtr + 1)]) === true && $tokens[($stackPtr + 1)]['code'] !== T_WHITESPACE) {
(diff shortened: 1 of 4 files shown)


CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowShortOpenTagSniff.php
```diff
@@ -72,7 +72,12 @@ public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
             $data  = array($token['content']);
             $fix   = $phpcsFile->addFixableError($error, $stackPtr, 'Found', $data);
             if ($fix === true) {
-                $phpcsFile->fixer->replaceToken($stackPtr, '<?php');
+                $correctOpening = '<?php';
+                if (isset($tokens[$stackPtr + 1]) && $tokens[$stackPtr + 1]['code'] !== T_WHITESPACE) {
```
jrfnl: You'll need to make the `isset($tokens[$stackPtr + 1])` explicitely test against `true` for the build to pass:
```if (isset($tokens[$stackPtr + 1]) === true && ...)```

ghunti:
@jrfnl The requested fix is pushed. Travis is failing with my commit, was it caused by some problem in my code? 
As a side note, `<?if` also works :-p So basically you don't need a whitespace after the short opening tag!

jrfnl:
> Travis is failing with my commit, was it caused by some problem in my code?

@ghunti Still a few more small code style issues to fix, nothing major - see the [travis build details](https://travis-ci.org/squizlabs/PHP_CodeSniffer/jobs/225501180#L219) for pointers.

ghunti:
@jrfnl Haven't realize that phpcs uses its own standard ^_^
It should be fine now (fingers crossed)

gsherwood:
Thanks a lot for finding and fixing this.

PR #1433, opened 2017-04-26
Add XML Schema Definition

renaatdemuynck:


jrfnl:
I haven't looked the code over in detail, but very much like the principle of this! Well done!

renaatdemuynck:
Thanks, I noticed PHPUnit has an XSD which I find very helpful so I started searching for one for PHP_CodeSniffer. I ended up writing one using the annotated sample file as an example.

**NOTE**: It would be best if a better URL for the scheme definition file was provided (e.g. [http://schema.squizlabs.com/3.0/phpcs.xsd](#)). For now it works using [https://raw.githubusercontent.com/renaatdemuynck/PHP_CodeSniffer/patch-1/phpcs.xsd](https://raw.githubusercontent.com/renaatdemuynck/PHP_CodeSniffer/patch-1/phpcs.xsd) but that's not a very user-friendly URL...

gsherwood:
Thanks a lot for doing this. I only found a couple of minor things that needed changing, but I also reformatted the XSD to remove some custom types (just fits my brain better for maintenance) and allowed tags in any order (was failing on some WordPress rulesets).

GaryJones:
@renaatdemuynck Would the intention be to suggest that `.phpcs.xml.dist` files can now start with something like:

```xml
<ruleset name="My Project"
	 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	 xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/squizlabs/PHP_CodeSniffer/master/phpcs.xsd">
```

(but with a nicer versioned URL)?

renaatdemuynck:
@GaryJones That is exactly how i did it. I tested it in Eclipse with the following URL: https://raw.githubusercontent.com/renaatdemuynck/PHP_CodeSniffer/patch-1/phpcs.xsd and it worked. But there should indeed be a nicer url like `https://schema.<domain>/<version>/<name>.xsd` or `https://<domain>/schemas/<name>_<version>.xsd` or something else (I based my example on the XSD URL of PHPUnit: http://schema.phpunit.de/4.5/phpunit.xsd).

GaryJones:
@gsherwood Is it possible to get the XSD under a friendlier versioned URL please?

glen-84:
A relative path also works ...

```xml
<ruleset
    name="example"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:noNamespaceSchemaLocation="../vendor/squizlabs/php_codesniffer/phpcs.xsd">
```

jrfnl:
@gsherwood Any plans yet on getting the XSD under a friendlier URL as suggested by @GaryJones ?

PR #1435, opened 2017-04-27
White space rules

diogocavilha:
- Spaces arround the assignment symbols.
- Single space between type cast and variable

### DIFF
--- a/CodeSniffer/Standards/PSR2/Sniffs/WhiteSpaces/AssignmentSniff.php
+++ b/CodeSniffer/Standards/PSR2/Sniffs/WhiteSpaces/AssignmentSniff.php
@@ -1,0 +1,66 @@
+<?php
+
+/**
+ * PSR2_Sniffs_WhiteSpace_AssignmentSpacingSniff.
+ *
+ * PHP version 5
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @copyright 2006-2014 Squiz Pty Ltd (ABN 77 084 670 600)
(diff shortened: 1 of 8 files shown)


gsherwood:
Thanks for the PR, but I've got a few problems with it.

Firstly, you've added two sniffs to the PSR2 standard that are checking things that PSR2 does not cover. PSR2 doesn't mention cast statements anywhere at all, and it actually intentionally omits operators and assignment from the standard. So merging this PR would instantly invalidate the included PSR2 standard.

New sniffs typically go into the Generic standard if they are not written for a specific coding standard. So that is where these ones could go. But I think there are already sniffs to cover both these cases.

Spaces around the assignment symbols is covered by `Squiz.WhiteSpace.OperatorSpacing`. You get errors like this:
```
ERROR | Expected 1 space before "="; 0 found (Squiz.WhiteSpace.OperatorSpacing.NoSpaceBefore)
ERROR | Expected 1 space after "="; 0 found (Squiz.WhiteSpace.OperatorSpacing.NoSpaceAfter)
```

Single space between type cast and variable is covered by `Generic.Formatting.SpaceAfterCast`. You get an error like this:
```
ERROR | A cast statement must be followed by a single space (Generic.Formatting.SpaceAfterCast.NoSpace)
```

All errors produced by those sniffs are also fixable and the sniffs have been around for years. So I'm not sure what functionality this PR is adding that PHPCS doesn't already have.

Can you explain what problem you were trying to solve?

diogocavilha:
Oh, I'm so sorry for that.

I'd like to create those sniffs just because I hadn't seen they already exists. This PR may be ignored so.
I'll delete this branch from my fork too.

gsherwood:
No worries, I'll close the PR. Thanks for getting back to me.

Issue #1421, opened 2017-04-11
Ternaries used in constant scalar expression for param default misidentified by tokenizer


jrfnl:
Since PHP 5.6 function param defaults (as well as class constants, class properties and static variables) can be [initialized with constant scalar expressions](http://php.net/manual/en/migration56.new-features.php#migration56.new-features.const-scalar-exprs), including [ternaries](https://wiki.php.net/rfc/const_scalar_exprs#supported_operations).

However, when using such a ternary in a function declaration for the default value:
* The `?` gets misidentified as `T_NULLABLE` while I would expect it to be `T_INLINE_THEN`
* The `:` gets misidentified as `T_COLON` while I would expect it to be `T_INLINE_ELSE`

Test code:
```php
function foo($c = ((BAR)?10:100)) {}
```

Debug output:
```
        *** START LEVEL MAP ***
        Process token 0 on line 1 [col:1;len:5;lvl:0;]: T_OPEN_TAG => <?php\n
        Process token 1 on line 2 [col:1;len:0;lvl:0;]: T_WHITESPACE => \n
        Process token 2 on line 3 [col:1;len:8;lvl:0;]: T_FUNCTION => function
        Process token 3 on line 3 [col:9;len:1;lvl:0;]: T_WHITESPACE =>
        Process token 4 on line 3 [col:10;len:3;lvl:0;]: T_STRING => foo
        Process token 5 on line 3 [col:13;len:1;lvl:0;]: T_OPEN_PARENTHESIS => (
        Process token 6 on line 3 [col:14;len:2;lvl:0;]: T_VARIABLE => $c
        Process token 7 on line 3 [col:16;len:1;lvl:0;]: T_WHITESPACE =>
        Process token 8 on line 3 [col:17;len:1;lvl:0;]: T_EQUAL => =
        Process token 9 on line 3 [col:18;len:1;lvl:0;]: T_WHITESPACE =>
        Process token 10 on line 3 [col:19;len:1;lvl:0;]: T_OPEN_PARENTHESIS => (
        Process token 11 on line 3 [col:20;len:1;lvl:0;]: T_OPEN_PARENTHESIS => (
        Process token 12 on line 3 [col:21;len:3;lvl:0;]: T_STRING => BAR
        Process token 13 on line 3 [col:24;len:1;lvl:0;]: T_CLOSE_PARENTHESIS => )
[...]

PR #1442, opened 2017-05-05
T_NULLABLE detection not working for nullable parameters and return type hints in some cases

kukulich:
The commit https://github.com/squizlabs/PHP_CodeSniffer/commit/46983c79ac8456c3a15fc32f63c5ddfb4a10430d broke T_NULLABLE detection. This should fix it and https://github.com/squizlabs/PHP_CodeSniffer/issues/1421 should still be fixed too.

The pull request is based on 2.9 branch however it's very easy to port this bugfix to 3.0 branch. You just have to change `PHP_CodeSniffer_Tokens::$emptyTokens` to `Utils\Tokens::$emptyTokens`.

It would be very nice if you could release versions 2.9.1 and 3.0.1 very quickly. Actually both versions are unusable for users that use nullable parameter and return type hints.

### DIFF
--- a/CodeSniffer/Standards/Squiz/Tests/WhiteSpace/OperatorSpacingUnitTest.inc
+++ b/CodeSniffer/Standards/Squiz/Tests/WhiteSpace/OperatorSpacingUnitTest.inc
@@ -189,3 +189,5 @@ function foo(string $bar, array $baz, ?MyClass $object) : MyClass {}
 declare(strict_types=1);
 
 function foo($c = ((BAR)?10:100)) {}
+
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks for the bug report, although I decided to code this in a slightly different way to avoid the extra loops and lines of code. I've committed a fix for versions 2 and 3, but I'm definitely not going to release without leaving the fixes in place for a while.

TomasVotruba:
Could you please release fix versions? This is blocker for long line of packages that depend on this.

See: https://github.com/slevomat/coding-standard/pull/151

TomasVotruba:
Thank you @gsherwood

PR #1448, opened 2017-05-07
Generic.Classes.OpeningBraceSameLine doesn't detect comment before opening brace

jrfnl:
There might be a comment between the opening brace and the last part of the class declaration.
The sniff currently did not take that into account.

Fixed including unit test demonstrating the issue.


gsherwood:
Thanks

PR #1450, opened 2017-05-07
Coding standard located under an installed_path with the same directory name throws an error

jrfnl:
If a "Standards" directory is called the same as one of the standards it contains and that standard is requested, an _"ERROR: the ... coding standard is not installed."_ error will be thrown and the sniffs will not run.

While I suppose this change in behaviour between PHPCS 3.x and PHPCS 2.x is intended to add support for "one standard" type of libraries, it breaks support for multi-standard libraries where one of the standards has the same name as the repository.

The change I'm proposing in this PR will maintain support for the PHPCS 2.x behaviour - thus not breaking existing registered standards - while at the same time supporting the new feature.

### DIFF
--- a/src/Util/Standards.php
+++ b/src/Util/Standards.php
@@ -164,10 +164,9 @@ class Standards
     {
         $installedPaths = self::getInstalledStandardPaths();
         foreach ($installedPaths as $installedPath) {
-            if (basename($installedPath) === $standard) {
+            $standardPath = $installedPath.DIRECTORY_SEPARATOR.$standard;
+            if (file_exists($standardPath) === false && basename($installedPath) === $standard) {
                 $standardPath = $installedPath;
-            } else {
-                $standardPath = $installedPath.DIRECTORY_SEPARATOR.$standard;
             }
 
             $path = Common::realpath($standardPath.DIRECTORY_SEPARATOR.'ruleset.xml');


gsherwood:
Just curious - what's the :fire: emoji represent in these recent PRs?

gsherwood:
Thanks a lot for finding and fixing this. Yes, it was to allow installed_paths to point directly to coding standards instead of having to point to the parent dir. Unfortunately my WordPress CS install dir is not called `WordPress`.

jrfnl:
Regarding the :fire: I've been playing a bit with commit messages lately to see if it helps in quick grasping of what a commit is all about, inspired by:
https://github.com/slashsBin/styleguide-git-commit-message
https://github.com/dannyfritz/commit-message-emoji

jrfnl:
Thanks for merging this!

> Unfortunately my WordPress CS install dir is not called WordPress.

I was running into it for both WP as well as PHPCompatibility, duh me I suppose, but it was an easy fix to make ;-)

gsherwood:
> I've been playing a bit with commit messages lately to see if it helps in quick grasping of what a commit is all about

Interesting concept. I'd personally have no hope of remembering all that :)

jrfnl:
Nor me, I use what seems to make sense at the time, though I generally find that emojis can be a powerful way to underline what a commit is about.

PR #1454, opened 2017-05-07
Squiz.WhiteSpace.OperatorSpacing is not checking spacing on either side of a short ternary operator

gmponos:
If the sniff has found a ternary statement it would return to quickly and so it would avoid doing some checks.

You can check the tests to see what I am talking about.


gsherwood:
The original commit for that skipping code is here: https://github.com/squizlabs/PHP_CodeSniffer/commit/17a165390e7321c5a4e823ff71ca990a404422e9

Looking at that change, the intention does seem to be that no space is checked _between_ the `?` and `:` operators, although the change also stops spacing being checked before and after `?:`. So I think the change in this PR corrects the intention of the sniff even though it's now checking spacing where it hasn't been checked in 3.5 years.

gsherwood:
Thanks a lot for finding and fixing this issue - and especially for the tests!

PR #1456, opened 2017-05-08
Fix some CS and docblocks

gmponos:
Not much to see here... minor changes mainly CS.

### DIFF
--- a/src/Sniffs/AbstractVariableSniff.php
+++ b/src/Sniffs/AbstractVariableSniff.php
@@ -16,7 +16,6 @@
 namespace PHP_CodeSniffer\Sniffs;
 
 use PHP_CodeSniffer\Files\File;
-use PHP_CodeSniffer\Exceptions\RuntimeException;
 
 abstract class AbstractVariableSniff extends AbstractScopeSniff
 {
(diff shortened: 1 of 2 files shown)


gsherwood:
I'd merge this in but you've got two changes in here.

The commenting ones are fine, except that I'd change `File` to `PHP_CodeSniffer\Files\File` because I use the fully qualified name everywhere (including in the same file you've changed). 

The second change to is_array doesn't come with any failing tests and so I'm not sure why this change is required here. If I was to change this code, I'd change it to use isset(), which I do for pretty much all comparisons now due to performance. So I'd do something like this:
```php
$phpReservedVars = array(
                    '_SERVER'              => true,
                    '_GET'                 => true,
                    '_POST'                => true,
                    '_REQUEST'             => true,
                    '_SESSION'             => true,
                    '_ENV'                 => true,
                    '_COOKIE'              => true,
                    '_FILES'               => true,
                    'GLOBALS'              => true,
                    'http_response_header' => true,
                    'HTTP_RAW_POST_DATA'   => true,
                    'php_errormsg'         => true,
                    );

// If it's a php reserved var, then its ok.
if (isset($phpReservedVars[$varName]) === true) {
    return;
}
```
If you'd like to make those changes, I'd be happy to merge them in.

gsherwood:
Thanks for doing that for me.

PR #1458, opened 2017-05-09
Added yield in Generic.LowerCaseKeyword

gmponos:


gmponos:
This is weird. The [current travis](https://github.com/squizlabs/PHP_CodeSniffer/pull/1458/files) failed for all PHP versions except from 7. I have 7 and it seems that's why it passed also locally.

gmponos:
Any help on how I can make the travis green?

src/Standards/Generic/Sniffs/PHP/LowerCaseKeywordSniff.php
```diff
@@ -117,7 +119,7 @@ public function process(File $phpcsFile, $stackPtr)
 
             $error = 'PHP keywords must be lowercase; expected "%s" but found "%s"';
             $data  = array(
-                      strtolower($keyword),
+                      $expected,
```
gmponos: Sorry for this change I could not hold myself 😛

jrfnl: You may also want to change this on line 128 below in that case.

jrfnl:
> Any help on how I can make the travis green?

I suspect that PHPCS does not backfill for the `yield` keyword which was only introduced in PHP 5.5, so in PHP 5.4 it would be a `T_STRING` with the content `yield` you'd need to look for.

src/Standards/Generic/Sniffs/PHP/LowerCaseKeywordSniff.php
gmponos: Done.. nice catch...

src/Standards/Generic/Tests/PHP/LowerCaseKeywordUnitTest.inc.fixed
```diff
@@ -16,4 +16,6 @@ goto a;
 function (array $a) {}
 const PRIVATE;
 HttpStatus::CONTINUE;
-function
```
gsherwood: This removes a very specific test case that is checking for an error that used to occur when the file ends with `function` (as happens when using check-as-you-type plugins). The commit is here: https://github.com/squizlabs/PHP_CodeSniffer/commit/88d3f29ffab05f68544529b63536d3ab82b7cb7f

When you see syntax errors in tests files, they are almost always there for a specific reason, so it's best to check the test file history just to be sure it's not a mistake.

gmponos: Done.

gmponos:
:( Closing since this https://github.com/squizlabs/PHP_CodeSniffer/pull/1492 covers more cases.

PR #1459, opened 2017-05-09
Fix exceptions uses in version 3.x

gmponos:
Fixed various places where the old exception of PHP_Codesniffer_Exception was used either as docblock either as an actual exception (found only one place... so it's mainly docblocks)


gsherwood:
Thanks for those fixes.

PR #1460, opened 2017-05-09
Fix docblocks about old PHP_Codesniffer_File

gmponos:
DocBlocks again 😄


gsherwood:
Thanks for these fixes.

PR #1461, opened 2017-05-09
Added scripts in composer.json

gmponos:
Before each push I had to go and find the command to run the `tests`, the `phpcs` and if I had errors for codestyle the command to fix them.

Most of the times I used to open `travis.yml` in order to find them but again I had to remember them or press CTRL+R on my command line in order to search them back.

I believe this adds simplicity. It's easy to remember and write the quickly.

So now you can execute the tests running: `composer tests`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,5 +36,10 @@
     "bin": [
         "bin/phpcs",
         "bin/phpcbf"
-    ]
+    ],
+    "scripts": {
+        "tests": "php vendor/bin/phpunit tests/AllTests.php",
+        "phpcs": "php bin/phpcs --no-cache --parallel=1",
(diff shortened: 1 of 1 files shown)


jrfnl:
`tests` sound very generic and could very likely conflict with other scripts added by other dependencies of a project....

gmponos:
> tests sound very generic and could very likely conflict with other scripts added by other dependencies of a project....

I am not sure I understand what do you mean....

gsherwood:
You don't need scripts to remember the commands. I have config files in the repo itself, so you just need to run `phpunit` if you want to run the tests, and `phpcs` if you want to check the coding standards.

PR #1464, opened 2017-05-11
Allowing no space before first variable declaration

apaarmann:
Allows first variable to be declared on a class without forcing an empty line before it

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/MemberVarSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/MemberVarSpacingSniff.php
@@ -86,7 +86,7 @@ class MemberVarSpacingSniff extends AbstractVariableSniff
 
         $prev       = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($first - 1), null, true);
         $foundLines = ($tokens[$first]['line'] - $tokens[$prev]['line'] - 1);
-        if ($foundLines === 1) {
+        if ($foundLines === 1 || ($foundLines === 0 && $tokens[$prev]['code'] === T_OPEN_CURLY_BRACKET)) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


apaarmann:
Current implementation forces to add a space even for the first member variable declared on the class, which in my opinion is not required. Otherwise, rule Squiz.WhiteSpace.MemberVarSpacing generates an error.

Some references where no space is added before the first member variable:
- https://github.com/symfony/http-foundation/blob/master/RequestMatcher.php
- https://github.com/thephpleague/container/blob/master/src/Container.php
- https://github.com/laravel/framework/blob/master/src/Illuminate/Routing/RouteCollection.php
- https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#42-properties

This pull request allows current scenario, of expecting single line space before variable along with allowing no line space for the first member variable.

gsherwood:
I couldn't merge this change in because it allows both 0 and 1 blank line before that first member var. This gives developers two different ways to format their code, which leads to inconsistency.

The way to do handle this in sniffs is to provide sniff properties, which can be changed in a ruleset.xml file. The default might require 1 blank line before the first member var, but that can be changed to 0 blank lines in a custom standard.

Normally I'd ask if you could create sniff properties for spacing, but in this case, there is already a PR for this exact thing. I haven't merge it in yet because I've been working on the 3.0 release, but I will be merging it in after I've released 3.0.1.

The PR is here: https://github.com/squizlabs/PHP_CodeSniffer/pull/1355

I'd encourage you to try out those changes and see if they work for you. If not, please comment on that PR so it can be improved before merging.

Thanks for taking the time to create this PR, even though I can't accept it this time. Hopefully I can merge some changes from you next :)

apaarmann:
Thanks Greg for the quick update on this. The reason for supporting both 0 and 1 blank lines was to have backward compatibility. Ideally, I think it should always be 0 (as used on few examples listed above).
But if you prefer to have a variable to control this, maybe we should default it to 0.

PR #1474, opened 2017-05-17
Fix slow input on stdin / require `-` for stdin

### LINKED ISSUE #993: phpcs reads from stdin which requires a terminal, and hangs if there are none.

### LINKED ISSUE #1417: Option to disable all checking of STDIN
I'm trying to run some script with bash and docker. I got some unexpected behavior, simplest way to reproduce it is running a bash script as follows:

```shell
#!/usr/bin/env bash

mkdir bug && echo "<?php phpinfo();" > bug/index.php
docker run -i --rm --name phpcs -v "$PWD":/usr/src/bug -w /usr/src/bug -e COMPOSER_ALLOW_SUPERUSER=1 php:7.1-cli bash <<EOC
    # Install tools
    apt-get update -q && apt-get install -y wget zip
    # Get composer 1.4.1
    wget https://raw.githubusercontent.com/composer/getcomposer.org/a68fc08d2de42237ae80d77e8dd44488d268e13d/web/installer -O - -q | php -- --quiet
    # Require PHP_CodeSniffer
    php composer.phar global require --dev --no-progress squizlabs/php_codesniffer
    # Run PHP_CodeSniffer
    ~/.composer/vendor/bin/phpcs -n -vv ./bug
    # Rest of the script
    echo "this could have been visible"
EOC
rm -rf bug/index.php
```

If you run this, the last part of the verbose output looks like this:
```
[...]
=> Ruleset processing complete; included 70 sniffs and excluded 0
	*** START PHP TOKENIZING ***
	Process token [0]: T_OPEN_TAG => <?php·
	Process token [1]: T_FUNCTION => function
	Process token [2]: T_WHITESPACE => ·
	Process token [3]: T_STRING => ?>
	*** END PHP TOKENIZING ***
	*** START PHP TOKENIZING ***
	Process token [0]: T_OPEN_TAG => <?php·
	Process token  1 : T_OPEN_PARENTHESIS => (
	Process token [2]: T_CLOSE_TAG => ?>
	*** END PHP TOKENIZING ***
	*** START PHP TOKENIZING ***
	Process token [0]: T_OPEN_TAG => <?php·
	Process token  1 : T_SEMICOLON => ;
	Process token [2]: T_CLOSE_TAG => ?>
	*** END PHP TOKENIZING ***
	*** START PHP TOKENIZING ***
	Process token [0]: T_OPEN_TAG => <?php·
	Process token [1]: T_FUNCTION => function
	Process token [2]: T_WHITESPACE => ·
	Process token [3]: T_STRING => ?>
	*** END PHP TOKENIZING ***
	*** START PHP TOKENIZING ***
	Process token [0]: T_OPEN_TAG => <?php·
	Process token  1 : T_OPEN_PARENTHESIS => (
	Process token [2]: T_CLOSE_TAG => ?>
	*** END PHP TOKENIZING ***
	*** START PHP TOKENIZING ***
	Process token [0]: T_OPEN_TAG => <?php·
	Process token [1]: T_ABSTRACT => abstract
	Process token [2]: T_WHITESPACE => ·
	Process token [3]: T_FUNCTION => function
	Process token [4]: T_WHITESPACE => ·
	Process token [5]: T_STRING => ?>
	*** END PHP TOKENIZING ***
	*** START PHP TOKENIZING ***
	Process token [0]: T_OPEN_TAG => <?php·
	Process token  1 : T_OPEN_PARENTHESIS => (
	Process token [2]: T_CLOSE_TAG => ?>
	*** END PHP TOKENIZING ***
	*** START PHP TOKENIZING ***
	Process token [0]: T_OPEN_TAG => <?php·
	Process token  1 : T_SEMICOLON => ;
	Process token [2]: T_CLOSE_TAG => ?>
	*** END PHP TOKENIZING ***
Creating file list... DONE (1 files in queue)
Changing into directory /usr/src/bug/bug
Processing index.php 
	*** START PHP TOKENIZING ***
	Process token [0]: T_OPEN_TAG => <?php·
	Process token [1]: T_STRING => phpinfo
	Process token  2 : T_OPEN_PARENTHESIS => (
	Process token  3 : T_CLOSE_PARENTHESIS => )
	Process token  4 : T_SEMICOLON => ;
	Process token [5]: T_WHITESPACE => \n
	*** END PHP TOKENIZING ***
	*** START TOKEN MAP ***
	*** END TOKEN MAP ***
	*** START SCOPE MAP ***
	*** END SCOPE MAP ***
	*** START LEVEL MAP ***
	Process token 0 on line 1 [col:1;len:6;lvl:0;]: T_OPEN_TAG => <?php·
	Process token 1 on line 1 [col:7;len:7;lvl:0;]: T_STRING => phpinfo
	Process token 2 on line 1 [col:14;len:1;lvl:0;]: T_OPEN_PARENTHESIS => (
	Process token 3 on line 1 [col:15;len:1;lvl:0;]: T_CLOSE_PARENTHESIS => )
	Process token 4 on line 1 [col:16;len:1;lvl:0;]: T_SEMICOLON => ;
	Process token 5 on line 1 [col:17;len:0;lvl:0;]: T_WHITESPACE => \n
	*** END LEVEL MAP ***
	*** START ADDITIONAL PHP PROCESSING ***
	*** END ADDITIONAL PHP PROCESSING ***
[PHP => 6 tokens in 1 lines]... 
DONE in 3ms (0 errors, 0 warnings)
Processing STDIN 
	*** START PHP TOKENIZING ***
	Process token [0]: T_INLINE_HTML => ····#·Rest·of·the·script\n····echo·"this·could·have·been·visible"\n
	*** END PHP TOKENIZING ***
	*** START TOKEN MAP ***
	*** END TOKEN MAP ***
	*** START SCOPE MAP ***
	*** END SCOPE MAP ***
	*** START LEVEL MAP ***
	Process token 0 on line 1 [col:1;len:24;lvl:0;]: T_INLINE_HTML => ····#·Rest·of·the·script\n
	Process token 1 on line 2 [col:1;len:39;lvl:0;]: T_INLINE_HTML => ····echo·"this·could·have·been·visible"\n
	*** END LEVEL MAP ***
	*** START ADDITIONAL PHP PROCESSING ***
	*** END ADDITIONAL PHP PROCESSING ***
[PHP => 2 tokens in 2 lines]... 
DONE in 1ms (0 errors, 0 warnings)
```

I would expect no STDIN to be read, as I provided a "./bug" directory to the phpcs script. Probably there is something going on while detecting the end of STDIN inside the HEREDOC syntax. I have tried this with some other tools as well (like phpstan, psalm and php-cs-fixer), and all of them are working like expected.

I have no clue at all as to what is going on, so if you have any idea to what I could try to pinpoint the misbehaving part, I'm happy to help.

* Edited, provided the complete log starting after ruleset processing
* Edited, removed the hardcoded xml code standard, not needed to reproduce

### LINKED ISSUE #1472: STDIN fails on slow input
I noticed that PHP CodeSniffer will fail if receiving the file on STDIN and the input on STDIN is slow.

I discovered this using [gulp-phpcs](https://github.com/JustBlackBird/gulp-phpcs) but I'm able to reproduce the issue just by piping the input (and using pv(1) to slow down the pipe).

This might be related to (some of) #993.

**Reproducing**
We'll use the coding standard from PHP CodeSniffer itself and we'll also check a file from PHP CodeSniffer. The problem seems to appear at file sizes above 8192 bytes. Reporter.php is ~11.7 kB.
```
$ wget https://github.com/squizlabs/PHP_CodeSniffer/raw/master/phpcs.xml.dist
$ wget https://github.com/squizlabs/PHP_CodeSniffer/raw/master/src/Reporter.php
```
Just cat'ing the file on standard in succeeds as expected:
```
$ cat Reporter.php | phpcs -
.

Time: 147ms; Memory: 4Mb
```
Making the pipe slow using `pv(1)` will make the check fail.
```
$ pv -qL 3k Reporter.php | phpcs -
E


FILE: STDIN
 12 | ERROR | [x] Expected 1 newline at end of file; 0 found

Time: 69ms; Memory: 4Mb

```

I have reproduced this both on my Macbook and a linux server.

I'm using "PHP_CodeSniffer version 3.0.0 (stable) by Squiz (http://www.squiz.net)." and also tried with current master (5847603911d5b49cf562b9a7368ec0f87a609139) with same result.

arnested:
Make it required to explicitly use `-` for reading from stdin.

Making it explicitly required to use will simplify the code by not needing a lot of edge cases.

Non-blocking reading from stdin was introduced in 80b156de8d440b6245e6b1dd7f8df777d99064fa because we had problems with auto detecting whether we should read from stdin or not. This is avoided by requiring `-`.

By going back to blocking read from stdin we can avoid the extra check introduced in 23bd0235c5c4ebed5d5605844335c73530028e43 for Windows.

We can also avoid the "read by line" and "usleep" introduced 8d669c50e4122c592a634580d9ea8b30c4edd0e1 and 63fafe07d6cb8588b787475a3cc041f52b05a4ac. They where introduced trying to solve what is possible the same problem still reported in #1472.

Fixes #1472.

I think this also fixes #993 and fixes #1417.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -6,7 +6,7 @@
  * and provides functions to access data stored in config files.
  *
  * @author    Greg Sherwood <gsherwood@squiz.net>
- * @copyright 2006-2015 Squiz Pty Ltd (ABN 77 084 670 600)
+ * @copyright 2006-2017 Squiz Pty Ltd (ABN 77 084 670 600)
  * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
  */
 
@@ -318,11 +318,9 @@ class Config
             $this->dieOnUnknownArg = $dieOnUnknownArg;
         }
 
-        $checkStdin = false;
         if (empty($cliArgs) === true) {
             $cliArgs = $_SERVER['argv'];
             array_shift($cliArgs);
-            $checkStdin = true;
         }
 
         $this->restoreDefaults();
@@ -354,23 +352,13 @@ class Config
         }//end if
 
         // Check for content on STDIN.
-        if ($checkStdin === true) {
-            $handle = fopen('php://stdin', 'r');
-            if (stream_set_blocking($handle, false) === true) {
-                $fileContents = '';
-                while (($line = fgets($handle)) !== false) {
-                    $fileContents .= $line;
-                    usleep(10);
-                }
+        if ($this->stdin === true) {
+            $handle       = fopen('php://stdin', 'r');
+            $fileContents = stream_get_contents($handle);
 
-                stream_set_blocking($handle, true);
-                fclose($handle);
-                if (trim($fileContents) !== '') {
-                    $this->stdin        = true;
-                    $this->stdinContent = $fileContents;
-                    $this->overriddenDefaults['stdin']        = true;
-                    $this->overriddenDefaults['stdinContent'] = true;
-                }
+            fclose($handle);
+            if (trim($fileContents) !== '') {
+                $this->stdinContent = $fileContents;
             }
         }
 


gsherwood:
I don't want people to have to use `-` just to pipe content into PHPCS.

This was the easy way out I could have taken, but it forces people to have to know an obscure CLI arg to do what I consider pretty basic stuff. Having PHPCS sit and wait for STDIN while someone types content on the screen is a serious edge case, so I added a switch for that.

What I want to do is find a way to, without any special options, have people able to use PHPCS with commands like `echo`, `cat`, and `xargs` for checking content.

The way to do this is possibly to change the behaviour so that STDIN is only ever looked at if PHPCS can't find anything to check on the command line. This gets more complex because the coding standard itself (normally inside phpcs.xml) might define files and folders to check, so there are times when PHPCS has things to check, but they are all sourced from a ruleset. It would need to wait a short time to see if anything can be found on STDIN before deciding to go ahead and process whatever the ruleset said to process. That "short time" is obviously going to be seen as "lag", so it would need to be really short.

I'm going to close this PR because this isn't how I want PHPCS to work. It would be better to have a discussion on the main issue (issue #1472), so I'll repost my comment there as well.

Issue #1488, opened 2017-05-27
--stdin-path with not existed path will ignore code checking


withwind8:
I'm using PHPCS to process data from STDIN.

When using --stdin-path to specify a existed path, everything is OK.
`cat tmp.php | phpcs --stdin-path=existed/path/to/php`

However, when a non-existent path is specified，phpcs will ignore the content from STDIN.
`cat tmp.php | phpcs --stdin-path=not/existed/path/to/php`

My requirement is to check the code that does not have real files.

In my opinion, it is not necessary to check the existence of stdin-path when processing data from STDIN. If you want, just specify the file instead of using STDIN.

I will submit a PR for it.

gsherwood:
Having a fake path ignored by the file filtering process is pretty easy, but it will then not apply any of the filters to it. And by not calling realpath on it, the path will still (potentially) be a relative path, which may or may not matter to the sniffs running over it.

I mentioned in the linked PR that I do not want to maintain code that attempts to turn a relative path into an absolute path without using the built-in realpath function, so it probably doesn't leave too many options.

The most obvious one is to say that if you specify a STDIN path, either on the CLI or inside the file itself using the phpcs_input_file: syntax, then:
1. Specify an absolute path if you need one because a relative path will not be converted to absolute
2. None of the filters (extensions, gitmodified etc) will apply, so the STDIN content will be processed regardless of your settings

I actually think this is fine because why would you specify STDIN content, and use a fake path, only to apply filters that exclude the content? It doesn't make a lot of sense to me.

Do you think those conditions are fair enough for your use case?

MichelHartmann:
I have the same problem and in my use case your conditions are totally acceptable.

gsherwood:
I ended up removing the realpath check from the filter because:
1. I don't think it is needed because all paths passed in would have already had realpath applied elsewhere.
2. This was causing ignore rules to be checked for symlink source paths instead of the target path, which is something that looks to have changed in 3.0
3. Local files were having their contents loaded up without realpath anyway, so the filter check felt inconsistent

The only good reason I could see for the realpath was that it ensured that the is_dir call was being made on a fully resolved file path, but that looks to be working fine without it anyway and was not the reason realpath was originally introduced.

gsherwood:
The end result of this change is that the filters you have set will be applied to whatever file path you have told PHPCS to use for STDIN. So STDIN will not be processed regardless - it will still have ignore patterns applied - but it can now use a fake file path.

PR #1489, opened 2017-05-27
Do not check the existence of stdin-path when processing data from STDIN 

withwind8:
#1488 

[realpath](http://php.net/manual/en/function.realpath.php) is  used to get canonicalized absolute pathname, whether to specify the file directly or to use STDIN with --stdin-path specified, and it will check the existence of the file.

In my opinion, it is not necessary to check the existence of stdin-path when processing data from STDIN. If you want, just specify the file instead of using STDIN.

This PR provides a new function to get canonicalized absolute pathname and not check existence. When processing data from STDIN with --stdin-path specified, use the new function.

### DIFF
--- a/src/Util/Common.php
+++ b/src/Util/Common.php
@@ -52,15 +52,57 @@ class Common
 
 
     /**
-     * CodeSniffer alternative for realpath.
-     *
-     * Allows for PHAR support.
+     * Get absolute path, do not check exsit
      *
      * @param string $path The path to use.
      *
      * @return mixed
      */
-    public static function realpath($path)
+    public static function absolutePath($path)
+    {
+        // Don't prefix absolute paths.
+        if (substr($path, 0, 1) !== '/') {
+            $path = getcwd().'/'.$path;
+        }
+
+        // Replace backslashes with forwardslashes.
+        $path = str_replace('\\', '/', $path);
+        // Combine multiple slashes into a single slash.
+        $path = preg_replace('/\/+/', '/', $path);
+
+        $parts = [];
+
+        foreach (explode('/', $path) as $part) {
+            // Ignore parts that have no value.
+            if (empty($part) === true || $part === '.') continue;
+            if ($part !== '..') {
+                // Cool, we found a new part.
+                array_push($parts, $part);
+            } else if (count($parts) > 0) {
+                // Going back up.
+                array_pop($parts);
+            } else {
+                // Path is outside of the root.
+                return false;
+            }
+        }
+
+        return '/'.implode('/', $parts);
+
+    }//end absolutePath()
+
+
+    /**
+     * CodeSniffer alternative for realpath.
+     *
+     * Allows for PHAR support.
+     *
+     * @param string  $path    The path to use.
+     * @param boolean $isStdin whether the file from STDIN.
+     *
+     * @return mixed
+     */
+    public static function realpath($path, $isStdin=false)
     {
         // Support the path replacement of ~ with the user's home directory.
         if (substr($path, 0, 2) === '~/') {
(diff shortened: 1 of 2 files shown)


src/Util/Common.php
```diff
@@ -52,15 +52,57 @@ public static function isPharFile($path)
...
+     */
+    public static function absolutePath($path)
+    {
+        // Don't prefix absolute paths.
+        if (substr($path, 0, 1) !== '/') {
+            $path = getcwd().'/'.$path;
+        }
+
+        // Replace backslashes with forwardslashes.
+        $path = str_replace('\\', '/', $path);
```
ktomk: Better use `strtr`

src/Util/Common.php
```diff
@@ -52,15 +52,57 @@ public static function isPharFile($path)
...
+        // Combine multiple slashes into a single slash.
+        $path = preg_replace('/\/+/', '/', $path);
```
ktomk: you could canonicalize backslashes with this as well in one. Combining multiple slashes is much appreceated.

src/Util/Common.php
```diff
@@ -52,15 +52,57 @@ public static function isPharFile($path)
...
+     *
+     * @param string $path The path to use.
+     *
+     * @return mixed
+     */
+    public static function absolutePath($path)
+    {
+        // Don't prefix absolute paths.
+        if (substr($path, 0, 1) !== '/') {
+            $path = getcwd().'/'.$path;
```
ktomk: getcwd will do a realpath (which might not be what you want according to the issue). You perhaps want to optionally inject that value as a parameter (which would allow to process on `getenv("PWD")` or some other OS-specific strategy to obtain the working directory).

src/Util/Common.php
```diff
@@ -52,15 +52,57 @@ public static function isPharFile($path)
...
+        // Combine multiple slashes into a single slash.
+        $path = preg_replace('/\/+/', '/', $path);
+
+        $parts = [];
+
+        foreach (explode('/', $path) as $part) {
+            // Ignore parts that have no value.
+            if (empty($part) === true || $part === '.') continue;
+            if ($part !== '..') {
+                // Cool, we found a new part.
```
ktomk: It's neither cool nor do we find anything here. Perhaps just `$parts[] = $path;`

src/Util/Common.php
```diff
@@ -52,15 +52,57 @@ public static function isPharFile($path)
...
+                array_push($parts, $part);
+            } else if (count($parts) > 0) {
```
ktomk: You mean `$part === '...'` I guess ....

src/Util/Common.php
```diff
@@ -52,15 +52,57 @@ public static function isPharFile($path)
...
+                // Going back up.
+                array_pop($parts);
+            } else {
+                // Path is outside of the root.
```
ktomk: No it is not. It actually is root then.

src/Util/Common.php
```diff
@@ -52,15 +52,57 @@ public static function isPharFile($path)
...
      * Allows for PHAR support.
      *
-     * @param string $path The path to use.
+     * @param string  $path    The path to use.
+     * @param boolean $isStdin whether the file from STDIN.
      *
      * @return mixed
      */
-    public static function realpath($path)
+    public static function realpath($path, $isStdin=false)
```
ktomk: Optional helper parameter, consider to instead create a second function and call that second function when intended (functional decomposition instead of conditional code).

gsherwood:
I don't like the idea of trying to expand references inside a fake file path like this. It also needs to be compatible with Windows paths and it doesn't look like it would be. I'm going to close this PR and attempt another solution, but I'll comment on the main issue.

PR #1490, opened 2017-05-27
Fix STDIN fails on slow input

### LINKED ISSUE #993: phpcs reads from stdin which requires a terminal, and hangs if there are none.

### LINKED ISSUE #1472: STDIN fails on slow input

arnested:
See #1472 for further discussion.

Fixes #993.
Fixes #1472.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,11 @@ matrix:
     - php: nightly
     - php: hhvm
 
+addons:
+  apt:
+    packages:
+      - pv
+
 before_install:
   # Circumvent a bug in the travis HHVM image - ships with incompatible PHPUnit.
   - if [[ $TRAVIS_PHP_VERSION == hhv* ]]; then composer self-update; fi
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -8,6 +8,7 @@
     <file>tests</file>
 
     <exclude-pattern>*/Standards/*/Tests/*\.(inc|css|js)</exclude-pattern>
+    <exclude-pattern>tests/*/File/*\.inc</exclude-pattern>
 
     <arg name="basepath" value="."/>
     <arg name="colors" />
--- a/src/Config.php
+++ b/src/Config.php
@@ -6,7 +6,7 @@
  * and provides functions to access data stored in config files.
  *
  * @author    Greg Sherwood <gsherwood@squiz.net>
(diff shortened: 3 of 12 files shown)


gsherwood:
I've merged in the main change, but not the unit tests. I'm not sure how I want to do CLI tests yet, so I've added some manual steps to my release process for the meantime.

Thanks a lot for this change. It seems to check out in my testing, but I was never able to replicate the problems other people were having, so hopefully I can get a few other people to test as well.

I'm going to close this as merged.

arnested:
Thank you for the merge.

I'll keep the CLI test around in a separate branch for a while: https://github.com/arnested/php_codesniffer/tree/cli-tests

PR #1491, opened 2017-05-30
[2.9] Added missing lowercase keywords

webimpress:
- function (closure)
  - parent
  - self
  - yield


jrfnl:
Unit tests are failing because the `yield` keyword did not exist yet in PHP < 5..5 and by the looks of it, is not backfilled by PHPCS ?

gsherwood:
Sorry, missed this 2.9 version.

I don't want to make changes to version 2 unless there are serious bugs, so I'm not going to merge this in as it changes behaviour. To work properly, it also needs the yield backfill, which I wont be doing in version 2 either.

I do have the 3.x PR on the todo though. Thanks.

PR #1493, opened 2017-05-30
[3.0] Fixes in abstract test case

webimpress:


gsherwood:
It looks like some of these changes were made via other commits. I've applied the remaining ones that were correct and fixed up setCliValues everywhere it is defined as this actually returns void (it used to return an array in 2.x). Thanks for pointing out these doc errors.

PR #1498, opened 2017-06-07
Sniff variables of nested anonymous classes

webimpress:
Member vars of nested classes (anonymous class in normal class) should be also sniffed.

Something is wrong with my changes, because now the error is reported twice in this line, but without my change the error is not reported at all. If I change order of the properties in anonymous class (first will be property without visibility) then all seems to be fine. Any ideas?

### DIFF
--- a/src/Sniffs/AbstractVariableSniff.php
+++ b/src/Sniffs/AbstractVariableSniff.php
@@ -41,6 +41,12 @@ abstract class AbstractVariableSniff extends AbstractScopeSniff
      */
     protected $currentFile = null;
 
+    /**
+     * The current scope that we are processing.
+     *
+     * @var int
+     */
+    protected $currentScope;
 
     /**
      * Constructs an AbstractVariableTest.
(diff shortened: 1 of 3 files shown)


webimpress:
I think it is now resolved. Please see my last commit. It is a bit more complex, but it works in my all cases.

webimpress:
@gsherwood Could you please reconsider this PR and add it to 3.1.0 release, please?
Is there any issue with it? Thanks a lot!

gsherwood:
I've scheduled it for 3.2.0 as I haven't even had time to review it.

gsherwood:
This PR deals with nested functions due to anon classes, but it doesn't actually process the variables it finds correctly. Any variable inside a function is consider a standard variable, so the member vars for the anon classes are not being processed as member vars.

I think an easier way to do this is just to check the scopes that a token is actually in rather than doing all that function tracking. I'll look into making that change and see if both these cases can be fixed at the same time.

Thanks for the test cases - they are very helpful.

gsherwood:
Sorry, should have clarified what is going on here given the tests pass. The same variable is being checked multiple times by this sniff due to it's multiple scopes. Once checks it as a member var, but also once as a normal var for each function it exists within.

gsherwood:
I've committed a change to support anon classes, stop duplicate checks, remove the function tracking, and support closures. I think it's all working, but it would be great if you could give it a test.

webimpress:
@gsherwood thanks for the fix, it seems to be fine for me right now :+1:

gsherwood:
@webimpress Thanks a lot for testing it.

PR #1506, opened 2017-06-14
Fix XMLReport not able to write filename with non-latin characters

### LINKED ISSUE #1189: XML report throws invalid character value error for some file paths
Im using PHP Storm and since last version (2016.2.1) phpcs was not working properly, giving me this warnkings:

phpcs: Warning: XMLWriter::writeAttribute(): invalid character value in C:\Projects\composer\vendorsquizlabs\php_codesniffer\CodeSniffer\Reports\Xml.php on line 66

PHP Warning:  XMLWriter::writeAttribute(): invalid character value in C:\Projects\composer\vendor\squizlabs\php_codesniffer\CodeSniffer\Reports\Xml.php on line 66

After some checks, o found the problem was with my temp path name. It turns out my user path has an accent (bad idea, many problems...): "C:\Users\Vinícius\AppData\Local\Temp... " 

For me, changing \Reports\XML.php line 66 to: 

$out->writeAttribute('name', utf8_encode($report['filename']));

...solved the problem

This is hapenning php_codesniffer 2.7.0

fMads:
This fixes #1189, where it wasn't possible to use PHP_CodeSniffer when the user's path (at least on Windows) included non-latin characters.
XMLWriter would fail with 
`phpcs: PHP Warning:  XMLWriter::writeAttribute(): invalid character value in vendor\squizlabs\php_codesniffer\CodeSniffer\Reports\Xml.php on line 66`

By doing utf8_encode, we make sure XMLWriter allows non-latin characters

### DIFF
--- a/src/Reports/Xml.php
+++ b/src/Reports/Xml.php
@@ -16,6 +16,23 @@ class Xml implements Report
 {
 
 
+	/**
+	 * Checks that the string is UTF-8 encoded, and converts it to UTF-8 if it isn't.
+	 *
+	 * @param string $string String to check and convert
+	 *
+	 * @return string
+	 */
+	private function makeSureStringIsUTF8($string)
+	{
+		if(mb_detect_encoding($string, 'UTF-8', true) === false)
+		{
+			$string	= utf8_encode($string);
+		}
+
+		return $string;
+	}
+
     /**
      * Generate a partial report for a single processed file.
      *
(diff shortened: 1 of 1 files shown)


gsherwood:
I'm not sure we can just call `utf_encode()` as it assumes too much about the encoding of the passed report path. That might be ok on the command line (I'm not 100% sure) but will presumably break any scripts wrapping PHPCS and setting the report paths with a different encoding. Possibly also automated build scripts.

Got any thoughts of that?

fMads:
Agree. I've updated the code, to check if the string is UTF-8 encoded, before doing `utf8_encode`

gsherwood:
This change would introduce a new dependency for PHPCS, so I'm not too keen on including it. I have not spent any time investigating alternate ways of approaching this.

ktomk:
Binary character encoding does not work with that magic in a stable manner. This should address specifically the problem as reported: Filenames on Windows.

Which character encoding is used for these filenames? Only if you know that, one could decide about how to address the issue. That information is missing so far with the report.

gsherwood:
Awaiting feedback of a potential fix on the main issue.

gsherwood:
Closing along with main issue.

PR #1507, opened 2017-06-14
Fix for not saving cache

kubawerlos:
After releasing `3.0.1` passing cache in command line is not working.

Reproducing:
```bash
git clone --depth=1 https://github.com/squizlabs/PHP_CodeSniffer.git
cd PHP_CodeSniffer/
composer install
bin/phpcs --cache=test.cache src/
```

No file `test.cache` will be written.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -815,7 +815,7 @@ class Config
             } else if (defined('PHP_CODESNIFFER_IN_TESTS') === false
                 && substr($arg, 0, 6) === 'cache='
             ) {
-                if ($this->cache === false || isset($this->overriddenDefaults['cacheFile']) === true) {
+                if (isset($this->overriddenDefaults['cacheFile']) === true) {
                     break;
                 }
 


gsherwood:
Thanks for finding this. I've fixed it now, but I couldn't use the exact fix you provided.

The intention of that code was to ensure the following test case worked:
- ruleset sets the `--cache=/path/to/cache arg`
- CLI uses `--no-cache`

When the `--no-cache` option is used, `$this->cache` is set to `false` but then the ruleset will override that. So what I actually forgot to add was the check of the overwritten defaults value, because the default value of the cache option is `false` to begin with.

Are you able to check this change and make sure it works for your case?

kubawerlos:
It's working (on master) although I'm getting:
```
Undefined index: cache in <path>/vendor/squizlabs/php_codesniffer/src/Config.php on line 829
```

it's `$this->overriddenDefaults['cache'] === true` condition, can't it be `isset($this->overriddenDefaults['cacheFile']) === true` as it is in line 831?

gsherwood:
> it's $this->overriddenDefaults['cache'] === true condition, can't it be isset($this->overriddenDefaults['cacheFile']) === true as it is in line 831?

That's exactly what it should have been. Thanks for testing it out and finding it - no idea what I was thinking.

I've pushed up that change to see if it works correctly for you now.

kubawerlos:
Yes, now it's working well, thanks @gsherwood.

Issue #1463, opened 2017-05-10
PHP_CodeSniffer does not register itself on composer autoload


rizkyramadhan:
I want to include PHP_CodeSniffer on my project, so I did this

 `composer require "squizlabs/php_codesniffer"`

just like any other php library and then create an index.php:
```
<?php

require("vendor/autoload.php");
var_dump(class_exists("PHP_CodeSniffer\Config"));
```

it returns false instead of true, but when I did this:

```
<?php

require("vendor/squizlabs/php_codesniffer/autoload.php");
var_dump(class_exists("PHP_CodeSniffer\Config"));
```
It returns true

So, I think PHP_CodeSniffer does not autoloading itself... 
is this intentional, or a bug ?

PatrickRose:
PHPCS doesn't register an autoloader via composer since designed to be
treated as if its a separate script

rizkyramadhan:
If so, can I use it this way ? any gotchas ?

gsherwood:
You currently need to manually include the PHPCS autoloader if you want to use it like a library.

The PHPCS autoloader is used to determine sniff codes for custom sniffs, where it has no idea what namespace is being used. So it is very important that the PHPCS autoloader be given the first chance to autoload a class. If the composer autoloader goes first, PHPCS can miss out on sniffs being loaded and it will have no idea what the fully qualified class name is, and thus what the sniff code is.

The autoloader does detect the presence of a composer autoloader and takes steps to ensure it is placed first in the autoload queue. It does this by unregistering and then reregistering the composer autoloader. It also uses the composer autoloader (if there is one) to help it figure out some file locations.

To get this working, I had to stop the composer autoloader from being used for PHPCS. If composer includes autoload.php for PHPCS too early, PHPCS' autoloader wont be able to register itself as the first autoloader in the queue.

It is possible that I might be able to allow composer to include the PHPCS autoloader with some more recent modifications that I've done, but it would take extensive testing and possibly break custom coding standards.

So for now, if you want to use PHPCS like a library, manually include its autoload.php file after you include the composer autoloader.

rizkyramadhan:
Great, thanks!

felixfbecker:
[...]

PR #1509, opened 2017-06-15
Add autoload.php to Composer autoloaded files

felixfbecker:
This makes autoloading classes work when using CodeSniffer programmatically.


gsherwood:
PHPCS can't rely on the composer autoloader, even when being used with composer. See my https://github.com/squizlabs/PHP_CodeSniffer/issues/1463#issuecomment-300637855 for an explanation.

PR #1510, opened 2017-06-16
[3.x] Generic/LineLength: add ability to ignore long comment lines completely

jrfnl:
Adds the ability to set a `$ignoreComments` property to turn off all notifications about long comment lines.

As suggested & discussed on twitter: https://twitter.com/MrDanack/status/874631725036523520

I've chosen a property-based implementation so as not to break BC for people who may have turned off the warning for long lines (so they only receive the error when the absolute limit has been reached).
If this would be implemented using a different errorcode for comments, these people who now start to see warnings about comments lines where they previously would not.
This property based implementation prevents that.

The property defaults to `false` which maintains the original behaviour of the sniff.

- [ ] Add property to the Customizable properties list in the wiki

----

Additionally, when adding the appropriate unit tests, I realized that the other custom properties for this sniff were not yet unit tested, so I've included a separate commit to fix that.

As the sniff uses the `T_OPEN_TAG` token and then walks through the whole file in one go, the properties need to be changed in the test file **_before_** the open tag is encountered.

In practice, this means that the unit tests for the changed properties need to be in separate files.


gsherwood:
Thanks a lot for contributing this. Docs updated here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#genericfileslinelength

jrfnl:
@gsherwood Thanks for merging this!

gsherwood:
This feature was added incorrectly. Instead of ignoring lines that only contain comments, using the new `$ignoreComments` property ends up ignoring all lines that end with comments, no matter how long the comment is.

Issue found when reviewing #2533

jrfnl:
Sorry about that. Clearly needed more test cases.

gsherwood:
I found the older code confusing in the sniff, so easy mistake to make. I missed it during the review as well - looked fine.

PR #1514, opened 2017-06-17
[2.x] Fix auto-fixer conflict between Squiz/ControlSignature and Generic/InlineControlStructure

jrfnl:
The WordPress project has (finally) decided to fix the remaining code style issues in the code base according to their own standard. See https://core.trac.wordpress.org/ticket/41057

An initial run of the auto-fixer yielded a number of bugs in the auto-fixers however. This PR is part of an effort to fix these issues.

As the WordPress Coding Standards are currently not (yet) compatible with PHPCS 3.x, it would be very much appreciated if this fix could (also) be merged into PHPCS 2.x.

This PR is the version for PHPCS 2.x. I'll be opening a sister-PR for the PHPCS 3.x/master branch.

----

#### Issue summary:
The autofixer for `else` / `do ... while` control structure signatures removes a new line which should be there when these control structures are used without braces.

#### Cause:
Conflict between the `Squiz.ControlStructures.ControlSignature.SpaceAfterKeyword` - _"Expected 1 space after ELSE keyword; newline found"_ - and the `Generic.ControlStructures.InlineControlStructure.NotAllowed` - _"Inline control structures are not allowed"_ - sniffs.
Basically, the Squiz sniff kicks in too early, removing the new line before the Generic sniff has had a chance to add the braces.

#### Fix:
I've fixed this by setting the `Squiz.ControlStructures.ControlSignature` sniff to ignore the "Expected 1 space after keyword" violation for inline control structures in combination with a new line.
This allows the `Generic.ControlStructures.InlineControlStructure` sniff to kick in in an early auto-fix pass to add the necessary braces.
In the next auto-fix pass, the `Squiz.ControlStructures.ControlSignature` will kick in (if still needed) as now the braces are in place.

#### Example code:
```php
if ( 'message' == $severity )
	$messages .= '	' . $error_message . "<br />\n";
else
	$errors .= '	' . $error_message . "<br />\n";
```

#### Was being fixed as:
```php
if ( 'message' == $severity ) {
	$messages .= '	' . $error_message . "<br />\n";
} else {       $errors .= '	' . $error_message . "<br />\n";
}
```

#### Expected fix /fix after this PR would be merged:
```php
if ( 'message' == $severity ) {
	$messages .= '	' . $error_message . "<br />\n";
} else {
	$errors .= '	' . $error_message . "<br />\n";
}
```

Related: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/971

### DIFF
--- a/CodeSniffer/Standards/Squiz/Sniffs/ControlStructures/ControlSignatureSniff.php
+++ b/CodeSniffer/Standards/Squiz/Sniffs/ControlStructures/ControlSignatureSniff.php
@@ -89,7 +89,7 @@ class Squiz_Sniffs_ControlStructures_ControlSignatureSniff implements PHP_CodeSn
             }
         }
 
-        if ($found !== 1) {
+        if ($found !== 1 && (isset($tokens[$stackPtr]['scope_opener']) === true || $found !== 'newline')) {
             $error = 'Expected 1 space after %s keyword; %s found';
             $data  = array(
                       strtoupper($tokens[$stackPtr]['content']),
--- a/CodeSniffer/Standards/Squiz/Tests/ControlStructures/ControlSignatureUnitTest.inc
+++ b/CodeSniffer/Standards/Squiz/Tests/ControlStructures/ControlSignatureUnitTest.inc
@@ -211,3 +211,20 @@ if (1 === $a) :
 else :
     echo 'not 1';
 endif;
+
+// Prevent whitespace after keyword fixes for inline control structures.
+if ( 'message' == $severity )
+	$messages .= '	' . $error_message . "<br />\n";
+else
+	$errors .= '	' . $error_message . "<br />\n";
+
+if ( ! is_subdomain_install() )
+	echo '<label for="blogname">' . __( 'Site Name:' ) . '</label>';
+elseif ( $a === $b )
+	echo '<label for="blogname">' . __( 'Site Domain:' ) . '</label>';
+else
+	echo '<label for="blogname">' . __( 'Site Domain:' ) . '</label>';
+
+do
+	$a = $b;
+while ( $a !== $b );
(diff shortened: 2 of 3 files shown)


jrfnl:
Darn! Found yet another example which is related to this issue, but not (yet) fixed by this PR. I'll see if I can improve on this solution some more.

Example code:
```php
while ( isset($menu[$ptype_menu_position]) || in_array($ptype_menu_position, $core_menu_positions) )
	$ptype_menu_position++;
```

Is currently being fixed as:
Example code:
```php
while ( isset( $menu[ $ptype_menu_position ] ) || in_array( $ptype_menu_position, $core_menu_positions ) ) {        $ptype_menu_position++;
}
```

gsherwood:
In my testing, your sample code is fixed like this:
```php
if ( 'message' == $severity ) {
    $messages .= '  ' . $error_message . "<br />\n";
} else {
$errors .= '    ' . $error_message . "<br />\n";
}
```
When I ran phpcbf with a standard containing only those two sniffs:
```xml
<?xml version="1.0"?>
<ruleset name="MyStandard">
  <description>My custom coding standard.</description>
  <rule ref="Squiz.ControlStructures.ControlSignature" />
  <rule ref="Generic.ControlStructures.InlineControlStructure" />
</ruleset>
```
The Generic sniff is kicking in before the Squiz one though, even though I've defined them in the opposite order.

The other thing that I'm wondering is why not just include additional sniffs to fix this code:
```php
if ( 'message' == $severity ) {
	$messages .= '	' . $error_message . "<br />\n";
} else {       $errors .= '	' . $error_message . "<br />\n";
}
```
So that it matches what you want? The Squiz and PSR2 standards will both fix that code to look the correct way, although the PEAR standard will not. You could pull individual rules from those standards to ensure that the code is fixed even if someone manually writes their code like that.

jrfnl:
Oh darn! Thanks for your response.

> The Squiz and PSR2 standards will both fix that code to look the correct way

That remark pointed me in the right direction and I think I've found the *real* culprit now. The `Squiz.ControlStructures.ControlSignature.NewlineAfterOpenBrace` error was excluded in the `WordPress-Core` ruleset. _sigh_

Sorry for wasting your (and my) time with this.

I've opened the discussion downstream to enable that check and will run some more tests.

For now, I'll close this issue.

gsherwood:
No problem

PR #1525, opened 2017-06-21
Tokenize comments starting from '/**' as doc-block comments

webimpress:
Currently comments starting form /** but without space after that
were not tokenized as doc-block comments.

For example:
```php
/**@var string */
```
was not recognized as doc-block comment.


gsherwood:
Thanks for the PR - good change.

PR #1532, opened 2017-06-25
Added missing scope_* values with T_START_NOWDOC token

### LINKED ISSUE #1531: Missing scope_* values with T_START_NOWDOC token
With `T_START_HEREDOC` token we have following values:
```
scope_condition = ...
scope_opener = ...
scope_closer = ...
```

I would expect to have the same also with `T_START_NOWDOC` token.

webimpress:
Resolves #1531

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -253,6 +253,13 @@ class PHP extends Tokenizer
                                                 'shared' => false,
                                                 'with'   => array(),
                                                ),
+                            T_START_NOWDOC  => array(
+                                                'start'  => array(T_START_NOWDOC => T_START_NOWDOC),
+                                                'end'    => array(T_END_NOWDOC => T_END_NOWDOC),
+                                                'strict' => true,
+                                                'shared' => false,
+                                                'with'   => array(),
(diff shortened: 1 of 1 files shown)


webimpress:
@gsherwood Can we have this hotfix in 3.0.2, please?

gsherwood:
> Can we have this hotfix in 3.0.2, please?

I was going to attempt to release today (assuming I find time), but I'll try.

gsherwood:
Merging this resulted in a fatal error when printing debug output due to the use of `token_name()` on a PHPCS-supplied token (`T_START_NOWDOC`). I have replaced the calls to a new Util method that supports both built-in and PHPCS-supplied tokens.

webimpress:
@gsherwood travis build didn't show any issues. You said the fatal error was on debug mode. Maybe then we should update somehow travis configuration to check also debug mode?

Should we use now everywhere new method from `Util` instead of `token_name()` ?

gsherwood:
> Should we use now everywhere new method from Util instead of token_name() ?

I made the change everywhere it was needed in my commit, so it should all be fine now.

PR #1540, opened 2017-07-03
Added missing function name tokens

webimpress:
Added to `functionNameTokens`:
- `T_SELF`
- `T_STATIC`

Registered in sniffs:
- `T_VARIABLE`
- `T_CLOSE_CURLY_BRACKET`
- `T_CLOSE_PARENTHESIS` (PHP7+)

Please see added tests for examples.


jrfnl:
I'm wondering whether adding these tokens to the `$functionNameTokens` array might not adversely affect other sniffs (both internal and in external standards).
These tokens could also be added to this sniff by grabbing the `$functionNameTokens` array and adding the extra tokens to it in the `register()` method (overloading the PEAR `FunctionCallSignatureSniff` `register()` method).

I haven't run any tests, this is just a question.

webimpress:
@jrfnl I know that adding these tags can have impact on some internals/external standards. I've checked all places where `$functionNameTokens` are used internally and I've added more test for all classes which using it. For me internally all is fine. The most tricky is of course `T_CLOSE_CURLY_BRACKET` and it could cause some issues. I wouldn't be worried about others.
Hm... but please note on the list is currently `T_STRING` and this token is used for many things, so always you need to check that it is actually function call. So ... all sniffs using `$functionNameTokens` should check next (non-empty) token and if it is not `T_OPEN_PARENTHESIS` they should skip the token.

In PHP 7.0+ we can have something even more complicated:
```php
<?php

class T {
    public function m() {
        return function ($a, $b) {
            echo $a, $b;
        };
    }
}

$t = new T;
$t->m()('a', 'b');
```
https://3v4l.org/qZ6Wk

Maybe we should consider also adding `T_CLOSE_PARENTHESIS` to the list of function name tokens...?

gsherwood:
Sorry, I've been busy and haven't commented on this.

The `functionNameTokens` array is there to provide a list of tokens that represent the text-based name of a function. This does not include variables or curly braces. These are not function name strings.

Sniffs that need to check function calls may absolutely need to include checks for these tokens, but they would need to start with the function name list and then add to it inside the sniff's register() method, as the FunctionCallArgumentSpacing sniff was already doing.

It looks to me like you've done all the hard bits already, so I'd just ask for the T_VARIABLE, T_CLOSE_CURLY_BRACKET, and T_CLOSE_PARENTHESIS tokens to be removed from the functionNameTokens array and placed directly into the sniffs. This also ensures backwards compatibility with custom sniffs and so allows me to merge it in with less risk.

If that sounds ok to you, please amend the PR and I'll target this for 3.1.0.

webimpress:
@gsherwood Sure, I'll amend my PR. What about `T_SELF` and `T_STATIC`? Should we have it in `functionNameTokens` or I should move it also?

gsherwood:
> Sure, I'll amend my PR. What about T_SELF and T_STATIC? Should we keep them in functionNameTokens or I should move them also?

My original thought was to keep them in the list. But I think the real solution would be to have PHPCS tokenize those specific cases as T_STRING so they don't get picked up when people are looking for function declarations. I'd need to have a bit of a think about if that's a good change, but either way, leave them in the list and remove them if they are changed to strings.

webimpress:
@gsherwood PR is updated now.

gsherwood:
I decided not to try changing the T_SELF and T_STATIC tokens at this time. Thanks for updating the PR and for all the tests.

PR #1542, opened 2017-07-04
Add metrics to the `PSR2.ControlStructures.SwitchDeclaration` sniff

### LINKED ISSUE #1541: Switch statement metrics
While analysing something for the WordPress Core CS project, I needed metrics on various bits of switch statements and realized that the most applicable sniff `PSR2.ControlStructures.SwitchDeclaration` does not record any metrics.

For my purposes, I've added these now locally, but I was wondering if there is any interest in (partially) adding these anyhow so they will be accessible for all analysed projects through https://squizlabs.github.io/PHP_CodeSniffer/analysis/

If so, I'd be happy to send in a PR based on the code I've already created.

jrfnl:
I've explicitely not added metrics for:
* lowercase keyword as those are already recorded by the `Generic.PHP.LowerCaseKeyword` sniff
* `break`/`return` not on a line by itself as that is already recorded by the `Generic.Formatting.DisallowMultipleStatements` sniff

Resolves #1541 

Please let me know if there is anything you'd like me to change (phrasing and such).


jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/33

PR #1545, opened 2017-07-06
Implement auto-fixing for Squiz.Operators.ValidLogicalOperators sniff

legoktm:


gsherwood:
This sniff doesn't do auto fixing because the fixed code is not identical to the original code. In this case, the operator precedence differs.

For this reason, auto-fixes tend to be limited to code formatting changes.

legoktm:
I didn't realize at the time that `&&` and `or` are not identical. In that case it doesn't make sense to auto-fix.

PR #1546, opened 2017-07-06
Squiz.PHP.LowerPHPFunctions - works incorrectly with namespaces

michalbundyra:
Yeah... I know it's crazy. Please see my tests.

Sniff should detect correctly internal PHP function even if they
preceded by `\`. Internal function names should be fixed also in imports,
but function names imported from other namespaces shouldn't be changed.


webimpress:
Please merge first #1548, this PR is based on it.

I've done a bit more in this sniff so not sure if it could be accepted.
Now I check all imported functions and all functions defined in the namespace to find out if the function is built-in or user-defined. Maybe we should check also autoloaded user-defined functions, but there, we don't know if used function has correct char case (user-defined functions with whole namespace are always lowercase - returned from method `get_defined_functions`.

webimpress:
@gsherwood Rebased, conflicts resolved. Any chance to include it in 3.3.0 release? Thanks!

gsherwood:
> Any chance to include it in 3.3.0 release?

There is too much to review in here regarding namespace detection for it to be released in 3.3.0.

michalbundyra:
@gsherwood is there any interest in this PR or should I close it? I've rebased it once already and I see it conflicted again. I'd like to have it fixed, I've provided all test cases and tests.... just review was missing. Please let me know what you think. Thanks.

PR #1548, opened 2017-07-06
Hotfix #1547 - Tokenize backslashes in import function

### LINKED ISSUE #1547: Wrong token type for backslash in use function
In case:
```php
use function \array_keys
```
backslash is tokenized as `T_STRING`.

In case:
```php
use function \MyNamespace\myFunction;
```
first backslash is tokenzed as `T_STRING` and the second one as `T_NS_SEPARATOR`.

---

If we import class:
   ```php
   use \DateTime;
   ```

or global constant:
   ```php
   use const \PHP_VERSION_ID;
   ```

or constant from other namespace:
   ```php
   use const \MyNamespace\MY_CONST;
   ```

all backslashes are tokenized as `T_NS_SEPARATOR`.

--- 

I would expect then the same behaviour for importing functions, so all backlashes should be there always tokenized as `T_NS_SEPARATOR`.

webimpress:
All backslashes in import function are now tokenized as `T_NS_SEPARATOR`.
Previously first backslash was tokenized incorrectly as `T_STRING`.
The same behaviour we have for classes and constants.

Resolves #1547

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -934,7 +934,9 @@ class PHP extends Tokenizer
                 so go forward and change the token type before it is processed.
             */
 
-            if ($tokenIsArray === true && $token[0] === T_FUNCTION) {
+            if ($tokenIsArray === true && $token[0] === T_FUNCTION
+                && $finalTokens[$lastNotEmptyToken]['code'] !== T_USE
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks a lot for this PR. Works well.

Issue #1550, opened 2017-07-07
Squiz.Commenting.FunctionComment false positive when function contains closure


josephzidell:
There is a false positive in `src\Standards\Squiz\Sniffs\Commenting\FunctionCommentSniff.php`

This code throws error: `Function return type is not void, but function is returning void here`
```php
/**
 * Test function
 *
 * @return array
 */
public function test()
{
	function () {
		return;
	};

	return [];
}
```

webimpress:
The same issue we have also in case:
```php
function test()
{
    new class() {
        public function test() {
            return;
        }
    };

    return [];
}
```

@josephzidell Here is my PR with fix for both cases: #1551

josephzidell:
Thanks for coming up with a fix so quickly

josephzidell:
@webimpress Found another, related issue

Code:
```php
/**
 * Test function
 *
 * @return void
 */
public function test()
{
	function () {
		return 4;
	};
}
```

Error: `Function return type is void, but function contains return statement`

webimpress:
@josephzidell I've added more tests https://github.com/squizlabs/PHP_CodeSniffer/pull/1551/commits/4a28555dd1536cb95bddc8ba09edf4a487c6154c
and all of them pass:
https://travis-ci.org/squizlabs/PHP_CodeSniffer/builds/252693542?utm_source=github_status&utm_medium=notification

It seems that my previous solution also covers these cases.

josephzidell:
:+1:

gsherwood:
This has now been fixed. Thanks a lot for reporting this.

PR #1551, opened 2017-07-08
Hotfix #1550 - Squiz.Commenting.FunctionComment - return type in doc-block

webimpress:
Please see #1550 for more information and see tests.


webimpress:
Rebased, conflicts resolved.

gsherwood:
Thanks for submitting this fix, but I ended up using a different method. There was already a closure and anon-class supported code block a few lines up in the file, so I used the same method that I was using there.

Thanks a lot for the test cases as well. They were very helpful.

PR #1552, opened 2017-07-09
Improve compatibility of Windows BAT files

JorisDebonnet:
The batch files didn't work for me because neither of those `@var_names@` existed (frankly I don't even know that syntax). I found several blog posts online about how to "fix" this by editing the bat file, but I figured it would be nice if this was simply included in this repository. Better out-of-the-box compatibility, yay!

So I extended the .bat files to support both the original logic, and the situation where "php" is available in the Environment Variables in Windows, which I believe is usually recommended to do. 

(Further, I think `@bin_dir@` is actually never required because we can use %~dp0, but I decided not to touch the orginal logic more than needed)

### DIFF
--- a/bin/phpcbf.bat
+++ b/bin/phpcbf.bat
@@ -5,10 +5,8 @@ REM @author    Greg Sherwood <gsherwood@squiz.net>
 REM @copyright 2006-2015 Squiz Pty Ltd (ABN 77 084 670 600)
 REM @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
 
-if "%PHPBIN%" == "" set PHPBIN=@php_bin@
-if not exist "%PHPBIN%" if "%PHP_PEAR_PHP_BIN%" neq "" goto USE_PEAR_PATH
-GOTO RUN
-:USE_PEAR_PATH
-set PHPBIN=%PHP_PEAR_PHP_BIN%
-:RUN
-"%PHPBIN%" "@bin_dir@\phpcbf" %*
+if "%PHP_PEAR_PHP_BIN%" neq "" (
+    set PHPBIN=%PHP_PEAR_PHP_BIN%
+) else set PHPBIN=php
+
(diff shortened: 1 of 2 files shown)


gsherwood:
Just FYI, the `@var_names@` are replaced during a PEAR install of PHPCS and are basically there because the script doesnt know where PHP is. In the main scripts, the hashbang line `#!/usr/bin/env php` does this and so it doesn't use these vars any more.

If you know how to change the bat files so that they will work no matter where PHP is installed or how PHPCS has been installed, that would be ideal.

JorisDebonnet:
I see, I had installed via composer instead of PEAR, that's why those variables were empty for me. 

Thanks for your feedback -- I do believe that this is what I have achieved with this patch. 
- If the 'PEAR-variables' are succesfully defined in the batch file, then it will just use those.
- If not, then: 
  - the bin-dir will default to [%~dp0](https://stackoverflow.com/questions/5034076/what-does-dp0-mean-and-how-does-it-work)
  - the php bin path will revert to just "php", which should work on most Windows installations because it is nearly always recommended to add the php bin location to Windows' Environment Variables' PATH.

gsherwood:
> I do believe that this is what I have achieved with this patch.

Sorry, I wasn't clear at all.

What I meant was: is it possible to have this file changed so it doesn't use any of the PEAR-specific `@var_names@` placeholders? That would bring it in line with the main scripts that also no longer use them.

You've probably given me enough information to play around with this myself and make sure it works when PEAR installed or not, but if you're able to provide any more guidance I would be grateful.

JorisDebonnet:
Aha, I see. Okay, I went ahead and simplified the .bat files. It's still using `%PHP_PEAR_PHP_BIN%` because I'm assuming that's a global variable set by PEAR? I just figure, if it's available anyway, why not use it; and in the other case, it will just try `php`, assuming most Windows users will have added that to their PATH.

gsherwood:
Thanks for making those changes - they worked well. And I agree, keeping that PEAR var in there makes sense if it gets replaced for some users.

Issue #1484, opened 2017-05-24
Runner::runPHPCS() is exiting


spranger:
I'd appreciate Runner::runPHPCS() returning to the caller with a return code instead of exiting the whole script.
I run Runner inside a testcase that should fail in case of codesniffer errors in my code.
But because of the exit(...) it never returns to the testcase.

gsherwood:
This would be a pretty easy change, but it's not actually going to work like that in all cases. Other parts of PHPCS also exit when they encounter errors, and there is no current way for them to instead pass up that code to the main caller so it can return an exit code. So the best I can do is just return the exit codes that the runPHPCS and runPHPCBF methods generate and you'll have to assume you don't have any config errors causing other exits.

If that still sounds ok, I'll make the change, but it's obviously not available to you right now.

If you want to check for PHPCS errors in a test case today, you're best bet is using a custom runner. Something like this: https://gist.github.com/gsherwood/aafd2c16631a8a872f0c4a23916962ac

Or copy the way the runPHPCS() method works but rip out all the bits about generators, explaining standards etc.

spranger:
I tried inheriting from the Runner-class, but failed because of several "private" functions. 
Making my own CustomRunner. I'd have to re-implement the whole file scanning stuff (or at least borrow the code from somewhere).
Normally, I'd assume an config-error etc. causing exceptions, potentially handled by the calling function.
A "main"-function can simply deal with exceptions and return codes and exit if needed.

gsherwood:
The two methods will now return their exit codes from version 3.0.1.

gsherwood:
From version 3.0.2, the exit codes will also work for internal errors. Instead of exiting, the code now throws a new DeepExit exception so that it can be caught higher up in the run methods.

hchow77:
Not sure if this should be a new issue or if it should be tacked onto this one, but noticed a version of the phpcs phar that I built from the build-phar.php script did not also return an exit code. Noticed the changes for this don't appear to have touched that script and at least from the looks of it, the stub section of build-phar.php that mirrors the code in the bin versions of phpcs/phpcbf may need to get updated as well in order to mirror this functionality for the built phar. Would it be reasonable to make that change as well or is there a specific reason why that was not added to the script?

Thanks for the consideration.

gsherwood:
@hchow77 I can't thank you enough for finding that before release - I just plain forgot to change the stub for the phars, but I've fixed that now.

[...]

PR #1562, opened 2017-07-15
Patch/fix config exception handling

ktomk:
While getting my own hands dirty with PHP Code Sniffer after a longer period of time, I stumbled over some things I'd like to address in this PR.

Please see the individual commit messages for a description of each, the commits are isolated by topic, but each commit should work clean on itself on top of master.

- fix config exception handling - little bug not importing the \Exception class
- add bootstrap file to phpunit configuration (extracted from the existing test-code)
- update on the contribution guide while I stumbled over it
- a single minor docblock class typehint while I stumbled over it in step-debugging within Phpstorm

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -12,6 +12,7 @@
 
 namespace PHP_CodeSniffer;
 
+use Exception;
 use PHP_CodeSniffer\Exceptions\RuntimeException;
 use PHP_CodeSniffer\Exceptions\DeepExitException;
 
@@ -1556,7 +1557,8 @@ class Config
             $output .= "\n?".'>';
 
             if (file_put_contents($configFile, $output) === false) {
-                return false;
+                $error = 'Failed to write to config file '.$configFile;
+                throw new RuntimeException($error);
             }
         }
 
@@ -1601,6 +1603,10 @@ class Config
         }
 
         include $configFile;
+        if (false === isset($phpCodeSnifferConfig)) {
+            $phpCodeSnifferConfig = null;
+        }
+
         self::$configData = $phpCodeSnifferConfig;
         return self::$configData;
 
--- a/tests/Core/AllTests.php
+++ b/tests/Core/AllTests.php
@@ -32,6 +32,7 @@ if (is_file(__DIR__.'/../../autoload.php') === true) {
 $tokens = new Tokens();
 
 require_once 'IsCamelCapsTest.php';
+require_once 'ConfigTest.php';
 require_once 'ErrorSuppressionTest.php';
 require_once 'File/GetMethodParametersTest.php';
 require_once 'File/FindExtendedClassNameTest.php';
@@ -62,6 +63,7 @@ class AllTests
     {
         $suite = new \PHPUnit_Framework_TestSuite('PHP CodeSniffer Core');
         $suite->addTestSuite('PHP_CodeSniffer\Tests\Core\IsCamelCapsTest');
+        $suite->addTestSuite('PHP_CodeSniffer\Tests\Core\ConfigTest');
         $suite->addTestSuite('PHP_CodeSniffer\Tests\Core\ErrorSuppressionTest');
         $suite->addTestSuite('PHP_CodeSniffer\Tests\Core\File\GetMethodParametersTest');
         $suite->addTestSuite('PHP_CodeSniffer\Tests\Core\File\FindExtendedClassNameTest');
--- a/tests/Core/ConfigTest.php
+++ b/tests/Core/ConfigTest.php
@@ -1,0 +1,204 @@
+<?php
+/**
+ * Tests for \PHP_CodeSniffer\Config
+ *
+ * @author    Tom Klingenberg <ktomk@users.github.com>
+ * @copyright 2017 Contributors
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ */
+
+namespace PHP_CodeSniffer\Tests\Core;
+
+use PHP_CodeSniffer\Config;
+
+class ConfigTest extends \PHPUnit_Framework_TestCase
+{
+
+
+    /**
+     * Flag config-file existence
+     *
+     * @var boolean
+     */
+    private $configExisted;
+
+
+    /**
+     * Configuration file path
+     *
+     * @var string
+     */
+    private $configFile;
+
+    /**
+     * Configuration file permissions
+     *
+     * @var null|int
+     */
+    private $configPerms;
+
+
+    /**
+     * Test set up
+     *
+     * @return void
+     */
+    public function setUp()
+    {
+        $this->configFile    = $this->getConfigPath();
+        $this->configExisted = file_exists($this->configFile);
+        if (true === $this->configExisted) {
+            $this->configPerms = fileperms($this->configFile);
+        }
+
+        parent::setUp();
+
+    }//end setUp()
+
+
+    /**
+     * Test tear down
+     *
+     * @return void
+     */
+    public function tearDown()
+    {
+
+        $fileExists = file_exists($this->configFile);
+        if (true === $fileExists && null !== $this->configPerms) {
+            chmod($this->configFile, ($this->configPerms & 0777));
+        }
+
+        if (false === $this->configExisted && true === $fileExists) {
+            unlink($this->configFile);
+        }
+
+        parent::tearDown();
+
+    }//end tearDown()
+
+
+    /**
+     * Test read-only config file throws a RuntimeException
+     *
+     * @return void
+     */
+    public function testReadonlyConfigThrowsPhpCodeSnifferRuntimeException()
+    {
+        $exception = $this->readonlyConfig(
+            function () {
+                Config::setConfigData('foo', 'bar');
+            }
+        );
+
+        $this->assertInstanceOf(
+            'PHP_CodeSniffer\Exceptions\RuntimeException',
+            $exception,
+            'Expected exception class'
+        );
+
+    }//end testReadonlyConfigThrowsPhpCodeSnifferRuntimeException()
+
+
+    /**
+     * Test read-only config file throws a RuntimeException
+     *
+     * @return void
+     */
+    public function testReadonlyConfigSettingValue()
+    {
+        $exception = $this->readonlyConfig(
+            function () {
+                $args = array(
+                         '--config-set',
+                         'foo',
+                         'bar',
+                        );
+                new Config($args);
+            }
+        );
+
+        $this->assertNotNull(
+            $exception,
+            'An expected exception was not thrown'
+        );
+
+        $this->assertInstanceOf(
+            'PHP_CodeSniffer\Exceptions\DeepExitException',
+            $exception
+        );
+
+    }//end testReadonlyConfigSettingValue()
+
+
+    /**
+     * Make callback operate on a read-only config file, helper function
+     * for tests.
+     *
+     * @param callable $callback to be run with a readonly config
+     *
+     * @return NULL|\Exception
+     */
+    private function readonlyConfig($callback)
+    {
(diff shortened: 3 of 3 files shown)


ktomk:
Btw, the fix in the exception handling should fix another flaw with exit code handling, ref https://github.com/squizlabs/PHP_CodeSniffer/issues/1484#issuecomment-308614032

CONTRIBUTING.md
```diff
@@ -1,13 +1,20 @@
...
-Before you contribute code to PHP\_CodeSniffer, please make sure it conforms to the PHPCS coding standard and that the PHP\_CodeSniffer unit tests still pass. The easiest way to contribute is to work on a checkout of the repository, or your own fork, rather than an installed PEAR version. If you do this, you can run the following commands to check if everything is ready to submit:
+Before you contribute code to PHP\_CodeSniffer, please make sure it conforms to the PHPCS
+coding standard and that the PHP\_CodeSniffer unit tests still pass. The easiest way to
+contribute is to work on a checkout of the repository, or your own fork, rather than an
+installed PEAR version. If you do this, you can run the following commands to check if
+everything is ready to submit:
 
     cd PHP_CodeSniffer
-    php scripts/phpcs
+    composer install
```
jrfnl: There is absolutely no need to run `composer install` unless you don't have PHPUnit installed on your system.

CONTRIBUTING.md
```diff
@@ -1,13 +1,20 @@
...
+    php bin/phpcs
 
-Which should give you no output, indicating that there are no coding standard errors. And then:
+Which should give you only informal output, indicating that there are no coding standard
+errors. And then:
 
-    phpunit
+    vendor/bin/phpunit
```
jrfnl: If you already have PHPUnit installed, the command will be `phpunit` not some composer path to phpunit.

tests/AllTests.php
```diff
@@ -11,24 +11,12 @@
...
-}
-
-if (defined('PHP_CODESNIFFER_CBF') === false) {
-    define('PHP_CODESNIFFER_CBF', false);
-}
-
-if (defined('PHP_CODESNIFFER_VERBOSITY') === false) {
-    define('PHP_CODESNIFFER_VERBOSITY', 0);
-}
+require_once __DIR__.'/bootstrap.php';
```
jrfnl: As you've put bootstrap file in the `phpunit.xml.dist`, there should be no need to require the file.

tests/Core/AllTests.php
```diff
@@ -11,23 +11,7 @@
...
-if (defined('PHP_CODESNIFFER_VERBOSITY') === false) {
-    define('PHP_CODESNIFFER_VERBOSITY', 0);
-}
-
-if (is_file(__DIR__.'/../../autoload.php') === true) {
-    include_once __DIR__.'/../../autoload.php';
-} else {
-    include_once 'PHP/CodeSniffer/autoload.php';
-}
+require_once __DIR__.'/../bootstrap.php';
```
jrfnl: Dito - this should not be needed.

tests/Standards/AllSniffs.php
```diff
@@ -14,23 +14,7 @@
...
-if (defined('PHP_CODESNIFFER_VERBOSITY') === false) {
-    define('PHP_CODESNIFFER_VERBOSITY', 0);
-}
-
-if (is_file(__DIR__.'/../../autoload.php') === true) {
-    include_once __DIR__.'/../../autoload.php';
-} else {
-    include_once 'PHP/CodeSniffer/autoload.php';
-}
+require_once __DIR__.'/../bootstrap.php';
```
jrfnl: Dito - this should not be needed.

src/Config.php
```diff
@@ -1601,6 +1603,10 @@ public static function getAllConfigData()
         }
 
         include $configFile;
+        if (false === isset($phpCodeSnifferConfig)) {
+            $phpCodeSnifferConfig = null;
```
jrfnl: The simpler way to do this is to turn the order around:
```php
$phpCodeSnifferConfig = null;
include $configFile;
```
... has the same result with less logic involved.

tests/Core/ConfigTest.php
```diff
@@ -0,0 +1,204 @@
...
+
+
+    /**
+     * Get path of the configuration file.
+     *
+     * @return string
+     */
+    private function getConfigPath()
+    {
+        return __DIR__.'/../../CodeSniffer.conf';
```
jrfnl: This presumes a stand-alone install and will not work when running the tests in a PEAR install.

CONTRIBUTING.md
ktomk: This is for the contribution guidelines and mirrors what will be executed in the CI environment, see .travis.yml. At least it comes closer.

CONTRIBUTING.md
ktomk: Same here. As I've been running over some kind of flaw, it might be good to add --no-cache and --parallels=1.

tests/AllTests.php
ktomk: This is for backwards compat, some users might have configured their own phpunit.xml which overrides the .dist and still need that. Also I didn't want to change the previous behavior, just add.

CONTRIBUTING.md
jrfnl: Eh... I'm a contributor and don't use Composer for contributing to PHPCS... got the phpunit.phar in my PATH variable.
All I'm saying is, don't presume everyone uses the same dev process as you do.

src/Config.php
ktomk: Unless the config file - as include - unsets the variable. Had it first like you write, then moved it before the access, that's all.

tests/AllTests.php
jrfnl: If a contributor is smart enough to have their own `phpunit.xml` (and yes, I do have one), they should be smart enough to keep up with what's being changed in the `phpunit.xml.dist` IMHO.

tests/Core/ConfigTest.php
ktomk: The test still works there as the file will be created if it does not exists (and removed afterwards). But you raise a good point, can you give the location for PEAR?

CONTRIBUTING.md
jrfnl: Those sound like good potential suggestions for debugging if/when needed. Not sure whether they should be the default though.

ktomk:
@jrfnl: Thanks for the feedback, much appreceated! I need some assistance with the config file location in PEAR testing setups it should work for both cases indeed!

CONTRIBUTING.md
ktomk: I don't, I can add a section making clear that this is applicable for Github workflow, e.g. git clone, composer install, change, push, PR triggers CI, PR gets review etc. I think that is quite common, but I don't have to press this on someone, it was more that the previous info was outdated.

CONTRIBUTING.md
ktomk: For me it was that I only became aware *after* the Travis build as it did flag an issue my local install didin't catch (w/o --parallels=1).

tests/AllTests.php
ktomk: And if you execute the tests really old-skool PEAR style, it is still needed IIRC. It's fine to keep it the same as earlier. I didn't want to refactor the whole test-suite only for that change which is safe IMHO.

tests/Core/ConfigTest.php
jrfnl: It's in the `/data/PHP_CodeSniffer` directory in a PEAR install.

Typical directory layout for the relevant directories:
```
- PHP
     - /data
          - /PHP_CodeSniffer       <- this is where the `.conf` file will be
     - /PEAR
          - /PHP
               - /CodeSniffer      <- this is the PHP_CodeSniffer root
     - /tests
          - /PHP_CodeSniffer      <- this is where the tests live
```

It might be good to get a second opinion on this though as I have to admit this is based PHPCS 2.x as I haven't updated my PEAR install for a while - I'm using a stand-alone git checkout instead myself.

src/Config.php
jrfnl: Interesting. Have you come across that situation ?
AFAIK the config file is only supposed to be created & written to by PHPCS, though someone can of course manually edit it. Adding an `unset()` in the file though, seems like a willful act of sabotage... but maybe I'm missing something ?

tests/AllTests.php
jrfnl: Good argument for keeping the include :+1:

gsherwood:
Thanks for the PR, but it too many different things for me to be able to merge it in. A seperate PR for each change makes it a lot easier to review.

The change I'm most interested in is what you describe as config exception handling and how it fixes a flaw in the exit code handling. I've already fixed the exit code issue (it will be released in 3.0.1) so I'm specifically interested in knowing more about the issue you have found and fixed.

If you'd like to amend this PR to describe and fix the Config class issue (do not include the unit tests as I've got more reservations about those) then I could look at merging this into an upcoming release. Or if you'd prefer to open a new issue, that works for me too.

jrfnl:
> (it will be released in 3.0.1)

@gsherwood Pst... Noticed this a couple of times before... don't you mean 3.0.2 ? or 3.1 ? If I remember correctly, you released 3.0.1 about a month ago :heart_eyes_cat:

gsherwood:
> @gsherwood Pst... Noticed this a couple of times before... don't you mean 3.0.2 ? or 3.1 ? If I remember correctly, you released 3.0.1 about a month ago 😻

I'm living in the past -_-

3.0.2 it is

ktomk:
@gsherwood: Sure, I've updated the PR so it covers only that change. The flaw is that PHPCS-RuntimeExceptions aren't caught one level up (by not importing \Exception into the file's namespace ) and not re-thrown as deep-exit exceptions when the config file is not readable or similar. See the single commit.

ktomk:
I still could benefit from some guidance regarding the testcase. It runs within Travis and on my local Setup but as it has been mentioned earlier in review, the location to the config-file when tests are run could vary. This must not break the testing but I have not run the tests with a Pear-like install.

gsherwood:
Thanks for the info. There isn't really any need to use the main Exception class for two statements, so I added the NS separators in there manually.

The problem of not being able to read the config file needs a bit more work than this I think. Looking into it, PHPCS isn't going to be able to exit cleanly without doing some additional checks before trying to include the file, or muting the include line.

Then the big decision is: does this read failure get reported to the user? If the config file is not readable, PHPCS should probably exit with an error. But that's behaviour I wouldn't change until at least a .1 release. It shouldn't really be a BC break because PHPCS would be throwing errors now anyway, but it needs a bit of thought.

Also, setting the config data to null breaks other parts of the code, especially when asking PHPCS to show config data. Setting it to an empty array works best in my testing, and stops the `getAllConfigData` method from running multiple times with the same result.

As for the unit tests, they need to work in a PEAR install and a git clone. I also wouldn't want them actually touching the real config data file, so PHPCS itself would need to be refactored to allow the unit tests to set up a fake config file just for testing, but that's a low priority for me. I'd rather just focus on deciding what to about unreadable conf files.

I'll add this to the 3.1.0 milestone and come back to it once I'm doing dev on that version.

ktomk:
Greg for the fix, this sounds fine to me. The flaw on it's own is rather limited and local and easy to fix. I just did the testing for getting my hands dirty and actually provoking that error. Some static analysis should be able to find it, too. I don't know if this needs in depth checking / fixing, sure injecting the config path would be nice, but most likely future material and of lower priority. I can see that, too. Just my 2 cents. Keep up the good work!

gsherwood:
The change I ended up making is to just exit if the config file cannot be read. I also fixed up the --config-set code so that it exits in the same way (is was throwing a RuntimeException previously).

Thanks again for help finding this issue.

ktomk:
Thank you for your ongoing support! I wish I could help more often.

PR #1563, opened 2017-07-17
Only include the custom autoload file once.

jrfnl:
I was running into trouble with a library containing several standards which all use the same custom autoload file, resulting in fatal `class already declared` errors for the autoloader class.

This simple change fixes that.


gsherwood:
Thanks for that. Should have been like that from the start.

jrfnl:
You're welcome & glad to see it merged so quickly :+1:

PR #1568, opened 2017-07-20
Patch Wrong sniff codes with "special namespace"

rhorber:
It is possible to use a namespace other than `StandardName\Sniffs`, i.e. `MyProject\PHPCS\Standard\Sniffs`. But then the sniff codes look like `Standard.Category.Code`, which imho is wrong.
This PR takes care of this situation. The gathered namespaces are saved with their standard name. When a sniff code is requested, it will be looked up with the namespace and then this name will be returned (not just a namespace level).

Edit/Addendum:
Probably its not the best solution with a static property, but I do not know the internas very well ;)
So feel free to adjust/correct things, as you like.

### DIFF
--- a/src/Util/Common.php
+++ b/src/Util/Common.php
@@ -444,6 +444,19 @@ class Common
             return $newName;
         }
 
+        // Handle special namespace.
+        $namespace = substr($newName, 0, $sniffPos);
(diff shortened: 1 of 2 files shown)


gsherwood:
I'm going to close this as a discussion is taking place on the main issue and this PR wouldn't be merge in as-is.

PR #1573, opened 2017-07-24
Squiz.WhiteSpace.LanguageConstructSpacing does not properly check for tabs and newlines

webimpress:
As we discussed in https://github.com/squizlabs/PHP_CodeSniffer/pull/1337#discussion_r129082516 I've created this PR to issues noted in PR #1337

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
@@ -51,6 +51,11 @@ class LanguageConstructSpacingSniff implements Sniff
     {
         $tokens = $phpcsFile->getTokens();
 
+        if (isset($tokens[($stackPtr + 1)]) === false) {
+            // Skip if there is no next token.
+            return;
+        }
+
         if ($tokens[($stackPtr + 1)]['code'] === T_SEMICOLON) {
             // No content for this language construct.
             return;
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
@@ -65,7 +65,7 @@ class LanguageConstructSpacingSniff implements Sniff
             $content = $tokens[($stackPtr + 1)]['content'];
             if ($content !== ' ') {
                 $error = 'Language constructs must be followed by a single space; expected 1 space but found "%s"';
-                $data  = array($tokenContent = Util\Common::prepareForOutput($content));
+                $data  = array(Util\Common::prepareForOutput($content));
                 $fix   = $phpcsFile->addFixableError($error, $stackPtr, 'IncorrectSingle', $data);
                 if ($fix === true) {
                     $phpcsFile->fixer->replaceToken(($stackPtr + 1), ' ');
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
@@ -11,6 +11,7 @@ namespace PHP_CodeSniffer\Standards\Squiz\Sniffs\WhiteSpace;
 
 use PHP_CodeSniffer\Sniffs\Sniff;
 use PHP_CodeSniffer\Files\File;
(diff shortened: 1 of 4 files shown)


webimpress:
@gmponos please have a look

src/Standards/Squiz/Tests/WhiteSpace/LanguageConstructSpacingUnitTest.inc
```diff
@@ -30,3 +30,10 @@ return;
 return $blah;
 return   $blah;
 return($blah);
+
+return	$tab;
+return
```
gmponos: I believe that this change here will raise a lot of issues. According to my understanding (maybe I am wrong) the goal of this sniff is to ensure that there are no two spaces between language constructs and $variables/functions etc.

So as I understand it the goal is to avoid this:
```
return  ['test' => 'value'];
```

and not this:
```
return  
    [
        'test' => 'value'
    ];
```

src/Standards/Squiz/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
```diff
@@ -50,17 +51,21 @@ public function process(File $phpcsFile, $stackPtr)
...
         if ($tokens[($stackPtr + 1)]['code'] === T_WHITESPACE) {
-            $content       = $tokens[($stackPtr + 1)]['content'];
-            $contentLength = strlen($content);
-            if ($contentLength !== 1) {
-                $error = 'Language constructs must be followed by a single space; expected 1 space but found %s';
-                $data  = array($contentLength);
+            $content = $tokens[($stackPtr + 1)]['content'];
+            if ($content !== ' ') {
+                $error = 'Language constructs must be followed by a single space; expected 1 space but found "%s"';
+                $data  = array($tokenContent = Util\Common::prepareForOutput($content));
```
gmponos: Personally I avoid assignments that can be tricky to the eye. I would've added the assignment on a different line.

gmponos: Now that I see it more carefully the `$tokenContent` is not used anywhere. So why is it assinged to a variable first?

webimpress: It was my mistake, the variable shouldn't be there, thanks. Now removed.

src/Standards/Squiz/Tests/WhiteSpace/LanguageConstructSpacingUnitTest.inc
webimpress: You have right, but sniff description says:
> Ensures all language constructs contain a single space between themselves and their content.

And I think that it is exactly what the sniff should do, we didn't have covered that case previously in tests, so I think @gsherwood is gonna decide what to do.

jrfnl: That could be avoided by checking for ' ' or `$phpcsFile->eolChar`

src/Standards/Squiz/Tests/WhiteSpace/LanguageConstructSpacingUnitTest.inc
```diff
@@ -30,3 +30,10 @@ return;
...
+$newLine;
+
```
gmponos: if this is gonna be merged I believe it should cover one more test case. What happens if there is a single space after return and then a new line.

webimpress: @gmponos I've added this test case and it seems to be fine 😄 Please see 5324d96

gmponos:
@gsherwood can you check on this PR I need for another one?

gsherwood:
The intention of this sniff is for a newline to be banned. That's why a return followed by `space` `newline` is auto-fixed by removing the newline character. It's just an old sniff I wrote before I knew better, so this fix for tabs and newlines looks correct.

gsherwood:
Thanks for this fix.

src/Standards/Squiz/Tests/WhiteSpace/LanguageConstructSpacingUnitTest.inc
glen-84: To work around this problem, you can use:

```php
return (
    [
        'test' => 'value'
    ]
);
```

PR #1583, opened 2017-07-31
Fixes the return codes for PHPCBF.

jcchavezs:
At the moment the logic for the return code is a bit confusing and there is a logical mistake in the `if`s:
```
/** line 203 **/ if ($this->reporter->totalFixed === 0) {
```
If this is `false`, meaning that the total fixed are greater than `0` there is no chance at all that  that this can be `true`:

```
/**line 214 **/ if ($this->reporter->totalFixable === 0) {
```
as of `$this->reporter->totalFixable >= $this->reporter->totalFixed` by definition.

Ping @gsherwood @jrfnl

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -149,7 +149,7 @@ class Runner
     /**
      * Run the PHPCBF script.
      *
-     * @return array
+     * @return int
      */
     public function runPHPCBF()
     {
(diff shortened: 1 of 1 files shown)


jrfnl:
I don't see anything particularly wrong with the code.

Your changes could be considered a simplification, but are based on an invalid assumption:

> as of $this->reporter->totalFixable >= $this->reporter->totalFixed by definition.

Based on the code, I get the impression that `totalFixable` is the total **after** the fixer has finished, i.e. the remaining fixable errors which for whatever reason did not get fixed.

I haven't tested the code or ran it through a debugger, but what with the fixer loops and one fix paving the way for another, I can imagine that `$this->reporter->totalFixed >= $this->reporter->totalFixable` is a quite likely situation.
And even when `totalFixed` > `totalFixable`, what with the loops, there might still be some `Fixables` which haven't been fixed.

To explain better what I mean:
* Initial run on the code reports 10 fixable errors/warnings.
* The first loop of phpcbf can fix 8 of these.
* In the next loop, the code now throws 6 fixable errors/warnings as the fixes made in the first loop, cause other sniffs to be triggered and throw errors.
* This second loop fixes all 6 errors.
* The third loop now throws 20 fixable errors as the fixes made in the second loop cause other sniffs to be triggered and throw errors.
* etc

gsherwood:
Exactly what @jrfnl said. With that in mind, does the code make more sense now?

jcchavezs:
@jrfnl I actually came up with this PR because when having `x` fixable errors and `x` being fixed, I kept receiving a code `2` but I did explore the sequence you mention.

jrfnl:
@jcchavezs Have you run `phpcs` after you've run `phpcbf` ? And if so, what does the report say ? Based on the `2` exit code, I would expect there to still be some fixable errors left over.

Issue #1577, opened 2017-07-27
Generic.InlineControlStructureSniff breaks with a comment between body and condition in do while loops


ht-bs:
A comment between the body and condition in do while loops, as illustrated below, breaks the Generic.InlineControlStructureSniff by producing the following error:

> Inline control structures are not allowed (Generic.ControlStructures.InlineControlStructure.NotAllowed)

```
do
{
}
// Comment
while (true);
```

Is this intended behavior? If not, a fix would be to replace the search for anything but a `T_WHITESPACE` token, with a search for anything but `Tokens::$emptyTokens` in line 86:
https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Generic/Sniffs/ControlStructures/InlineControlStructureSniff.php#L86

gsherwood:
Closed by PR #1587

PR #1587, opened 2017-08-03
Allow comments between body and condition in do while loops.

ht-bs:
Fixes issue #1577.


gsherwood:
Thanks a lot for fixing this, and for the test.

I like to credit contributors in the changelog, so please let me know your name if you'd be happy to be included. If not, no problem, and thanks again.

Issue #1592, opened 2017-08-06
Wiki: Additional clarification needed for the exclude-pattern


jrfnl:
> The ignore patterns can also be complete regular expressions. If you do specify a regular expression, be aware that * is converted to .* for the convenience in simple patterns, like those used in the example above. So use * anywhere you would normally use .*.

Ref: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-files-and-folders

I believe it would be good to also mention that this also means that any `.` characters will need to be escaped - `\.` - to be seen as a literal `.`.

I ran into this when running PHPCS over a new sniff I am looking to pull here and couldn't figure out why PHPCS did not examine the file.
```bash
phpcs -p -s ./src/Standards/Generic/Tests/CodeAnalysis/AssignmentInConditionUnitTest.php
```

The reason was, of course, that the exclude pattern in the PHPCS own ruleset did not escape the dot in the pattern:
```xml
<exclude-pattern>*/Standards/*/Tests/*.(inc|css|js)</exclude-pattern>
```
which caused the pattern to match on:
**./src/Standards/Generic/Tests/CodeAnalysis/AssignmentInC**onditionUnitTest.php

I will send in a PR to fix this, but hope that clarifying the wiki instructions will help prevent this for others.

If you like, I can also look into fixing this within the `Filter::shouldIgnorePath()` method, but that will become more complicated as the code would need to verify if a `.` is already escaped or not to determine whether or not to escape it.

gsherwood:
Good idea. I've made the change to that note.

PR #1593, opened 2017-08-06
Fix `<exclude-pattern>` in PHPCS native ruleset

jrfnl:
See issue #1592 for the rationale.

And apparently the file I was testing wasn't the only one being accidentally excluded, so fixed the minor code style issues being thrown up for previously excluded files in a separate commit.


gsherwood:

jrfnl:
You're welcome ;-)

PR #1594, opened 2017-08-06
Add new `AssignmentInCondition` sniff to `Generic`

jrfnl:
Most of the time variable assignments being made in conditions are unintentional and are in actual fact typos which were intended as a comparison.

This sniff will warn when any such a variable assignment is found in a condition.

Note: the sniff does currently **not** detect variable assignments in the conditional part of ternaries.

Inspired by [this twitter thread](https://twitter.com/jrf_nl/status/893891839903838209), or rather: the responses I received to my tweet.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -167,6 +167,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="InlineControlStructureStandard.xml" role="php" />
        </dir>
        <dir name="CodeAnalysis">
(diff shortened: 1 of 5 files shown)


gsherwood:
Nice idea, and seems to be working well. I think this sniff wants me to change some `while` loops with assignments into `do/while` loops, which is probably a good thing. I'm not going to though 😉  too busy at the moment.

djibarian:
@gsherwood Not an easy thing to convert `while($row = getNextRow()){}` to a do/while.

jrfnl:
@djibarian That's why assignments in `while` conditions have a separate error code... See #1725

PR #1606, opened 2017-08-12
Print progress to stderr

exussum12:
When redirecting output to save it, The progress info is also sent.

Progress info should be send to stderr so it can either be redirected
seperatly or just left ot print the progress.

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -731,7 +731,7 @@ class Runner
 
         // Show progress information.
         if ($file->ignored === true) {
-            echo 'S';
+            fwrite(STDERR, 'S');
         } else {
             $errors   = $file->getErrorCount();
             $warnings = $file->getWarningCount();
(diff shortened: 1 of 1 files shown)


jrfnl:
Why ?

exussum12:
For when you are directing the real report to save to a file. Progress output is mixed in with this and not what is expected. Calling without redirection this pr makes 0 difference as both streams are shown

jrfnl:
Interesting. I would expect one to use the `--report-full=./myfilename.txt` command line option to save the real report to file instead of redirecting it. Would this PR have any influence of that (and similar) cl options ?

exussum12:
It would be identical to using the report-full option, With much less syntax.

this pr would not change existing behavior , unless someone is parsing out the progress from the standard output,

gsherwood:
This is not a change I want to make right now. There are already options to send different reports to different files, so anyone already capturing STDOUT clearly wants all the debug output they are asking for. Plus, it is a BC break.

I may consider something like this for a major version, but I'd expect all non-report output (including verbose output, debug output, memory/time usage output) to be sent to STDERR as well.

gsherwood:
I'm not going to accept this PR so I'll close it, but I'd be happy to discuss this change on a new issue if you'd like to create one.

PR #1607, opened 2017-08-12
Generic/DisallowTabIndent: multi line comments + error message/code fix

jrfnl:
Sister-PR to #1599.

----

### Use the correct error message and code

The "TabsUsed" error message and code were only set before the loop and not reset within the loop.
 In practice this meant that as soon as the first non-indent error is encountered, the `$error` and `$errorCode` would be overruled and for any indent errors found after that, the wrong message and error code would be thrown.

### Allow the sniff to fix tab indents for multi-line /* */ comments

This is the same as done in #1599 for the DisallowSpaceIndent sniff.

### Metrics

I also noticed that the metrics were not in line with how they are being recorded in the other sniff as the improvements which were previously added there in #1404 were not yet applied here, quite apart from the metrics needing some love for the `/* */` comment.

So I've fixed that too.

For in-depth details of the effect on the metrics of this fix, please see: https://gist.github.com/jrfnl/9a5824916969dc6cf27a24d750d78319

~~While the metrics for both sniffs are now better than they were before and more in line with each other, there is still a difference between them.~~

~~The indentation metrics created by the `DisallowSpaceIndentSniff` are probably best to use for the historic metrics overviews as - except for comments - the `DisallowTabIndentSniff` will report "mixed" when it encounters tab indents combined with precision spacing, while the `DisallowSpaceIndentSniff` can distinguish between the precision spacing and non-precision spacing and will adjust the metrics accordingly.~~

**[Edit] By combining the improvements I'd already made with @gsherwood's improvements in https://github.com/squizlabs/PHP_CodeSniffer/commit/bbe55a50b79f99a2a196ab827e5cdffa1a754207 and https://github.com/squizlabs/PHP_CodeSniffer/commit/2aa0e09c962b29ea2a04a595c554d8a014a50c97, the metrics should now be quite similar for both sniffs, though I haven't ran any checks to verify.**


jrfnl:
As a side-note (not covered by this PR), I believe that the "nonIndentTab" part (i.e. tabs found for mid-line alignment or in comment texts) should not be covered by this sniff, as:
1. It's an undocumented "feature" of the sniff, i.e. unexpected behaviour.
2. It is not in line with the sister-sniff which does not do this.
3. Indentation and mid-line alignment are two different things and projects/frameworks may have different rules for each.

For WPCS, I actually created a separate sniff to address "no tabs for mid-line alignment", while tabs are enforced for indentation. I'd be happy to pull that sniff here if you're interested and would suggest - though probably in a major version as it is a BC break - removing that part from this sniff.

jrfnl:
Rebased for merge conflicts & updated to allow for the changes made related to #1702

jrfnl:
Rebased for merge conflicts & code style...

gsherwood:
Thanks for pointing me here, and thanks for the PR.

jrfnl:
You're welcome & glad to help.

PR #1611, opened 2017-08-15
Resolve problem: illegal character in Tokenizer

chybaDapi:
When in file we have polish characters (eq: "Błąd") or Cyrillic, and we run Sniffer via PHPStorm IDE Sniffer will stop and we will get error: 
```iconv_strlen(): Detected an illegal character in input string in /vendor/squizlabs/php_codesniffer/src/Tokenizers/Tokenizer.php on line 193```
This proposal will fix this issue.

### DIFF
--- a/src/Tokenizers/Tokenizer.php
+++ b/src/Tokenizers/Tokenizer.php
@@ -189,8 +189,7 @@ abstract class Tokenizer
             ) {
                 // There are no tabs in this content, or we aren't replacing them.
                 if ($checkEncoding === true) {
-                    // Not using the default encoding, so take a bit more care.
-                    $length = @iconv_strlen($this->tokens[$i]['content'], $this->config->encoding);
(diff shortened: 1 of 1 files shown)


mourawaldson:
Hey,
I've just noticed that on line 304 there's still a call for "iconv_strlen", besides there are 2 other places where a condition on function_exitsts is looking for iconv_strlen as well, I guess that should be changed as well.

mourawaldson:
By the way, thanks for the initiative to fix this.

Mopolo:
Hi,

I have the same issue and this change resolves it for me.

jougene:
Please merge this issue to master. We all need it.

gsherwood:
> Please merge this issue to master. We all need it.

I am no going to merge in a change that modifies the PHPCS requirements so heavily.

I must have missed this PR, so I'm going to close it because this issue is not going to be fixed in this way.

PR #1614, opened 2017-08-15
Fix file level doc for test class

178inaba:
Fix DisallowLongArraySyntax to DisallowShortArraySyntax.


gsherwood:
Thanks for fixing that.

PR #1630, opened 2017-08-27
Added new sniff that checks about multiple Class/Interface/Trait on a file

### LINKED ISSUE #1627: Generic.Files.OneClass/Interface/TraitPerFile sniffs don't work for full PSR-4 compliance
Hi, we're currently working on migrating the MediaWiki codebase to adhere to [PSR-4](http://www.php-fig.org/psr/psr-4/). PSR-4 requires that each file contain only one class/trait/interface. So I added the following to our ruleset:
```xml
	<rule ref="Generic.Files.OneClassPerFile" />
	<rule ref="Generic.Files.OneInterfacePerFile" />
	<rule ref="Generic.Files.OneTraitPerFile" />
```

However, these three sniffs together still allow you to have one file with a class AND a trait in it. Is it possible to have a sniff that ensures there's only one class or one interface or one trait per file?

gmponos:
That sounded interesting and I thought that I could help here.

closes #1627

### DIFF
--- a/src/Standards/Generic/Sniffs/Files/OneObjectStructurePerFileSniff.php
+++ b/src/Standards/Generic/Sniffs/Files/OneObjectStructurePerFileSniff.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * Checks that only one object structure is declared per file.
+ *
+ * @author    Andy Grunwald <andygrunwald@gmail.com>
+ * @copyright 2010-2014 Andy Grunwald
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ */
+
+namespace PHP_CodeSniffer\Standards\Generic\Sniffs\Files;
+
+use PHP_CodeSniffer\Sniffs\Sniff;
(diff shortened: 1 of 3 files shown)


src/Standards/Generic/Sniffs/Files/OneObjectStructurePerFileSniff.php
```diff
@@ -0,0 +1,55 @@
...
+ * @copyright 2010-2014 Andy Grunwald
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ */
+
+namespace PHP_CodeSniffer\Standards\Generic\Sniffs\Files;
+
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Files\File;
+
+class OneObjectStructurePerFileSniff implements Sniff
```
gmponos: I am not sure about the naming @gsherwood can you advise here.

gsherwood:
Thanks for doing this. I think the name is fine.

PR #1633, opened 2017-08-30
Squiz.PHP.EmbeddedPhp: two bug fixes for when there's only a comment

jrfnl:
### Squiz.PHP.EmbeddedPhp: prevent adding superfluous semi-colon

When a snippet of embedded PHP contains only a comment, the `Squiz.PHP.EmbeddedPhp` sniff would add a semi-colon before the comment which is unnecessary and makes for untidy code.

Includes unit tests.

Without this fix, the unit test snippet would be "fixed" to look like:
```php
<?php ;/* translators: My sites label */ ?>
```

While the change looks more extensive because of the change in indentation, in reality all I did was wrap the relevant code in an `if ($prev !== $stackPtr) {}`.

----

### Squiz.PHP.EmbeddedPhp: prevent removing embedded PHP when no space between comment and close tag

When there is no space between a `/* .. */` comment and the close tag, PHPCS would report an `Empty embedded PHP tag found (Squiz.PHP.EmbeddedPhp.Empty)` error and the fixer would accidentally remove a complete line of embedded PHP.

Includes unit tests.

This bug is related to the inconsistency in `findNext()` as previously reported in #1319

### DIFF
--- a/src/Standards/Squiz/Sniffs/PHP/EmbeddedPhpSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/EmbeddedPhpSniff.php
@@ -341,31 +341,33 @@ class EmbeddedPhpSniff implements Sniff
         }
 
         $prev = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($closeTag - 1), $stackPtr, true);
-        if ((isset($tokens[$prev]['scope_opener']) === false
-            || $tokens[$prev]['scope_opener'] !== $prev)
-            && (isset($tokens[$prev]['scope_closer']) === false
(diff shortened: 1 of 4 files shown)


GaryJones:
> While the change looks more extensive because of the change in indentation, in reality all I did was wrap the relevant code in an if ($prev !== $stackPtr) {}.

[Diff with whitespace ignored](https://github.com/squizlabs/PHP_CodeSniffer/pull/1633/files?w=1).

jrfnl:
@GaryJones Neat trick :sunglasses:

gsherwood:
Thanks a lot for finding and fixing those issues.

jrfnl:
You're welcome!

Issue #1584, opened 2017-07-31
Bad merge


andreydiveev:
https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Squiz/Tests/PHP/HeredocUnitTest.inc

jrfnl:
Duplicate of #1314. This is a valid test case.

Issue #1314, opened 2017-01-27
Merge conflict in master branch


zookee1:
Hej Team,
you've got a merge-conflict on master in this file:
/master/CodeSniffer/Standards/Squiz/Tests/PHP/HeredocUnitTest.inc

Cheers,
Nick

jrfnl:
@zookee1 No, that's a valid test case. See the actual commit which added the test case. 8129f9b034f1868c76e9552db5469e08c9a4839c

zookee1:
I see.. My PHPStorm just found it and I thought this was an actual error. :)
Thanks for the clarification!

PR #1634, opened 2017-08-30
Removed commited git confilict

JarJak:


jrfnl:
Duplicate of #1314, #1584.
Not to discourage you from contributing, but this is a valid test case.

JarJak:
Oh, OK :)

30.08.2017 2:51 PM "Juliette" <notifications@github.com> napisał(a):

gmponos:
Since there are already two other issues I would've suggest to add a comment there in order not to be confusing.

gsherwood:
I added the comment. Thanks.

PR #1640, opened 2017-09-04
Squiz.Strings.DoubleQuoteUsage replaces tabs with spaces when fixing

jrfnl:
### Squiz/DoubleQuoteUsage: Use original string when testing and fixing double quotes

If a string contains tabs, they were previously - unintentionally - replaced by spaces while that is not the responsibility of this sniff.

Includes unit test,

--- 

While examining the sniff code to fix the above issue, I also noticed two other optimizations which could benefit this sniff:

### Squiz/DoubleQuoteUsage: Skip past subsequent tokens in a multi-line string

As the last token of the multi-line string is known, it can be returned to skip past subsequent tokens in a multi-line string instead of checking each string token.

### Squiz/DoubleQuoteUsage: Make check for double quoted string a little more specific and strict

Only check strings which are finished double quotes strings.
This offers additional protection against examining unfinished strings during live code checking.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Strings/DoubleQuoteUsageSniff.php
+++ b/src/Standards/Squiz/Sniffs/Strings/DoubleQuoteUsageSniff.php
@@ -44,11 +44,6 @@ class DoubleQuoteUsageSniff implements Sniff
     {
         $tokens = $phpcsFile->getTokens();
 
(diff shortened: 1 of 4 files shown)


gsherwood:
Thanks for fixing this + those two optimisations were a good idea.

jrfnl:
Thanks for merging this in :+1:

PR #1641, opened 2017-09-04
Docker support

korchasa:
Docker support for docker-based CI/CD environments. 

Usage:
  docker run --rm -v /full/loca/path/to/code:/app <image_name> phpcs ./src 
  docker run --rm -v $PWD:/app <image_name> phpcs ./src ./tests --report=junit
  docker run --rm -v $PWD:/app <image_name> phpcbf ./src

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM alpine:latest
+
+WORKDIR /phpcs
+
+COPY . /phpcs
+
+RUN apk --no-cache add \
+        ca-certificates \
+        php7 \
+        php7-ctype \
+        php7-tokenizer \
+        php7-simplexml \
+        php7-xmlwriter \
+    && ./bin/phpcs ./src ./tests \
+    && ln -s /phpcs/bin/phpcs /usr/local/bin/phpcs \
+    && ln -s /phpcs/bin/phpcbf /usr/local/bin/phpcbf
+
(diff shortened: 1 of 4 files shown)


gsherwood:
It's normally better to include a single feature in a PR so we can discuss it more easily, so I'll comment on each feature separately.

## Docker Support
I don't know why PHPCS should include a dockerfile. PHPCS is already available in a number of docker containers, along with a wide range of other PHP QA tools. What is the benefit of including a dockerfile directly in this project?

## Teamcity Report
I'm not familiar with Teamcity, but it's obviously a Java build tool like Jenkins. If that's the case, I would have thought the more generic checkstyle report format could be used, like in Jenkins, rather than needing a custom reporting format just for Teamcity. But if that's not the case, I'm still reluctant to include this report format in the core as I've never had any requests for it and I don't use the tool, so I don't really want to maintain it.

But you don't need the report included in the core to use it. If you save your report class somewhere, you can tell PHPCS to use it when generating reports. So instead of running `phpcs /path/to/code --report=teamcity` you would instead run `phpcs /path/to/code --report=/path/to/Teamcity.php`.

If I have more requests for direct Teamcity support, I will certainly consider adding the new report format. But I added the ability to use 3rd-party report classes for cases just like this, so I'd encourage you to use that feature for now.

PR #1643, opened 2017-09-06
Fix only when fixer is enabled

webimpress:
I have found unused variable `$fix` in two places, and I think there were missing conditions to run fixer only when `$fix === true`.


gsherwood:
Thanks for fixing this. Thankfully, these commands don't really do anything unless you are fixing the file, but they are wasting time by being called and so should always be wrapped in an IF statement.

PR #1644, opened 2017-09-06
Generic/SubversionProperties: minor efficiency fix

jrfnl:
Don't examine this file again after the first PHP open tag has been passed.


gsherwood:
Thanks. I had forgotten this sniff even existed :)

PR #1647, opened 2017-09-08
Generic.PHP.RemovedFunctions sniff

stevegrunwell:
The `Generic.PHP.DeprecatedFunctions` sniff works well for detecting when a function has been deprecated within PHP, but fails to identify when a function has been removed from PHP. This new sniff is meant to fill in that gap by checking for functions that have been removed.

Unfortunately, this requires maintaining a list of removed functions, which is populated by parsing the PHP changelog. Fortunately, functions don't get removed very regularly, so it should be pretty maintainable moving forward (especially considering they're explicitly keyed to the version in which they were removed).

This work could be extended by filling in more functions that were deprecated earlier in the PHP release cycle and their alternatives, but for an initial submission the list is limited to those removed in PHP 7.0 (my team was recently bit by someone running `mysql_connect()` on a PHP 7.0 server).


jrfnl:
Just my two cents: I don't think this sniff belongs in PHPCS. There is a dedicated external PHPCS standard which can check code for PHP cross-version compatibility which covers this and much much more.
See: https://github.com/wimg/PHPCompatibility

stevegrunwell:
Ah, wasn't aware of that package. Does seem reasonable to keep this in a separate package for those concerned with it.

Thanks for the heads up!

jrfnl:
@stevegrunwell If you see anything which can be improved/is missing in the PHPCompatibility standard, please feel invited to contribute to it. Pull requests welcome :heart_eyes_cat:

PR #1650, opened 2017-09-09
Feature - function comment throw tag - process functions without doc comment

webimpress:
This PR contains feature and fix which I find out on working on the feature...

The feature is possibility to enable processing functions without doc comment.
Previously sniff was processing only functions with doc comment before. I've added new variable which allows to change that behaviour. By default the variable is set to `true` so we skip functions without doc comment, to keep backward compatibility. Basically, previous functionality is not changed here, and if user would like to enable this feature must do it by setting the variable in his ruleset.

The bug was with anonymous classes, sniff suggest adding throw tag on wrong function, when function contains an anonymous function with method which throw an exception:

```php
class A
{
  public function b()
  {
    return new class {
      public function c()
      {
        throw new \Exception();
      }
    }
  }
}
```

so in above example throw should be documented for method `c` of anonymous class, not on method `A::b`.

Another one, maybe shouldn't be considered as a bug, is the following:
```php
function x() {
    try {
        throw new \Exception();
    } catch(\Exception $e) {
    }
}
```
In that case `\Exception` is not going to be forced to be documented anymore.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
@@ -63,9 +63,7 @@ class FunctionCommentThrowTagSniff extends AbstractScopeSniff
                 return;
             }
 
-            if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG
-                && $tokens[$commentEnd]['code'] !== T_COMMENT
-            ) {
+            if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG) {
                 // Function doesn't have a doc comment.
                 return;
             }
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
@@ -15,6 +15,12 @@ use PHP_CodeSniffer\Util\Tokens;
 
 class FunctionCommentThrowTagSniff extends AbstractScopeSniff
 {
+    /**
+     * If functions without doc comment should be skipped.
+     *
+     * @var boolean
+     */
+    public $skipFunctionsWithoutDocComment = true;
 
 
     /**
@@ -51,16 +57,20 @@ class FunctionCommentThrowTagSniff extends AbstractScopeSniff
         $find[] = T_WHITESPACE;
 
         $commentEnd = $phpcsFile->findPrevious($find, ($currScope - 1), null, true);
-        if ($tokens[$commentEnd]['code'] === T_COMMENT) {
-            // Function is using the wrong type of comment.
-            return;
-        }
+        if ($this->skipFunctionsWithoutDocComment === true) {
+            if ($tokens[$commentEnd]['code'] === T_COMMENT) {
+                // Function is using the wrong type of comment.
+                return;
+            }
 
-        if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG
-            && $tokens[$commentEnd]['code'] !== T_COMMENT
-        ) {
-            // Function doesn't have a doc comment.
-            return;
+            if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG
+                && $tokens[$commentEnd]['code'] !== T_COMMENT
+            ) {
+                // Function doesn't have a doc comment.
+                return;
+            }
+        } else if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG) {
+            $commentEnd = $currScope;
         }
 
         $currScopeEnd = $tokens[$currScope]['scope_closer'];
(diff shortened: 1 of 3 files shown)


src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
```diff
@@ -15,6 +15,12 @@
 
 class FunctionCommentThrowTagSniff extends AbstractScopeSniff
 {
+    /**
+     * If functions without doc comment should be skipped.
+     *
+     * @var boolean
+     */
+    public $skipFunctionsWithoutDocComment = true;
```
webimpress: Default value is `true` to keep the previous functionality unchanged.

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
```diff
@@ -51,16 +57,18 @@ protected function processTokenWithinScope(File $phpcsFile, $stackPtr, $currScop
         $find[] = T_WHITESPACE;
 
         $commentEnd = $phpcsFile->findPrevious($find, ($currScope - 1), null, true);
-        if ($tokens[$commentEnd]['code'] === T_COMMENT) {
-            // Function is using the wrong type of comment.
-            return;
-        }
+        if ($this->skipFunctionsWithoutDocComment === true) {
```
webimpress: These changes are looking a bit confusing, but there is only added that condition and for the body is added one more indent.

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
```diff
@@ -51,16 +57,18 @@ protected function processTokenWithinScope(File $phpcsFile, $stackPtr, $currScop
...
+            if ($tokens[$commentEnd]['code'] === T_COMMENT) {
+                // Function is using the wrong type of comment.
+                return;
+            }
 
-        if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG
-            && $tokens[$commentEnd]['code'] !== T_COMMENT
```
webimpress: The second part of the condition has been removed as redundant, it's always true, because in line 61 we already return on opposite condition.

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
```diff
@@ -51,16 +57,18 @@ protected function processTokenWithinScope(File $phpcsFile, $stackPtr, $currScop
...
-        ) {
-            // Function doesn't have a doc comment.
-            return;
+            if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG) {
+                // Function doesn't have a doc comment.
+                return;
+            }
+        } else if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG) {
+            $commentEnd = $currScope;
```
webimpress: The `else if` statement is added here - we use `$commentEnd` to add errors, in case when we do not skip functions without doc comment and we are processing function without doc comment errors are going to be added on `$currScope` which is `function`.

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
```diff
@@ -70,7 +78,7 @@ protected function processTokenWithinScope(File $phpcsFile, $stackPtr, $currScop
         $currPos          = $stackPtr;
         $foundThrows      = false;
         while ($currPos < $currScopeEnd && $currPos !== false) {
-            if ($phpcsFile->hasCondition($currPos, T_CLOSURE) === false) {
+            if ($this->isLastScope($tokens[$currPos]['conditions'], $currScope) === true) {
```
webimpress: The previous condition was not sufficient because it was checking _any_ closure. It doesn't work correctly in case nested closures, and doesn't also work for anonymous classes.

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
```diff
@@ -160,21 +168,23 @@ protected function processTokenWithinScope(File $phpcsFile, $stackPtr, $currScop
...
-                continue;
-            }
-
-            if ($tokens[($tag + 2)]['code'] === T_DOC_COMMENT_STRING) {
-                $exception = $tokens[($tag + 2)]['content'];
-                $space     = strpos($exception, ' ');
-                if ($space !== false) {
-                    $exception = substr($exception, 0, $space);
+        $throwTags = array();
+        if ($commentEnd < $currScope) {
```
webimpress: Above condition is added - the condition is satisfied always when we skip functions without doc comment and if false when we process functions without doc comment. There is no any other changes, just added one more indent.

gsherwood:
Sorry, just getting to this. What's the benefit of this sniff processing functions with no comments? It's a commenting sniff.

webimpress:
@gsherwood As I said in the PR description, would be nice to have possibility to report errors on method without doc-comment that throws some exception, so `@throws` annotation is missing. Currently we don't have any other sniff which will allow it, and the functionality (of detecting throws in the function) will be common if we have another sniff. If you are not going include this here I hope you will be able to fix bugs I've noted (see my examples in the description). Thanks!

gsherwood:
> As I said in the PR description, would be nice to have possibility to report errors on method without doc-comment that throws some exception

If you've got a missing doc comment, it just feels like you'd get loads of errors. None of the other commenting sniffs process empty docblocks by throwing all errors, so I'm not really that keen on getting this one to do it.

Maybe consider including `PEAR.Commenting.FileComment.Missing` in your ruleset to error when function docblocks are missing.

> I hope you will be able to fix bugs I've noted

I've fixed the anon class issue. The sniff was written in a bit of a weird way, searching for throws tags inside functions, so I've changed it to work like a normal commenting sniff.

I've haven't done anything about catching an exception your throwing yourself. I suspect that actually requires a bit more work to ensure it is correct in all cases (including re-throwing) and I don't have the time to do that right now. If you'd like to work on it and submit another PR, I'll certainly take a look.

I'm going to close this PR. Thanks for submitting it, and for reporting that bug.

webimpress:
> Maybe consider including `PEAR.Commenting.FileComment.Missing` in your ruleset to error when function docblocks are missing.

But what in case the doc block is needed only with throw tag?
For example in ZF we decided to not include in doc block params and return type if these are specified in method signature, but we always have to include throws tags., for example:
```
public function foo(DateTime $dt, string $s) : int
{
}
```
Does not require any doc block, but if I add throw inside I would need one.
As you can see then `PEAR.Commenting.FileComment.Missing` is not a solution for me.

My changes were backward compatible, so you can enable them only if you really want.

PR #1651, opened 2017-09-11
Squiz.WhiteSpace.SuperfluousWhitespace - Multi Line String check

nigelhough:
# Superflous whitespace check for multi line string
Have extended the SuperfluousWhitespace check to be able to detect and fix whitespace at the EOL in a multi line string deceleration.

Have included an option to ignore this check and set the default value to true so this is a non breaking change.

## Why
This is in my teams coding standards and something that we weren't able to get reported by Code Sniffer in our build pipeline. It is often found in multi line SQL definitions where a space has been left after the SELECT, FROM or WHERE.

I originally thought of adding a new custom rule but thought it fitted well within the existing SuperfluousWhitespace so decided to extend.

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
@@ -39,6 +39,15 @@ class SuperfluousWhitespaceSniff implements Sniff
      */
     public $ignoreBlankLines = false;
 
+    /**
+     * If TRUE, whitespace rules are not checked for multi line strings.
+     *
+     * Multi line strings are string definitions that spans multiple lines.
+     *
+     * @var boolean
+     */
+    public $ignoreMultiLineStrings = true;
+
 
     /**
      * Returns an array of tokens this test wants to listen for.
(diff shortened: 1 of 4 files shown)


gsherwood:
(Sorry it has taken so long to review this)

This is a tricky one because it sits outside what PHPCBF is designed to do, which is detect and fix formatting issues. The fixes that PHPCBF applies should just be syntax adjustments and never change the actual behaviour of an application. In this case, it would be modifying the content of a variable or function argument, which isn't something I want to take responsibility for.

Of course, this might (and, I assume, does) work perfectly fine in your application, so it's probably really helpful. But once I include it in the PHPCS core sniffs, I need to deal with any problems that arise if someone enables this option.

I think the most I'd be comfortable with this sniff doing is to throw a warning for multiline strings that contain whitespace at the end of a line. I don't even think that needs to be an option - it just needs to have a different error code so people can turn it off, or turn it into an error instead of a warning. I'd make that particular error code unfixable as well.

If you think that would still be useful to your, I'd be happy to merge in a PR that does that.

nigelhough:
Hey @gsherwood 

Thanks for you comments on this pull request. I agree with your points and suggestions.

I am currently travelling at the moment and not back until August. When I get back I will re-work the changes as suggested and either update or re-raise this pull request.

Thanks

gsherwood:
> I am currently travelling at the moment and not back until August.

Not a problem. Enjoy.

PR #1654, opened 2017-09-13
Travis: run the build tests against PHP 7.2 as well

jrfnl:
The new Trusty images as per Sept 7 include an image for PHP 7.2 (even though it hasn't been released yet) and nightly is now PHP 7.3-dev.


gsherwood:
Thanks

PR #1656, opened 2017-09-15
Using the '--sniffs' argument has a problem with case sensitivity

Fischer-Bjoern:
When using the '--sniffs'  argument with phpcs you have to enter the the sniffs case sensitive for the sniff to work.
However when using any other casing (like all lowercase) of the same letters, phpcs will not fail.
Instead it will find the sniff and run it, but it will never report any errors.
If different letters are used, phpcs obviously complains directly that it can't find the sniff.

This is since the matching of errors to sniffs is done with in_array(), which is case sensitive. But the loading of sniffs uses strtolower() in various places.

So this patch makes the '--sniffs' argument fully case sensitive.

### DIFF
--- a/src/Ruleset.php
+++ b/src/Ruleset.php
@@ -200,15 +200,15 @@ class Ruleset
 
         $sniffRestrictions = array();
         foreach ($restrictions as $sniffCode) {
-            $parts     = explode('.', strtolower($sniffCode));
-            $sniffName = $parts[0].'\sniffs\\'.$parts[1].'\\'.$parts[2].'sniff';
+            $parts     = explode('.', $sniffCode);
+            $sniffName = $parts[0].'\Sniffs\\'.$parts[1].'\\'.$parts[2].'Sniff';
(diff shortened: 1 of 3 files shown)


jrfnl:
I've run into this a number of times as well. While I would personally prefer the solution to be for PHPCS to handle this in a case-insensitive manner (fault tolerant), any solution which will not make PHPCS silently fail on this is an improvement.

gsherwood:
The intention was to make this string case insensitive, but I obviously stuffed that up during the file rewrite. I've made a change to the File class so that sniff error codes are compared in a case insensitive way.

I think there is still a fair bit of work left to make rules inside ruleset.xml files case insensitive, but that would need to be left for a feature release if it is deemed worth doing. Rulesets are a bit different because they are fairly static, but CLI args tend to be written quickly and benefit from case insensitive checks where possible.

Thanks for letting me know about this, even though I didn't end up using the PR.

Fischer-Bjoern:
Sure. What matters is that the problem is fixed.

PR #1658, opened 2017-09-15
Enable process substitution

exussum12:
When using bash process substitution leaves a file descriptor to read
from. PHP needs its own stream for this, so this translates the two.

Example

`phpcs --file-list=<(git diff origin/master --name-only -- '*.php')`

To only run code sniffer for the changed files

### DIFF
--- a/src/Util/Common.php
+++ b/src/Util/Common.php
@@ -70,6 +70,11 @@ class Common
             }
         }
 
+        // Check for process substitution
+        if (strpos($path, '/dev/fd') === 0) {
+            return str_replace('/dev/fd', 'php://fd', $path);
+        }
+
         // No extra work needed if this is not a phar file.
         if (self::isPharFile($path) === false) {
             return realpath($path);
--- a/src/Util/Common.php
+++ b/src/Util/Common.php
@@ -70,7 +70,7 @@ class Common
             }
         }
 
-        // Check for process substitution
+        // Check for process substitution.
         if (strpos($path, '/dev/fd') === 0) {
             return str_replace('/dev/fd', 'php://fd', $path);
         }


gsherwood:
That command runs for for me on using bash. Can you provide any more info as to what the problems is you are trying to fix and how to replicate it?

exussum12:
What os ? 

What does echo <(echo test) 

Show ?

It should show something like /dev/FD/5

Php can't read file descriptors like that and instead needs the conversation to happen which this pr does

exussum12:
In fact. Make a change which introduces an error. Then run the above command without the patch nothing will return. With the path just that one file will be run though phpcs

exussum12:
Without this patch

phpcs --file-list=<(git diff origin/master --name-only -- '*.php')
ERROR: The specified file list "" does not exist

Usage: phpcs [-nwlsaepqvi] [-d key[=value]] [--colors] [--no-colors] [--stdin-path=<stdinPath>]


with this patch

just the changed files are run though phpcs (MUCH faster than running the whole lot)

It saves doing the following

git diff origin/master --name-only -- '*.php' > fileListToCheck.txt
phpcs --file-list=fileListToCheck.txt
rm fileListToCheck

gsherwood:
> What os ?
macOS 10.12

> What does echo <(echo test)
> 
> Show ?
```
$ echo <(echo test)
/dev/fd/63
```

> Make a change which introduces an error. Then run the above command without the patch nothing will return.

That's exactly how I tested it, and it worked fine. I put some whitespace at the end of my Runner.php file and ran your exact command. This is what I got:
```
$ phpcs --file-list=<(git diff origin/master --name-only -- '*.php')
E 1 / 1 (100%)



FILE: src/Ruleset.php
 14 | ERROR | [x] Whitespace found at end of line

Time: 418ms; Memory: 23.25Mb
```

If I run PHPCS normally, I get a full run and my error at the end:
```
$ phpcs
.....................................E......................  60 / 496 (12%)
............................................................ 120 / 496 (24%)
............................................................ 180 / 496 (36%)
............................................................ 240 / 496 (48%)
............................................................ 300 / 496 (60%)
............................................................ 360 / 496 (73%)
............................................................ 420 / 496 (85%)
............................................................ 480 / 496 (97%)
....S.S.S.S.....                                             496 / 496 (100%)



FILE: src/Ruleset.php
 14 | ERROR | [x] Whitespace found at end of line

Time: 16.98 secs; Memory: 44.5Mb
```

What version of PHPCS are you running?

Have you tried the built-in Git modified filter to see if that works?
```
$ phpcs --filter=gitmodified
E 1 / 1 (100%)



FILE: src/Ruleset.php
 14 | ERROR | [x] Whitespace found at end of line

Time: 471ms; Memory: 23.25Mb
```

exussum12:
Ok in that case Mac must make /dev/fd/63 a real file instead of a file descriptor 

This is not how Linux works 
https://bugs.php.net/bug.php?id=53465
https://unix.stackexchange.com/questions/74604/how-to-open-process-substituted-file-from-php

The modified filter does not work as I need it to. This allows the CI server to only run the changed files in a single pr for large code bases this greatly speeds up the process. I imagine this works with or without the patch on Mac though I don't have one to confirm. 
Using the php stream wrapper should work everywhere

gsherwood:
Thanks for providing that info. You're obviously right about macOS, but I was able to replicate and confirm the fix on a linux box and also confirm everything is still working on macOS.

Thanks for the PR.

PR #1665, opened 2017-09-19
Feature - define arrays in ruleset

webimpress:
Currently if we want override array property in ruleset we have to do:
```xml
<property name="forbiddenFunctions" type="array"
    value="print=>echo,create_function=>null" />
```

It's fine as long as the array content is short.

With this feature the new method to define arrays in the ruleset is introduced:
```xml
<property name="forbiddenFunctions" type="array">
    <element key="print" value="echo"/>
    <element key="create_function" value="null"/>
</property>
```

### DIFF
--- a/src/Ruleset.php
+++ b/src/Ruleset.php
@@ -945,21 +945,37 @@ class Ruleset
                     if (isset($prop['type']) === true
                         && (string) $prop['type'] === 'array'
                     ) {
-                        $value  = (string) $prop['value'];
                         $values = [];
-                        foreach (explode(',', $value) as $val) {
-                            list($k, $v) = explode('=>', $val.'=>');
-                            if ($v !== '') {
-                                $values[trim($k)] = trim($v);
-                            } else {
-                                $values[] = trim($k);
+                        if (isset($prop->element) === true) {
+                            foreach ($prop->element as $element) {
+                                if ($this->shouldProcessElement($element) === false) {
+                                    continue;
+                                }
+
+                                $value = (string) $element['value'];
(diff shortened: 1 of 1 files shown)


gsherwood:
A change to the format like this would need to be postponed until a 4.0, but I think it is a good idea.

webimpress:
@gsherwood With this PR I *have not* removed previous version of defining array, both are working fine. 
I think then you can release it in 3.x and deprecate old method, and in 4.0 you will just remove old method. What do you think?

gsherwood:
> I think then you can release it in 3.x and deprecate old method, and in 4.0 you will just remove old method. What do you think?

I would prefer one method and one set of docs for how to use it.

jrfnl:
> I think then you can release it in 3.x and deprecate old method, and in 4.0 you will just remove old method. 

> I would prefer one method and one set of docs for how to use it.

I kind of like the idea of having a change-over period available for such a big change in how array property values are passed (even if the new method would not be documented in the wiki until the 4.x release).

webimpress:
@gsherwood Any chance to have it in 3.x release? This PR just add new ability to define array values which is more clear. I see similar change has been made in 3.x release with "@codingStandards" comments - deprecated old notation and introduced the new one.

gsherwood:
This has now been merged in and will be in 3.3.0. The old syntax is now deprecated and will be removed in version 4. Docs were updated.

Thanks for this PR - it's a much better syntax. I did change the way the debug info prints though because a `print_r()` broke the formatting.

Issue #1666, opened 2017-09-19
Infinite loop in Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php


louisl:
When using this sniff in a linting environment -  If I have the code something like the following as I'm typing:
```
$somthing = array(
'a' => 'b',
'c' =>
);
```
PHP goes to into an infinite loop.

I've tracked it down to line 486
```
$nextContent = $phpcsFile->findNext(
                    Tokens::$emptyTokens,
                    ($nextToken + 1),
                    $arrayEnd,
                    true
                );
```

After that I've added:

```
if ($nextContent === false) {
    break;
}
```
So I'm exiting the loop if there's no next content.

orlangur:
Is PHPCS supposed to work for files with parse errors at all? What is the point in running PHPCS so frequently, before even the file become valid?

louisl:
PHPCS doesn't care about parse errors, it just tokenises then checks code format against rules, this particular issue raises the cpu to 100% and brings all the fans on so needs addressing imo. I don't see a problem displaying errors and warnings as you type rather than on save.

orlangur:
> PHPCS doesn't care about parse errors, it just tokenises then checks code format against rules

Surely, but how many more sniffs are not parse-error-tolerant?

> this particular issue raises the cpu to 100% and brings all the fans on so needs addressing imo

Is this some standard workflow/PhpStorm plugin? Just out of curiosity as I never used such approach.

louisl:
> Is this some standard workflow/PhpStorm plugin? Just out of curiosity as I never used such approach.

I use SublimeText 3 with SublimeLinter and SublimeLinter-phpcs plugin among others.

> Surely, but how many more sniffs are not parse-error-tolerant?

Having used the above setup for the last 4 or 5 years this is the first time I've come across this, probably because not many standards have sniffs regarding arrays because they're difficult to write. I have a modified version of that sniff in a standard I'm writing that brought it to my attention.

gsherwood:
I've run PHPCS over the sample code you provided, and I can see that `$nextContent` becomes `false`, but that does not cause an infinite loop for me. The value of `$nextContent` is just placed into an array for later use and doesn't affect the `for` loop that the call is within.

Are you sure it is this sniff causing the error? Are you able to replicate with the built-in Squiz standard using your code snippet?

louisl:
[...]

PR #1667, opened 2017-09-19
Fix for issue #1666 infinite loop

louisl:
Exit the loop if there's no content.


gsherwood:
Awaiting feedback to help reproduce from #1666

PR #1670, opened 2017-09-20
PEAR.Files.IncludingFile: remove excess whitespace along with parentheses

JDGrimes:
See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1153:

> The fixer for `PEAR.Files.IncludingFile.BracketsNotRequired` turns this:
> 
> ```php
> require_once( '/path/to/file.php' );
> ```
> 
> Into this:
> 
> ```php
> require_once  '/path/to/file.php' ;
> ```
> 
> - The space before the closing parenthesis is left.
> - A new space is added after `require_once`, in addition to the space that already followed the opening parenthesis.
> 
> In other words, the fixer assumes that no extra whitespace will be here. 

This PR changes the fixer to take existing whitespace into account, so that it will play nicer with the whitespace conventions of the WordPress coding standards.

So the above example would now be fixed correctly:

```php
require_once '/path/to/file.php';
```


src/Standards/PEAR/Sniffs/Files/IncludingFileSniff.php
```diff
@@ -57,12 +57,16 @@ public function process(File $phpcsFile, $stackPtr)
...
                 $phpcsFile->fixer->beginChangeset();
                 $phpcsFile->fixer->replaceToken($tokens[$nextToken]['parenthesis_closer'], '');
-                if ($tokens[($nextToken - 1)]['code'] !== T_WHITESPACE) {
+                if ($tokens[($nextToken - 1)]['code'] !== T_WHITESPACE && $tokens[($nextToken + 1)]['code'] !== T_WHITESPACE) {
                     $phpcsFile->fixer->replaceToken($nextToken, ' ');
                 } else {
                     $phpcsFile->fixer->replaceToken($nextToken, '');
                 }
 
+                if ($tokens[($tokens[$nextToken]['parenthesis_closer'] - 1)]['code'] === T_WHITESPACE) {
```
jrfnl: Should whitespace after the parenthesis closer be removed as well ?

src/Standards/PEAR/Tests/Files/IncludingFileUnitTest.inc
```diff
@@ -97,3 +97,6 @@ if ($var == false) {
 
 require_once ('blank.inc');
 include_once ('blank.inc');
+
+require_once( 'blank.inc' );
+include_once( 'blank.inc' );
```
jrfnl: Maybe also add unit test case for:
```php
require_once ( 'blank.inc' );
include_once ( 'blank.inc' );
```
i.e. whitespace both before and after the parenthesis.

JDGrimes: This got me thinking more about this, and I'm wondering if modifying this sniff is really the way to go. It seems to be adding a lot of whitespace-related complexity to the fixer, when the sniff really has nothing to do with whitespace. There was already a precedent within the sniff, but checking on both sides of both parentheses is 6 different checks. It seems like the fixing should really be handled by sniffs specifically for whitespace, like `Squiz.WhiteSpace.LanguageConstructSpacing` and `Squiz.WhiteSpace.SemicolonSpacing`. The one whitespace check that was already part of this sniff makes sense, but I think maybe we should just add those two sniffs to the WordPres ruleset rather than adding more whitespace handling to this sniff.

If you agree @jrfnl, I'll close this PR, and we'll open one on WPCS instead to add those sniffs to the ruleset.

jrfnl: @JDGrimes Good call. I concur with your analysis and would be happy for those two sniffs to be added to WPCS.

PR #1673, opened 2017-09-22
Squiz/ArrayBrackets: Add `fixed` file

jrfnl:
The fixed file highlights that the fixer does not only remove spaces, but also removes comments.

Whether this is intentional or not, is unclear to me.

If it is unintentional, I'd be happy to add a second commit to fix this.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1109,6 +1109,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
       <dir name="Tests">
        <dir name="Arrays">
         <file baseinstalldir="PHP/CodeSniffer" name="ArrayBracketSpacingUnitTest.inc" role="test" />
+        <file baseinstalldir="PHP/CodeSniffer" name="ArrayBracketSpacingUnitTest.inc.fixed" role="test" />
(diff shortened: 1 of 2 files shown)


gsherwood:
The problem with comments is that there really isn't a good place for them to go. If you have any ideas, I'm happy to talk about it. But I'll merge this in as-is to ensure the current behaviour is documented. Thanks.

jrfnl:
> The problem with comments is that there really isn't a good place for them to go. If you have any ideas, I'm happy to talk about it. 

@gsherwood  Removing comments, to me, makes this a risky fixer to run as those comments are normally there for a reason.

I'd suggest leaving comments alone and just making sure there is no whitespace between the bracket and the comment.

If that sounds good to you, I'll open a separate PR to fix the sniff.

gsherwood:
> I'd suggest leaving comments alone and just making sure there is no whitespace between the bracket and the comment.
> 
> If that sounds good to you, I'll open a separate PR to fix the sniff.

Sounds like a good plan.

jrfnl:
@gsherwood I've opened PR #1694 which should sort this out.

PR #1674, opened 2017-09-22
`getTokensAsString()` method: Add option to respect original content

jrfnl:
When a file uses tabs and a `tab-width` is set, the `content` will have spaces instead of tabs.
This could lead to situations where the output of `getTokensAsString()` when used to move content about would unintentionally replace tabs with spaces.

This small change allows for retrieving the original content by passing a third parameter.

I've chosen to add a parameter rather than change the default function behaviour to prevent breaks for existing sniffs relying on this behaviour, like the Squiz/ArrayDeclaration sniff: https://github.com/squizlabs/PHP_CodeSniffer/blob/2ca6cf56420616cb47a88249eb5e3520f0e31ac7/src/Standards/Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php#L477-L480

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -1679,12 +1679,14 @@ class File
      * Returns the content of the tokens from the specified start position in
      * the token stack for the specified length.
      *
-     * @param int $start  The position to start from in the token stack.
-     * @param int $length The length of tokens to traverse from the start pos.
+     * @param int $start       The position to start from in the token stack.
+     * @param int $length      The length of tokens to traverse from the start pos.
+     * @param int $origContent Whether the original content or the tab replaced
+     *                         content should be used.
      *
      * @return string The token contents.
      */
-    public function getTokensAsString($start, $length)
+    public function getTokensAsString($start, $length, $origContent=false)
     {
         $str = '';
         $end = ($start + $length);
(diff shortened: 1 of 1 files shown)


jrfnl:
@gsherwood Could you please have a look at this ? This PR is a prerequisite for a PSR-12 PR I have waiting in the wings.

gsherwood:
Thanks for this change.

Just an FYI that I've slowed down on PSR-12 now that it's been moved back into draft after all my (mostly unanswered) questions on the mailing list. So don't be offended if I get to some other stuff before PSR-12 PRs.

jrfnl:
@gsherwood Thanks for the merge.

> Just an FYI that I've slowed down on PSR-12 now that it's been moved back into draft after all my (mostly unanswered) questions on the mailing list.

I hadn't realized it had gone back into draft. Good to know.

> So don't be offended if I get to some other stuff before PSR-12 PRs.

No worries. More than anything, having the PR open for a PSR-12 sniff is intended to signal that you won't need to write that sniff yourself (unless the sniff is not good enough of course, but I'm happy to receive feedback & improve) ;-)

I guess I should have add a "PSR-12" prefix to the LowerCaseTypeDeclaration PR #1685, but the ruleset wasn't being worked on yet when I originally pulled it :pensive:

PR #1682, opened 2017-10-01
PSR2.ControlStructures.ControlStructureSpacing should check declare statements

mhujer:


src/Standards/PSR2/Tests/ControlStructures/ControlStructureSpacingUnitTest.inc
```diff
@@ -45,3 +45,9 @@ foreach (  $something as $blah => $that  ) {}
 // @codingStandardsChangeSetting PSR2.ControlStructures.ControlStructureSpacing requiredSpacesBeforeClose 0
 
 $binary = b"binary string";
+
+declare(ticks=1) {} // OK
```
mhujer: I used `declare(ticks=1)` instead of `declare(strict_types=1)`, because I wasn't sure if it is OK for the test file to be _invalid_.

jrfnl: It's ok for a test file to be invalid. It's tokenized after all, not parsed.

jrfnl:
Just my two pennies for what it's worth: I do not think this PR should be accepted as the sniff deals with _ControlStructures_ and `declare()` is not a control structure.
What about adding a completely new sniff to `Generic` dedicated to checking whitespace for `declare()` statements ? I'd love to see it be flexible/configurable as well, so people with low space as well as space rich standards could use it.

mhujer:
I'm not sure, but in the [documentation](http://php.net/declare) the `declare` is listed in the _Control Structures_ section.

jrfnl:
> I'm not sure, but in the documentation the declare is listed in the Control Structures section.

Interesting. Still, `return` and `require_once` are also listed in that section, but are IMHO - like `declare` - very different type of structures too.

mhujer:
Rebased on current master.

mhujer:
Rebased on current master to resolve merge conflicts.

PR #1683, opened 2017-10-01
PSR2.ControlStructures.ControlStructureSpacing - add newlineAllowedAfterOpen option

mhujer:
Generally, I prefer to use zero spaces after opening bracket, but in some cases, it makes sense to have a newline there (because it improves readability).

This PR adds an option to allow newline after opening bracket (`false` by default to maintain backwards-compatibility).

It can improve `if` readability:
```php
if (    // <--- newline after opening bracket
	$priceWithoutVatFrom === null ||
	$priceWithoutVatTo === null ||
	$priceWithVatFrom === null ||
	$priceWithVatTo === null
) {
	return new PriceFilterConfig(0, 0, 0, 0);
}
```

or in multi-catch:

```php
} catch (     // <--- newline after opening bracket
	\MyNamespace\FullyQualified\Long\Make\Exception\MakeNotFoundBySlugException |
	\MyNamespace\FullyQualified\Long\Model\Exception\ModelNotFoundBySlugAndMakeException |
	\MyNamespace\FullyQualified\Long\ModelLine\Exception\ModelLineNotFoundBySlugAndModelException
	$e
) {
	// ...
}
```

The tests are missing, because I couldn't find a way to test a sniff with an option (see how it can be done [in Slevomat Coding Standard](https://github.com/slevomat/coding-standard/blob/b5d47e0c74fd54ec03a54fe2846fe8b14ebb2df1/tests/Sniffs/TypeHints/DeclareStrictTypesSniffTest.php#L142)). Can you point me to an example how it is tested in PHPCS?


jrfnl:
> Can you point me to an example how it is tested in PHPCS?

See https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Generic/Tests/Functions/OpeningFunctionBraceBsdAllmanUnitTest.inc#L162-L180

mhujer:
@jrfnl Thanks! I can't believe I didn't see it.

src/Standards/PSR2/Tests/ControlStructures/ControlStructureSpacingUnitTest.inc
```diff
@@ -45,3 +45,49 @@ foreach (  $something as $blah => $that  ) {}
...
+try {
+    throw new \Exception();
+} catch (
+    \MyNamespace\FullyQualified\Long\Make\Exception\MakeNotFoundBySlugException |
+    \MyNamespace\FullyQualified\Long\Model\Exception\ModelNotFoundBySlugAndMakeException |
+    \MyNamespace\FullyQualified\Long\ModelLine\Exception\ModelLineNotFoundBySlugAndModelException
+    $e
+) {
+    // ...
+}
```
jrfnl: Please reset the property to the default after the tests targeting it, to make sure it will not have an impact on additional unit tests being added later.

jrfnl:
@mhujer Just a question: the way you have implemented this now is to _allow_ a new line after the open parenthesis.
What about _enforcing_ a new line after the open parenthesis if the condition is multi-line ? That would mean changes to the fixer too (and more unit tests), but I'm just wondering if that should be covered.

src/Standards/PSR2/Tests/ControlStructures/ControlStructureSpacingUnitTest.inc
mhujer: fixed

mhujer:
The motivation behind this PR was that we have a few (exactly three) occurrences of a multi-line condition with a newline. Not having an option to allow them prevented us from using this sniff.

I found one example of ~multiline condition in our codebase which should not start with a newline:

```php
if (!$this->elasticClient->indices()->exists([
	'index' => $indexName,
])) {
	throw new \Exception(sprintf('Index %s does not exist', $indexName));
}
```
(I know that it can refactored to a separate variable or handled in sniff by detecting the `&&` or `||` etc. )

But it makes sense to me to leave the decision if the newline is necessary to the developer's discretion. (And to be honest, I don't want to develop a feature I'm sure I won't be using)

mhujer:
Rebased on current master.

mhujer:
Rebased on current master to resolve merge conflicts.

mhujer:
Rebased on current `master`.

PR #1685, opened 2017-10-03
New Generic/LowerCaseTypeDeclaration sniff

### LINKED ISSUE #1684: Lowercase scalar type hints
When using PHP 7 it is possible to hint scalar type parameters but it is not required that they are lowercase. The following examples are valid PHP but I would like to enforce that the `string` type hint is always lowercase:
```php
function (string $name) {
  echo $name
}
```
```php
function (String $name) {
  echo $name
}
```
I usually just stick with PSR standards and I'm having a hard time trying to find all the available rules.
Is there an existing rule that comes with phpcs that can sniff this? This seems like a common thing to check so I'm not sure why it's hard to find a sniff that fits.

jrfnl:
New sniff to verify that all non-classname based type declarations are lowercase, both for parameter type declarations as well as return type declarations.

Includes auto-fixer.

Fixes #1684

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -269,6 +269,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="ForbiddenFunctionsStandard.xml" role="php" />
         <file baseinstalldir="PHP/CodeSniffer" name="LowerCaseConstantStandard.xml" role="php" />
         <file baseinstalldir="PHP/CodeSniffer" name="LowerCaseKeywordStandard.xml" role="php" />
+        <file baseinstalldir="PHP/CodeSniffer" name="LowerCaseTypeDeclarationStandard.xml" role="php" />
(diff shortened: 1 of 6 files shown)


jrfnl:
N.B.: If that would be preferred, I could easily split the sniff in two independent sniffs as the errors are completely independent of each other anyway.

bramcordie:
I didn't mention return types in #1684 but I definitely want to check that as well. The most practical solution for me is checking both with one sniff.

jrfnl:

jrfnl:
@gsherwood A bit of shame that double work has been done, but it looks like the `Generic.PHP.LowerCaseType` sniff added by you in https://github.com/squizlabs/PHP_CodeSniffer/commit/5a989395a4318475590c6515512f17e2b8e4df8f covers this as well.

I'll run some tests to see if anything from this sniff should still be added to the `Generic.PHP.LowerCaseType` sniff, but other than that, I think this PR can probably be closed, as well as issue #1684.

Issue #1671, opened 2017-09-20
Request : Issue with the 'TestCase' class alias in bootstrap file


El-Sam:
Hi, first of all, thanks for this library, it's of a great help.

I'm using PHPUnit 6 for my unit tests. 

In the `bootstrap.php` file in `PHP_CodeSniffer/tests` , a class alias is created, this alias causes `Multiple definitions exist for class TestCase` issue in my IDE (PHPStorm).

This breaks a lot of features that speed up my testing like code completion. So, I think you should remove this alias or handle it in another way that you find convenient.

gsherwood:
I might need a bit more information. Is this just an issue with your IDE thinking that multiple class definitions exist when they actually don't at runtime? If so, isn't this an issue with your IDE and not an issue with PHPCS?

The reason the class alias is there is so that the unit tests will run with both old and new versions of PHPUnit, so it can't just be removed. Do you have a suggestion for another way to achieve PHPUnit backwards compatibility?

orlangur:
Caused by a hacky https://github.com/squizlabs/PHP_CodeSniffer/pull/1384/files.

```
+if (class_exists('PHPUnit_Framework_TestCase') === true && class_exists('PHPUnit\Framework\TestCase') === false) {
 +    class_alias('PHPUnit_Framework_TestCase', 'PHPUnit\Framework\TestCase');
 +}
```

So, PhpStorm ignores this condition?

https://stackoverflow.com/a/23079082/8640867 doesn't help? Even if `PHPUnit_Framework_TestCase` does not exist which is the case for PHPUnit 6 PhpStorm is complaining?

orlangur:
> The reason the class alias is there is so that the unit tests will run with both old and new versions of PHPUnit, so it can't just be removed. Do you have a suggestion for another way to achieve PHPUnit backwards compatibility?

@gsherwood sorry if I missed a previous explanation somewhere, but why `PHP_CodeSniffer` needs to explicitly support [unsupported](https://github.com/sebastianbergmann/phpunit/wiki/Release-Process) PHPUnit versions?

I remember you explained reasoning behind supporting unsupported PHP versions but PHPUnit is just a tool to test `PHP_CodeSniffer` itself, it was tested just fine without such a hack https://travis-ci.org/squizlabs/PHP_CodeSniffer/jobs/254679990. Why not just let Composer to install whatever PHPUnit version is suitable for the particular platform?

gsherwood:
[...]

PR #1686, opened 2017-10-05
Possible fix for #1671

svycka:
fix for https://github.com/squizlabs/PHP_CodeSniffer/issues/1671 should not break alias but also should fix IDE problems at least phpStorm works. lets see if tests pass

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -30,17 +30,17 @@ $tokens = new \PHP_CodeSniffer\Util\Tokens();
 
 // Compatibility for PHPUnit < 6 and PHPUnit 6+.
 if (class_exists('PHPUnit_Framework_TestSuite') === true && class_exists('PHPUnit\Framework\TestSuite') === false) {
-    class_alias('PHPUnit_Framework_TestSuite', 'PHPUnit\Framework\TestSuite');
+    class_alias('PHPUnit_Framework_TestSuite', 'PHPUnit'.'\Framework\TestSuite');
 }
 
 if (class_exists('PHPUnit_Framework_TestCase') === true && class_exists('PHPUnit\Framework\TestCase') === false) {
-    class_alias('PHPUnit_Framework_TestCase', 'PHPUnit\Framework\TestCase');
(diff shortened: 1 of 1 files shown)


svycka:
okay, PHP_CodeSniffer is not happy about this workaround but maybe we could ignore this or implement in PHP_CodeSniffer friendly way? Anyway is this good way to fix this?

svycka:
Also if we don't want to fix this then maybe we can inverse this and create aliases, not for new phpunit versions but for old then only people with legacy versions will have problems lets hope there are fewer people

jrfnl:
> Also if we don't want to fix this then maybe we can inverse this

It's been set up so that it will be easy to drop support for older PHPUnit versions when the time comes. Inversing the aliases breaks that.

svycka:
@jrfnl don't know how this would be more difficult to drop support older versions but maybe don't know everything. Then what about this workaround of a workaround? :)

jrfnl:
> Then what about this workaround of a workaround?

For what it's worth: IMHO this is a shortcoming/issue with the IDE and not something which should be solved in PHPCS.
If something should be done in PHPCS at all about this, I'd be in favour of adding the tests back into the `.gitattributes` file so they don't ship by default.

svycka:
I understand that this is not a phpcs problem but this issue is really annoying also easy to fix and don't expect this to be fixed in all IDE's soon. But I am +1 to adding tests to .gitattributes.

gsherwood:
I've merged this in. Is anyone able to test and see if it is working.

Before anyone decides to comment, I'm aware this is a hacky fix. I really don't have the time to put effort into something like this. As long as it works, I'm happy for now.

svycka:
dev-master no longer shows duplicate implementations so works for me. But That's my PR so guess this doesn't count...

gsherwood:
> That's my PR so guess this doesn't count...

If you're the only one who confirms it, it counts a lot! Thanks for testing.

gsherwood:
Confirmed a second time on main report.

PR #1689, opened 2017-10-07
Custom property arrays: be lenient about space usage

### LINKED ISSUE #1688: customAutoEscapedFunctions doesn't allow spaces between items
In your ruleset file, `customAutoEscapedFunctions` allows comma-separated lists of values. However, if you put `, ` instead of just `,`, it ignores the second value.

Fails, allowing only `foo`:
```
<property name="customAutoEscapedFunctions" value="foo, bar" type="array" />
```

Works as expected:
```
<property name="customAutoEscapedFunctions" value="foo,bar" type="array" />
```

jrfnl:
While it is debatable whether it is a bug or a feature that custom property arrays are unforgiving about spaces surrounding the individual keys and values, all custom array properties in PHPCS itself do not expect spaces around them.

This minor change removes spaces around array keys and values before the property is passed to the sniff.

This means that each of the below examples will now result in the same property value being set:
```xml
        <properties>
            <property name="forbiddenFunctions" value="print=>echo,create_function=>null" type="array" />
            <property name="forbiddenFunctions" value="print=>echo, create_function=>null" type="array" />
            <property name="forbiddenFunctions" value="print => echo, create_function => null" type="array" />
        </properties>
</rule>
```

Fixes #1688

### DIFF
--- a/src/Ruleset.php
+++ b/src/Ruleset.php
@@ -950,9 +950,9 @@ class Ruleset
 
                             list($k,$v) = explode('=>', $val.'=>');
                             if ($v !== '') {
-                                $values[$k] = $v;
+                                $values[trim($k)] = trim($v);
                             } else {
-                                $values[] = $k;
+                                $values[] = trim($k);
                             }
                         }
 


gsherwood:
Thanks for this.

PR #1691, opened 2017-10-08
Squiz/SemicolonSpacing: improve handling of semicolons in `for` control structures

### LINKED ISSUE #1690: Squiz/SemicolonSpacing: semi-colons in for() control structures
The `Squiz.WhiteSpace.SemicolonSpacing` sniff checks that there is no whitespace before a semi-colon.

This might however conflict with rules regarding spacing within a `for()` control structure:
```php
for ($i = 1; ; $i++) {}
for (; ; ) {}
```

See http://php.net/manual/en/control-structures.for.php examples 2 and 3.

I would like to fix this, but would like some guidance on what would be the preferred fix:
* Selectively exclude semi-colons found within a `for()` condition when they are preceded by the open parenthesis or by another `;`.
* Exclude semi-colons found within a `for()` condition completely from the sniff.
* Include them, but have a separate error code for semi-colons found within a `for()` condition so they can be modularly excluded.

@gsherwood What would you prefer ? Or maybe there is yet a fourth option I haven't thought of which would be an even better way to fix this ?

Related issue: https://github.com/squizlabs/PHP_CodeSniffer/pull/1670

jrfnl:
This is a solution for issue #1690 in which I've basically implemented the first of the three options mentioned in the issue.

> Selectively exclude semi-colons found within a for() condition when they are preceded by the open parenthesis or by another ;.

Includes unit tests.

Fixes #1690

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/SemicolonSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/SemicolonSpacingSniff.php
@@ -58,7 +58,25 @@ class SemicolonSpacingSniff implements Sniff
         }
 
         $nonSpace = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($stackPtr - 2), null, true);
-        if ($tokens[$nonSpace]['code'] === T_SEMICOLON) {
+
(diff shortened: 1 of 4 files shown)


gsherwood:
Thanks for doing this.

Issue #1580, opened 2017-07-28
Allow single line closures


CarsonF:
This code:
```php
$func = function ($item) { return $item['foo']; };
```
Triggers:
- `Generic.Formatting.DisallowMultipleStatements.SameLine`
- `PEAR.WhiteSpace.ScopeClosingBrace.Line`
- `Squiz.WhiteSpace.ScopeClosingBrace.ContentBefore`
- `Squiz.Functions.MultiLineFunctionDeclaration.ContentAfterBrace`

---

IMO the `DisallowMultipleStatements` error is a false positive. The code above is not the same as
```php
$a = ''; $b = '';
```
----

For the rest of the errors, it would be nice to the have an option to allow single statement methods to be on one line (PHPStorm has this option). Maybe another option to further restrict it to only closures.

tgr:
Referenced in https://phabricator.wikimedia.org/T154789

gsherwood:
> would you entertain a PR for this?

I can't see how this could be done with the existing sniffs. This is a specific rule for closures that does not exist in any of the included standards and is probably best served using a custom sniff. Including a rule like "Allow single line closures if they only have one statement" in these sniffs is far too specific.

gsherwood:
Closing due to lack of additional interest in a new sniff.

PR #1692, opened 2017-10-08
Allow empty {} on single line in Squiz.WhiteSpace.ScopeClosingBrace.ContentBefore

tgr:
Allows non-awkward single line syntax for empty functions/classes.

Fixes one of the cases mentioned in #1580.


gsherwood:
Sorry, I clearly missed this PR when it first came out.

But closing now as this is not the intention of the Squiz standard and so should not be changed there. A config option in the sniff, or maybe a different error code that can be ignored, would be preferred.

PR #1694, opened 2017-10-09
Squiz.Arrays.ArrayBracketSpacing is removing some comments during fixing

jrfnl:
As discussed in https://github.com/squizlabs/PHP_CodeSniffer/pull/1673#issuecomment-335048664, this PR addresses the `Squiz.Arrays.ArrayBracketSpacing` sniff unintentionally removing comments which were placed adjacent to the brackets.
This should not happen anymore once this PR has been merged.

Includes adjusted and additional unit tests.

While looking at the sniff, I noticed two other improvements begging to be made :heart_eyes_cat: 
* The sniff still had some checks in it to prevent it acting on short array syntax. As the short array open/close brackets have their own tokens now, these checks are no longer needed.
* The sniff had a sanity check for parse errors/live coding for the close bracket, but not for the open bracket.

Both of these additional issues have been fixed in separate commits.


gsherwood:
Thanks a lot for doing that.

PR #1701, opened 2017-10-12
New Generic.WhiteSpace.ArbitraryParenthesesSpacing sniff

jrfnl:
While working on the big WP Core auto-fixing project, I realized there was no sniff available to check the whitespace around arbitrary parentheses. So, here goes.

This new sniff checks the spacing on the **_inside_** of arbitrary parentheses.
A conscious choice has been made not to check on the outside as this would very quickly conflict with other sniffs.

Arbitrary parentheses, for the purposes of this sniff, are all parentheses which do not belong to a control structure, function declaration, function call or language construct.
This is to prevent conflicts and duplication with rules for those type of constructs.

**Notes:**
* The sniff will throw (fixable) errors when an incorrect amount of whitespace (too much/too little) has been found on the inside or the parenthesis.
* The sniff will throw a warning when an empty set of arbitrary parentheses is encountered.
* The desired amount of whitespace on the inside of the parentheses can be configured using the `spacingInside` property. Default: `0`.
* Whether or not to allow new lines can be configured using the `ignoreNewlines` property. Default: `false`.

#### Todo once the sniff has been merged:

- [ ] Add info on the custom properties to the wiki

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -296,6 +296,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="SubversionPropertiesStandard.xml" role="php" />
        </dir>
        <dir name="WhiteSpace">
+        <file baseinstalldir="PHP/CodeSniffer" name="ArbitraryParenthesesSpacingStandard.xml" role="php" />
         <file baseinstalldir="PHP/CodeSniffer" name="DisallowSpaceIndentStandard.xml" role="php" />
         <file baseinstalldir="PHP/CodeSniffer" name="DisallowTabIndentStandard.xml" role="php" />
         <file baseinstalldir="PHP/CodeSniffer" name="ScopeIndentStandard.xml" role="php" />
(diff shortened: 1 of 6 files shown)


gmponos:
> Includes extensive unit tests.

That's true it has a lot of test but I was wondering, how is it possible for a developer to test if the options inside a sniff work? (like [here](https://github.com/squizlabs/PHP_CodeSniffer/pull/1701/files#diff-92d9b0dbdc4f42b82a56c0248f8c0f24R34))

jrfnl:
> That's true it has a lot of test but I was wondering, how is it possible for a developer to test if the options inside a sniff work? 

I'm not sure I understand your question. Have you looked at the included unit test files ? You can see how it's done there. I [change the public properties](https://github.com/squizlabs/PHP_CodeSniffer/pull/1701/files#diff-343fd6a4fce283c7371574affe8db877R135) a couple of times inline (and [reset after](https://github.com/squizlabs/PHP_CodeSniffer/pull/1701/files#diff-343fd6a4fce283c7371574affe8db877R147)). The combination of other lines on which to expect errors based on the value property + a different expected fix in the `.fixed` file, make sure that the properties are handled correctly.

gmponos:
Ahhhh... Haven't noticed this and I was not aware that an option like this exists

    // @codingStandardsChangeSetting Generic.WhiteSpace.ArbitraryParenthesesSpacing spacingInside 1

Nice to know.. thank you and ignore my previous comment.

jrfnl:
Rebased & updated code style...

gsherwood:
Thanks a lot for this. I ended up changing the `spacingInside` property to `spacing`, only because all other documented sniff properties using `spacing` unless they have multiple spacing settings (at which time they use `spacing` and then something else).

jrfnl:
@gsherwood You're welcome :smile: 

Side-note: I'd just named it slightly differently as some of the other `spacing` properties are about spacing _inside_ and _outside_ and I wanted to prevent confusion, but I'm perfectly happy with the change in property naming.

PR #1706, opened 2017-10-14
Squiz/NonExecutableCode: fix undefined index during live coding reviews

jrfnl:
First fix in a series to fix the issues found.

Error encountered: `Undefined index: scope_closer in phpcs/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php on line 238`

Error occurs when an unfinished function or OO structure is encountered.

While looking at the code, I saw three more minor improvements which I have included in this PR:
* Account for modern PHP OO structures.
    The sniff is designed to skip over declarations, but only accounted for functions, classes and interfaces.
    This has been updated to include closures, traits and anonymous classes.
* Add the semi-colon to the ignore list for reporting.
    In this context a 'lone' semi-colon is most often encountered at the end of a closure or anonymous class, so this might as well be skipped over.
* The sniff can thrown two or more errors on the same line when several "exit" commands have been found above the line.
    I've adjusted the error message to also report on the line where the "exit" command was encountered to make it easier to fix/debug these type of errors.

Includes unit tests.

To reproduce/test: take the updated unit test file and just run the original sniff over it.


jrfnl:

PR #1708, opened 2017-10-14
Generic/DisallowLongArray: fix undefined index when running fixer

jrfnl:
Third fix in a series to fix the issues found.

Error encountered: `Undefined index:  in phpcs/src/Fixer.php on line 524`

Error occurred when due to, for instance, a git merge conflict artifact, the array closer has not been determined.

Includes unit tests.

One of the already existing unit tests results has been adjusted. a found `array` keyword without parentheses was previously being fixed to short array syntax.
However, as this is "unfinished" code, it could just as easily have been the start of an array function call like `array_merge()`, making this a risky fix.

For this case as well as the new case, the error will still be reported, but the fixer will no longer run for these.

To reproduce/test: take the updated unit test file and just run the original sniff over it.


jrfnl:

PR #1710, opened 2017-10-15
Squiz/InlineComment: fix fixer conflict when comment found at end of function

### LINKED ISSUE #1709: Squiz: fixer conflict for inline comment at end of function
Given this snippet:
```php
$a = function ()
{
    // body
};
```

There is a fixer conflict between the `Squiz.Commenting.InlineComment` sniff and the `Squiz.WhiteSpace.FunctionClosingBraceSpace` sniff.

* The first demands no blank line beneath an inline comment.
* The second demands a blank line at the end of a function.

Working up a PR to fix this now.

jrfnl:
Fourth fix in a series to fix the issues found.

When an inline comment is found at the end of a function, removing a blank line after it conflicts with the `Squiz.WhiteSpace.FunctionClosingBraceSpace` sniff which demands a blank line at the end of a function.

To reproduce the issue, run (against `master`):
`phpcbf -p -s --standard=Squiz ./src/Standards/Squiz/Tests/Commenting/InlineCommentUnitTest.inc -vv`

I have chosen to fix this by adding a specific error code for that situation and excluding that error code from the `Squiz` ruleset.

This way, the backward-compatibility break for other standards using the `Squiz.Commenting.InlineComment` sniff will be smallest.
Only if a standard explicitly in/excluded the `Squiz.Commenting.InlineComment.SpacingAfter` errorcode will this have any effect on them.

The unit tests already contain a test covering this. As the errorcode is excluded via the ruleset, the results of the unit tests do not change.

Fixes #1709

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/InlineCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/InlineCommentSniff.php
@@ -292,6 +292,19 @@ class InlineCommentSniff implements Sniff
                 return;
             }
 
+            $errorCode = 'SpacingAfter';
+
+            if (isset($tokens[$stackPtr]['conditions']) === true) {
+                $type         = end($tokens[$stackPtr]['conditions']);
+                $conditionPtr = key($tokens[$stackPtr]['conditions']);
+
+                if (in_array($type, [T_FUNCTION, T_CLOSURE], true) === true
+                    && $tokens[$conditionPtr]['scope_closer'] === $next
+                ) {
+                    $errorCode = 'SpacingAfterAtFunctionEnd';
+                }
+            }
+
             for ($i = ($stackPtr + 1); $i < $phpcsFile->numTokens; $i++) {
                 if ($tokens[$i]['line'] === ($tokens[$stackPtr]['line'] + 1)) {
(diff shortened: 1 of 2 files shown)


jrfnl:

jrfnl:
Rebased (again) for merge conflicts...

gsherwood:
Your solution seems like a good one to me as well. Thanks for this.

jrfnl:
You're welcome :smiley: 

FYI: I'm just now doing a new fixer conflict run to see where we're at.

Initial findings:
* The fixer conflict for 1 particular test file seems to have been fixed along the way.
* 6 files with new fixer conflicts.... these may well all be related to one sniff, but I haven't dug that deep yet. I wouldn't be surprised if it has to do with the new PHPCS annotations, but need to have a detailed look to be sure.

Current status based on running `phpcbf` for each standard against all test case files:
* **PSR1**: -- does not contain sniffs with fixers.
* **PSR2**: **3** test case files result in fixer conflicts
* **PEAR**: **9** test case files result in fixer conflicts
* **Squiz**: **41** test case files result in fixer conflicts
* **Zend**: **3** test case files result in fixer conflicts

(MySource can not be tested as the ruleset includes and exclude pattern for the test directory)

PR #1813 will solve five of these.

PR #1714, opened 2017-10-15
PEAR/FunctionDeclaration: fix fixer conflict in handling of unfinished closures

jrfnl:
Seventh fix in a series to fix the issues found.

Closures - both JS as well as in PHP - have their own token, but only if the structure is finished.
For live coding/parse errors, the structure will be identified as `T_FUNCTION`.

This caused a fixer conflict within the `PEAR.Functions.FunctionDeclaration` sniff and by extension in the `Squiz.Functions.MultiLineFunctionDeclaration` sniff, between the `SpaceAfterFunction` and the `SpaceBeforeOpenParen` error codes.

Unfinished closures can be distinguished from functions by checking for the `scope_opener` having been set.

Includes unit tests.


gsherwood:
Thanks

PR #1707, opened 2017-10-14
Zend/ValidVariableName: fix undefined index error when exception is encountered

jrfnl:
Came across this error when running fixer conflict checks for the various standards. (See https://github.com/squizlabs/PHP_CodeSniffer/pull/1645#issuecomment-336314794 )
Second fix in a series to fix the issues found.

Error encountered: `Undefined index: scope in phpcs/src/Standards/Zend/Sniffs/NamingConventions/ValidVariableNameSniff.php on line 142`

Error occurred when a variable was encountered in an interface.

Includes unit test

PR #1715, opened 2017-10-15
Tests: Add unit tests for the `File::getMemberProperties()` method

jrfnl:
Adds dedicated unit tests for the `File::getMemberProperties()` method.

Also minor documentation update for the method itself.

Related to #1707

-----

### Updated

I realized after I'd pulled it that I'd forgotten to add the new tests to the `tests/Core/AllTests.php` file. 

Once I added them, a bug with PHP4 `var $property` type declarations started showing.
I fixed that in a separate commit.

I also realized that the tests didn't yet cover "_group property declarations_". While these aren't very common in most PHP projects, they are supported by the language and therefore should be correctly handled by the `File::getMemberProperties()` method.

Once I added the additional unit tests, it became clear that _group property declarations_ which included value definitions were not (yet) supported by the method, so that has been fixed in an additional separate commit.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -136,6 +136,8 @@ http://pear.php.net/dtd/package-2.0.xsd">
       <file baseinstalldir="" name="FindExtendedClassNameTest.php" role="test" />
       <file baseinstalldir="" name="FindImplementedInterfaceNamesTest.inc" role="test" />
       <file baseinstalldir="" name="FindImplementedInterfaceNamesTest.php" role="test" />
+      <file baseinstalldir="" name="GetMemberPropertiesTest.inc" role="test" />
+      <file baseinstalldir="" name="GetMemberPropertiesTest.php" role="test" />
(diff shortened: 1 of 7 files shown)


jrfnl:

gsherwood:
Thanks for those tests, and the fix.

gsherwood:
Just noticed that this broke detection of member vars that are commented, like this:
```php
class TestClass
{
    /**
     * Comment goes here.
     *
     * @var array
     */
    private static $_foo = [];
}
```
Looks to be due to the change to looking for the start of the statement and then stopping once we get to docblock opening tags. I'll add some more tests and put in a fix. Looks to be an easy one.

jrfnl:
@gsherwood Sorry about that. Your fix looks totally sensible to me and at the very least there are now unit tests to safeguard this in the future :sunglasses:

PR #1716, opened 2017-10-15
Generic/MultipleStatementAlignment: fix fatal error

jrfnl:
Eighth fix in a series to fix the issues found.

In certain cases, it was possible for the `$expected` padding to become less than zero, leading to the following fatal error during fixing:
`str_repeat(): Second argument has to be greater than or equal to 0 in phpcs/src/Standards/Generic/Sniffs/Formatting/MultipleStatementAlignmentSniff.php on line 293`

This PR fixes that.

Includes unit tests.


jrfnl:

PR #1717, opened 2017-10-16
Squiz/BlockComment: fix fixer conflict and more

### LINKED ISSUE #1918: Squiz/BlockComment sniff: comment closer is misidentified when comment contains PHPCS annotation
When tags are used in a block comment which also includes a PHPCS annotation, the comment closer is misidentified.

```php
	/*
	 * Some text
	 * See: {@link https://url.com/link/}
	 * phpcs:disable Standard.Category.Sniff
	 */
```

Will result in:
```
 Comment closer must be on a new line
 (Squiz.Commenting.BlockComment.CloserSameLine)
```

Interestingly, this only happens when the PHPCS annotation is not prefixed with a `@`, so it may actually be a tokenizer issue.

```php
	/*
	 * Some text
	 * See: {@link https://url.com/link/}
	 * @phpcs:disable Standard.Category.Sniff
	 */
```

I will look into this. If it turns out to be an issue with the BlockComment sniff, I will add the fix as an additional commit to PR #1717 which addresses a number of other issues in the same sniff.

If it's the tokenizer, I will open a separate PR.

jrfnl:
Ninth fix in a series to fix the issues found.

To start with, this PR fixes a fixer conflict where the `Squiz.Commenting.BlockComment` sniff conflicts with itself.

To reproduce, run the following against the first commit (additional unit tests):
`phpcbf -p -s ./src/Standards/Squiz/Tests/Commenting/BlockCommentUnitTest.inc --standard=Squiz --sniffs=Squiz.Commenting.BlockComment`.

The fixer conflict is caused by the sniff seeing the second block comment of the additional unit test as both:
a) an individual block comment and
b) as part of the first block comment, which leads to conflicting directives on what the indent should be.

Fixed by the second commit.

While working on this I noticed a number of other things which could use some love:
1. The second block comment of the additional unit test was not recognized as an empty comment. This was caused by tabs for the indentation not being taken into account. Fixed in the third commit.
2. For block comment lines which are prefixed with stars, the sniff would require too much indentation. (4 instead of 1). Fixed in the fourth commit.
3. The sniff made a presumption that `tab-width` would always be 4. This would cause a fixer conflict with itself if `tab-width` was set to, for instance, 3. The fixer made the presumption that the _expected indent_ would always be cleanly dividable by the `tab-width` and would end up continuously making the same fix as the indent used by the fixer would not be the same as the expected indent. Fixed in the fifth commit.
4. And while I was at it, I noticed that the fixer for the `LastLineIndent` was missing while all other line indent error codes did have a fixer attached to it. Added in the sixth commit.

As a final commit I've made some minor efficiency fixes, reducing the number of (duplicate) function calls the sniff was making.

-------

### Updated

I've added one more commit to fix another fixer conflict.

To reproduce, run the following against `master`:
`phpcbf -p -s ./src/Standards/Squiz/Tests/WhiteSpace/OperatorSpacingUnitTest.inc --standard=Squiz`

Now the sniff will check whether a one-line `/* */` comment is the last content on a line and only auto-fix it if it is. If not, it will throw an error, but won't auto-fix.

------

### Updated (again)

I've added one more commit to fix a problem the sniff was having with the new PHPCS annotations.

See the commit message and the issue in which I reported the problem for more details.
I've (of course) added unit tests to demonstrate the issue and safeguard against it in the future.

For the builds to pass, I've had to rebase the PR to take advantage of fixes to the tokenizing of PHPCS annotations added between when this PR was originally pulled and now (PR #1901 and #1827).

Fixes #1918


jrfnl:

gsherwood:
Thanks a lot for all these fixes and improvements.

PR #1720, opened 2017-10-16
Squiz/ScopeKeywordSpacing: improve sniff + fix fixer conflict

jrfnl:
Tenth PR in the series to fix fixer conflicts.

`phpcbf -p -s ./src/Standards/Squiz/Tests/NamingConventions/ValidVariableNameUnitTest.inc --standard=Squiz`

To fix this, I reviewed the whole sniff.

Notes:
* The sniff did not have a `fixed` file, so added that.
* The check for whether or not the scope keyword is used for a property or in another capacity, should be done in a style-independent manner. As it was, the sniff could trigger in undesired situations. Fixed + added unit tests to safeguard against this.
* When there would be more than one whitespace token between the keyword and the next content, only one whitespace token would be addressed by the fixer. The next whitespace token would fall through to the next loop. This was the cause of the fixer conflict. Improved the fixer to deal with all whitespace tokens in one go.
* Related to #1715: the sniff would also indiscriminately kick into action on multi-line multi-property definitions. IMHO those should be handled by a separate sniff if so desired. For now, I've fixed this by bowing out when it's indisputable that a multi-line multi-property definition is detected. Includes unit tests.
* Lastly, I've added some minor safeguard against the sniff kicking into action too early during live coding.

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/ScopeKeywordSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/ScopeKeywordSpacingSniff.php
@@ -44,19 +44,23 @@ class ScopeKeywordSpacingSniff implements Sniff
     {
         $tokens = $phpcsFile->getTokens();
 
+        if (isset($tokens[($stackPtr + 1)]) === false) {
+            return;
+        }
+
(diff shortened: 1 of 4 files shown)


jrfnl:

jrfnl:
Rebased & adjusted for merge conflicts after b55fd82f4f6335a519bd787703810551a2dfb9ad ...

gsherwood:
Thanks

PR #1721, opened 2017-10-16
 Squiz/MemberVarSpacing: improved handling of multi-line property declarations

jrfnl:
Related to #1715 and #1720, the other sniff involved in the fixer conflict mentioned in #1720.

**TL;DR>** Adds unit test cases with multi-line multi-property declarations and unit tests for class importing traits.

For the multi-line multi-property declarations I've adjusted the sniff to handle this more graciously as it would previously:
* correct the number of lines above the line for each variable
* not correct the number of lines above the line containing the scope keyword

IMHO this is incorrect and only the lines above the line containing the scope keyword should be checked & fixed.

By returning the `$stackPtr` of the semi-colon at the end of the property declaration, we skip past the additional variables.

As the abstract classes this class extends from did not allow for passing a "skip-to pointer" - in contrast to `Sniff` -, those have been adjusted as well to facilitate this.

For classes importing traits, the first property will currently not be recognized as such and will not receive special treatment.
This behavior has now been documented with unit tests. This behavior has not been changed.

### DIFF
--- a/src/Sniffs/AbstractScopeSniff.php
+++ b/src/Sniffs/AbstractScopeSniff.php
@@ -121,7 +121,10 @@ abstract class AbstractScopeSniff implements Sniff
      * @param int                         $stackPtr  The position in the stack where this
      *                                               token was found.
      *
-     * @return void
+     * @return void|int Optionally returns a stack pointer. The sniff will not be
+     *                  called again on the current file until the returned stack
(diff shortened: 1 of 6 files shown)


jrfnl:

jrfnl:
Rebased for merge conflict with merged PR #1841

gsherwood:
Thanks for this.

PR #1724, opened 2017-10-17
New Generic.VersionControl.GitMergeConflict sniff

jrfnl:
New sniff to detect merge conflict artifacts left in files.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -352,6 +352,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="UnnecessaryStringConcatSniff.php" role="php" />
        </dir>
        <dir name="VersionControl">
+        <file baseinstalldir="PHP/CodeSniffer" name="GitMergeConflictSniff.php" role="php" />
         <file baseinstalldir="PHP/CodeSniffer" name="SubversionPropertiesSniff.php" role="php" />
        </dir>
        <dir name="WhiteSpace">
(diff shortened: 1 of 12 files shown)


gmponos:
@jrfnl personally I believe that this is a great job you did here.

I also wanted to create a sniff like this but I didn't for the following reasons.


 - I was afraid that it might conflict with sniffs about `heredoc`.
 - I didn't know how to do it. First of all I was not sure how to detect the merge conflicts. If the merge conflict starts from a comment I believe that using codesniffer it is not detectable.

So if you ever happen to have a case like this:

``` php
/**
<<<<<<< HEAD
 * @var string $bar
============
 * @var string $bar
>>>>>>> master
 */
```

Furthermore what happens to cases that the above extends to further than comments like this:

``` php
/**
<<<<<<< HEAD
 * @var string $bar
 */
public function foo($bar){ }

/**
============
 * @var string $bar
>>>>>>> master
*/
```

- After thinking about the above case I thought that I must scan the whole file in order to find conflicts. So that leads me to the third reason why I dropped the idea of creating this sniff. Because the way I see it since I can't detect cases like the above and I have to scan the whole file I prefer not to use codesniffer to detect conflicts and use `php lint`.

if you have a case in your code like this then php lint will fail.

``` php
public function foo($bar) {
<<<<<<< HEAD
$x = 1 + 2;
============
$x = 1 + 1;
>>>>>>> master
}
```

So I can easily add a hook before committing my changes and run only the staged files like `php -l myfile.php`. Afterward I can do a grep about  `<<<<<<< HEAD` or `=========`  or `>>>>>>>`

This way I *save* a lot of processing time and not needed commits since php lint is way faster.

Although when it comes to OSS you can't control what everyone will have at their git hooks. So you could also add `php lint` in your continuous integration or add `php lint` as a sniff (although I believe it already exists, but I am not sure).

PS. My comment concerns only the PHP and not JS, and CSS.

jrfnl:
> add php lint as a sniff (although I believe it already exists, but I am not sure).

It does, it's called `Generic.PHP.Syntax`.

And yes, I prefer to lint code as the first thing in travis scripts, but that's another matter.

I'll try and have a look at the two cases. All the same, as I said in https://github.com/squizlabs/PHP_CodeSniffer/pull/1726#issue-266095442, it would actually be better if PHPCS could detect merge conflicts in the tokenizer and would skip the file altogether if those are found as whatever comes after the merge conflict will be tokenized incorrectly (for PHP an CSS) and could lead to unexpected results.

TomasVotruba:
This is great! I merge lot of various open source projects and this would save me over dozens of times

jrfnl:
Rebased for updated code style...

jrfnl:
@gmponos I've reworked the sniff taking your comments into account. All situation you pointed out as well as some additional ones I thought up are now accounted for. Let me know what you think.

gmponos:
Would you be interested in adding some test cases where the developer has forgotten back only one of this (most of the times this happens in comments for me):
```
<<<<<<< HEAD
```
Or 
```
>>>>>>> master
```

Rather than this it seems like a great work.

jrfnl:
> Would you be interested in adding some test cases where the developer has forgotten back only one of this

I could, but it wouldn't test anything new. The reworked sniff looks for openers/closers at the beginning of a line and no longer tries to match a closer to an opener.

gmponos:
oh, OK it's fine by me then.

jrfnl:
@gsherwood This subject recently came up again in the context of WPCS and related projects. One of the related projects has now added their own - no where near as comprehensive - sniff for this, false positives and all.

It would be so much better (for them) if they could just use this sniff instead, so it would be great if this sniff could become generally available.
With that in mind, would it be possible to earmark this PR for PHPCS 3.4.0 ?

In the mean time, I've just had another look at this sniff and improved it yet some more:
* Rebased against current master to ensure that it still complies with the current coding standard for PHPCS.
* Added support for detecting merge conflict boundaries within Heredoc/Nowdoc content.
* Improved support for detecting merge conflicts in CSS when the conflict opener is in a comment and the delimiter and closer are outside the comment.
* Added support for detecting the `=======` delimiter and throw a separate error for that.
* Added the "type" of boundary found to the the error message.
* Improved the inline documentation in the unit test case files to explain why there are so many of them and what all those tests which look like duplicates of each other are actually testing.
* Added yet more unit tests.
* Squashed the commits into one.

TomasVotruba:
@gsherwood Better late than never ;)  thanks for merging.

@jrfnl Thank you, can't wait to try it :+1:

jrfnl:
@gsherwood No matter, at least it'll be available now.

@TomasVotruba :+1: I look forward to hearing any feedback you may have.

PR #1725, opened 2017-10-17
Generic/AssignmentInCondition: add separate errorcode for assignment found in while

jrfnl:
A `while()` condition is the only control structure in which it is sometimes legitimate to use an assignment in condition.

Having a separate error code will enable people to selectively exclude warnings for this.

### DIFF
--- a/src/Standards/Generic/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
+++ b/src/Standards/Generic/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
@@ -150,10 +150,15 @@ class AssignmentInConditionSniff implements Sniff
             }
 
             if ($hasVariable === true) {
+                $errorCode = 'Found';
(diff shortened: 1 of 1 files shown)


jrfnl:
@gsherwood This is a very small, but quite useful change. Could this please be tagged for `3.2.0` ?

jrfnl:
@gsherwood Could this tiny change please be merged ?

gsherwood:
> Could this tiny change please be merged ?

Not in a bug fix release. If anyone is muting the previous error code, they'll now start getting an error coming back for while loops.

I know this is only a single code in a single sniff and they should ignore the whole sniff, but I know people don't always do that.

jrfnl:
Understood. Thanks for milestoning it for 3.3.0. :+1:

gsherwood:
Finally merged in :) Thanks.

jrfnl:
Thank you!

PR #1726, opened 2017-10-17
Improve handling of git merge conflicts in code

jrfnl:
Eleventh PR in the series to fix fixer conflicts.

`phpcbf -p -s ./src/Standards/Squiz/Tests/PHP/HereDocUnitTest.inc --standard=Squiz`

To fix this, I have:
* Added a new `isMergeConflictBoundary()` method to the `File` class to distinguish between an operator and an operator token which is not actually an operator, but part of a git merge conflict boundary.
* Added a new static `$gitBoundaryTokens` array to the `Util\Tokens` class.
* Added some code to the `Squiz.Formatting.OperatorBracket` and the `Squiz.WhiteSpace.OperatorSpacing` sniffs to check if an operator is in fact a git merge boundary using the new `File::isMergeConflictBoundary()` method.
* Added unit tests to both sniffs to verify that it all works as it should + that it will stay that way ;-)

This fixer conflict was also the inspiration for PR #1724.

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -2266,4 +2266,80 @@ class File
     }//end findImplementedInterfaceNames()
 
 
+    /**
+     * Determine if the passed token is part of a merge conflict boundary.
+     *
+     * Used to distinguish whether a token is an operator or part of a git
+     * merge conflict boundary.
+     * For PHP and CSS, only the T_SL token is checked.
+     * For JS, any of the following tokens will be checked: T_SL, T_ZSR,
+     * T_LESS_THAN, T_GREATER_THAN, T_IS_IDENTICAL and T_EQUAL.
+     *
+     * If a merge conflict boundary is found, it will return a stackPtr to
+     * to the end of the boundary to enable sniffs which listen for these
+     * operators to skip over the rest of the merge conflict boundary.
+     *
+     * @param int $stackPtr The position of the target token.
+     *
+     * @return boolean|integer Boolean false if the token is not part of
+     *                         a merge conflict boundary.
+     *                         An integer stackPtr to the end of the merge
+     *                         conflict boundary if it is.
+     */
+    public function isMergeConflictBoundary($stackPtr)
+    {
+        if ($this->tokens[$stackPtr]['code'] !== T_SL
+            && ($this->tokenizerType !== 'JS'
+            || isset(Util\Tokens::$gitBoundaryTokens[$this->tokens[$stackPtr]['code']]) === false)
+        ) {
+            return false;
+        }
+
+        $startOfLine = $stackPtr;
+        if ($this->tokens[$stackPtr]['column'] !== 1) {
+            for ($i = ($stackPtr - 1); $i >= 0; $i--) {
+                if ($this->tokens[$i]['column'] === 1) {
+                    $startOfLine = $i;
+                    break;
+                }
+            }
+        }
+
+        $endOfLine = $stackPtr;
+        if (isset($this->tokens[($stackPtr + 1)]) === true) {
+            for ($i = ($stackPtr + 1); $i < $this->numTokens; $i++) {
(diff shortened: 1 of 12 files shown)


gsherwood:
My feeling on this one is that the core shouldn't have git-specific (or any VCS really) code in it because it's not going to be used by enough sniffs, or enough other places in the core.

In this specific case, I understand that you're trying to sort out the fixer conflicts, but I think that having git merge conflicts inside source files is going to be pretty rare, so I don't think there is too much of a problem having the sniff include this merge-detection code itself. If we find that it gets repeated too often, we could always consider moving it into the core then.

> As a side-note, it would be great if the tokenizers would handle merge conflict artefacts better or at least more consistently.

Maybe this is the permanent solution, but it requires detection of merge conflicts from a range of vcs products and it actually might be more effort than it is worth. Failures of PHPCS due to merge conflicts isn't something I have reported to me. Plus, it is best if a file with merge conflicts doesn't get fixed at all (even if you ask PHPCBF to fix it) so maybe the solution is to use the vcs to detect a file in conflict before it is checked rather than have PHPCS handle a potentially tough piece of code to parse.

gmponos:
@jrfnl personally I believe that this is a great job you did here.

I also wanted to create a sniff like this but I didn't for the following reasons.


 - I was afraid that it might conflict with sniffs about `heredoc`.
 - I didn't know how to do it. First of all I was not sure how to detect the merge conflicts. If the merge conflict starts from a comment I believe that using codesniffer it is not detectable.

So if you ever happen to have a case like this:

``` php
/**
<<<<<<< HEAD
 * @var string $bar
============
 * @var string $bar
>>>>>>> master
 */
```

Furthermore what happens to cases that the above extends to further than comments like this:

``` php
/**
<<<<<<< HEAD
 * @var string $bar
 */
public function foo($bar){ }

/**
============
 * @var string $bar
>>>>>>> master
*/
```

- After thinking about the above case I thought that I must scan the whole file in order to find conflicts. So that leads me to the third reason why I dropped the idea of creating this sniff. Because the way I see it since I can't detect cases like the above and I have to scan the whole file I prefer not to use codesniffer to detect conflicts and use `php lint`.

if you have a case in your code like this then php lint will fail.

``` php
public function foo($bar) {
<<<<<<< HEAD
$x = 1 + 2;
============
$x = 1 + 1;
>>>>>>> master
}
```

So I can easily add a hook before committing my changes and run only the staged files like `php -l myfile.php`. Afterward I can do a grep about  `<<<<<<< HEAD` or `=========`  or `>>>>>>>`

This way I *save* a lot of processing time and not needed commits since php lint is way faster.

Although when it comes to OSS you can't control what everyone will have at their git hooks. So you could also add `php lint` in your continuous integration or add `php lint` as a sniff (although I believe it already exists, but I am not sure).

PS. My comment concerns only the PHP and not JS, and CSS.

jrfnl:
> My feeling on this one is that the core shouldn't have git-specific (or any VCS really) code in it because it's not going to be used by enough sniffs, or enough other places in the core.

I understand and actually agree, but couldn't think of another way to solve this without lots of duplicate code.

> In this specific case, I understand that you're trying to sort out the fixer conflicts

If I remember correctly, there was only one test case file with a merge conflict boundary in it, so as an alternative to this PR, I could just add that one file to an `ignore` directive for the fixer test.

I'm a bit wary of that solution as it opens the way to just adding more and more files to an `ignore` list for the fixer test which goes against the principle of the test, but I do think this is the one case where it is legitimate to make an exception.

@gsherwood If you agree with the alternative I propose, you can just close this PR and I will add the ignore to the future PR which will add the fixer conflict test to travis.

gsherwood:
> If you agree with the alternative I propose, you can just close this PR and I will add the ignore to the future PR which will add the fixer conflict test to travis.

Sounds ok to me. I also don't want to just exclude every file that is hard to fix in testing, but this one does feel like an exception, so I think we can live with it.

PR #1728, opened 2017-10-17
New Generic.Arrays.ShortArraySpacing sniff

### LINKED ISSUE #1727: ArrayBracketSpacingSniff doesn't catch `[ 1, 2, 3 ]`
My expectation is that`ArrayBracketSpacing` enforces that *that there are no spaces around square brackets.*

However, given the following file *array.php*:

```php
<?php
$array = [ 1, 2, 3 ];

```

The following command does *not* fail (for me):
`phpcs array.php --standard=Squiz --sniffs=Squiz.Arrays.ArrayBracketSpacing`

Why is that?

caugner:
Fixes #1727

### DIFF
--- a/src/Standards/Generic/Sniffs/Arrays/ShortArraySpacingSniff.php
+++ b/src/Standards/Generic/Sniffs/Arrays/ShortArraySpacingSniff.php
@@ -1,0 +1,98 @@
+<?php
+/**
+ * Ensure that there are no spaces around short array square brackets.
+ *
+ * @author    Greg Sherwood <gsherwood@squiz.net>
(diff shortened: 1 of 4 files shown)


caugner:
Problem: It doesn't handle multi-line short arrays correctly (yet).

gsherwood:
My intention is to provide a range of sniffs that can help enforce array definition and formatting rules in a future version, and to retire the one from the Squiz standard. This sniff is a bit too specific when considered within that context, so I feel like it might be removed during future array sniff development so I'm hesitant to merge it in without knowing exactly what array sniffs are going to be written.

For reference, the issue about breaking up the array sniff is issue #582.

caugner:
Closing, as #582 should solve it eventually.

Issue #1586, opened 2017-08-03
Generic.WhiteSpace.DisallowSpaceIndent.SpacesUsed shouldn't show waring if used in an statement


JoryHogeveen:
When using multiline `if` statements etc. this check generates an error when formatted like this:

```php
	if ( CONDITION &&
	     CONDITION &&
	     CONDITION
	) {
		//...
	}
```
The whole block is indented with tabs but within the if statement shouldn't it be indented as spaces for alignment dispite your tabs width settings?

gsherwood:
That code doesn't generate any errors for me, but I am assuming you've indented the 2nd and 3rd condition lines with tabs and then a single space to align the condition text. Like this:
```
if·($foo)·{
»   if·(·CONDITION·&&
»   »   ·CONDITION·&&
»   »   ·CONDITION
»   )·{
»   »   //...
»   }
}
```

If you've instead done something like this:
```
if·($foo)·{
»   if·(·CONDITION·&&
»   ·····CONDITION·&&
»   ·····CONDITION
»   )·{
»   »   //...
»   }
}
```

Then the sniff is going to produce errors because some of those spaces can be replaced by a single tab.

If you can tell me exactly how your code is formatted, I might be able to help figure out if there is a bug in the sniff.

JoryHogeveen:
Hi @gsherwood ,

I indeed indent like your second example. Using tabs to for the indentation of the full `if` block and spaces if after that.
I also think this _should_ be the correct way since it allows for fixed alignment of conditions. If you use tabs the alignment will change when you change the tab size.

For example (using the example that is correct according to the current standards):

```
if·($foo)·{
»   if·(·CONDITION·&&
»       ·CONDITION·&&
»       ·CONDITION
»   )·{
»   »   //...
»   }
}
```

When you change the tab size to 8 (like GitHub does) it would look like this:
```
if·($foo)·{
»       if·(·CONDITION·&&
»               ·CONDITION·&&
»               ·CONDITION
»       )·{
»       »       //...
»       }
}
```

JoryHogeveen:
Oh, same thing for long string alignments:

Not the the last `.` is actual PHP.
```
»   $var = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut sagittis, '
»   ........ 'neque a venenatis gravida, tellus nisi dictum justo, ut euismod nisi justo placerat libero. '
[...]

PR #1736, opened 2017-10-29
Allow DisallowSpaceIndentSniff to support multiline alignment.

### LINKED ISSUE #1735: Generic_Sniffs_WhiteSpace_DisallowSpaceIndentSniff cannot deal with alignment vs indentation
The sniff `Generic_Sniffs_WhiteSpace_DisallowSpaceIndentSniff` does not properly recognize the difference between _indentation_ (denoting the change of scope, usually done with tabs) and _alignment_ (positioning within same scope, usually done with spaces). I'm using these terms as they are visually represented [in this blog article](https://dmitryfrank.com/articles/indent_with_tabs_align_with_spaces).

Here's an example, with a multi-line conditional:
```php
<?php
function my_function() {
	if ( $first_condition === true
	     || $second_condition === true
	     || $answer !== 42 ) {
		// Do something here.
	}
}
```

The above uses 1 tab in all three lines within the function to denote that all conditions are being processed at the same level.
However, the second and third one are also preceded by 5 spaces, to align them with the beginning of the first one.

The current implementation of the sniff has the concept of "precision indentation", which I assume is meant to relate to _alignment_. However, it assumes that this precision indentation should always be smaller in number of spaces than the `$_tabWidth`.

With the default tab width of 4, this makes the code above fail, while it would not fail with a tab width of 6. So, the value of differentiating between indentation and alignment is moot, as both are directly depending on the tab width again.

Alignment with spaces (that follows after tabs and is not mixed with them) should be acceptable regardless of the number of space chars, as the needed alignment depends on the preceding code that it should be aligning to.

schlessera:
The `Generic/Sniffs/WhiteSpace/DisallowSpaceIndentSniff` erroneously tries to manipulate spaced alignment over multiple lines, where the number of spaces exceeds `$tabWidth`.

This pull request fixes this issue by trying to be smart about remembering the indentation of the previous line and ignoring the spaces that are supposed to be for alignment only. They need to meet the following criteria:

* Same tabbed indentation than on previous line.
* No tabs after spaces.

Fixes #1735

### DIFF
--- a/src/Standards/Generic/Sniffs/WhiteSpace/DisallowSpaceIndentSniff.php
+++ b/src/Standards/Generic/Sniffs/WhiteSpace/DisallowSpaceIndentSniff.php
@@ -74,11 +74,15 @@ class DisallowSpaceIndentSniff implements Sniff
                         T_DOC_COMMENT_WHITESPACE => true,
                        );
 
-        $tokens = $phpcsFile->getTokens();
+        $tokens        = $phpcsFile->getTokens();
+        $currentIndent = 0;
         for ($i = ($stackPtr + 1); $i < $phpcsFile->numTokens; $i++) {
             if ($tokens[$i]['column'] !== 1 || isset($checkTokens[$tokens[$i]['code']]) === false) {
                 continue;
             }
+
+            $previousIndent = $currentIndent;
+            $currentIndent  = 0;
 
             // If tabs are being converted to spaces by the tokeniser, the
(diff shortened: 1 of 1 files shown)


gsherwood:
PR references issue #1735. This is a duplicate of #1586 so this PR would be targeting a change for that.

schlessera:
I started looking into rebasing this to latest `master`, but it's a bit more involved, so I'll need to take a deeper look as soon as I find the time.

PR #1737, opened 2017-10-30
Squiz.CSS.EmptyStyleDefinition sees comment as style definition and fails to report error

jrfnl:
The `Squiz.CSS.EmptyStyleDefinition` sniff checks for CSS styles being defined without value.

This sniff would fail to report an issue if there is no value, but a comment is encountered instead.

Includes unit tests.


jrfnl:

gsherwood:
Thanks for this fix

jrfnl:
You're welcome ;-)

PR #1739, opened 2017-10-30
Squiz/OperatorBracket: fix fixer conflict

jrfnl:
PR thirteen in the series to fix fixer conflicts.

The `Squiz.Formatting.OperatorBracket` sniff would conflict with itself if it could not identify correctly the start or end of the arithmetic expression which would cause it to go into a fixer loop.

The fix in this commit checks if that's the case before throwing the error and if so, does throw the error, but doesn't try to auto-fix it anymore.

Includes additional unit test highlighting the issue.


jrfnl:

gsherwood:
Thanks

PR #1749, opened 2017-11-18
Add link to download the phar files in the README

mnapoli:
There was a link pointing to the Phar documentation, which I confused with the link to download the actual phars. I think this link is confusing and useless.

Instead it makes much more sense to have a link to the files to download directly (I spent several minutes trying to find where to download those files).

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ PHP\_CodeSniffer requires PHP version 5.4.0 or greater, although individual snif
 Installation
 ------------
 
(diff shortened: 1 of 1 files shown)


gsherwood:
The reason that link is there is because this was original just a PEAR package and proceeded Phars and Composer, so they were still new-ish install methods when I wrote that. But people know what those are now, so I think I'm just better off removing the links all together to remove confusion.

Thanks for letting me know.

gsherwood:
I've made that README change now. I also moved the composer install steps above the PEAR ones because composer installs are more common than PEAR installs for PHPCS.

PR #1762, opened 2017-11-26
Generic.WhiteSpace.Disallow[Space/Tab]Indent not inspecting content before open tag

jrfnl:
Both these sniffs include `T_INLINE_HTML` in the `$checkTokens` array, but would only start checking the file **after** the first PHP open tag, so if a file - such as a view - would start with inline HTML and only have the first PHP open tag on line 5, the first 4 lines would not be inspected, nor fixed for tabs vs spaces.

I have not added unit tests for this as it would require to either renumber all the unit tests òr to add a second unit test file for both sniffs. If so desired, I'd be happy to make those changes, but it seemed a bit over the top for such a small change.


gsherwood:
Thanks for this fix.

jrfnl:
You're welcome ;-)

PR #1768, opened 2017-12-04
Removed old PHP version checks

webimpress:
Repository supports PHP 5.4+

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -229,12 +229,6 @@ class Runner
      */
     public function checkRequirements()
     {
-        // Check the PHP version.
-        if (PHP_VERSION_ID < 50400) {
-            $error = 'ERROR: PHP_CodeSniffer requires PHP version 5.4.0 or greater.'.PHP_EOL;
-            throw new DeepExitException($error, 3);
-        }
-
         if (extension_loaded('tokenizer') === false) {
             $error = 'ERROR: PHP_CodeSniffer requires the tokenizer extension to be enabled.'.PHP_EOL;
             throw new DeepExitException($error, 3);
--- a/src/Standards/Generic/Tests/PHP/DisallowShortOpenTagUnitTest.php
+++ b/src/Standards/Generic/Tests/PHP/DisallowShortOpenTagUnitTest.php
@@ -52,13 +52,6 @@ class DisallowShortOpenTagUnitTest extends AbstractSniffUnitTest
     {
         switch ($testFile) {
         case 'DisallowShortOpenTagUnitTest.1.inc':
-            if (PHP_VERSION_ID < 50400) {
-                $option = (boolean) ini_get('short_open_tag');
-                if ($option === false) {
-                    // Short open tags are off and PHP isn't doing short echo by default.
-                    return [];
-                }
-            }
             return [
                 5  => 1,
                 6  => 1,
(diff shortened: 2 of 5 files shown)


src/Runner.php
```diff
@@ -229,12 +229,6 @@ public function runPHPCBF()
      */
     public function checkRequirements()
     {
-        // Check the PHP version.
-        if (PHP_VERSION_ID < 50400) {
-            $error = 'ERROR: PHP_CodeSniffer requires PHP version 5.4.0 or greater.'.PHP_EOL;
```
jrfnl: This one looks like it should not be removed. This is the error which will be thrown if for whatever reason someone run PHPCS 3.x on PHP 5.3.

src/Standards/Generic/Tests/PHP/DisallowShortOpenTagUnitTest.php
```diff
@@ -52,13 +52,6 @@ public function getErrorList($testFile='')
     {
         switch ($testFile) {
         case 'DisallowShortOpenTagUnitTest.1.inc':
-            if (PHP_VERSION_ID < 50400) {
```
jrfnl: I think more changes are needed for the DisallowShortOpenTag sniff. I'll have a look.

src/Runner.php
webimpress: The requirement is set in composer and I think it should be enough.

src/Standards/Generic/Tests/PHP/DisallowShortOpenTagUnitTest.php
webimpress: @jrfnl Did you have chance to check it? I don't know what do you mean...

src/Runner.php
jrfnl: @webimpress By far not all people use PHPCS via Composer, so I don't necessarily agree.

src/Standards/Generic/Tests/PHP/DisallowShortOpenTagUnitTest.php
jrfnl: @webimpress Sorry, I'm at a conference at the moment, I'll try and get to it over the next few days.

src/Runner.php
gsherwood: This check needs to stay.

webimpress: Ok then, should I revert it or you will do it on merge?

gsherwood: I've just noticed the autoloader fails for < 5.4, so leave it. I think I need to move the requirement check higher up.

src/Standards/Generic/Tests/PHP/DisallowShortOpenTagUnitTest.php
jrfnl: Ok. Finally had a chance to look at this properly. I was confusing this sniff with the sister-sniff `DisallowAlternativePHPTags`. I'm sending you a PR with a very minor addition to this PR.

Other than that, I noticed another issue regarding the `DisallowShortOpenTag` sniff (and it's sister), but that should be handled in a separate PR - see #1784.

And another issue regarding the testing of the `DisallowAlternativePHPTags` sniff, but, again, that should be handled in another PR - see #1786.

TomasVotruba:
:+1:

gsherwood:
I ended up just putting the version check back into `checkRequirements()` because I have up trying to get a PHP 5.3 install running again to test/fix the autoloader. Having that version check there will at least remind me to update it if I change the min version again.

Thanks for finding and fixing the other checks.

PR #1771, opened 2017-12-06
Added new sniff Generic.PHP.DeclareStrictTypes

### LINKED ISSUE #911: Add new sniff for strict _types declaration
New sniff for php 7 is needed to allow to enforce `declare(strict_types=1)`

webimpress:
Configuration option:

- ~`omitCommentWithTags` - tags names, comment with one of these tags will be omitted and declaration will be added below the comment (default: `['@author', '@copyright', '@license']`,~
- `format` - how declaration should look like (default `declare(strict_types=1);`).
- ~`linesBefore`~ `spacesBefore` - number of blank lines before the declaration (default `1`)
- ~`linesAfter`~ `spacesAfter` - number of lines after the declaration and before the script (default `1`)

Resolves #911 

/cc @Xerkus


src/Standards/Generic/Sniffs/PHP/DeclareStrictTypesSniff.php
```diff
@@ -0,0 +1,211 @@
...
+
+
+    /**
+     * Returns an array of tokens this test wants to listen for.
+     *
+     * @return array
+     */
+    public function register()
+    {
+        return [T_OPEN_TAG];
```
gmponos: Do you believe that it would be nice to have a check not to run this if the PHP version is below php 7.

I think that if someone uses this sniff inside a ruleset and has a travis that he checks his code against php 5 and 7 it will be hard for him to include/exclude this sniff for the specific version

webimpress: @gmponos I don't think it's needed, because this sniff should be included in your ruleset only when your project supports PHP 7.0+. Otherwise it has no sense.

gmponos: 🤦‍♂️ you are correct.

jrfnl:
@webimpress :+1: 

Just wondering whether it would be better to have two sniffs:
* one to require a declare strict statement
* one to check the formatting

The reason would be that some standards may require a strict statement, but haven't got a strict guideline on how it should be formatted.
Or similarly, a standard where doesn't require strict typing for every file, but does want the statement to be formatted in a certain way if it exists.

Of course, the other errorcodes could be silenced or - if using PHPCS 3.x - the `NotFound` errorcode could be included on its own, but having two separate sniffs might be simpler and more discoverable ?

Xerkus:
I could have sworn strict types declaration was required to be the first statement in the file. Apparently other `declare()`s is an exception

webimpress:
@jrfnl Thanks for your great feedback. I've addressed almost all of your suggestion in my last update.
I had to remove one test with different configuration option, because now I check also if declare statement is first statement in the file. I don't know how else we can test different configuration options?

I need to check on more case:
```
<?php
// comment

declare(strict_types=1);
```

I think in that case, also declaration should be moved before the comment.

Could you have a look please on my changes and let me know what do you think? Thanks!

gsherwood:
Sorry, I've just got to this on my todo list.

I actually don't like the idea of either the `omitCommentWithTags` or `format` properties. The `omitCommentWithTags` mixes this sniff up with documentation sniffs, and the `format` property assumes everyone is going to want to format their declare line in a very different way. I don't think I'd merge this in with either of those options, so I'll need to know how important they really are.

If I was writing a strict types sniff, I'd write one to enforce the format outlined in PSR-12 because I imagine that will be the way developers will write these statements when they don't have any strong opinion about how to format them or where to place them.

So to be merged in, I think this sniff should only be checking the format of the declare statement, to ensure it looks exactly like `declare(strict_types=1);` and has a blank line before and after, depending on if there is any content there. The properties would probably be `spacingBefore` `spacingAfter` for the number of blank lines (consistent with props used by other sniffs) and maybe even `requiredSpacesAfterOpen` and `requiredSpacesBeforeClose` if we think people will want spaces inside the parenthesis.

PSR-12 should be able to make use of this sniff, and then ensure other rules (like the fact this must come *after* a file comment) are followed using new sniffs.

I'm going to remove this from the 3.3.0 roadmap, but will re-add if we can agree on a new version of the sniff before that release.

webimpress:
@gsherwood I've updated this PR:
- rebased and resolved conflicts;
- removed  `omitCommentWithTags` property;
- renamed `linesBefore` and `linesAfter` properties to `spacesBefore` and `spacesAfter` accordingly;

I didn't add `requiredSpacesAfterOpen` and `requiredSpacesBeforeClose`. I can't see reason why someone would like to have space after open and no space before close. I think `format` gives us more flexibility and we know what we should insert by default (if the strict declaration is missing).
We can make this property protected/private if you don't want this flexibility.

Is there any chance to bring it back to 3.3.0 release?

src/Standards/Generic/Sniffs/PHP/DeclareStrictTypesSniff.php
```diff
@@ -310,7 +310,7 @@ public function process(File $phpcsFile, $stackPtr)
 
                 $this->checkOtherDeclarations($phpcsFile, $next);
 
-                return (count($tokens) + 1);
+                return $phpcsFile->numTokens;
```
jrfnl: If I remember correctly, elsewhere `return ($phpcsFile->numTokens + 1)` is used. Shouldn't really make a difference, but just thought I'd mention it for consistency sake.

webimpress:
@gsherwood any chance to re-add it into 3.3.0 release?

WarriorXK:
Is there any progress on this? I would like to include this in our ruleset

player259:
@WarriorXK You can use SlevomatCodingStandard.TypeHints.DeclareStrictTypes from https://github.com/slevomat/coding-standard. There are many useful sniffs.

floverdevel:
How can i try the `Generic.PHP.DeclareStrictTypes` ?
Usually i run phpcs with the `--exclude=Generic.Files.LineLength --standard=PSR2` options.
Is there a way to run it so i can test this PR?
Maybe something like this :
`vendor/bin/phpcs -p  --extensions=php --exclude=Generic.Files.LineLength --sniffs=Generic.PHP.DeclareStrictTypes --standard=PSR2 src/ tests/`

gsherwood:
@floverdevel To test, you'll need to merge this PR, then create a `ruleset.xml` file as you can't do what you are after on the command line alone. Make the contents of the file:
```xml
<?xml version="1.0"?>
    <ruleset name="MyStandard"> 
    <rule ref="PSR2">
        <exclude name="Generic.Files.LineLength"/>
    </rule>
    <rule ref="Generic.PHP.DeclareStrictTypes"/>
</ruleset>
```
Then run PHPCS like this: `vendor/bin/phpcs -p --extensions=php --standard=/path/to/ruleset.xml src/ tests/`

floverdevel:
@gsherwood thank you very much !!!

gsherwood:
This sniff still has a public sniff property called `format`, but that's not how the other sniffs set their rules. Maybe we need to list out how the format of the declare statement could change and how this sniff wants to allow it to change.

Maybe we need a setting for how many spaces there are inside the parenthesis. Maybe another for spaces around the equals sign? Or maybe we release it to enforce the most common format first and add settings later if they are requested.

floverdevel:
@gsherwood just a quick question : what prevents us from merging this pull-request?
I understand that the code may not be perfect, but it could help many people to use this sniff ... can we merge the code and refactore thereafter?

gsherwood:
> can we merge the code and refactore thereafter?

The changes relate to public sniff properties. Changes to these would be BC breaks that can only be made in a major version, so this needs to be right before it is released.

floverdevel:
Any news on this PR ?
I see that one job failed ...
https://travis-ci.org/squizlabs/PHP_CodeSniffer/jobs/483778385

gsherwood:
> Any news on this PR ?

No changed has been made to remove the `format` property and instead provide sniff properties to configure the format, so the PR won't be merged. I'd even be happy with sniff enforcing a single format only.

I also don't think this sniff should be inserting the declare statement if it doesn't exit. The PHPCS fixers don't modify the behaviour of code - only the formatting of the code.

webimpress:
@gsherwood Changes  has been made in this PR very long time ago. Property `format` is private so it is not possible to configure it. 

Anyway, I am closing this PR as I can't see any interest of including it in that library.

Everyone who is interested in using sniff like that can use my library:
https://github.com/webimpress/coding-standard
and sniff: `WebimpressCodingStandard.Sniffs.DeclareStrictTypes`

Dr10s:
You no need configure format property.
You can configure declaration by other  public properties:
```
<rule ref="vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/TypeHints/DeclareStrictTypesSniff.php" >
        <properties>
            <property name="spacesCountAroundEqualsSign" value="0"/>
        </properties>
</rule>
```

PR #1784, opened 2017-12-14
Build/Travis: bring back testing with alternative ini settings for PHP 5.x

jrfnl:
When PHPCS 3.x was being prepared, the builds for PHP < 5.4 were removed from the Travis matrix. However, that also meant that the build against alternative ini settings was removed, while it shouldn't have been.

Re-added against an arbitrary PHP 5.x version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ matrix:
   include:
     - php: 5.4
     - php: 5.5
+    - php: 5.5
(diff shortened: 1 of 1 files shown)


gsherwood:
I had the custom ini in there for PHP7, so I probably just skipped over PHP5 in my head. Thanks for re-adding.

jrfnl:
No worries. It happens :smile:

PR #1786, opened 2017-12-14
Generic/DisallowAlternativePHPTags: improve unit tests

jrfnl:
The current unit tests did not test that the `warning`s were correctly being thrown when the `asp_tags` ini setting was set to `Off`.
This needs to be tested using a separate test file as the warnings cannot be auto-fixed, so the unit tests would fail if this was tested against unit test file 2.

This change in this commit ensures that the warnings are now also unit tested.

N.B.: The actual testing will only be done once PR #1784 has been merged!


gsherwood:
Thanks for doing this.

PR #1787, opened 2017-12-14
Remove unused test file

jrfnl:
This file was replaced by the `DisallowShortOpenTagUnitTest.[1-3].inc` files and was already removed from the `package.xml` file.


gsherwood:
Good catch. Thanks.

Issue #1788, opened 2017-12-14
3.2 hangs forever reading from stdin on our Jenkins/Docker CI


stephank:
Hi! There's a bit of a complicated issue going on between the stdin detection changes and the way Jenkins with the Docker pipeline plugin runs things.

We basically have a project with phpcs as a dev dependency, and a Dockerfile from which we build the project. Then we run phpcs inside that image. In our Jenkinsfile that looks like:

```groovy
let image = docker.build('foobar')
image.inside {
  sh '''#!/bin/sh -xe
    ./vendor/bin/phpcs -p --standard=PSR2 --extensions=php src/
  '''
}
```

Even though we don't specify stdin on the commandline, phpcs appears to hang reading from stdin, which blocks indefinitely. If I tack on `< /dev/null` at the end of the phpcs command, things work again. 3.1.1 also works fine.

I'll try to get some more info. It should be possible to reproduce this with Docker alone.

stephank:
It appears I cannot run phpcs in Docker on my local macOS with Docker for Mac setup either. I tried:

```sh
docker run --rm -it -v $PWD:/src -w /src php ./vendor/bin/phpcs -p --standard=PSR2 --extensions=php src/
```

Or just starting a shell in a PHP container, and running it from there:

```sh
docker run --rm -it -v $PWD:/src -w /src php /bin/sh
# ./vendor/bin/phpcs -p --standard=PSR2 --extensions=php src/
```

But strangely, the `< /dev/null` trick doesn't work here, whereas it does in Jenkins.

It does respond to `--help` and also shows usage when you run it simply without args.

kissifrot:
It seems it also hangs when used from PhpStorm:
https://youtrack.jetbrains.com/issue/WI-39190
https://stackoverflow.com/questions/47797797/error-with-php-code-sniffer-and-phpstorm

gsherwood:
Any chance you could test master? I've committed a fix for the PHPStorm issue already (just needs some more testing) and it would be good to see if that's the same issue. If that fix checks out for you as well, I can do a 3.2.1 release to resolve it.

eclipxe13:
I tested locally using PHPStorm and it works fine! thanks @gsherwood

Running on Debian GNU linux Buster (testing)
Phpstorm 2017.3
PHP_CodeSniffer version 3.2.1 (at commit 6172062f0)

gsherwood:
@eclipxe13 Thanks for testing.

[...]

PR #1792, opened 2017-12-15
fix(stdin) set non-blocking stream when cli stdin are used

Pierozi:
Address issue #1788 
Sequel of #1490 

I believe the set non-blocking mode as been missed during the refactoring.
@arnested Could you confirm ?

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -365,6 +365,7 @@ class Config
         }
 
         $handle = fopen('php://stdin', 'r');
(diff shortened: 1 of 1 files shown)


arnested:
Sure, I'll have a look later today or Saturday or Sunday.

Pierozi:
Actually, already been fixed by @gsherwood adding timeout of 200ms using `stream_select` instead of `stream_get_contents`

PR #1810, opened 2017-12-25
Generic/OpeningBraceKernighanRitchie: minor efficiency fix

jrfnl:
No need to call the `recordMetric()` method again if the metric has already been recorded.


gsherwood:
Thanks

PR #1813, opened 2017-12-26
PEAR/ScopeClosingBrace: fix fixer conflict

jrfnl:
Another small PR to fix a fixer conflict between the `PEAR.WhiteSpace.ScopeClosingBrace` and the `PEAR.WhiteSpace.ScopeIndent` (extended from `Generic.WhiteSpace.ScopeIndent`) sniffs.

If a `case` or `default` statement uses braces, the scope indent and scope closing brace would fight over where the brace should be placed.

The unit test file already contains a test case highlighting the issue.
https://github.com/squizlabs/PHP_CodeSniffer/blob/b55fd82f4f6335a519bd787703810551a2dfb9ad/src/Standards/PEAR/Tests/WhiteSpace/ScopeClosingBraceUnitTest.inc#L88-L90
To reproduce the issue, run the following command on the `master` branch:
`phpcbf -p -s -vv ./src/Standards/PEAR/Tests/WhiteSpace/ScopeClosingBraceUnitTest.inc --standard=PEAR`

Note: this change *does* change the behaviour of the sniff in that particular case. If that's undesired, another solution will need to be found.

### DIFF
--- a/src/Standards/PEAR/Sniffs/WhiteSpace/ScopeClosingBraceSniff.php
+++ b/src/Standards/PEAR/Sniffs/WhiteSpace/ScopeClosingBraceSniff.php
@@ -137,8 +137,9 @@ class ScopeClosingBraceSniff implements Sniff
         }
 
         $fix = false;
-        if ($tokens[$stackPtr]['code'] === T_CASE
-            || $tokens[$stackPtr]['code'] === T_DEFAULT
+        if (($tokens[$stackPtr]['code'] === T_CASE
+            || $tokens[$stackPtr]['code'] === T_DEFAULT)
+            && $tokens[$scopeEnd]['code'] !== T_CLOSE_CURLY_BRACKET
         ) {
             // BREAK statements should be indented n spaces from the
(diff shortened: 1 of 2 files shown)


gsherwood:
I don't like the change of behaviour for PEAR.WhiteSpace.ScopeClosingBrace.BreakIndent with this, but there is a clear conflict here.

I think the only way to fix this is figure out how to change PEAR.WhiteSpace.ScopeIndent to not report errors for the closing brace of CASE and DEFAULT statements. But I don't really want to do that with a new sniff property because it's pretty specific, so that might mean (finally) refactoring the main scope indent sniff so that some of it can be overridden in the PEAR sniff.

jrfnl:
> that might mean (finally) refactoring the main scope indent sniff so that some of it can be overridden in the PEAR sniff.

@gsherwood Yes, that's one I haven't dared to touch yet, but I'm seeing more conflicts which involve that sniff. I still need to find the time to dig in deeper to come up with good test cases though.
Will report back when I have.

PR #1820, opened 2017-12-30
Squiz/EmbeddedPhp: Allow for ignore comments

jrfnl:
This is basically the same fix as previously pulled & merged in #1554, but now to allow for `// phpcs:ignore` comments as they have their own token.

> The Squiz.PHP.EmbeddedPhp sniff will create a fixer conflict with itself when it encounters a // comment just before the PHP close tag.
> 
> The tokenizer will tokenize whitespace at the end of a // comment as part of the T_COMMENT token, so the fixer just keeps adding more spaces between the T_COMMENT and the T_CLOSE_TAG which in the next fixer round is then tokenized again as being part of the T_COMMENT resulting in the loop.

### DIFF
--- a/src/Standards/Squiz/Sniffs/PHP/EmbeddedPhpSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/EmbeddedPhpSniff.php
@@ -373,7 +373,7 @@ class EmbeddedPhpSniff implements Sniff
         if ($tokens[($closeTag - 1)]['code'] === T_WHITESPACE) {
             $trailingSpace = strlen($tokens[($closeTag - 1)]['content']);
         } else if (($tokens[($closeTag - 1)]['code'] === T_COMMENT
-            || $tokens[($closeTag - 1)]['code'] === T_PHPCS_IGNORE)
+            || isset(Tokens::$phpcsCommentTokens[$tokens[($closeTag - 1)]['code']]) === true)
             && substr($tokens[($closeTag - 1)]['content'], -1) === ' '
         ) {
             $trailingSpace = (strlen($tokens[($closeTag - 1)]['content']) - strlen(rtrim($tokens[($closeTag - 1)]['content'])));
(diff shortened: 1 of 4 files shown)


gsherwood:
Do you think it would be best to use `Tokens::$commentTokens` instead? It includes docblock comments as well, but I don't think that's a problem, although I haven't looked into it properly yet.

jrfnl:
> Do you think it would be best to use Tokens::$commentTokens instead? It includes docblock comments as well, but I don't think that's a problem, although I haven't looked into it properly yet.

@gsherwood Maybe.. let me think this through.

The problem this PR and the previous one address, is that a `// comment     `  with whitespace after it will be tokenized including the whitespace, which is why the the sniff will never find whitespace between a `//` comment and the closing tag and goes into a fixer loop because of it.

The closing tag of `/* comment */` and `/** docblock */` does not include any whitespace after it - whitespace after it will be tokenized as `T_WHITESPACE`, so none of the other comment tokens should have that problem.

Having said that, this also means that the `substr() === ' '` condition will never match, so it would probably do no harm.

Maybe it would be useful to have a separate "Tokens::$phpcsCommentTokens" variable ? That way, this sniff and other sniffs could just look for `T_COMMENT` + `Tokens::$phpcsCommentTokens` in these type of situations.

gsherwood:
> Maybe it would be useful to have a separate Tokens::$phpcsCommentTokens variable ?

Yeah, that's probably a better idea.

jrfnl:
@gsherwood I've added a second commit implementing that.

jrfnl:
@gsherwood A quick search of the sniffs tells me there are a couple more sniffs which could benefit from using the new `$phpcsCommentTokens` array.
Would you like me to add a additional commit fixing up those sniffs too or would you rather I do that in a separate PR ?

gsherwood:
> Would you like me to add an additional commit fixing up those sniffs too or would you rather I do that in a separate PR ?

Sorry, I merged this in before I saw that, but it's probably better to review them separately anyway. Thanks for doing that change.

jrfnl:
@gsherwood No worries, I've started pulling the various fixes now. Sending in separate PRs per sniff. All are very small PRs and should be easy to review/merge.

jrfnl:
Thanks for merging those. I expect to be able to pull the rest tomorrow.

PR #1823, opened 2018-01-02
ValidVariableName sniffs: remove superfluous property setting

jrfnl:
Looks like the property is no longer used in either of these sniffs or the parent sniffs they extend.

A quick look at the history reveals that the properties were introduced when the sniffs were written in 2007, but it looks like the use of the property was already removed a few days later (still in 2007) and that the property was never used in a released version.

Removing the properties does not seem to have any negative effects. The unit tests still pass without the property and the sniffs do not throw errors about an "undefined property".


gsherwood:
2007 :) Nice find. Thanks.

PR #1828, opened 2018-01-04
Squiz.WhiteSpace.SuperfluousWhiteSpace ignoreBlankLines property ignores whitespace after single line comments

jrfnl:
Setting `ignoreBlankLines` to `true`, would also cause the sniff to ignore single-line comments with superfluous whitespace at the end.

Includes unit test.

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
@@ -192,6 +192,7 @@ class SuperfluousWhitespaceSniff implements Sniff
 
             // Ignore blank lines if required.
             if ($this->ignoreBlankLines === true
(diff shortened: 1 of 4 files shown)


src/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
```diff
@@ -192,6 +192,7 @@ public function process(File $phpcsFile, $stackPtr)
 
             // Ignore blank lines if required.
             if ($this->ignoreBlankLines === true
+                && trim($tokens[$stackPtr]['content']) === ''
```
gsherwood: Might be easier to just use `&& $tokens[$stackPtr]['code'] === T_WHITESPACE`

jrfnl: Done. (Commit amended)

gsherwood:
Thanks for that change

PR #1830, opened 2018-01-04
Squiz/NonExecutableCode: support goto as a structure leaving command

jrfnl:
`goto` is a valid way to leave exit a control structure and was not (yet) taken into account by the `Squiz.PHP.NonExecutableCode` sniff.

This PR fixes that. Includes unit tests.

When adding the unit tests, I realized that some 50-something lines at the end of the case file would always have to be at the end of the case file. This meant that any new unit tests for this sniff would need to be added above that last test and would mean that the line numbers for the errors expected from that last test would need to be renumbered.

To prevent this for the future and to make it easier to add new tests, I've now split the unit test case file in two, moving the last test (those 50-something lines) to a secondary unit test case file.

For easier review, please look at the commits separately.


gsherwood:
Thanks for this. Good idea re: the second text file

PR #1831, opened 2018-01-04
Generic/LineLength: ignore new PHPCS annotations for the metrics

jrfnl:
Ignore the new `// phpcs:` annotation comments completely when they are on a line by themselves.
Do not record metrics about lines only containing these type of comments.

Errors would not be thrown anyway as lines containing these comments are ignored when throwing errors. Still, this does save the sniff from executing the rest of the logic when its not necessary.

Includes unit test.


jrfnl:
The Travis failure is unrelated to this PR. Looks like the PHP 7.2 build just needs to be restarted.

gsherwood:
Makes sense. Thanks for this.

jrfnl:
You're welcome.

PR #1832, opened 2018-01-04
Squiz/CSS/Indentation: minor efficiency fix

jrfnl:
This makes the sniff more efficient by skipping the logic for errors which will not be reported anyway.

Includes unit test.


gsherwood:
👍  Thanks 👍

PR #1833, opened 2018-01-04
PSR2/SwitchDeclaration: bug fix - ignore PHPCS annotation tokens

jrfnl:

Includes adjusted unit test.

This fixes a bug where when a `// phpcs:ignore` comment is at the end of a line, it would be incorrectly "fixed" by moving it down a line, making the comment apply to a different line altogether.


jrfnl:
The Travis failure is unrelated to this PR. Looks like two of the builds just needs to be restarted.

gsherwood:
👍 👍  Thanks

PR #1834, opened 2018-01-05
PSR2/ClassDeclaration: bug fix - ignore PHPCS annotation tokens

jrfnl:

Includes unit test.

This fixes a bug where an error would be reported for a `// phpcs:` comment being on the same line as the class close brace.


gsherwood:
Thanks 👍 👍

PR #1835, opened 2018-01-05
Squiz/ClassDeclaration: bug fix - ignore PHPCS annotation tokens

jrfnl:

Includes adjusted unit test.

This fixes a bug where an error would be reported for a `// phpcs:` comment being on the same line as the class close brace, as well as the fixer for that error moving the `// phpcs:` comment, while it shouldn't.


jrfnl:
The Travis failure is unrelated to this PR. Looks like one of the builds just needs to be restarted.

gsherwood:
👍  Thanks

Issue #1817, opened 2017-12-28
Blank line not enforced after control structure if comment on same line as closing brace


josephzidell:
```php
<?php

$true = true;
if ($true) {
	echo 'hi 1';
}
if ($true) {
	echo 'hi 2';
}//end if
if ($true) {
	echo 'hi 3';
}
```

Throws an error after the first `if`, but not after the second. When there's a closing comment, a blank line is not enforced.

This likewise affects `foreach`s etc...

gsherwood:
Thanks for reporting this

josephzidell:
Thanks @gsherwood really appreciate the quick fix

PR #1836, opened 2018-01-05
Squiz/ControlStructureSpacing: bug fix - allow for PHPCS annotation tokens

jrfnl:

Includes unit test.

Related to #1817, which was recently fixed in a7d4118b9ca880acd2b9a7d70b7861e9228756eb. This commit builds onto that fix.

Without the fix, the `NoLineAfterClose` error would not be reported when there is a `phpcs:` comment after a control structure close brace and the needed fix would not be applied.


gsherwood:
Thanks 👍

PR #1837, opened 2018-01-05
Squiz/FunctionSpacing: bug fix - ignore PHPCS annotation tokens

jrfnl:

Includes unit test.

Without this fix, the fixer would report an error for the spacing above the function and add two blank lines between the `// phpcs:` comment and the function declaration.


gsherwood:
Thanks

PR #1838, opened 2018-01-05
Expand ForbiddenFunctionsSniff

carusogabriel:
Based on PHP's [List of Function Aliases](http://php.net/manual/en/aliases.php#aliases).

Should I add tests to all of them, or no, is overwork?

### DIFF
--- a/src/Standards/Generic/Sniffs/PHP/ForbiddenFunctionsSniff.php
+++ b/src/Standards/Generic/Sniffs/PHP/ForbiddenFunctionsSniff.php
@@ -29,6 +29,13 @@ class ForbiddenFunctionsSniff implements Sniff
     public $forbiddenFunctions = [
         'sizeof' => 'count',
         'delete' => 'unset',
+        'doubleval' => 'floatval',
+        'is_double' => 'is_float',
+        'is_real' => 'is_float',
+        'is_integer' => 'is_int',
+        'is_long' => 'is_int',
+        'is_writeable' => 'is_writable',
+        'join' => 'implode',
     ];
 
     /**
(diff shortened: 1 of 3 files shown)


gsherwood:
You closed this without comment, but I've finally got to this issue in my todo list so I thought I'd comment anyway.

The ForbiddenFunctions sniff is mostly used by customising the list of forbidden functions inside a ruleset.xml file, as described here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#genericphpforbiddenfunctions

So a lot more functions could be added, but they wouldn't help anyone who is already overriding the property value, so I tend to not modify the Generic version of this sniff.

MarkMaldaba:
@gsherwood I think that this sniff should actually be separated into two:

1. A new `ForbiddenAliases` sniff should be added, which contains all the PHP alias functions, recommending that the canonical function be used instead.  Ideally, each function could be individually ignorable in ruleset.xml, so if your coding-style happens to allow `is_integer()`, say, you can easily allow this whilst still checking for the other aliases.  Note that there might be some functions in that list that shouldn't be included, as they are very prevalent, e.g. `die()`.
2. This sniff, `ForbiddenFunctions`, should be available as a tool for projects that recommend against built-in functions, e.g. because they have a custom alternative that they use instead.  This sniff should be empty by default, but should otherwise work as it does currently.

Thoughts?  I can create an issue for this if you think it makes sense.

gsherwood:
@MarkMaldaba I'm not really interested in maintaining a list of alias functions within PHPCS. Given my workload, it feels like one more list of things I wont be able to maintain, especially when some developers will want to allow some of the aliases to be used.

PR #1839, opened 2018-01-05
Squiz/CommentedOutCode: ignore PHPCS annotations + minor other fixes

jrfnl:
Another one in the series to make sure sniffs handle the new PHPCS annotations correctly.

PHPCS annotations in commented out code would be tokenized and would throw the "commented out" % off, making the sniff underreport commented out code.

This PR fixes that.

While reviewing the sniff to figure out how to fix this in the best way, I came across numerous other bugs and inefficiencies and have fixed each one in a separate commit.

Includes unit tests covering the bugs.

The PR looks bigger than it is. It will be easiest to review by looking at each commit separately.

-------
### Bug fix (minor): Prevent parsing out valid code.

The regex used to parse out comment fencing was a little too overactive and would change the below commented out code:
```php
//eval('$var = 4;');
```
to
```php
//eval('$var - ;');
```

Now only real comment fencing for which this regex was intended will be replaced.

### Bug fix: Treat multi-line indented `//` comments as one

Multi-line comments are intended to be treated as one comment by this sniff.

Indented `//` style comments tokenize as `T_WHITESPACE` + `T_COMMENT`.
This meant that the compare for multi-line comments would fail for these and that each comment would be analyzed separately instead, causing false negatives.

Includes unit test.

### Efficiency fix (minor) - don't re-tokenize empty comments

Empty comments don't need to be re-tokenized and examined, so the sniff can bow out early in that case.

Covered by the existing unit tests on line 41 and 43 of the test case file.

### Fix two unit tests

As the sniff doesn't listen anymore to docblock comment tokens since 638f98d, these unit tests were no longer covering anything.

Changing these to `/* */` style comments re-activates the unit tests. This does change the result of the analysis for one of these two comments.

### Bug fix: ignore PHPCS annotations in comment blocks

The new PHPCS annotations could throw off the detection of commented out code.
Removing them before the comment is re-tokenized fixes that.

### DIFF
--- a/src/Standards/Squiz/Sniffs/PHP/CommentedOutCodeSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/CommentedOutCodeSniff.php
@@ -160,15 +160,6 @@ class CommentedOutCodeSniff implements Sniff
 
         ini_set('error_reporting', $oldErrors);
 
-        $emptyTokens = [
-            T_WHITESPACE              => true,
-            T_STRING                  => true,
-            T_STRING_CONCAT           => true,
-            T_ENCAPSED_AND_WHITESPACE => true,
-            T_NONE                    => true,
-            T_COMMENT                 => true,
-        ];
-
         $numTokens = count($stringTokens);
 
         /*
@@ -197,6 +188,15 @@ class CommentedOutCodeSniff implements Sniff
         ) {
             return;
         }
+
+        $emptyTokens = [
+            T_WHITESPACE              => true,
+            T_STRING                  => true,
+            T_STRING_CONCAT           => true,
+            T_ENCAPSED_AND_WHITESPACE => true,
+            T_NONE                    => true,
+            T_COMMENT                 => true,
+        ];
 
         $numComment  = 0;
         $numPossible = 0;
--- a/src/Standards/Squiz/Sniffs/PHP/CommentedOutCodeSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/CommentedOutCodeSniff.php
@@ -122,6 +122,9 @@ class CommentedOutCodeSniff implements Sniff
                 $tokenContent = substr($tokenContent, 1);
             }
 
+            // Remove PHPCS annotations.
+            $tokenContent = preg_replace('`[@]?(?:phpcs:(?:enable|disable|set|ignore)|codingStandards(?:Ignore|Change)).*$`', '', $tokenContent);
+
             $content     .= $tokenContent.$phpcsFile->eolChar;
             $lastLineSeen = $tokens[$i]['line'];
         }//end for
--- a/src/Standards/Squiz/Sniffs/PHP/CommentedOutCodeSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/CommentedOutCodeSniff.php
@@ -78,11 +78,7 @@ class CommentedOutCodeSniff implements Sniff
             return;
         }
 
-        $content = '';
-        if ($phpcsFile->tokenizerType === 'PHP') {
-            $content = '<?php ';
-        }
-
+        $content      = '';
         $lastLineSeen = $tokens[$stackPtr]['line'];
         for ($i = $stackPtr; $i < $phpcsFile->numTokens; $i++) {
             if ($tokens[$i]['code'] === T_WHITESPACE) {
@@ -130,12 +126,6 @@ class CommentedOutCodeSniff implements Sniff
             $lastLineSeen = $tokens[$i]['line'];
         }//end for
 
-        $content = trim($content);
-
-        if ($phpcsFile->tokenizerType === 'PHP') {
-            $content .= ' ?>';
-        }
-
         // Quite a few comments use multiple dashes, equals signs etc
         // to frame comments and licence headers.
         $content = preg_replace('/[-=#*]{2,}/', '-', $content);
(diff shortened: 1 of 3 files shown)


gsherwood:
How did the annotations throw off the detection of commented out code? I couldn't figure that out from the unit tests.

I'm also a bit worried about forgetting to maintain that annotation regex as I change things, so I'd rather do filtering using the new token types, but I'm just not sure how to get the sniff to fail to see if that works.

gsherwood:
Forgot to say that the rest all looks good - thanks for that.

jrfnl:
> How did the annotations throw off the detection of commented out code? I couldn't figure that out from the unit tests.

This unit tests demonstrates it:

https://github.com/jrfnl/PHP_CodeSniffer/blob/348d66143f05a871f857b29850f6847a4824c650/src/Standards/Squiz/Tests/PHP/CommentedOutCodeUnitTest.inc#L97-L107

When the sniff would be run as-is (i.e. just check out the unit test commit), it will not be recognized as commented out code as the `$percentCode` would be 29%.
Stripping out the PHPCS annotations, fixes that and the code percentage will be 46%.

In detail for that particular unit test:

Commit | `$numCode` | `$numComment` | `$numPossible` | `$numTokens` | `$percentCode`
-- | -- | -- | -- | -- | --
Unit test commit | 21 | 50 | 5 | 75 | 29%
After annotation correction commit | 19 | 23 | 1 | 42 | 46%


> I'm also a bit worried about forgetting to maintain that annotation regex as I change things, so I'd rather do filtering using the new token types, but I'm just not sure how to get the sniff to fail to see if that works.

The problem with that is, that those tokens aren't tokenized as PHPCS annotations. 

In the initial file tokenizing, the comment is a `/* */` style comment and the annotations use `//` comment style within that, so they are just tokenized as normal lines in the comments.

That goes for both of the unit tests with the annotations:
```
Process token 125 on line 97 [col:1;len:2;lvl:0;]: T_COMMENT => /*\n
Process token 126 on line 98 [col:1;len:29;lvl:0;]: T_COMMENT => function myFunction( $param )\n
Process token 127 on line 99 [col:1;len:1;lvl:0;]: T_COMMENT => {\n
Process token 128 on line 100 [col:1;len:57;lvl:0;]: T_COMMENT => \t// phpcs:disable Standard.Category.Sniff -- for reasons.\n
Process token 129 on line 101 [col:1;len:45;lvl:0;]: T_COMMENT => \tif ( preg_match( '`[abc]`', $param ) > 0 ) {\n
Process token 130 on line 102 [col:1;len:17;lvl:0;]: T_COMMENT => \t\tdo_something();\n
Process token 131 on line 103 [col:1;len:2;lvl:0;]: T_COMMENT => \t}\n
Process token 132 on line 104 [col:1;len:56;lvl:0;]: T_COMMENT => \t// phpcs:enable Standard.Category.Sniff -- for reasons.\n
Process token 133 on line 105 [col:1;len:0;lvl:0;]: T_COMMENT => \n
Process token 134 on line 106 [col:1;len:19;lvl:0;]: T_COMMENT => }//end myFunction()\n
Process token 135 on line 107 [col:1;len:2;lvl:0;]: T_COMMENT => */
Process token 136 on line 107 [col:3;len:0;lvl:0;]: T_WHITESPACE => \n
Process token 137 on line 108 [col:1;len:0;lvl:0;]: T_WHITESPACE => \n
Process token 138 on line 109 [col:1;len:1;lvl:0;]: T_WHITESPACE => \t
Process token 139 on line 109 [col:2;len:2;lvl:0;]: T_COMMENT => /*\n
Process token 140 on line 110 [col:1;len:90;lvl:0;]: T_COMMENT => \t * function myFunction($param ) // @phpcs:ignore Standard.Category.Sniff -- for reasons.\n
Process token 141 on line 111 [col:1;len:5;lvl:0;]: T_COMMENT => \t * {\n
Process token 142 on line 112 [col:1;len:3;lvl:0;]: T_COMMENT => \t *\n
Process token 143 on line 113 [col:1;len:23;lvl:0;]: T_COMMENT => \t * }//end myFunction()\n
Process token 144 on line 114 [col:1;len:4;lvl:0;]: T_COMMENT => \t */
Process token 145 on line 114 [col:5;len:0;lvl:0;]: T_WHITESPACE => \n
```

Then when they get re-tokenized, the second unit test with annotations does tokenize the (end) annotation comment correctly, but the first unit test with annotations will have had the `//` stripped from the beginning already before the snippet gets retokenized and will now tokenize as a `goto` label with a number of `T_STRING`s after it.

I could change [the part of the sniff where the comments get prepared for retokenization](https://github.com/squizlabs/PHP_CodeSniffer/blob/35cd2080abbac35ba31eeaddf3f01a9add2106b2/src/Standards/Squiz/Sniffs/PHP/CommentedOutCodeSniff.php#L78-L115) to take the comment style into account and only strip off `//` and `#` at the start of a line when it's not a `/* */` style comment.

That way the annotations would get correctly tokenized when the snippets get retokenized and I could use the token type to discard them.

@gsherwood Does this explain it better ? What do you think ?

jrfnl:
@gsherwood I've added two new commits to the PR.

The first one addresses your concerns regarding the regex and implements what I suggest in my previous comment.

The second one addresses another potential case involving PHPCS annotations in multi-line `//` style comments which was not yet correctly handled. Instead of being seen as one comment block, these lines would be treated as stand-alone comments. It includes a new unit test demonstrating the issue.

gsherwood:
I'm an idiot -_-

I didn't see that the entire function was commented out, which is also why I thought the token types could be used. Thanks for taking the time to explain it all out. I'll take a look at your suggested change to detect the inner-most annotations.

gsherwood:
Thanks for those changes. All looks good.

PR #1842, opened 2018-01-05
Generic/Todo & Fixme: minor efficiency fix

jrfnl:
As "errors" in the PHPCS annotations will not be reported anyway, no need to examine them for these sniffs.


gsherwood:
Makes sense :) Thanks.

PR #1843, opened 2018-01-05
Generic/UpperCaseConstantName: bug fix - ignore PHPCS annotation tokens

jrfnl:

Includes unit test.

This fixes a bug where non-uppercase constants defined using the `const` keyword would not be reported.

It also fixes some outdated code. Looking at the history, originally this sniff also tried to detect _usage_ of non-uppercase constants.
As of commit 4af13118b1fedd89faadd78b9bccf858ce21dbc9 / October 2013, this is no longer the case and only definition of non-uppercase constants is reported.
It looks like both the sniff as well as the unit test case file still contains quite some code related to the old behaviour which could be further simplified, but that is outside the scope of this PR.


gsherwood:
Thanks for this.

I'm not sure why I didn't remove that old code when constant usage detection was removed, but I've done it now. Thanks for pointing that out as well.

PR #1846, opened 2018-01-08
Support for @inheritdoc with and without curly brackets.

bjdelange:


gsherwood:
Sorry, I've just gotten around to reviewing this.

This PR looks to just detect the use of @inheritdoc and then skip all checks. From reading the docs, my understanding is that the use of this tag does not exclude you from overriding the descriptions and other tags of a class/property/method etc. It can even be used to just import descriptions, as per their example:
```php
/**
 * Makes chocolate bars
 *
 * There are two aspects to this class.
 * {@inheritdoc }  In addition, the foo class
 * makes the bars chocolate
 */
class foo extends bar
{
}
```
So this PR isn't actually adding support for the @inheritdoc tag. It's making an assumption that using it means you don't want the content of your comments checked in any way, which seems like it would be a real problem.

If I'm misunderstanding, can you please provide some more info for me. Otherwise I won't be able to merge this in because it would exclude valid comments from being checked.

gsherwood:
Closing due to no feedback and that I can't accept this PR as-is.

PR #1848, opened 2018-01-11
MultipleStatementAlignment: do not break on comments within arrays.

jrfnl:
Multi-line arrays interspersed with comments were treated differently than multi-line arrays without comments.

This aligns the behaviour of both to be the same.

Includes unit tests.


jrfnl:
Rebased for merge conflict with commit f65239d7f8be86fac1917304c49f53559ba46a9d which basically does the same as this PR does for arrays, but then for closures and anonymous classes.

jrfnl:
@gsherwood I believe this one should go in the `3.2.3` release as it addresses an issue with the new annotations (though the same issue existed before then for comments and as reported in #1870 for closures and the likes).

gsherwood:
Thanks

PR #1852, opened 2018-01-12
Travis: speed up build times by disabling Xdebug

jrfnl:
As suggested by @johnbillion in https://johnblackbourn.com/reducing-travis-ci-build-times-for-wordpress-projects/ , when not creating code coverage reports, Xdebug is not needed and disabling it will speed up the build.

Based on the build for this PR, the difference is approximately 30% (faster).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ matrix:
     - php: hhvm
 
 before_install:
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks a lot

PR #1853, opened 2018-01-14
Travis/build: validate composer.json file

jrfnl:
Similar to the PEAR `package.xml` file being validated, it would be a good safeguard to validate the `composer.json` file when doing the build testing on Travis.


gsherwood:
Good idea. Thanks.

jrfnl:
Looks like this got merged, but GH is not picking up on. Close it ?

PR #1860, opened 2018-01-20
Travis: More stable solution for removing Xdebug when not needed

jrfnl:
Builds onto #1852

As per https://twitter.com/kelunik/status/954242454676475904

When newer images of PHP versions become available, Xdebug isn't always installed.
Using this little titbit, the builds won't break because of it.


gsherwood:
Thanks.

PR #1865, opened 2018-01-27
Add function name on missing function doc comment

cixtor:
### Old Error Messages

```
10 | ERROR | [ ] Missing function doc comment
20 | ERROR | [ ] Missing function doc comment
30 | ERROR | [ ] Missing function doc comment
```

### New Error Messages

```
10 | ERROR | [ ] Missing function doc comment: lorem()
20 | ERROR | [ ] Missing function doc comment: ipsum()
30 | ERROR | [ ] Missing function doc comment: dolor()
```

### Sample File

```
<?php

class Foo
{
    public function lorem()
    {
    }

    private function ipsum()
    {
    }
}

function dolor()
{
}
```

### DIFF
--- a/src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
@@ -61,7 +61,7 @@ class FunctionCommentSniff implements Sniff
         ) {
             $function = $phpcsFile->getDeclarationName($stackPtr);
             $phpcsFile->addError(
-                'Missing docblock for function "%s()"',
+                'Missing function doc comment for function "%s()"',
                 $stackPtr,
                 'Missing',
                 [$function]
--- a/src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
@@ -59,7 +59,8 @@ class FunctionCommentSniff implements Sniff
         if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG
             && $tokens[$commentEnd]['code'] !== T_COMMENT
         ) {
-            $phpcsFile->addError('Missing function doc comment', $stackPtr, 'Missing');
+            $function = $phpcsFile->getDeclarationName($stackPtr);
+            $phpcsFile->addError('Missing function doc comment: '.$function.'()', $stackPtr, 'Missing');
             $phpcsFile->recordMetric($stackPtr, 'Function has doc comment', 'no');
             return;
         } else {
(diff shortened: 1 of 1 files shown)


src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -59,7 +59,8 @@ public function process(File $phpcsFile, $stackPtr)
         if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG
             && $tokens[$commentEnd]['code'] !== T_COMMENT
         ) {
-            $phpcsFile->addError('Missing function doc comment', $stackPtr, 'Missing');
+            $function = $phpcsFile->getDeclarationName($stackPtr);
+            $phpcsFile->addError('Missing function doc comment: '.$function.'()', $stackPtr, 'Missing');
```
jrfnl: 1. IMHO, the error message would read more fluently as `Missing function doc comment for function "%s()"`

2. PHPCS offer a `$data` parameter to the `addError()` method to pass in arbitrary values. That should be used instead of concatenating the error message.

Refs:
http://pear.php.net/package/PHP_CodeSniffer/docs/3.2.2/apidoc/PHP_CodeSniffer/File.html#methodaddError

cixtor: I think saying _"function"_ twice is redundant and also makes the line unnecessarily long:

```
123 | ERROR   | [ ] Missing function doc comment for function "loremIpsumDolor()"
```

The 80-char-limit can be kept with any of these two options:

```
123 | ERROR   | [ ] Missing docblock for function "loremIpsumDolor()"
123 | ERROR   | [ ] Missing function docblock for "loremIpsumDolor()"
```

I have updated the pull-request with the first one, let me know if you agree.

jrfnl: :+1: Only thing which comes to mind is that, to stay in line with other similar messages, it should maybe read "Missing doc comment for function ....". What do you think ?

cixtor: When the DocBlock is missing from the File and Class, these are the messages:

```
2 | ERROR | Missing file doc comment
3 | ERROR | Missing class doc comment
```

So I guess your suggestion makes sense, I will change it to:

```
4 | ERROR | Missing function doc comment for function "lorem()"
```

So the messages for the sample file that I mentioned before, will return this:

```
  2 | ERROR | Missing file doc comment
  3 | ERROR | Missing class doc comment
  5 | ERROR | Missing function doc comment for function "lorem()"
  9 | ERROR | Missing function doc comment for function "ipsum()"
 14 | ERROR | Missing function doc comment for function "dolor()"
```

jrfnl: I only meant to highlight `docblock` vs `doc comment`. I agree with your earlier remark saying that "mentioning the word function twice is redundant".

gsherwood:
Thanks for this change. I tweaked the message just a little and arrived at `Missing doc comment for function foo()`

PR #1873, opened 2018-02-02
Add standards path resolution for files served over http(s).

### LINKED ISSUE #1872: Standard/Ruleset over HTTP
It would be nice to be able to load a ruleset over HTTP(S). For example, if I wanted everyone in my organization to follow certain coding standards, I could serve up a file at `https://www.myorganization.com/phpcs/ruleset.xml`. Then developers in my organization could use that file like:

```
$ phpcs --standard=https://www.myorganization.com/phpcs/ruleset.xml file.php
```

Right now if you do this you get a `the coding standard is not installed` error.

Are there any reservations about this? If not, I'll go ahead and begin working on a pull request.

smsalisbury:
`Common::realpath()` passes through paths with an http(s) protocol.
`Standards::isInstalledStandard()` checks if the file exists and is an xml file for standards with an http(s) protocol.

This should resolve #1872 

I did not see any unit tests associated with these methods, so I didn't add any testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .idea/*
 /vendor/
 composer.lock
+/src/Standards/HttpCache/
--- a/src/Ruleset.php
+++ b/src/Ruleset.php
@@ -154,7 +154,35 @@ class Ruleset
         }
 
         foreach ($standardPaths as $standard) {
-            $ruleset = @simplexml_load_string(file_get_contents($standard));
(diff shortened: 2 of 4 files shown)


gsherwood:
Thanks for your work on this, but I've decided that this isn't a feature I want to support in PHPCS at this time, so I'm going to close the PR and related issue. I haven't received any additional requests for this functionality and I don't want to open PHPCS up to any additional security issues.

Issue #1756, opened 2017-11-22
Documentation for testing custom sniffs?


rmccue:
From what I can see, the bootstrap and AllSniffs tests can be used for third-party custom sniffs, but there's no documentation on how to actually run these with custom standards (apart from a [reference in the 3.1.0 release](https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.1.0)). Would be very useful to have a wiki page on how to set this up. :)

louisl:
This is how I do it since 3.1.0,


`cd /Path/To/YourStandardProjectFolder`

`composer install squizlabs/php_codesniffer`
`composer install phpunit/phpunit`

Verify PHP_Codesniffer works.

`./vendor/bin/phpcs -i`

 Check files match the PHPCS standard.

`./vendor/bin/phpcs --ignore=*/Tests/* ./MyStandard/ --standard=./vendor/squizlabs/php_codesniffer/phpcs.xml.dist`

Set installed standard to MyStandard.

`./vendor/bin/phpcs --config-set installed_paths /Path/To/YourStandardProjectFolder/MyStandard`

Verify the MyStandard standard is installed.

`./vendor/bin/phpcs -i`

Run unit tests.

`./vendor/bin/phpunit --debug --filter MyStandard`

sirbrillig:
I've found it much easier and more flexible to write custom tests rather than using the `AbstractSniffUnitTest`. Something like the following (the boilerplate can be abstracted into a base test case or a helper).

```php
```
use PHPUnit\Framework\TestCase;
use PHP_CodeSniffer\Files\LocalFile;
use PHP_CodeSniffer\Ruleset;
use PHP_CodeSniffer\Config;

class DisallowExtractSniffTest extends TestCase {
	public function testDisallowExtractSniff() {
		$fixtureFile = __DIR__ . '/fixture.php';
		$sniffFiles = [__DIR__ . '/../../../StandardName/Sniffs/Extract/DisallowExtractSniff.php'];
		$config = new Config();
		$ruleset = new Ruleset($config);
		$ruleset->registerSniffs($sniffFiles, [], []);
		$ruleset->populateTokenListeners();
		$phpcsFile = new LocalFile($fixtureFile, $ruleset, $config);
		$phpcsFile->process();
		$foundErrors = $phpcsFile->getErrors();
		$lines = array_keys($foundErrors);
		$this->assertEquals([7], $lines);
	}
}
```

[...]

PR #1877, opened 2018-02-07
Improve testability of rulesets

timoschinkel:
By extracting the retrieval of the installed standards a third party ruleset/sniff developer, me, can easily create an extension of `AllSniffs` that only runs the sniffs of the ruleset in question.

This simple extraction will prevent doing cli magic like `--filter` and allow full configurability in a `phpunit.xml`.

Adds to #1756


gsherwood:
Thanks for the PR. If it helps you test custom standards, I'm very happy.

PR #1883, opened 2018-02-07
Squiz/ValidVariableName: don't hide one message behind another

### LINKED ISSUE #1880: ValidVariableName doesn't test a private member when PrivateNoUnderscore is excluded
Private members are only tested when the name contains a leading underscore.
```xml
<rule ref="Squiz.NamingConventions.ValidVariableName">
    <exclude name="Squiz.NamingConventions.ValidVariableName.PrivateNoUnderscore" />
</rule>
```
```php
class Test
{
    private $private_member;
}
```

jrfnl:
No reason not to throw both a message about a leading underscore as well as about camelCaps.

Fixes #1880

-----

[Edit]: I've added a second commit which basically does the same for the `Zend.NamingConventions.ValidVariableName` sniff which was suffering from the same issue.

~~Working on fixing a few more sniffs suffering from the same.~~ All done for now.


gsherwood:
Thanks a lot for this. Saved me a lot of time.

jrfnl:
> Saved me a lot of time.

@gsherwood Open source software is every users responsibility ;-)

PR #1887, opened 2018-02-08
use json_encode in JSON reports to fix invalid outputs

marvasDE:
I had issues with a sniffer, which use class names in the source. These class names had (not escaped) namespaces included. This make the json output invalid.

After looking at the code, I was wondering, why PHP_CodeSniffer doesn't use json_encode for this task.

### DIFF
--- a/src/Reports/Json.php
+++ b/src/Reports/Json.php
@@ -59,7 +59,7 @@ class Json implements Report
                     $messagesObject->column = $column;
                     $messagesObject->fixable = $fixable;
 
-                    $messages .= json_encode($messagesObject);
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for the PR. PHPCS used to use json_encode but it was changed some time in the past due to older PHP version compatibility (5.1). But that's not an issue any more, so I've merged this in to restore that.

I was getting double encoding due to the manual encoding of error messages, so I had to remove that. I was also getting the fixable flag coming out as a string instead of a boolean, so I changed that as well. Other than those things, the change looks to work well.

marvasDE:
Thank you for merging :-)

PR #1888, opened 2018-02-08
AbstractVariableSniff: add list of PHP reserved variables

jrfnl:
Duplicate code removal.

The list of PHP reserved variables was up to now defined in four places in two separate sniffs.
Both these sniffs extend the `AbstractVariableSniff`, so it makes sense to define the list as a property of that sniff and to re-use it in the individual sniffs instead.

Also changes the array format in a way that the more efficient `isset()` can be used instead of `in_array()`.


gsherwood:
Thanks for doing this.

PR #1893, opened 2018-02-10
PSR2/UseDeclaration: allow comments after the last USE statement

### LINKED ISSUE #1891: Comment on last USE statement causes false positive for PSR2.Namespaces.UseDeclaration.SpaceAfterLastUse
# Overview
If the last USE statement has a comment, an incorrect `PSR2.Namespaces.UseDeclaration.SpaceAfterLastUse` error is reported.

# Steps to reproduce
Run `phpcs --standard=PSR2` on a file with the following contents:
```php
<?php
use Foo\Bar; //comment

// This should pass.
```

## Expected output
None. This should validate correctly.

## Actual output
```
FILE: /Users/matt/Projects/PHP_CodeSniffer/tests/TestCommentAfterLastUse.php
-------------------------------------------------------------------------------------
-------------------------------------------------------------------------------------
 2 | ERROR | [x] There must be one blank line after the last USE statement; 0 found;
-------------------------------------------------------------------------------------
-------------------------------------------------------------------------------------

Time: 114ms; Memory: 6Mb
```

iammattcoleman:
Resolves #1891

### DIFF
--- a/src/Standards/PSR2/Sniffs/Namespaces/UseDeclarationSniff.php
+++ b/src/Standards/PSR2/Sniffs/Namespaces/UseDeclarationSniff.php
@@ -142,7 +142,7 @@ class UseDeclarationSniff implements Sniff
             return;
         }
 
-        $next = $phpcsFile->findNext(T_WHITESPACE, ($end + 1), null, true);
+        $next = $phpcsFile->findNext(Tokens::$emptyTokens, ($end + 1), null, true);
 
         if ($next === false || $tokens[$next]['code'] === T_CLOSE_TAG) {
             return;
--- a/src/Standards/PSR2/Tests/Namespaces/UseDeclarationUnitTest.8.inc
+++ b/src/Standards/PSR2/Tests/Namespaces/UseDeclarationUnitTest.8.inc
@@ -1,0 +1,4 @@
+<?php
+use Foo\Bar; //comment
+
+// This should pass.


src/Standards/PSR2/Sniffs/Namespaces/UseDeclarationSniff.php
```diff
@@ -142,7 +142,7 @@ public function process(File $phpcsFile, $stackPtr)
             return;
         }
 
-        $next = $phpcsFile->findNext(T_WHITESPACE, ($end + 1), null, true);
+        $next = $phpcsFile->findNext([T_WHITESPACE, T_PHPCS_IGNORE, T_COMMENT], ($end + 1), null, true);
```
jrfnl: You may want to use `Tokens::$EmptyTokens` instead for completeness.

iammattcoleman: Thanks. Updated!

gsherwood:
Thanks a lot for fixing this, and for the test.

dhensby:
This appears to have caused a regression, please see #1898

dhensby:
OK - the regression here is that this allows you to do:

```php
use Class; // comment
```

but it now causes this to fail:

```php
use Class;

//comment
```

Because the test fixture tried to evaluate both at once it *appeared* to fix the problem, but it actually caused a regression instead

PR #1900, opened 2018-02-15
Allow PHPUnit 7.0 in composer.json

webimpress:


gsherwood:
I only just managed to support PHPUnit 7 in master, so you'll need master (or eventually 3.2.3) to have PHPCS work with v7.

Thanks for the PR.

PR #1901, opened 2018-02-17
Fixed PHPCS annotations in multi-line tab-indented comments + not ignoring whole line for phpcs:set

jrfnl:
This PR should go into `3.2.3` as it fixes two bugs in how the tokenizer deals with the new PHPCS annotations.

This PR will be easiest to review by looking at each commit individually.

**Bug 1**:
PHPCS annotations in tab-indented multi-line block comments were not recognized nor tokenized correctly.

**Bug 2**:
A `phpcs:set` would always cause the whole line to be ignored, even when its used as an end-comment.

As `phpcs:set` annotations are not covered in the unit test file, I wasn't sure were to add the tests for this.

**Other**:
While staring at the code I saw a few minor other optimizations/typos, those have been fixed in separate commits.

### DIFF
--- a/src/Tokenizers/Tokenizer.php
+++ b/src/Tokenizers/Tokenizer.php
@@ -320,8 +320,11 @@ abstract class Tokenizer
                     }
 
                     if (substr($commentTextLower, 0, 9) === 'phpcs:set') {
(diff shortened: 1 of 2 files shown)


gsherwood:
Thanks for both of these fixes

PR #1904, opened 2018-02-17
PEAR/MultiLineCondition: bug fix - deal correctly with PHPCS annotations

jrfnl:
Another one in the series to fix the sniffs to deal correctly with the new PHPCS annotations. Three more sniffs to go.

This PR should be tagged for the `3.2.3` release,

To review the PR, please look at the individual commits.

This PR fixes three issues in the `PEAR.ControlStructures.MultiLineCondition` sniff when it encountered PHPCS annotations within multi-line conditions:
1. It didn't allow for trailing PHPCS annotations.
2. When PHPCS annotations were encountered within a multi-line condition, it could occur that the same `StartWithBoolean` error would be thrown twice for the same line.
3. When a condition was spread out across multiple lines interspersed with comments and/or PHPCS annotations, the fixer would remove the comments.

The PR can be tested by checking out the first commit (unit tests) and running both `phpcs` as well as `phpcbf` against it to see the identified bugs in practice.


gsherwood:
Thank you.

Issue #1905, opened 2018-02-18
Problem when putting another tag before a param tag in a group


phil-davis:
Make a doc block like this:
```
	/**
	 * Some text about this step/method
	 *
	 * @param string $name
	 * @return void
	 * @When I create a file called :name
	 */
```

``phpcs``, for the ``When`` line, spits out the error ``Tag cannot be grouped with parameter tags in a doc comment``
which is correct - if I like, I can suppress this in the ``standards`` settings, and put all the tags without blank lines between each tag-type.

Move the ``When`` tag above the ``param`` tag:
```
	/**
	 * Some text about this step/method
	 *
	 * @When I create a file called :name
	 * @param string $name
	 * @return void
	 */
```

``phpcs``, for the ``param`` line, spits out the error ``Parameter tags must be grouped together in a doc comment``

That is incorrect - the ``param`` lines (in this case only 1 ``param`` line)  are "grouped together". The problem is that there is a non-param tag in the group.

It should spit out, for the ``When`` line, the error ``Tag cannot be grouped with parameter tags in a doc comment``

The code processing this case is not understanding that the non-param tag could appear in a``param`` group before the first ``param`` tag line is found in the group.

phil-davis:
I noticed this because I want to suppress ``Tag cannot be grouped with parameter tags in a doc comment`` and keep reporting ``Parameter tags must be grouped together in a doc comment``

But if I have that combination and write a doc block like the 2nd example above, then I get a fail when it should pass.

gsherwood:
> That is incorrect - the param lines (in this case only 1 param line) are "grouped together". The problem is that there is a non-param tag in the group.

When the sniff says "grouped together" it means there should be a blank line before and after the grouping of param tags. Like this:
```php
/**
 * Some text about this step/method
 *
 * @When I create a file called :name
 *
 * @param string $name
 *
 * @return void
 */
```
That's what the sniff is trying to enforce, so the error is correct.

phil-davis:
I want to suppress ``Tag cannot be grouped with parameter tags in a doc`` so in my ``phpcs.xml`` is:
```
	<exclude name="Generic.Commenting.DocComment.NonParamGroup" />
```
and then I [want|expect] any of the following doc blocks to pass:

Example 1:
```
	/**
	 * Some text about this step/method
	 *
	 * @param string $name
	 * @param string $folder
	 * @return void
	 * @When I create a file called :name in :folder
	 */
```

Example 2:
[...]

PR #1906, opened 2018-02-18
Report correct problem when another tag appears before param tag

phil-davis:
Fixes issue #1905 

A tag group that contains a ``param`` tag could have other tags either before or after it in the group. When we see the first ``param`` tag, there might have already been some non-param tags already processed in the group. 

The existing code, when it came across a ``param`` tag and it was already in "the middle" of processing a tag group that was not a ``param`` group, would complain about ``Parameter tags must be grouped together in a doc comment`` - that was not correct. Actually it can just decide that it has now come across the ``param`` group.

Then the processing to report about ``Tag cannot be grouped with parameter tags in a doc comment`` needs to be done later, during another pass through the tag groups. Then that can detect other tags in a ``param`` tag group occurring either before or after the ``param`` tag(s).


gsherwood:
See discussion on #1905

gsherwood:
I tried this change with this sample file:
```php
<?php
/**
 * Some text about this step/method
 *
 * @see    Bar
 * @param  string $name
 * @see    Foo
 * @param  string $folder
 * @return void
 */
```
Previously, you'd get these errors (I'm only focusing on the param grouping ones):
```
  6 | ERROR   | [ ] Parameter tags must be grouped together in a doc comment (Generic.Commenting.DocComment.ParamGroup)
  7 | ERROR   | [ ] Tag cannot be grouped with parameter tags in a doc comment (Generic.Commenting.DocComment.NonParamGroup)
  7 | ERROR   | [ ] Tags must be grouped together in a doc comment (Generic.Commenting.DocComment.TagsNotGrouped)
  8 | ERROR   | [ ] Tags must be grouped together in a doc comment (Generic.Commenting.DocComment.TagsNotGrouped)
  9 | ERROR   | [ ] Tag cannot be grouped with parameter tags in a doc comment (Generic.Commenting.DocComment.NonParamGroup)
```
With this change, I now get:
```
  5 | ERROR   | [ ] Tag cannot be grouped with parameter tags in a doc comment (Generic.Commenting.DocComment.NonParamGroup)
  7 | ERROR   | [ ] Tag cannot be grouped with parameter tags in a doc comment (Generic.Commenting.DocComment.NonParamGroup)
  7 | ERROR   | [ ] Tags must be grouped together in a doc comment (Generic.Commenting.DocComment.TagsNotGrouped)
  8 | ERROR   | [ ] Tags must be grouped together in a doc comment (Generic.Commenting.DocComment.TagsNotGrouped)
  9 | ERROR   | [ ] Tag cannot be grouped with parameter tags in a doc comment (Generic.Commenting.DocComment.NonParamGroup)
````
So you end up losing the `Parameter tags must be grouped together in a doc comment` error message, which is the one I know you want to keep here. You do retain the `Tags must be grouped together in a doc comment` error for the second `@param` tag, but I'm wondering if losing that first error message is the desired outcome of this change.

**Comment by phil-davis** [comment]
That looks good to me. In my mind, ``Parameter tags must be grouped together in a doc comment`` is about seeing all the ``param`` tags in a single block of tags (in whatever mess of an order like your example. Then withing the block that has the ``param`` tags, there can be the other problems:

``Tags must be grouped together in a doc comment`` - which applies equally well to the ``param`` tag or to other tags, like ``see`` in your example. Each tag name (``see``, ``param``, ``return`` in the example) must be be grouped together.

``Tag cannot be grouped with parameter tags in a doc comment`` - which applies to all the non-``param`` tags that appear wherever in a block of tags that contains any ``param`` tag.

So I can suppress ``Tag cannot be grouped with parameter tags in a doc comment`` and get:

```
  7 | ERROR   | [ ] Tags must be grouped together in a doc comment (Generic.Commenting.DocComment.TagsNotGrouped)
  8 | ERROR   | [ ] Tags must be grouped together in a doc comment (Generic.Commenting.DocComment.TagsNotGrouped)
```

and then I can sort the tags out "properly":
```
<?php
/**
 * Some text about this step/method
 *
 * @see    Bar
 * @see    Foo
 * @param  string $name
 * @param  string $folder
 * @return void
 */
```
and (for me) this doc-block will now pass.

phil-davis:
I added a test case also.

gsherwood:
Thanks for the feedback and test case. This change will be in the 3.3.0 release.

PR #1914, opened 2018-02-26
CommentedOutCode: ignore small comments and annotations from other tools

### LINKED ISSUE #1910: False positives for Squiz.PHP.CommentedOutCode.Found in 3.2.3
When updating to 3.2.3, I noticed several new warnings for `Squiz.PHP.CommentedOutCode.Found`. Looking at the code in question, those were all triggered by lines like this:

```php
return false; // @codeCoverageIgnore
```

While `@` could be construed as the error suppression operator, annotations like this should be ignored by the sniff, as the were in previous versions up to 3.2.2.

### LINKED ISSUE #1912: Invalid "commented out code" warnings after upgrading to 3.2.3
Updated quizlabs/php_codesniffer updated from 3.2.2 to 3.2.3

Getting a lot of: "This comment is 50% valid code; is this commented out code?" warnings on things I don't consider code. 

Here are some examples of the lines it's complaining about:

```
// Ok!
/* Go! */
// ISO-639-3
````

I disagree?

jrfnl:
This implements option 3 and 5 as discussed in #1910:
* Ignore comments which only consist of a silence operator + one alphanumeric word.
    This will prevent false positives on most (not all) ignore annotations from other tools.
* Ignore comments which result in only two or less non-whitespace tokens.
    This will prevent false positives on `// Ok!` type of comments.

    To implement this, I've elected to remove the open/close tags from the token array of the retokenized comment to prevent having to make the calculations more complicated.

Fixes #1910
Fixes #1912

### DIFF
--- a/src/Standards/Squiz/Sniffs/PHP/CommentedOutCodeSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/CommentedOutCodeSniff.php
@@ -129,6 +129,11 @@ class CommentedOutCodeSniff implements Sniff
             $lastCommentBlockToken = $i;
         }//end for
 
+        // Ignore typical warning suppression annotations from other tools.
+        if (preg_match('`^\s*@[A-Za-z]+\s*$`', $content) === 1) {
+            return ($lastCommentBlockToken + 1);
+        }
+
         // Quite a few comments use multiple dashes, equals signs etc
         // to frame comments and licence headers.
         $content = preg_replace('/[-=#*]{2,}/', '-', $content);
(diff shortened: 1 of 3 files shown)


gsherwood:
You're a dead set legend. Thanks for doing this.

jrfnl:
You're welcome :blush:

jrfnl:
@gsherwood This merged PR is missing the milestone (3.3.0). Should it be added ?
Looking at the list of recently merged PRs, there are a few more missing a milestone.

gsherwood:
> This merged PR is missing the milestone (3.3.0). Should it be added ?

Yes, thanks. I fixed the others too.

PR #1923, opened 2018-03-06
Fix fatal 'class already declared' errors when using external standards

jrfnl:
I've seen similar issues now in several external standards, where people are receiving a `Cannot declare class Standard\Sniffs\Category\SomethingSniff, because the name is already in use` error.

I've not been able to reproduce the issues, but noticed that - at least in some of the cases - this involved external sniffs which extend other external sniffs and use a `use` statement for the parent sniff.

I suspect that the fact that I cannot reproduce this, may be due to sniffs not loading in a predefined order, but depending on when the directive to include the sniff/standard is encountered in the ruleset.

I imagine that if a child sniff is included first and a parent sniff second, this causes the fatal error (though I haven't been able to reproduce it that way either).

Using `include_once` instead of `include` for typical class based files should solve this problem.

To that end, I've reviewed all uses of `include/require` in PHPCS and the three addressed in this PR are the result of that review.

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -76,7 +76,7 @@ if (class_exists('PHP_CodeSniffer\Autoload', false) === false) {
                 if (strpos(__DIR__, 'phar://') !== 0
                     && file_exists(__DIR__.'/../../autoload.php') === true
                 ) {
-                    self::$composerAutoloader = include __DIR__.'/../../autoload.php';
+                    self::$composerAutoloader = include_once __DIR__.'/../../autoload.php';
                     if (self::$composerAutoloader instanceof \Composer\Autoload\ClassLoader) {
                         self::$composerAutoloader->unregister();
                         self::$composerAutoloader->register();
@@ -164,7 +164,7 @@ if (class_exists('PHP_CodeSniffer\Autoload', false) === false) {
             $interfaces = get_declared_interfaces();
             $traits     = get_declared_traits();
 
-            include $path;
+            include_once $path;
 
             $className  = null;
             $newClasses = array_reverse(array_diff(get_declared_classes(), $classes));
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -313,7 +313,7 @@ class Runner
 
         // Include bootstrap files.
         foreach ($this->config->bootstrap as $bootstrap) {
-            include $bootstrap;
+            include_once $bootstrap;
         }
 
         if ($this->config->stdin === true) {


gsherwood:
I merged this, but have now reverted it. I missed the change of include to include_once in the load() method when I reviewed it. Sorry about that.

If PHPCS attempts to load a class and it has never loaded it before, then it does an include_once, it's not going to find the proper class/interface/trait name and will probably set the value to NULL. That will then mean that anywhere in PHPCS where someone tries to get the name of the class, they'll get NULL and produce some sort of error.

So I think that all this does is push the error elsewhere. Without being able to reliably replicate the issue and understand it, I don't want to make core changes like this one.

jrfnl:
@gsherwood I understand and that makes sense.

I just don't know how else this could be solved. I haven't been able to reproduce the issue, but I suspect it might have something to do with Composer loading files before PHPCS kicks in.

Bright ideas welcome. Some of the issues I've seen are linked here, so you can see the full conversations about it.

gsherwood:
> I suspect it might have something to do with Composer loading files before PHPCS kicks in

Yes, probably. Although if you run PHPCS directly, this shouldn't happen because the autoloader kicks in so early. But if there is something else loading classes sooner, then PHPCS wont see those classes get loaded and it wont be able to translate sniff codes into namespaced classes. It's really important that PHPCS gets a first chance to load those classes up.

jrfnl:
@gsherwood I know and I get it. I'm just trying to figure out how to help these users running into fatal errors.

CamelKing:
Hi, I am facing the same problem.

Any follow up on this? Thanks.

CamelKing:
Interestingly, add in one line actually "fix" the problem?

	<rule ref="WordPress.Files.FileName">
		<properties>
			<property name="is_theme" value="true"/>                                <---- Add in this line
			<property name="custom_test_class_whitelist" type="array">
				<element value="Titan_TestCase"/>
			</property>
		</properties>
	</rule>

Without that line, running phpcs will crash with error message as per others.
No idea why this work, but sure hope this helps your debugging. Cheers.

jrfnl:
@CamelKing Could you open an issue in the [WordPressCS repo](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/) about this ? Preferably with a backtrace of the error message and test code or a POC repo with which the reproduce the issue ?

That line really shouldn't make a difference, so I'd to take a closer look.

PR #1924, opened 2018-03-06
Generic/MultipleStatementAlignment: fix bug/list syntax before assignment operator

### LINKED ISSUE #1870: Invalid warning in multiple assignments alignment with closure or anon class
The following example is marked as invalid by phpcs
```
1 $foobarbaz = array_map(
2    function ($n) {
3         return $n * $n;
4     },
5     [1, 2, 3]
6 );
7 $foo       = 5;
```
This causes the following warning on line 7:
```
Generic.Formatting.MultipleStatementAlignment.IncorrectWarning
```
Basically, the `$foo` is considered to be a new assignment block. The reason for this, I think, is the semicolon in the function closure on line 3. Phpcs thinks the assignment block has ended because the semicolon is in a different context. 

The reason I think this should not happen is because PHPStorm automatically aligns the equals sign like the example given above. This is because of the code style option `Assignment statement -> Align consecutive assignments` in PHPStorm.

Hence my question whether this can be added/fixed in phpcs?

### LINKED ISSUE #1922: Equal sign alignment check broken when list syntax used before assignment operator
After updating from 3.2.2 to 3.2.3, the code fixer produces the following patches due to the codestyle check failing all of the sudden. It seems to only be when using `[] =` (http://php.net/manual/en/function.list.php)

```diff
      */
     public function invoicesOverviewAction(Request $request, Client $client)
     {
-        $handler = $this->handler_factory->create(InvoicePaymentHandlerType::class);
+        $handler                       = $this->handler_factory->create(InvoicePaymentHandlerType::class);
 
         [$payable_invoices, $invoices] = $this->getInvoices($client);
```

```diff
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $transformer = new ContractTransformer($options['contracts']);
-        $types       = ['support.contact.question' => ContactData::QUESTION];
+        $transformer                                    = new ContractTransformer($options['contracts']);
+        $types                                          = ['support.contact.question' => ContactData::QUESTION];
 
         [$important, $questions, $incidents, $requests] = $this->createContractBuckets($options['contracts']);

```

My best guess is that the blank line is now ignored, while it shouldn't be.

jrfnl:
This simplifies the fixes which were previously committed to fix #1870 and #1848, which were released in `3.2.3`.

Both those fixes try to prevent the sniff getting confused over (potentially) multi-line value assignments, such as closures and arrays.

The simplification removes the previously added code in favour of skipping to the end of a statement as soon as an assignment operator has been found.

This might also make the sniff marginally faster as the tokens between the assignment operator and the end of the statement will no longer be examined.

N.B.: The reason for going one back from the end of the assignment (line 241), is that otherwise the code addressing the issue covered by the unit test on line 220 / added in commit 579d39cac3163bdf05ce1b50a7e470acd452735b would not be triggered, leading to incorrect results.

Fixes #1922

Side-note: I think it would be a good idea to add unit tests for the `File::findEndOfStatement()` method.

### DIFF
--- a/src/Standards/Generic/Sniffs/Formatting/MultipleStatementAlignmentSniff.php
+++ b/src/Standards/Generic/Sniffs/Formatting/MultipleStatementAlignmentSniff.php
@@ -116,30 +116,6 @@ class MultipleStatementAlignmentSniff implements Sniff
 
         for ($assign = $stackPtr; $assign < $phpcsFile->numTokens; $assign++) {
             if (isset($find[$tokens[$assign]['code']]) === false) {
-                if ($tokens[$assign]['code'] === T_CLOSURE
-                    || $tokens[$assign]['code'] === T_ANON_CLASS
-                ) {
-                    $assign   = $tokens[$assign]['scope_closer'];
-                    $lastCode = $assign;
-                    continue;
-                }
-
-                // Skip past the content of arrays.
-                if ($tokens[$assign]['code'] === T_OPEN_SHORT_ARRAY
-                    && isset($tokens[$assign]['bracket_closer']) === true
-                ) {
-                    $assign = $lastCode = $tokens[$assign]['bracket_closer'];
-                    continue;
-                }
-
-                if ($tokens[$assign]['code'] === T_ARRAY
-                    && isset($tokens[$assign]['parenthesis_opener']) === true
-                    && isset($tokens[$tokens[$assign]['parenthesis_opener']]['parenthesis_closer']) === true
-                ) {
-                    $assign = $lastCode = $tokens[$tokens[$assign]['parenthesis_opener']]['parenthesis_closer'];
-                    continue;
-                }
-
                 // A blank line indicates that the assignment block has ended.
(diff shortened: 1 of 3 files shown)


iltar:
Would it be an idea to add this fixture as well, just to show that if there is _no_ blank line, it should still be aligned?
```php
public function buildForm(FormBuilderInterface $builder, array $options)
{
    $transformer                                    = new ContractTransformer($options['contracts']);
    $types                                          = ['support.contact.question' => ContactData::QUESTION];
    [$important, $questions, $incidents, $requests] = $this->createContractBuckets($options['contracts']);
}
```

jrfnl:
> Would it be an idea to add this fixture as well, just to show that if there is no blank line, it should still be aligned?

I've added the additional unit test as a safe-guard.

Just so you know: that test case would not be affected by the fix I've now implemented anyway as the whole "check for arrays" bit has been removed, so this is more a safeguard for the future than that it actually tests something now.

iltar:
Fair enough! :+1:

PR #1930, opened 2018-03-11
Squiz/ControlSignature: fix bug icw trailing comment after closing brace

### LINKED ISSUE #1857: try/catch
PSR 2 should require that `catch` be on the same line as the closing `}` of `try`. If a comment is accidentally added after the closing `}` as below, we don't see an error.

```
try {
    $x = 1;
} // stray comment here
catch (Exception $exception) {
    $x = 2;
}
```

Is this expected behavior?

### LINKED ISSUE #1929: PSR2 reporting the same issue twice/superfluous new ClosingBracePlacement sniff
Commit https://github.com/squizlabs/PHP_CodeSniffer/commit/e1d3a5b097887e8f6eabf1e229bdbe17f9ba7dd8 added a new sniff called `PSR2.ControlStructures.ClosingBracePlacement` in response to issue #1857.

I'm not sure what purpose the new sniff serves. The PSR2 ruleset already includes the `Squiz.ControlStructures.ControlSignature` sniff which reports on the same issue.

AFAICS, the only problem was that that sniff did not correctly handle trailing comments after a brace.

Take the following example code:
```php
if ($something) {
}
else {
}

if ($something) {
}
// comment
else {
}

if ($something) {
} // comment
else {
}
```

This will now result in:
```
------------------------------------------------------------------------------------------
FOUND 5 ERRORS AFFECTING 5 LINES
------------------------------------------------------------------------------------------
  4 | ERROR | [x] Expected 1 space after closing brace; newline found
  5 | ERROR | [x] Control structure keyword must be on the same line as the closing brace
    |       |     from the earlier body
  9 | ERROR | [x] Expected 1 space after closing brace; newline found
 11 | ERROR | [ ] Control structure keyword must be on the same line as the closing brace
    |       |     from the earlier body
 16 | ERROR | [ ] Control structure keyword must be on the same line as the closing brace
    |       |     from the earlier body
```

In other words, all the messages are now thrown in duplicate, except for the test case with a trailing comment behind the closing brace of the `if`.

jrfnl:
This PR addresses the double error messages for the same issue as reported in #1929.

It:
* Removes the new `PSR2.ControlStructures.ClosingBracePlacement` sniff.
* Adds unit tests relating to issue #1857 to the `Squiz.ControlStructures.ControlSignature` sniff.
* Fixes the `Squiz.ControlStructures.ControlSignature` sniff not reporting an error for the secondary keyword not being on the same line as the closing brace of the primary keyword when there is a trailing comment after the closing brace

Additionally:
* The `finally` keyword was previously not taken into account for the "close brace on same line as secondary keyword" check. This has also been fixed.

I've run the adjusted `Squiz.ControlStructures.ControlSignature` sniff as submitted in this PR, over the unit tests which were originally added with the `PSR2.ControlStructures.ClosingBracePlacement` sniff and while the error is reported on a different line (line of close brace vs line of keyword), the same issues *are* being reported and the same fixes are now correctly made.

Fixes #1929 
Fixes #1857


gsherwood:
I have no idea why I ended up writing that new sniff. I remember going to look for an existing one and didn't thought it was being excluded in PSR2 for some reason.

Thanks for fixing this up the right way.

jrfnl:
@gsherwood Thanks for merging this.
I can imagine that it's hard to keep track of all the sniffs. I just noticed it as a build running against master started to break (custom ruleset based on PSR2, but excluding that specific error message).

Issue #1931, opened 2018-03-12
Generic opening brace placement sniffs do not correctly support function return types


jrfnl:
The auto-fixer in the `Generic.Functions.OpeningFunctionBraceKernighanRitchie` sniff currently will break when a function has a return type declaration which is not on the same line as the closing parenthesis of the function declaration.

Example code:
```php
// This works fine:
function myFunction(): Something
{
    return null;
}

// This will create a fixer conflict:
function myFunction($a, $lot, $of, $params)
    : array
{
    return null;
}
```

gsherwood:
I think the solution is to make sure this code doesn't produce an error:
```php
function myFunction($a, $lot, $of, $params)
    : array {
    return null;
}
```
Because the opening brace is technically on the same line as the declaration, but the declaration is split over multiple lines, which is the job of another sniff to check.

That will also make it consistent with this code, which also produces no error:
```php
function myFunction(
    $a,
    $lot,
    $of,
    $params
) {
    return null;
}
```

gsherwood:
Should be fixed now. Thanks for reporting it.

jrfnl:
@gsherwood Thanks for working on this.

Unfortunately, the fix has issues with trailing comments:
```php
function myFunction($a, $lot, $of, $params)
    : array // comment
{
    return null;
}

function myFunction($a, $lot, $of, $params)
    : array // phpcs:ignore Standard.Category.Sniff -- for reasons.
{
    return null;
}
```

When run against `Generic.Functions.OpeningFunctionBraceKernighanRitchie`, I would expect this to be fixed as (or an error thrown, but not auto-fixed):
```php
function myFunction($a, $lot, $of, $params)
    : array { // comment
    return null;
}

function myFunction($a, $lot, $of, $params)
    : array { // phpcs:ignore Standard.Category.Sniff -- for reasons.
    return null;
}
```

In reality, this will now be fixed (broken) as:
```php
function myFunction($a, $lot, $of, $params)
    : array {
 // comment
    return null;
}

function myFunction($a, $lot, $of, $params)
    : array {
 // phpcs:ignore Standard.Category.Sniff -- for reasons.
    return null;
}
```

gsherwood:
[...]

PR #1938, opened 2018-03-15
Generic opening brace placement sniffs incorrectly move PHPCS annotations

jrfnl:
As discussed in #1931, moving a trailing `phpcs:ignore` annotation to the next line changes its meaning, which is what currently happens when the `ContentAfterBrace` error is triggered by one of the new PHPCS annotations in the `Generic.Functions.OpeningFunctionBraceKernighanRitchie` and `Generic.Functions.OpeningFunctionBraceBsdAllman` sniffs.

For the `Generic.Functions.OpeningFunctionBraceBsdAllman` sniff, this also impacts the fixer of the `BraceOnSameLine` errror which could also move the comment down.

This PR fixes that.

For this fix, I've chosen to ignore *all* the PHPCS annotations, not just the `phpcs:ignore` one. If so desired, I can change the PR to only make an exception for a trailing `phpcs:ignore` comment and to move `phpcs:disable/enable/set` comments to the next line, just like "normal" comments.

Includes unit tests.


gsherwood:
> For this fix, I've chosen to ignore all the PHPCS annotations

I think that makes sense. Thanks a lot for fixing this.

PR #1939, opened 2018-03-19
phpcs:set annotations do not cause the line they are on to be ignored

jrfnl:
### Tokenizer: minor bug fix

The ignore setting for PHPCS annotations on their own line uses `['all' => true]` everywhere else, except for the `phpcs:set` annotation.

This has now been made consistent.

----
### SuperfluousWhitespace: fix fixer bug

The change in the tokenizer exposed a bug in the `Squiz.WhiteSpace.SuperfluousWhitespace` sniff, where if the last non-whitespace token in a file is a PHPCS annotation, the fixer would not kick in.

Moving the fixer start token determination for JS/CSS files into the fixer code block, fixes this.

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/SuperfluousWhitespaceSniff.php
@@ -162,15 +162,15 @@ class SuperfluousWhitespaceSniff implements Sniff
                 ) {
                     return;
                 }
-
-                if ($phpcsFile->fixer->enabled === true) {
-                    $prev     = $phpcsFile->findPrevious(T_WHITESPACE, ($stackPtr - 1), null, true);
-                    $stackPtr = ($prev + 1);
-                }
             }//end if
 
             $fix = $phpcsFile->addFixableError('Additional whitespace found at end of file', $stackPtr, 'EndFile');
             if ($fix === true) {
+                if ($phpcsFile->tokenizerType !== 'PHP') {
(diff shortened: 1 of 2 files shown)


gsherwood:
Looks like there are more problems here. Muting that whole line means that this code wont report an error with that sniff:
```php
<?php
/* phpcs:set foo bar */ ?>
```
Although it still fixes it for some reason, I think (still need to look into that).

But also, no error for start of file whitespace will be reported for this:
```php

<?php /* phpcs:set foo bar */
?>
```

So I thought I'd fix the check to make sure the comment is actually on a line by itself and found that check only looks towards the front of the line and doesn't check if there is any content after the comment. But it also ignores open tags before the comment, so it sees `<?php /* phpcs:set foo bar */` as being on a comment on a line by itself and so would never report errors on the open tag, which is pretty bad given a lot of sniffs do that.

More to look into.

jrfnl:
@gsherwood Just checking: Are you looking into this or would you like me to dig deeper ?

gsherwood:
> Are you looking into this or would you like me to dig deeper ?

I'm looking into it. Have an idea I'm testing out.

gsherwood:
I had to differentiate between lines that had no other tokens on them, and lines that had no other important tokens on them. The phpcs:set annotation needed to only ignore the line if it was really the only thing on there, which the other ones (especially phpcs:ignore) could ignore the line they are on even if there are PHP tags.

Tests are all passing, so I think it's working.

jrfnl:
@gsherwood Nice, though things are getting more and more complicated. Happy to hear this is fixed now though.

PR #1944, opened 2018-03-20
Replace slow in_array with array_key_exists

thiemowmde:
The arrays that are used in these calls are all intentionally constructed as maps with the array keys repeating the values.


gsherwood:
Yep, these were not converted over when I did that big change to move towards `isset()`. There are a few other sniffs that could be converted over as well, if anyone wants to spend some time doing it. Thanks for the PR, even though I went a different way.

thiemowmde:
isset() is not really faster than array_key_exists(), see http://maettig.com/1397246220. isset() is less expressive and more error-prone. array_key_exists() tells a better story.

jrfnl:
> isset() is not really faster than array_key_exists(),

According to the article you link to, they are in actual fact nearly twice as fast.

thiemowmde:
I wrote the article. ;-) It's faster, but not twice, and the difference does not matter in real life situations. When I profiled a 20 seconds PHPCS run I found 12ms are spend on 100,000 array_key_exists() calls. Sure, replacing them with isset() might save a millisecond or two.

PR #1948, opened 2018-03-21
Allow requesting a custom report using the report name or FQN

### LINKED ISSUE #1942: [Feature request] Allow requesting a custom report using the report FQN
Ok, I've been looking into creating some custom reports for external rulesets.

As far as I can see, you can either use one of the reports shipped with PHPCS or point PHPCS to a file which contains a custom report class using `--report=./path/to/report-class.php` on the command-line or using `<arg name="report" value="./path/to/report-class.php"/>` in a ruleset.

Now, the problem with that is that this relies on the path to the custom report file being stable and predictable.

* An absolute path will work, but is not portable.
* A relative path will work but only when the path is relative **_to the directory from which PHPCS is called_** - not when the path is relative to the ruleset/standard which adds the report -.
    - This is fine in a project Composer based install where you can do `./vendor/org/external-standard/Reports/CustomReport.php`.
    - This is problematic for registered _global_ Composer installed standards.
    - This is problematic for registered external rulesets installed in any other way.

PHPCS 3.0.1 via commit 1a598514a37ed6203247ebb4cfde44a2b777c2cd introduced the ability for external rulesets to add a custom autoload file in a ruleset using:
`<autoload>./path/to/relative/file.php</autoload>`.

This now allows for some flexibility with regards to loading non-sniff files where PHPCS does not need to translate sniff names to file and class names.

#### So, having said all that, to make the ability to have custom report formats more portable, I would very much like to suggest to add the option to allow loading custom report classes using their fully qualified classname.

For an external standard to offer a new custom report format, they would need to use the custom `autoload` option to sort out the loading of the file, but the `Reporter.php` class could then just instantiate the class based on the passed class name.

To show it in code:

In the external ruleset:
```xml
<autoload>./autoload.php</autoload>

<!-- Optionally add the report if it should be the default for this external standard. -->
<!-- Alternatively, an end-user could choose to use the report using
    `--report=\StandardNamespace\Reports\MyCustomReport` on the command line -->
<arg name="report" value="\StandardNamespace\Reports\MyCustomReport"/>
```

In the `autoload.php` file for the external standard, something along the lines of:
```php
spl_autoload_register(function ($class) {
    // Only try & load our own classes.
    if (stripos($class, 'StandardNamespace') !== 0) {
        return;
    }

    $file = realpath(__DIR__) . DIRECTORY_SEPARATOR . strtr($class, '\\', DIRECTORY_SEPARATOR) . '.php';

    if (file_exists($file)) {
        include_once $file;
    }
});
```

To allow for this, the report loading/instantiation logic in the `Reporter.php` class would need only minimal changes:
* It would need to detect when a FQN is passed as the report name
* It should trigger the autoloading of the file using a `class_exists()` and throw an exception if the class is not found.

https://github.com/squizlabs/PHP_CodeSniffer/blob/aace22f9e11eacebbac3ccfd5bd2299071a4be3e/src/Reporter.php#L108-L124

@gsherwood What do you think ? Would you be open to adding this option ?

If there is a green light, I'd be happy to make the proposed changes to the `Reporter.php` file.

jrfnl:
This allows for external standards to provide portable custom reports while installed in an arbitrary directory.

* The external standard has to make sure it includes an `<autoload>` directive in the ruleset and sort out the loading of the custom report classes through a `spl_autoload_register()`-ed autoloader.
* The name of the custom report class can be provided either on the command line using `--report=...` or via the ruleset using `<arg name="report" value=".."/>`.
* The name of the custom report class can be either the Fully Qualified report class name òr the standard relative class name (i.e. without the top-level namespace for the standard).
    - The FQN does not have to be prefixed with an `\`, but things will work just fine if it is.
    - If just the class name is provided, PHPCS will try to find the report in any of the namespaces as registered by the installed standards.
    - If the report class is not found in any of these, a `DeepExitException` will be thrown.

So, as an example, let's take an external standard with the following ruleset.xml file:
```xml
<?xml version="1.0"?>
<ruleset name="MyStandard" namespace="MyStandard\CS">
    <autoload>./../autoload-bootstrap.php</autoload>
    <rule ref="MyStandard"/>
</ruleset>
```

... which would provide a custom report class `MyStandard\CS\Custom` and an autoloader which would load the custom report

... an end-user could now request the custom report using any of the following names either from the command-line or from their own custom ruleset:
* `\MyStandard\CS\Custom`
* `MyStandard\CS\Custom`
* `\Custom`
* `Custom`

Existing functionality such as requesting several reports in one go `--report=source,Custom` and writing the report to a file using `--report-file=./path/to/file.txt` will work with this new feature as well.

Fixes #1942

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -218,6 +218,18 @@ if (class_exists('PHP_CodeSniffer\Autoload', false) === false) {
 
 
         /**
+         * Retrieve the namespaces and paths registered by external standards.
+         *
+         * @return array
+         */
+        public static function getSearchPaths()
+        {
+            return self::$searchPaths;
+
+        }//end getSearchPaths()
+
+
+        /**
          * Gets the class name for the given file path.
          *
          * @param string $path The name of the file.
--- a/src/Reporter.php
+++ b/src/Reporter.php
@@ -105,6 +105,7 @@ class Reporter
                 $output = $config->reportFile;
             }
 
+            $reportClassName = '';
             if (strpos($type, '.') !== false) {
                 // This is a path to a custom report class.
                 $filename = realpath($type);
@@ -114,8 +115,32 @@ class Reporter
                 }
 
                 $reportClassName = Autoload::loadFile($filename);
-            } else {
+            } else if (class_exists('PHP_CodeSniffer\Reports\\'.$type) === true) {
+                // PHPCS native report.
                 $reportClassName = 'PHP_CodeSniffer\Reports\\'.$type;
+            } else if (class_exists($type) === true) {
+                // FQN of a custom report.
+                $reportClassName = $type;
+            } else {
+                // OK, so not a FQN, try and find the report using the registered namespaces.
+                $registeredNamespaces = Autoload::getSearchPaths();
+                $trimmedType          = ltrim($type, '\\');
+
+                foreach ($registeredNamespaces as $nsPrefix) {
+                    if ($nsPrefix === '') {
+                        continue;
+                    }
+
+                    if (class_exists($nsPrefix.'\\'.$trimmedType) === true) {
+                        $reportClassName = $nsPrefix.'\\'.$trimmedType;
+                        break;
+                    }
+                }
+            }//end if
+
+            if ($reportClassName === '') {
+                $error = "ERROR: Class file for report \"$type\" not found".PHP_EOL;
+                throw new DeepExitException($error, 3);
             }
 
             $reportClass = new $reportClassName();


jrfnl:
FYI - Here's an external standard which can be seen as an example implementation for this PR: https://github.com/jrfnl/QA-WP-Projects

Autoloader: https://github.com/jrfnl/QA-WP-Projects/blob/3f237c9e99f2d9253c9bde7768aea866117ffc28/autoload-bootstrap.php#L27-L40
Custom report: https://github.com/jrfnl/QA-WP-Projects/blob/master/Reports/WPQA.php
Report set in the ruleset: https://github.com/jrfnl/QA-WP-Projects/blob/3f237c9e99f2d9253c9bde7768aea866117ffc28/WP-QA-Basic/ruleset.xml#L27

jrfnl:
@gsherwood Is there any chance this PR could be earmarked for the 3.3.0 version ? The change itself is quite small and straighforward, but this would be a big step forward in making custom reports so much easier to use.

gsherwood:
Thanks a lot for this - a really great change. I went looking for the docs to update and realised I don't have any docs for custom report classes yet, so I'll have to get on that later when I get some time.

jrfnl:
Glad you approve & thanks for the merge.

PR #1950, opened 2018-03-21
Avoid calling Config::__get() in a loop

thiemowmde:
I was surprised to see the impact this minor optimization does have. In the example run I profiled Config::__get() was called about 88,000 times, and these calls accounted for about 4.6% of the total runtime.

With this optimization applied Config::__get() is now called 5,000 times, and these calls now account for 0.3% of the total runtime.

Yes, I'm talking total runtime. This means a run of 20 seconds total will be about a second faster.

Before:
image

After:
image


gsherwood:
Thanks for this PR.

PR #1951, opened 2018-03-21
Avoid calling File::hasCondition() twice

thiemowmde:
This does not have a measurable effect on runtime, but immensely reduces the number of calls to this function. In the run I profiled hasCondition() was called about 18,000 times before and 10,000 times after this patch. The total runtime for these calls went from 1.3% to 1.1%.

Note there are a few more sniffs where the exact same optimization can be applied.


gsherwood:
Thanks

PR #1965, opened 2018-03-24
SquizMultiLineFunctionDeclaration: fix bug when param has trailing comment/PHPCS annotation

### LINKED ISSUE #1959: SquizMultiLineFunctionDeclaration error when param has trailing comment
Hi,

I'm encountering issue with ``Squiz.Functions.MultiLineFunctionDeclaration.OneParamPerLine`` snif.

The following works perfectly:

```php
    public function __construct(
        $foo,
        $bar
    ) {
```

while the following wont:

```php
    public function __construct(
        $foo,   // This is foo
        $bar
    ) {
```

and triggers ``error - Multi-line function declarations must define one parameter per line (Squiz.Functions.MultiLineFunctionDeclaration.OneParamPerLine)``

Please note that the following is working:

```php
    public function __construct(
        $foo /* this is foo */,
        $bar // this is bar
    ) {
```

jrfnl:
Very easy fix ;-)
Includes unit tests.
Should also solve any issues this particular check in the sniff has/had with trailing PHPCS annotations.

_Note: this sniff most likely does still have an issue with PHPCS annotations when fixing an incorrect multi-line function declaration._

Fixes #1959

### DIFF
--- a/src/Standards/Squiz/Sniffs/Functions/MultiLineFunctionDeclarationSniff.php
+++ b/src/Standards/Squiz/Sniffs/Functions/MultiLineFunctionDeclarationSniff.php
@@ -179,7 +179,7 @@ class MultiLineFunctionDeclarationSniff extends PEARFunctionDeclarationSniff
                 continue;
             }
 
-            $next = $phpcsFile->findNext(T_WHITESPACE, ($i + 1), null, true);
+            $next = $phpcsFile->findNext(Tokens::$emptyTokens, ($i + 1), null, true);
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks a lot for fixing this.

gsherwood:
> Note: this sniff most likely does still have an issue with PHPCS annotations when fixing an incorrect multi-line function declaration.

Given there are two params on the line for an error to occur, it's pretty hard to figure out which param the PHPCS annotation would be trying to silence, or where is is trying to start a change of a property etc. I think what it does now (keeps the comment with the param is follows) is a good enough solution.

PR #1967, opened 2018-03-26
Replace slow Tokens::tokenName() with define()

thiemowmde:
I actually profiled this and found Tokens::tokenName() is called hundreds of thousands of times here. Even if the method implementation appears to be trivial, it is expensive. The line touched in this patch always converts the same integer constant to a string, and compares the string then. In the profile run I did this line alone accounts for 3% of the total runtime. PHPs define() is much faster, even if called as often as it is the case here.

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -1146,7 +1146,7 @@ class PHP extends Tokenizer
                 as T_HASHBANG while PHP proper uses T_INLINE_HTML.
             */
 
-            if ($tokenIsArray === true && defined('T_HASHBANG') && $token[0] === T_HASHBANG) {
+            if ($tokenIsArray === true && defined('T_HASHBANG') === true && $token[0] === T_HASHBANG) {
                 $finalTokens[$newStackPtr] = [
                     'content' => $token[1],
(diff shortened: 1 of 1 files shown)


legoktm:
That code block is just compat for HHVM 3.5-3.6, which are extremely old and very much unsupported by upstream. Could we just remove it?

gsherwood:
> This code style is a pain
@thiemowmde Maybe you could try running PHPCS and PHPUnit, as I've put in the contributing doc. It's only 2 steps and then travis wont complain about code style to you, and you don't have to complain about code style to me :)

thiemowmde:
Hehe, sorry. ;-) Please feel free to squash that commit message away.

I'm aware of PHPCS' capabilities – that's why I'm here in the first place, right? I don't even mind getting Travis reports. I just think that a codestyle that leads to a codebase containing more than 2200 `=== true` and `=== false` in situations that are guaranteed to be boolean is actually painful. Even if I understand the benefit this strict rule does have, I think it's not worth it. But this is your codebase. So all I can do is complain a little bit. ;-)

gsherwood:
Thanks for the PR.

PR #1968, opened 2018-03-26
Added coveralls configuration - generate Code Coverage report

webimpress:
It would be nice to have Code Coverage report with the repository. I'm adding here integration with coveralls, but we can use any other service - whatever you prefer.

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,2 @@
+coverage_clover: clover.xml
+json_path: coveralls-upload.json
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+.coveralls.yml export-ignore
 .travis.yml export-ignore
 package.xml export-ignore
 phpunit.xml.dist export-ignore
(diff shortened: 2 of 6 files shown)


gsherwood:
Thanks for taking the time to try this, but I'm going to close this because I would never use the code coverage reports generated by the build so this is not something I want to maintain going forward.

webimpress:
@gsherwood What about adding ability to generate coverage on dev/local env? I think it is very useful especially when you write new sniff to check if you have covered all cases. Because of that I noticed some issues in code before.

I can't generate it easily right now because whitelist is not specified in the `phpunit.xml.dist`. Of course I can have my local `phpunit.xml` but this is annoying when I have to add it always when I check-out the repository without defined whitelist to generate coverage.
We don't need to run it in the build if you don't want to. But still - it's some kind of measure. We don't need to chase numbers. Coveralls also show changes in coverage (per file) on PRs. So it is easy to notice what has been changed and if we have some dead code.

PR #1973, opened 2018-03-28
Added missing function visibility

webimpress:


gsherwood:
Thanks for this. That check was missing from the PHPCS standard, so I've added it in as well.

PR #1974, opened 2018-03-29
Allow "file" in "rule" - phpcs.xsd hotfix

webimpress:
It should be possible to define rule just for some files, so:
```xml
<rule ref="Sniff">
    <exclude-pattern>*</exclude-pattern>
    <file>FILE_TO_CHECK</file>
</rule>
```

### DIFF
--- a/phpcs.xsd
+++ b/phpcs.xsd
@@ -53,6 +53,7 @@
             <xs:element name="exclude-pattern" type="patternType" maxOccurs="unbounded" minOccurs="0"></xs:element>
             <xs:element name="include-pattern" type="patternType" maxOccurs="unbounded" minOccurs="0"></xs:element>
             <xs:element name="properties" type="propertiesType" maxOccurs="1" minOccurs="0"></xs:element>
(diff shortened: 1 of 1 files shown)


webimpress:
@gsherwood Yes, I know I can use `include-pattern` but it also works with `file`, but it's not allowed by xsd schema.

Is it going to stop working on some point then? If no, I think the schema should be updated, as I did in that PR.

gsherwood:
> Yes, I know I can use include-pattern but it also works with file

It shouldn't work the same as using `file`, and it doesn't when I test it. Those `file` tags are ignored when inside `rule` tags. I can't find any code that would even attempt to process them.

Are you sure your file tags are being processed?

webimpress:
> Are you sure your file tags are being processed?

I've tested it again and you are right - these are __not__ processed. Obviously I did something wrong before... Sorry. Thanks for your help!

gsherwood:
No problem. Thanks for re-testing.

PR #1976, opened 2018-03-30
Fix tokenizing T_OPEN_TAG with comment straight after

webimpress:
In case we have:
```php
<?php/** comment */
```

it is tokenized as:
T_OPEN_TAG = <?
T_STRING = php
...

We should have here only
T_OPEN_TAG = <?php
...

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -736,6 +736,20 @@ class PHP extends Tokenizer
             }//end if
 
             /*
+                <?php/** is tokenized incorrectly as T_OPEN_TAG and T_STRING
+            */
+
+            if ($tokenIsArray === true
+                && $token[0] === T_OPEN_TAG
+                && $token[1] === '<?'
+                && $tokens[($stackPtr + 1)][0] === T_STRING
+                && strtolower($tokens[($stackPtr + 1)][1]) === 'php'
+            ) {
+                $token[1] .= $tokens[($stackPtr + 1)][1];
+                $tokens[($stackPtr + 1)] = null;
+            }
+
+            /*
                 Before PHP 7.0, the "yield from" was tokenized as
                 T_YIELD, T_WHITESPACE and T_STRING. So look for
                 and change this token in earlier versions.


gsherwood:
This sample code is tokenized like this for me:
```
*** START PHP TOKENIZING ***
Process token [0]: T_INLINE_HTML => <?php/**·comment·*/\n
*** END PHP TOKENIZING ***
```
The reason is that the sample code is not valid PHP code and the T_INLINE_HTML token is coming straight for the built-in PHP tokenizer.

Is this real code that is working for you?

webimpress:
@gsherwood I got confused because it is "good code" for PHPStorm but actually it's not:
https://3v4l.org/7G76E

However that code is tokenized as I said:
https://3v4l.org/e2oYn
(T_OPEN_TAG, T_STRING, ...) on PHP 5.6-7.2 so I don't know why you get `T_INLINE_HTML`...

This is my result on the latest `master` for code:
```php
<?php/** comment */
```

```
        *** START PHP TOKENIZING ***
        Process token [0]: T_OPEN_TAG => <?
        Process token [1]: T_STRING => php
        Process token [2]: T_DOC_COMMENT => /**·comment·*/
                *** START COMMENT TOKENIZING ***
                Create comment token: T_DOC_COMMENT_OPEN_TAG => /**
                Create comment token: T_DOC_COMMENT_WHITESPACE => ·
                Create comment token: T_DOC_COMMENT_STRING => comment·
                Create comment token: T_DOC_COMMENT_CLOSE_TAG => */
                *** END COMMENT TOKENIZING ***
        Process token [3]: T_WHITESPACE => \n
        *** END PHP TOKENIZING ***
```

webimpress:
@gsherwood Sorry, closed by accident, I'm reopening it and you'll decide if you want to appy this fix or not. It's not valid PHP code, so I understand if you decide not to.

gsherwood:
> so I don't know why you get T_INLINE_HTML...

I think you have short open tags on. So while PHP can't match `<?php`, it can match `<?` and so it does. From that point, `php` looks like a regular string.

> It's not valid PHP code, so I understand if you decide not to.

Because of what I said above, I absolutely can't merge it in. It would make it look like the opening tag was not a short open tag even though PHP used one, which would make it impossible to detect the usage of short open tags by a sniff.

Thanks for posting this though. It was interesting to look into.

PR #1979, opened 2018-04-02
Tokenizer does not change heredoc to nowdoc token if the start tag contains spaces

webimpress:
`<<< 'EOF'` is a valid nowdoc start tag (there could be more than one space, or even tab ...)


gsherwood:
Thanks for reporting this, although I fixed it using a strpos rather than using a regex because it is faster and PHP wont allow things like newlines and comments as part of that tag. If you've think I've made a mistake there, please let me know.

webimpress:
@gsherwood Solution with `strpos` instead of regular expression looks good for me! Thanks!

PR #1985, opened 2018-04-11
PHPCS annotations: prevent "noise" from comment punctuation checks with --ignore-annotations

jrfnl:
When PHPCS is run with `--ignore-annotations`, there are three sniffs which would create a lot of "noise" complaining about PHPCS annotation comment lines not starting with a capital letter or ending in a full stop.

The three commits contained in this PR prevent those errors being thrown when a PHPCS annotation is encountered.

For the `Generic.Commenting.DocComment` sniff, this also fixes the sniff *not* throwing a "missing short description" error when the first thing in a docblock is a PHPCS annotation (when run with `-ignore-annotations`).

Includes unit tests in separate files to allow for the config settings for those files to be different from the main test file(s).


gsherwood:
Looking at this, I'm now wondering if it is actually better to tokenize those comments even if ignoring annotations, but just don't set any ignore rules. I thought it would be better not to, but the duplicate code it creates isn't great. I'll have a look.

gsherwood:
> I'm now wondering if it is actually better to tokenize those comments even if ignoring annotations, but just don't set any ignore rules

I decided to go ahead and do this. Have a look at the latest master when you get some time, or let me know if you think of some reasons to keep the tokenizing the way it was. Thanks.

gsherwood:
> I seem to remember we talked about this before and didn't reach a decision at that time.

Same, although I think I was leaning on the side of not tokenising them. But with all that noise being generated, I'm also happy with this change.

Issue #1989, opened 2018-04-12
When MissingShort is suppressed then param tag messages not emitted


phil-davis:
Suppress the requirement for a short description in the PHPdoc block, like:
```

	<rule ref="PEAR">
		<exclude name="Generic.Commenting.DocComment.MissingShort" />
	</rule>
```
then write a doc block like:
```
	/**
	 * 
	 * @param string $myParameter
	 * @return boolean
	 */
```
and it passes ``phpcs`` inspection.

But I expect ``phpcs`` to complain with a message like:
```
-----------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------------------------------
 56 | ERROR | Tag cannot be grouped with parameter tags in a doc comment (Generic.Commenting.DocComment.NonParamGroup)
-----------------------------------------------------------------------------------------------------------------------
```
If I do write a short description, e.g.:
```
	/**
	 * Short description of method
	 * 
	 * @param string $mailhogUrl
	 * @return ResponseInterface
	 */
```
then ``phpcs`` does complain about ``NonParamGroup``

Also note that when ``MissingShort`` is enabled, then ``phpcs`` complains about the missing short description (as expected), but it does not go on to complain about ``NonParamGroup`` (or other problems in the doc block).

I guess that the doc block parsing is getting terminated when no short description is found, and in that case is not going on to look for other problems.

gsherwood:
Closed by PR #1990

PR #1990, opened 2018-04-12
Process doc tags when short description is missing

phil-davis:
Fixes issue #1989 

In the existing code, when the short description was missing, then ``DocCommentSniff`` ``process()`` would return immediately. The parsing of the tag problems comes later, and so that did not happen.

With this PR, the various problems related to the description lines put inside an ``else`` of the ``Check for a comment description`` ``if`` statement - so that they are only processed if there is at least a description found. The "early" return is removed. This allows the code to continue, examining matters related to tags.

### DIFF
--- a/src/Standards/Generic/Tests/Commenting/DocCommentUnitTest.php
+++ b/src/Standards/Generic/Tests/Commenting/DocCommentUnitTest.php
@@ -49,8 +49,8 @@ class DocCommentUnitTest extends AbstractSniffUnitTest
             95  => 1,
             156 => 1,
             158 => 1,
-            170 => 3,
-            171 => 3,
+            170 => 4,
+            171 => 4,
             179 => 1,
             183 => 1,
             184 => 2,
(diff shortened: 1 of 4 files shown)


src/Standards/Generic/Tests/Commenting/DocCommentUnitTest.php
```diff
@@ -49,8 +49,8 @@ public function getErrorList()
             95  => 1,
             156 => 1,
             158 => 1,
-            170 => 3,
-            171 => 3,
+            170 => 4,
```
phil-davis: This line should have always had an extra complaint about the tag being on the opening line, not on its own line. Now the complaint comes out - so the count is bumped here.

src/Standards/Generic/Tests/Commenting/DocCommentUnitTest.php
```diff
@@ -49,8 +49,8 @@ public function getErrorList()
...
+            171 => 4,
```
phil-davis: Same as above.

gsherwood:
Thanks for this PR. Looks like a good change.

PR #1998, opened 2018-04-19
Adding downloading via wget

Qussayyon:
Some systems might not have curl, so using `wget` is good option.


Qussayyon:
great.
thanks.

PR #1999, opened 2018-04-20
Ability to set array values using phpcs:set

webimpress:
Property name should be with `[]` and values must be comma separated.
If we want provide also array keys we need to separate them with `=>`.
So the format is the same as the one possible format in ruleset, i.e.:

```php
// phpcs:set Squiz.PHP.SniffName property[] key=>value,key2=>value2
```


webimpress:
@gsherwood Any chance to have it merged into 3.3.0 release? I need this to test my custom sniff with array properties. Or maybe there is another way to do it? Thanks!

webimpress:
Because of hotfix #2004 probably this is not gonna work to set empty value. Maybe we should allow just:
```php
// phpcs:set Squiz.PHP.SniffName property[]
```
(so no value provided) and it will mean to set empty value (`[]`)?
Thoughts?

webimpress:
Rebased with base #2004.
No it is also possible to set empty value for the property - empty array for arrays:
```php
// phpcs:set Squiz.PHP.SniffName property[]
```

and `null` for others:
```php
// phpcs:set Squiz.PHP.SniffName property
```

(so there is no value at all, it means to set empty value).

gsherwood:
Thanks a lot for this change

PR #2003, opened 2018-04-23
Use array_key_exists instead of in_array + array_keys

carusogabriel:


src/Standards/Squiz/Sniffs/Operators/ComparisonOperatorUsageSniff.php
```diff
@@ -157,7 +157,7 @@ public function process(File $phpcsFile, $stackPtr)
 
         for ($i = $start; $i <= $end; $i++) {
             $type = $tokens[$i]['code'];
-            if (in_array($type, array_keys(self::$invalidOps[$tokenizer])) === true) {
+            if (array_key_exists($type, self::$invalidOps[$tokenizer]) === true) {
```
jrfnl: Using `isset()` would probably be more stable, as `array_key_exists()` will still return `true`, even when the key has no value.

gsherwood:
Thanks for pointing this out, although I decided to use `isset()` as I use it throughout the code base for stuff like this. I also changed `$validOps` so that I could use `isset()` for that as well.

PR #2004, opened 2018-04-24
Fixed counting elements in array in if statements

webimpress:
`explode` function returns array and we should count number of elements in that array.
Comparison `ARR >= 3` always returns `true`.

Code (https://3v4l.org/NkdIT):
```php
var_dump([] >= 3);
var_dump([1] >= 3);
var_dump([1, 2] >= 3);
var_dump([1, 2, 3] >= 3);
var_dump([1, 2, 3, 4] >= 3);
```

Results:
```
bool(true)
bool(true)
bool(true)
bool(true)
bool(true)
```


gsherwood:
Thanks a lot for finding this. It was not intentional for those `count()`s to be missing.

PR #2018, opened 2018-05-15
Generic.Formatting.MultipleStatementAlignment does see PHP close tag as end of statement block

jrfnl:
The case where PHP and HTML is mixed was not taken into account by the sniff up to now.

Includes unit test.


gsherwood:
Thanks for fixing this.

jrfnl:
You're welcome.

PR #2022, opened 2018-05-22
[PSR12] Introduce NullableWhitespace sniff

timoschinkel:
A sniff, including docs and tests, to check and fix the formatting of nullable type hinting.

> ## [4.5 Method and function Arguments](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#45-method-and-function-arguments)
>
> ...
> 
> In nullable type declarations there MUST NOT be a space between the question mark and the type.

References #750


src/Standards/PSR12/Sniffs/Methods/NullableWhitespaceSniff.php
```diff
@@ -0,0 +1,57 @@
...
+     * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+     * @param int                         $stackPtr  The position of the current token in the
+     *                                               stack passed in $tokens.
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+        if (isset($tokens[($stackPtr + 1)]) === true
```
jrfnl: Looks like you are only checking (and fixing) the first token after the `?`.

What about if there is `whitespace new-line whitespace` ? or what about if there is a comment or PHPCS annotation in between ?
```php
function ( ?
    string $varA,
    ? /* comment about why it is nullable */ array $varB,
    ? // phpcs:ignore Standard.Cat.Sniff -- for reasons
    int $varC
) {}
```
I'm not saying that sort of code is clean by any standard, but IMHO this sniff should not break on it.

src/Standards/PSR12/Tests/Methods/NullableWhitespaceUnitTest.inc
```diff
@@ -0,0 +1,41 @@
...
+class MyClass {
+    public function validReturnType(): ?MyObject
+    {
+    }
+
+    public function validReturnTypeForThisSniff():?MyObject
+    {
+    }
+
+    public function invalidReturnTypeTooLittleWhitespace():? MyObject
```
jrfnl: `TooLittleWhitespace` => `TooMuchWhitespace`

src/Standards/PSR12/Tests/Methods/NullableWhitespaceUnitTest.inc
```diff
@@ -0,0 +1,41 @@
...
+    {
+    }
+
+    public function invalidReturnTypeWhitespaceAfterQuestionMark(): ? MyObject
```
jrfnl: Checking same thing as test above.

src/Standards/PSR12/Tests/Methods/NullableWhitespaceUnitTest.inc
```diff
@@ -0,0 +1,41 @@
+<?php
```
jrfnl: * I don't see any tests for nullable parameter type declarations (can be combined with the existing tests).
* Why do you have separate test for `interface`s ? You are sniffing for `T_NULLABLE`, not for `class`/`interface` or `trait` for that matter, so the test is not testing anything new.
    May be it is an idea to mix the tests up a bit: have one test with a global function, one with a closure, one with a method in a class, one with an interface method etc ?

timoschinkel:
@jrfnl Thank you for the feedback. I will review my PR based on your comments.

jrfnl:
@timoschinkel All sounds good. Would you like me to take another look ?

timoschinkel:
@jrfnl Yes please :)

src/Standards/PSR12/Sniffs/Functions/NullableTypeDeclarationSniff.php
```diff
@@ -0,0 +1,69 @@
...
+     *
+     * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+     * @param int                         $stackPtr  The position of the current token in the
+     *                                               stack passed in $tokens.
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $nextValidTokenPtr = $phpcsFile->findNext([T_STRING, T_ARRAY_HINT, T_RETURN_TYPE], ($stackPtr + 1), null);
```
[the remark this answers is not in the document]
timoschinkel: Good point. I missed that situation and I was not aware these tokens are no longer used. Will change this.

src/Standards/PSR12/Sniffs/Functions/NullableTypeDeclarationSniff.php
```diff
@@ -0,0 +1,69 @@
...
+        if ($nextValidTokenPtr !== ($stackPtr + 1)) {
+            $nonWhitespaceTokenPtr = $phpcsFile->findNext([T_WHITESPACE], ($stackPtr + 1), $nextValidTokenPtr, true);
+
+            if ($nonWhitespaceTokenPtr === false) {
+                if ($phpcsFile->addFixableError(
```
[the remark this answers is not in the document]
timoschinkel: I'm happy to comply to the patterns used within the project for consistency.

timoschinkel:
@jrfnl Thank you again for the review. Shows I'm still a but of a novice in writing sniffs. I will try to apply the comments.

jrfnl:
> Shows I'm still a but of a novice in writing sniffs

@timoschinkel  I wouldn't say that at all, you should see some of the sniffs I've reviewed... :smile:

src/Standards/PSR12/Sniffs/Functions/NullableTypeDeclarationSniff.php
```diff
@@ -0,0 +1,73 @@
+<?php
+/**
+ * Verifies that nullable typehints are lacking superfluous whitespace, e.g. ?int
+ *
+ * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @copyright 2006-2015 Squiz Pty Ltd (ABN 77 084 670 600)
```
caugner: Shouldn't this be `2006-2018`?

src/Standards/PSR12/Sniffs/Functions/NullableTypeDeclarationSniff.php
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+     * @param int                         $stackPtr  The position of the current token in the
+     *                                               stack passed in $tokens.
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $nextValidTokenPtr = $phpcsFile->findNext([T_STRING, T_NS_SEPARATOR], ($stackPtr + 1), null);
```
jrfnl: Further on my previous comment about this line: https://github.com/squizlabs/PHP_CodeSniffer/pull/2022#discussion_r191965848

This still will not work as intended.

Additional test cases which should be added and demonstrate the problem:
```php
function testSelf( ? self $self ) : ? self {}
function testParent( ? parent $parent ) : ? parent {}
function testCallable( ? callable $callable ) : ? callable {}
```

timoschinkel: I've added the additional test cases and the issue appears not to be the reverse check - the additional cases *are* marked as violations - but not taking into account all the allowed/expected type hints and therefor not marking them as fixable. 

What I did is restructured the code to the reverse check and cleaned it up a bit so the maximum nesting depth of conditionals is two. I added three additional tokens `T_CALLABLE`, `T_SELF` and `T_PARENT`. I tried to find an exhaustive list of allowed type hint tokens to make sure I'm not forgetting any tokens. 

I also updated the copyright comment.

src/Standards/PSR12/Sniffs/Functions/NullableTypeDeclarationSniff.php
```diff
@@ -0,0 +1,85 @@
...
+        ];
+
+        $nextNonEmptyPtr = $phpcsFile->findNext([T_WHITESPACE], ($stackPtr + 1), null, true);
+        if ($nextNonEmptyPtr === false) {
+            // Parse error or live coding.
+            return;
+        }
+
+        $tokens          = $phpcsFile->getTokens();
+        $validTokenFound = in_array($tokens[$nextNonEmptyPtr]['code'], $validTokens) === true;
```
jrfnl: This can be made more efficient by:
1. Putting the `$validTokens` array in a class property - it's not going to change during the run anyway.
2. Changing the `$validTokens` array to use a `TOKEN_CODE => true` format and
3. Changing the call to `in_array` to `isset()` on this line.

src/Standards/PSR12/Sniffs/Functions/NullableTypeDeclarationSniff.php
```diff
@@ -0,0 +1,85 @@
...
+                }
+            }
+
+            return;
+        }
+
+        // Non-whitespace tokens found; trigger error but don't fix.
+        $phpcsFile->addError('Unexpected characters found after nullable', ($stackPtr + 1), 'UnexpectedCharactersFound');
+
+        return;
```
jrfnl: Return not needed here as you're at the end of the function anyhow.

src/Standards/PSR12/Sniffs/Functions/NullableTypeDeclarationSniff.php
jrfnl: > I tried to find an exhaustive list of allowed type hint tokens to make sure I'm not forgetting any tokens.

Looking at the current list, I'm pretty sure you're ok now.

timoschinkel:
Vacation caused some delay. My apologies for that. I updated the code to follow the remarks. The assignment of the `$nextNonEmptyCode` to a separate variable is purely for readability purposes.

gsherwood:
Finally merged :)

This turned out to be a nice sniff. Thanks @jrfnl for doing those reviews, and thanks @timoschinkel for making all those changes.

I did change the error messages so they reflect the text of the PSR-12 standard because it's just something I normally do with PSR-namespaced sniffs.

PR #2023, opened 2018-05-23
3.3.0: Generic/LowerCaseType: allow for nullable types and more

jrfnl:
This PR ports over various things which were already covered by PR #1685, but not (yet) covered by the new `Generic.PHP.LowerCaseType` sniff.

@gsherwood As the `Generic.PHP.LowerCaseType` sniff is new to PHPCS 3.3.0, it would be advisable to merge this PR before releasing PHPCS 3.3.0.

All significant changes are accompanied by unit tests.

### Generic/LowerCaseType: add missing types

* `parent` is a valid type declaration and has been supported since the beginning, though only properly since PHP 5.2.
* `void` was added in PHP 7.1 (return type only).
* `object` was added in PHP 7.2.

### Generic/LowerCaseType: have unique error messages and codes

Differentiate the error messages and error codes to enable modular usage of the sniff.

### Generic/LowerCaseType: indicate the precise line of the error

For (long) multi-line function declarations, it will be more helpful throwing the error on the line where the error actually occurs instead of on the line of the function keyword.

### Generic/LowerCaseType: various efficiency fixes

* Bow out early if there are no function parameters.
* No need to execute the exact same function call multiple times.

### Generic/LowerCaseType: minor terminology fix

### DIFF
--- a/src/Standards/Generic/Docs/PHP/LowerCaseTypeStandard.xml
+++ b/src/Standards/Generic/Docs/PHP/LowerCaseTypeStandard.xml
@@ -1,17 +1,17 @@
 <documentation title="Lowercase PHP Types">
     <standard>
     <![CDATA[
-    All PHP types used for type hints and return types should be lowercase.
+    All PHP types used for parameter type and return type declarations should be lowercase.
     ]]>
     </standard>
     <code_comparison>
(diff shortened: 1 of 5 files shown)


gsherwood:
Thanks for pulling out those changes - they were all great.

PR #2024, opened 2018-05-23
Drop HHVM testing

photodude:
HHVM LTS 3.24 is the last version to support PHP 5 compatibility
the third paragraph of the announcement states "HHVM will not aim to target PHP7" and further down "We do not intend to be the runtime of choice for folks with pure PHP7 code." (HHVM will only support HACK going forward)
https://hhvm.com/blog/2017/09/18/the-future-of-hhvm.html

### DIFF
--- a/src/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
+++ b/src/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
@@ -61,7 +61,7 @@ class SyntaxSniff implements Sniff
         }
 
         $fileName = escapeshellarg($phpcsFile->getFilename());
-        $cmd      = escapeshellcmd($this->phpPath)." -l $fileName 2>&1";
+        $cmd      = escapeshellcmd($this->phpPath)." -l -d display_errors=1 -d error_prepend_string='' $fileName 2>&1";
         $output   = shell_exec($cmd);
         $matches  = [];
         if (preg_match('/^.*error:(.*) in .* on line ([0-9]+)/m', trim($output), $matches) === 1) {
--- a/src/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
+++ b/src/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
@@ -61,14 +61,9 @@ class SyntaxSniff implements Sniff
         }
 
         $fileName = escapeshellarg($phpcsFile->getFilename());
-        if (defined('HHVM_VERSION') === false) {
-            $cmd = escapeshellcmd($this->phpPath)." -l -d display_errors=1 -d error_prepend_string='' $fileName 2>&1";
-        } else {
-            $cmd = escapeshellcmd($this->phpPath)." -l $fileName 2>&1";
-        }
-
-        $output  = shell_exec($cmd);
-        $matches = [];
+        $cmd      = escapeshellcmd($this->phpPath)." -l $fileName 2>&1";
+        $output   = shell_exec($cmd);
+        $matches  = [];
         if (preg_match('/^.*error:(.*) in .* on line ([0-9]+)/m', trim($output), $matches) === 1) {
             $error = trim($matches[1]);
             $line  = (int) $matches[2];
--- a/src/Standards/Generic/Tests/Arrays/DisallowLongArraySyntaxUnitTest.php
+++ b/src/Standards/Generic/Tests/Arrays/DisallowLongArraySyntaxUnitTest.php
@@ -25,6 +25,7 @@ class DisallowLongArraySyntaxUnitTest extends AbstractSniffUnitTest
     protected function getTestFiles($testFileBase)
     {
         $testFiles = [$testFileBase.'1.inc'];
+        $testFiles[] = $testFileBase.'2.inc';
 
         return $testFiles;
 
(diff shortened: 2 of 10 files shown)


jrfnl:
Oh... there are a number of (mostly unit test) files which have special conditions for HHVM, those can be removed as well if HHVM support is being dropped.

* `src/Standards/Generic/Sniffs/PHP/SyntaxSniff.php`
* `src/Standards/Generic/Tests/Arrays/DisallowLongArraySyntaxUnitTest.php`
* `src/Standards/Generic/Tests/Files/EndFileNewlineUnitTest.php`
* `src/Standards/Generic/Tests/Files/EndFileNoNewlineUnitTest.php`
* `src/Standards/Generic/Tests/PHP/DisallowAlternativePHPTagsUnitTest.php`
* `src/Standards/Generic/Tests/PHP/DisallowShortOpenTagUnitTest.php`
* `src/Standards/PSR2/Tests/Files/EndFileNewlineUnitTest.php`
* `src/Standards/Squiz/Tests/Commenting/FileCommentUnitTest.php`

photodude:
Odd, the custom ini for PHP 5.5 and php 7 are throwing errors. Everything else seems fine, just not those two tests.

src/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
```diff
@@ -61,14 +61,9 @@ public function process(File $phpcsFile, $stackPtr)
...
         $fileName = escapeshellarg($phpcsFile->getFilename());
-        if (defined('HHVM_VERSION') === false) {
-            $cmd = escapeshellcmd($this->phpPath)." -l -d display_errors=1 -d error_prepend_string='' $fileName 2>&1";
-        } else {
-            $cmd = escapeshellcmd($this->phpPath)." -l $fileName 2>&1";
-        }
-
-        $output  = shell_exec($cmd);
-        $matches = [];
+        $cmd      = escapeshellcmd($this->phpPath)." -l $fileName 2>&1";
```
jrfnl: AFAICS, this should be `$cmd = escapeshellcmd($this->phpPath)." -l -d display_errors=1 -d error_prepend_string='' $fileName 2>&1";`

src/Standards/Generic/Tests/Arrays/DisallowLongArraySyntaxUnitTest.php
```diff
@@ -26,12 +26,6 @@ protected function getTestFiles($testFileBase)
     {
         $testFiles = [$testFileBase.'1.inc'];
 
-        // HHVM doesn't tokenize any of the file after a git
```
jrfnl: `if (defined('HHVM_VERSION') === false) {` === not HHVM, so this whole method can be removed.

src/Standards/Generic/Tests/PHP/DisallowShortOpenTagUnitTest.php
```diff
@@ -24,14 +24,9 @@ class DisallowShortOpenTagUnitTest extends AbstractSniffUnitTest
      */
     protected function getTestFiles($testFileBase)
     {
-        $testFiles = [$testFileBase.'1.inc'];
-
-        $option = (boolean) ini_get('short_open_tag');
-        if ($option === true || defined('HHVM_VERSION') === true) {
```
jrfnl: The condition and the `if`/`else` logic needs to stay, just the ` || defined('HHVM_VERSION') === true` part needs to be removed.

src/Standards/Generic/Tests/Arrays/DisallowLongArraySyntaxUnitTest.php
photodude: Oops "only run this check on non-HHVM versions" test needs to be added back, but the check can be removed

src/Standards/Generic/Tests/PHP/DisallowShortOpenTagUnitTest.php
photodude: Oops, should have caught that. Fixed now, thanks.

src/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
jrfnl: I figured as much ;-)

photodude:
Should be ready to go now. I expect everything should be passing with these last two changes.

src/Standards/Generic/Tests/Arrays/DisallowLongArraySyntaxUnitTest.php
jrfnl: If the `getTestFiles()` method is not there, PHPCS automatically will test all files which start with `SniffNameUnitTest....`, so if there is no logic needed to determine which files to test - i.e. what we're left with now -, the whole method can be removed.

photodude: Oh, I see. I understand now what you are saying. I'll make that improvement in the morning.

photodude: fixed

gsherwood:
@photodude Thanks for putting this together, and thanks for reviewing @jrfnl. All looked good to me.

I also removed some PHP tokenizer exceptions for HHVM that I've been dying to get rid of, so I don't think there is any HHVM code anywhere in PHPCS any more now.

photodude:
@gsherwood You are welcome. 
@jrfnl Thanks for the review and the suggestions.

PR #2028, opened 2018-05-25
PEAR/ValidFunctionName: fix uninitialized string offset error

### LINKED ISSUE #2027: PEAR.NamingConventions.ValidFunctionName error when function name includes double underscore
When running PHPCS on an existing codebase, the following error occurred:

 An error occurred during processing; checking has been aborted. The error message was: Uninitialized string offset: 0 in /usr/share/pear/PHP/CodeSniffer/src/Standards/PEAR/Sniffs/NamingConventions/ValidFunctionNameSniff.php on line 239 (Internal.Exception)

This occurred on the following function name:  send_system_email__to_user

After removing the double underscore, the error goes away.

Example code
```
<?php

function send_system_email__to_user($body, $recipient){
        echo 'test';
}

```
I am using version 3.2.3 of PHP_CodeSniffer  and PHP v 5.6.36

jrfnl:
When a function name would contain more than one subsequent underscore, the `explode()` call will create entries containing an empty string in the `$nameBits` array.

The proposed fix will remove those empty entries before testing each `$bit`.
([`array_filter()` called without callback](http://php.net/manual/en/function.array-filter.php#example-6090) will remove everything which after juggling would evaluate to `false`).

**Note:** The resulting suggested `$newName` for the function will no longer contain a double underscore either.

If it would be desired to allow the double underscore, the alternative would be to change the `if ($bit{0} !== strtoupper($bit{0}))` condition to `if ($bit !== '' && $bit{0} !== strtoupper($bit{0}))`. This would also require an adjustment in the inner loop which creates the new package name.

Includes unit test.

Fixes #2027

### DIFF
--- a/src/Standards/PEAR/Sniffs/NamingConventions/ValidFunctionNameSniff.php
+++ b/src/Standards/PEAR/Sniffs/NamingConventions/ValidFunctionNameSniff.php
@@ -232,6 +232,7 @@ class ValidFunctionNameSniff extends AbstractScopeSniff
         if ($packagePart !== '') {
             // Check that each new word starts with a capital.
             $nameBits = explode('_', $packagePart);
+            $nameBits = array_filter($nameBits);
             foreach ($nameBits as $bit) {
                 if ($bit{0} !== strtoupper($bit{0})) {
                     $newPackagePart = '';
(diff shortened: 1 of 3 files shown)


gsherwood:
> If it would be desired to allow the double underscore

I had a look through the PEAR coding standard again underscores are only used for namespacing. Function names aren't actually supposed to be defined using underscores instead of camelCase, although the sniff would have to check directory names to see if the underscore were truely being used for namespacing.

So I don't think any additional changes would be needed for this PR.

gsherwood:
Thanks a lot for fixing this

PR #2032, opened 2018-05-27
Squiz/FunctionSpacing: fix fixer conflict - ignore function at start of embedded PHP

jrfnl:
[Fixer conflicts series PR]

`phpcbf -p -s ./src/Standards/Generic/Tests/Files/LineEndingsUnitTest.inc --standard=Squiz`

Basically, the `Squiz.WhiteSpace.FunctionSpacing` sniff demands 2 blank lines between functions, while the `Squiz.PHP.EmbeddedPHP` sniff demands no blank lines at the start of an embedded PHP block, which causes a fixer conflict if a function is declared at the top of an embedded PHP block.

To fix this, I've given precedence to the `EmbeddedPHP` sniff - i.e. no blank lines between the PHP open tag and the start of the function declaration -.
The `FunctionSpacing` will now ignore functions at the start of an embedded PHP block.

Includes unit test.

----

Side-note: the docblock at the top of the sniff actually states `Checks the separation between methods in a class or interface.`, but the sniff does not actually check if the function it is examining is within a class or an interface.
Adding that check could be considered an alternative solution for the fixer conflict, but would change the existing behaviour of the sniff significantly.

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/FunctionSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/FunctionSpacingSniff.php
@@ -68,6 +68,16 @@ class FunctionSpacingSniff implements Sniff
      */
     public function process(File $phpcsFile, $stackPtr)
     {
+        $tokens           = $phpcsFile->getTokens();
+        $previousNonEmpty = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($stackPtr - 1), null, true);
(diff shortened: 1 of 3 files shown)


gsherwood:
Giving EmbeddedPHP precedence makes sense to me too. Thanks.

> the docblock at the top of the sniff actually states Checks the separation between methods in a class or interface., but the sniff does not actually check if the function it is examining is within a class or an interface.

Fixed now. I guess it used to be limited to methods at one stage.

PR #2040, opened 2018-06-04
Squiz/SuperfluousWhitespace: recognize unicode whitespace at start/end of file

### LINKED ISSUE #2038: Whitespace before PHP open tag sniff ?
Would there be interest in a `WhitespaceBeforePHPOpenTag` sniff ?

I'm creating one for an external standard with the aim to help devs avoid the dreaded "headers already send" error and would be happy to pull it here instead.

The `Generic.PHP.CharacterBeforePHPOpenTag` sniff already exists, but that sniff also demands that the first PHP open tag is the first thing in a file (except for maybe a shebang), so is not usable if the project the sniff is used on would contain views without a file docblock within PHP tags at the top of the file.

jrfnl:
As unicode regexes can be even more expensive than regular regexes, I've moved the check for non-whitespace out of the loop.

Includes unit tests. The `5` file is the one with unicode. The other ones are just more confirmation that the sniff works as expected.

Fixes #2038


gsherwood:
Thanks a lot for this.

PR #2064, opened 2018-06-16
Config.php: minor code simplification

jrfnl:
Came across this snippet which looked like it could be simplified. Not sure if there was a specific reason why you wrote it like this. I couldn't find any relevant commit message about this in the Git history.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -1292,9 +1292,7 @@ class Config
             $error .= $this->printShortUsage(true);
             throw new DeepExitException($error, 3);
         } else {
-            $files       = $this->files;
-            $files[]     = $file;
-            $this->files = $files;
+            $this->files[] = $file;
             self::$overriddenDefaults['files'] = true;
         }
 


gsherwood:
It's because the config class doesn't have any direct members - it uses the __set magic method to set values as it made things a lot easier for me during the redevelopment.

But the problem when you do this is that __set doesn't support arrays. So instead of being able to just append a value to an array member, you need to first get it, then append, then set it again.

I *think* it may be possible to achieve this if __get returns by reference, but it's not normally what I want.

jrfnl:
@gsherwood Interesting. I hadn't had the time yet to look into the Travis failure, but that explains it.

I'm going to close this PR in favour of leaving things as they are, though I'd suggest adding an inline comment to the code for future reference.

gsherwood:
> I'd suggest adding an inline comment to the code for future reference.

Good idea. Have done that now.

PR #2073, opened 2018-06-27
Reports/Summary: sort files based on directory

jrfnl:
The original sort order of the file names in the summary report was alphabetic and did not take subdirectories into account.

This could result in a list like:
```
dir/afile/subdir/file.php
dir/afile.php
dir/b/subdir/file.php
dir/cfile.php
```

The change proposed in this PR sorts the file names in a way that the directory layout is taken into account, so the above list would now display as:
```
dir/afile.php
dir/cfile.php
dir/afile/subdir/file.php
dir/b/subdir/file.php
```

### DIFF
--- a/src/Reports/Summary.php
+++ b/src/Reports/Summary.php
@@ -95,6 +95,27 @@ class Summary implements Report
             $maxLength = max($maxLength, $fileLen);
         }
 
+        uksort(
+            $reportFiles,
+            function ($keyA, $keyB) {
+                $pathPartsA = explode(DIRECTORY_SEPARATOR, $keyA);
+                $pathPartsB = explode(DIRECTORY_SEPARATOR, $keyB);
+
+                do {
+                    $partA = array_shift($pathPartsA);
+                    $partB = array_shift($pathPartsB);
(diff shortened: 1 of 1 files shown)


gsherwood:
I think this is a good change, but I'm wondering if this sorting should go up into the main Reporter class so that the order is consistent for all reports that use filenames. Thoughts?

jrfnl:
> I'm wondering if this sorting should go up into the main Reporter class so that the order is consistent for all reports that use filenames. Thoughts?

At this moment, I've only looked at the reports I use most often, so I'm not actually sure which other reports use filenames.

What about if I change this to a named method in the main Reporter class ?
Depending on whether they need it, child classes can then just reference the method as the callback for a user defined sort, but the call to the sort function itself stays with the child class allowing for different sort orders depending on the report type.

gsherwood:
Ignore me. Most reports are just going to print in the order that the files were checked - very few do the work to print their data in a way that is sortable.

I think this can be merged as is.

jrfnl:
You're very welcome ;-)

PR #2074, opened 2018-06-27
Reports/Source: sort sources on count and source

jrfnl:
The original sort order of the sources in the source report was based on the error/warning count alone.
This would make the resulting order for errors with the same count unpredictable and different on each run, making it more difficult to compare between runs.

The change proposed in this PR sorts the list first on the error/warning count and secondary on the source code, stabilizing the sort order.


gsherwood:
Really good idea. Thanks.

PR #2075, opened 2018-06-27
Reports/Source: remove some code which is never used

jrfnl:
Probably a copy/paste artifact or a left over from an old(er) implementation.


gsherwood:
👍 Thanks

PR #2081, opened 2018-07-04
Improvements

carusogabriel:
Together with PHPStorm and https://github.com/kalessil/phpinspectionsea, here are some improvements that we can make to PHP_CodeSniffer :smile:

### DIFF
--- a/src/Files/LocalFile.php
+++ b/src/Files/LocalFile.php
@@ -48,7 +48,7 @@ class LocalFile extends File
                 fclose($handle);
 
                 if (strpos($firstContent, '@codingStandardsIgnoreFile') !== false
-                    || strpos(strtolower($firstContent), 'phpcs:ignorefile') !== false
+                    || stripos($firstContent, 'phpcs:ignorefile') !== false
                 ) {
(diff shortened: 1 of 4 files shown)


src/Files/File.php
```diff
@@ -776,11 +776,7 @@ public function addFixableError(
         $severity=0
     ) {
         $recorded = $this->addError($error, $stackPtr, $code, $data, $severity, true);
-        if ($recorded === true && $this->fixer->enabled === true) {
-            return true;
-        }
-
-        return false;
+        return $recorded === true && $this->fixer->enabled === true;
```
gsherwood: This isn't how I write code. I prefer to use the IF statement for my own readability.

gsherwood:
If you want to adjust the PR, I'd merge in stripos change, the fixed var name, and the unused imports. But I write my return statements like they are very munch on purpose, so I wouldn't merge those.

carusogabriel:
@gsherwood Done :+1:

gsherwood:
Thanks for doing that for me.

PR #2085, opened 2018-07-06
Remove PHP binary searching, always use PHP_BINARY

willemstuursma:
I have a setup with multiple PHP interpreters installed (e.g. PHP 5.6, 7.0, 7.1 etc). When running `phpcs` with the non standard binary (not `/usr/bin/php`, files get linted with `/usr/bin/php`, leading to false positive syntax error violations. 

Since the minimum supported PHP version is 5.4.0, we can just always use `PHP_BINARY` and not bother with finding the PHP binary ourselves.


jrfnl:
@willemstuursma You are missing the point of that piece of code. PHPCS can be run on a different PHP version than the PHP version targetted by the subject code.

See: https://github.com/squizlabs/PHP_CodeSniffer/pull/1104

willemstuursma:
It can't be linted for a different PHP version unless it has a binary for that right? I'll have a closer look.

If I noticed it correctly, it would just always use the php binary found by which before anyway.

willemstuursma:
So looking at the code again, it is possible to lint against a different version if the `php_path` config property is set.

However, in the `master` if this property is not set, `which php` is executed and the result is returned.

Instead, it should return `PHP_BINARY` if there is no `php_path` set.

I'll update the PR to reflect this.

willemstuursma:
Thanks @jrfnl for pointing this out, I think I got it now.

jrfnl:
@willemstuursma Looks like you're trying to solve this at the `Config` level now, rather than at the `Sniff` level.
In that case, if I remember correctly, there are a number of other sniffs which use this config setting and should probably be adjusted.

I don't have an opinion on whether it will work as expected - I haven't run any tests. 
Have you tested with:
* config set in `CodeSniffer.conf` ?
* config set using a `<config>` directive in a custom ruleset  ?

It looks like the unit tests also could use some adjustment. The logic currently used in `shouldSkipTest()` looks superfluous what with PHP 5.4 being the minimum PHP version.
Also, looks like the unit test will always error for having a parse error. The "no parse error" test is not tested. This would require a secondary unit test `inc` file to be added.
IMHO it would also be great to have a unit test which will use the `php_path` from the config and could pass/fail based on that, to make sure that the `php_path` is always respected properly.

If moving the logic handling to `Config` wasn't your intention, for improving the code at the sniff level, all that would be needed would be to removedthe inner if/else in favour of setting `$this->phpPath = PHP_BINARY;`.

willemstuursma:
> Looks like you're trying to solve this at the Config level now, rather than at the Sniff level.

I think the changes in the Config are appropriate, now it knows where to find the php binary (no longer needs `which php`). 

I agree on the tests, I'll see what I can come up with.

willemstuursma:
> In that case, if I remember correctly, there are a number of other sniffs which use this config setting and should probably be adjusted.

I checked, there are no other sniffs that need the PHP executable's path.

I removed the `shouldSkipTest` and added a unit test for the Config behaviour.

The other problems you pointed out, while valid, probably deserve a dedicated issue and PR.

src/Config.php
```diff
@@ -179,7 +179,7 @@ class Config
      *
      * @var array<string, string>
      */
-    private static $executablePaths = [];
+    private static $executablePaths = ["php" => PHP_BINARY];
```
gsherwood: I'd prefer an exception for PHP to be made directly in the getExecutablePath() method so that when (inevitably) code moves around in there, this change doesn't cause issues.

src/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
```diff
@@ -48,20 +41,8 @@ public function register()
...
-                    $this->phpPath = PHP_BINARY;
-                } else {
-                    return;
-                }
-            }
-        }
-
         $fileName = escapeshellarg($phpcsFile->getFilename());
-        $cmd      = escapeshellcmd($this->phpPath)." -l -d display_errors=1 -d error_prepend_string='' $fileName 2>&1";
+        $cmd      = escapeshellcmd(Config::getExecutablePath('php'))." -l -d display_errors=1 -d error_prepend_string='' $fileName 2>&1";
```
gsherwood: As the same sniff object is used for every file, this change results in calling Config::getExecutablePath('php') for every single file being checked. You removed all the path caching that this sniff had in place to improve performance, so you'll need to put the private member var back, along with the IF statement above. It can just be made far simpler now.

tests/Core/AllTests.php
```diff
@@ -46,6 +47,7 @@ public static function main()
     public static function suite()
     {
         $suite = new TestSuite('PHP CodeSniffer Core');
+        $suite->addTestSuite('PHP_CodeSniffer\Tests\Core\ConfigTest');
```
gsherwood: I don't think this test is required - it can be removed. I would prefer any tests to be well thought out and submitted in a different PR (if you want to do that testing) rather than a quick adhoc test case. I really appreciate you doing it, but I don't think it is required for this change.

jrfnl: Sorry, that's probably due to a remark I made earlier which was misunderstood.

I meant testing that the config is respected just for this sniff.

src/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
willemstuursma: My motivation for not putting it back, is that the `Config` class already caches it. Its kinda like that class's responsibility to cache config values, right?

tests/Core/AllTests.php
willemstuursma: It's useful now that you asked me to move the PHP_BINARY constant around.

src/Standards/Generic/Sniffs/PHP/SyntaxSniff.php
gsherwood: Config:: getExecutablePath() will call Config::getConfigData() which will call Config::getAllConfigData() and an isset(). This will happen for every file being checked.

It doesn't sound like a lot, but it has an impact at scale. When I get time to do profiling, these are the sort of caches I put in to try and cut down the number of overall function calls and lines of code executed. I'd prefer the cache to be left in place to ensure a consistent performance profile across versions.

willemstuursma:
@gsherwood the requested changes have been made.

gsherwood:
> anything I need to do to move this forward?

No. I just need to find time.

gsherwood:
Thanks for making those changes.

PR #2086, opened 2018-07-08
Fix compatibility with PHP 7.3/nightly

jrfnl:
As of PHP 7.3, using `continue` to target a `switch` control structure will throw an `E_WARNING`.

Applying `continue` to a `switch` is equivalent to using `break` and more often than not, a `continue` targeting a higher level control structure is actually intended.

To target the higher level control structure, a numeric argument has to be passed to `continue`.

Refs:
* http://php.net/manual/en/control-structures.continue.php
* https://github.com/php/php-src/pull/3364
* https://wiki.php.net/rfc/continue_on_switch_deprecation

This change has been recently committed to PHP itself and will break the build against `nightly` for the current `master` already.
See: https://travis-ci.org/jrfnl/PHP_CodeSniffer/jobs/401533444 (this branch, but without commits, so equal to the current PHPCS `master`).

This PR fixes the instances of this in PHPCS so [the build against `nightly` will pass again](https://travis-ci.org/jrfnl/PHP_CodeSniffer/jobs/401536827).

Notes:
* All these switches are nested in other control structures, but end the switch at the scope closer of the outer control structure, so using `break` in these cases is equivalent to using `continue 2`.

---
FYI: The [PHPCompatibility](https://github.com/wimg/PHPCompatibility) standard will include a sniff to detect this in the near future. This PR is the result of some tests I've been running on the WIP PHPCompatibility sniff :wink:

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -1408,7 +1408,7 @@ class File
                 // If it's null, then there must be no parameters for this
                 // method.
                 if ($currVar === null) {
-                    continue;
+                    continue 2;
                 }
 
                 $vars[$paramCount]            = [];
--- a/src/Tokenizers/CSS.php
+++ b/src/Tokenizers/CSS.php
@@ -390,7 +390,7 @@ class CSS extends PHP
 
                     // Needs to be in the format "url(" for it to be a URL.
                     if ($finalTokens[$x]['code'] !== T_OPEN_PARENTHESIS) {
-                        continue;
+                        continue 2;
                     }
 
                     // Make sure the content isn't empty.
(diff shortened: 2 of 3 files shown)


gsherwood:
> Applying continue to a switch is equivalent to using break and more often than not, a continue targeting a higher level control structure is actually intended.

And I'm sure that's what I intended in all these cases. It's just lucky that there is no code after the switch in all of them.

I know that using break here is the equivalent of what the code actually does, but it's not what the intention of the developer was. So I think using `continue(2)` might be the way to go here so that I can later differentiate between times I wanted to only break out of the switch, and times when I really wanted the loop to continue.

jrfnl:
@gsherwood I've updated (amended) the PR to reflect your remark. Should be good to go.

gsherwood:
Thanks a lot for finding and fixing these. I think that PHPCompat sniff is going to come in handy.

jrfnl:
> Thanks a lot for finding and fixing these. I think that PHPCompat sniff is going to come in handy.

You're welcome ;-)

The PR for the PHPCompatibility sniff has been opened: https://github.com/wimg/PHPCompatibility/pull/687 (in case other people are looking for it),

PR #2088, opened 2018-07-10
added errors attribute with value zero to junit report testsuites

nicholascus:
'errors' attribute is required by xunit jenkins plugin (that's weird cause junit itself seems to define that parameter optional). Still it would be good to have PHP_CodeSniffer junit xml report compatible with Jenkins xunit plugin.

https://github.com/jenkinsci/xunit-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/xunit/types/model/xsd/junit-10.xsd

### DIFF
--- a/src/Reports/Junit.php
+++ b/src/Reports/Junit.php
@@ -39,6 +39,7 @@ class Junit implements Report
 
         $out->startElement('testsuite');
         $out->writeAttribute('name', $report['filename']);
+        $out->writeAttribute('errors', 0);
 
         if (count($report['messages']) === 0) {
             $out->writeAttribute('tests', 1);


nicholascus:
image
currently we are forced to use workaround and update generated xmls with `sed -i -- 's/failures=/errors="0" failures=/g' php53comp-www-xunit.xml`
pipeline operation failing without this fix is
```
    xunit(
        testTimeMargin: '600000',
        thresholdMode: 2,
        thresholds: [
            [$class: 'FailedThreshold', failureNewThreshold: '0', failureThreshold: '100', unstableNewThreshold: '0', unstableThreshold: '100'],
            [$class: 'SkippedThreshold', failureNewThreshold: '15', failureThreshold: '20', unstableNewThreshold: '10', unstableThreshold: '15']],
        tools: [JUnit(deleteOutputFiles: true, failIfNotNew: true, pattern: 'php53comp-www-xunit.xml', skipNoTestFiles: true, stopProcessingIfError: false)])
```

gsherwood:
Thanks a lot for reporting and fixing this. I also added errors to the testsuites tag (even though it is optional) for consistency.

PR #2104, opened 2018-07-28
Unit test runner: ignore `bak` files

jrfnl:
Just a pet peeve.

Some IDE/code editors create `.bak` files when a file has been edited which cause the unit test runs to fail as these files would be tested as if they were intentional test case files.

This minor fix prevents this from happening.


gsherwood:
Good idea. Thanks.

PR #2105, opened 2018-07-29
Generic: fix unintentional errors in test case files

jrfnl:
[Fixer conflicts series PR]

While looking through the remaining fixer conflicts, I did a quick check to see if any of them may be caused by parse or compile errors in the test case files. _Hint: Yes, some are._

In my opinion, sniffs should be resilient against parse errors caused by live coding so they can be used in an IDE, but do not need to account for every single parse/compile error possible which could prevent it from working.
Some inaccurate results when a sniff is run on a file with a parse/compile error should be expected and acceptable.

So to this end, I've linted the test case files and fixed any parse errors and compile errors in these _which were not explicitly testing for the handling of parse errors by the sniff_.

I've chosen to err on the side of caution, only fixing parse/compile errors which were clearly not testing anything in the sniff and left parse/compile errors in when in doubt.

This is the part of a series of PRs doing this for the test case files in each standard.

Most parse errors were of the following types:
* Missing semi-colons after the close brace of anonymous functions/classes.
* Missing comma's after array items.
* Classes with parentheses after the class name `class ABC() {}`.
* Visibility keywords before global function declarations.

Compile errors were mostly:
* Duplicate function names used.
* Missing required arguments for magic methods.
* Declararing protected/private methods in interfaces.

This PR will be easiest to review by looking at the individual commits in combination with the more detailed commit messages.

This PR also contains a few QA improvements to some test case files where the unit tests weren't actually testing anything (new) or only testing invalid code. These kind of improvements are each in separate commits.


src/Standards/Generic/Tests/CodeAnalysis/UnusedFunctionParameterUnitTest.inc
```diff
@@ -77,4 +77,4 @@ function bar($x)
 
 function ($a, $b) {
     return $a * 2;
-}
+};
```
gsherwood: Is this change needed?

src/Standards/Generic/Tests/PHP/LowerCaseKeywordUnitTest.inc
```diff
@@ -1,25 +1,25 @@
...
 GOTO a;
-function (Array $a) {}
+function (Array $a) {};
 const PRIVATE;
 HttpStatus::CONTINUE;
 Function ($f) {
     Yield $f;
     Yield From fun();
-}
+};
```
gsherwood: This one might have been better fixed with `$foo = Function`...

jrfnl: That wouldn't fix the parse error.

src/Standards/Generic/Tests/CodeAnalysis/UnusedFunctionParameterUnitTest.inc
jrfnl: Both this one as well as the one below are closures. If the `;` is missing after the closing brace, it is considered a parse error by PHP.

gsherwood: /facepalm I completely missed the lack of function name in there.

src/Standards/Generic/Tests/PHP/LowerCaseKeywordUnitTest.inc
gsherwood: Sorry, I meant assigning to a var *as well as* the semicolon on the end - just to make it look somewhat closer to real code. That's what tripped me up with the one above.

But honestly, it's not big deal. I just started reviewing stuff because I had a few spare minutes and realised I got through a single commit before I ran out of time.

jrfnl: @gsherwood I've added a new commit which addresses this for all such test code involving closures for the `Generic` sniffs.

PR #2106, opened 2018-07-29
PEAR: fix unintentional errors in test case files

jrfnl:
[Fixer conflicts series PR]

While looking through the remaining fixer conflicts, I did a quick check to see if any of them may be caused by parse or compile errors in the test case files. _Hint: Yes, some are._

In my opinion, sniffs should be resilient against parse errors caused by live coding so they can be used in an IDE, but do not need to account for every single parse/compile error possible which could prevent it from working.
Some inaccurate results when a sniff is run on a file with a parse/compile error should be expected and acceptable.

So to this end, I've linted the test case files and fixed any parse errors and compile errors in these _which were not explicitly testing for the handling of parse errors by the sniff_.

I've chosen to err on the side of caution, only fixing parse/compile errors which were clearly not testing anything in the sniff and left parse/compile errors in when in doubt.

This is the part of a series of PRs doing this for the test case files in each standard.

Most parse errors were of the following types:
* Missing semi-colons after the close brace of anonymous functions/classes.
* Missing comma's after array items.
* Classes with parentheses after the class name `class ABC() {}`.
* Visibility keywords before global function declarations.

Compile errors were mostly:
* Duplicate function names used.
* Missing required arguments for magic methods.
* Declaring protected/private methods in interfaces.

This PR will be easiest to review by looking at the individual commits in combination with the more detailed commit messages.

This PR also contains a few QA improvements to some test case files where the unit tests weren't actually testing anything (new) or only testing invalid code. These kind of improvements are each in separate commits.


gsherwood:
> Duplicate function names used.

Don't make me use unique function names in test classes... please :)

jrfnl:
> Don't make me use unique function names in test classes... please :)

I figured that might be contentious, which is why it was in a separate commit. :sunglasses:  I'll remove that particular commit from both this PR as well as the other one.

jrfnl:
FYI: I've removed the commit from both PRs & force pushed the branches, so should be good, or at least, better now ;-)

jrfnl:
You're welcome. Will try to do the same for the PSR1/2/12/Squiz and Zend standards in the near future.

PR #2109, opened 2018-07-31
Generic.Functions.CallTimePassByReference false positive for bitwise and used in function argument

Birzenbae:
For example, "someFunction( someConst & $var )" will not be detected as passed reference anymore.

### DIFF
--- a/src/Standards/Generic/Sniffs/Functions/CallTimePassByReferenceSniff.php
+++ b/src/Standards/Generic/Sniffs/Functions/CallTimePassByReferenceSniff.php
@@ -130,6 +130,7 @@ class CallTimePassByReferenceSniff implements Sniff
                     || $tokenCode === T_CLOSE_PARENTHESIS
                     || $tokenCode === T_CLOSE_SQUARE_BRACKET
                     || $tokenCode === T_LNUMBER
+                    || $tokenCode === T_STRING
                     || isset(Tokens::$assignmentTokens[$tokenCode]) === true
                 ) {
                     continue;
--- a/src/Standards/Generic/Tests/Functions/CallTimePassByReferenceUnitTest.inc
+++ b/src/Standards/Generic/Tests/Functions/CallTimePassByReferenceUnitTest.inc
@@ -1,5 +1,7 @@
 <?php
 class myclass extends yourclass implements someint {
+    const MY_CONST = 0777;
+
     function myfunc($var) {
         echo $var;
     }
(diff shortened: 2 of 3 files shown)


jrfnl:
@Birzenbae Please help me out as I'm trying to understand, but if I read your code sample above, I see a "bitwise and", not a passing by reference. What am I missing ?

Birzenbae:
@jrfnl Okay, so without this fix "someFunction( someConst & $var )" and "someFunction( MyClass::someConst & $var )" are detected as passing a reference into a function call.
This is a false positive because it is a "bitwise and". Therefore the additional T_STRING in the blacklist to prevent this behavior.
The new test lines would fail without this fix.

I'm sorry, i guess my commit message is a bit confusing.

jrfnl:
@Birzenbae Ah, thanks! That explains it. I haven't looked at the sniff in detail, but I know there is a `File::isReference()` method. Might be worth looking into whether that method is/can be used and if that needs improvement (too).

jrfnl:
Ok, just had a closer look at the sniff and [its sister in the PHPCompatibility standard](https://github.com/PHPCompatibility/PHPCompatibility/blob/master/PHPCompatibility/Sniffs/PHP/ForbiddenCallTimePassByReferenceSniff.php) and ran both sniffs against each others unit tests.

The PHPCompatibility sniff already took the test case @Birzenbae added into account as it *does* use the `$phpcsFile->isReference()` method.
All the same, it didn't catch the `$foo(&$myvar);` test case (line 25 PHPCS unit test case file) and it gave a false positive for `Hooks::run( [ &$updates ] );` (line 30 PHPCS unit test case file).
Both these issues will be fixed by downstream PR https://github.com/PHPCompatibility/PHPCompatibility/pull/723.

@Birzenbae's fix is good for the test case added in this PR.
The PHPCS sniff, however will still generate false positives for comparisons with references when passed as function call parameters such as `efg( true === &$b );` (line 74, 75 of the [PHPCompatibility unit test case file](https://github.com/PHPCompatibility/PHPCompatibility/blob/8c41eac07a053b8c78cd9dc38c498ffb6687170f/PHPCompatibility/Tests/sniff-examples/call_time_pass_by_reference.php#L74-L75)).

gsherwood:
I couldn't get the patch for this PR, but I ended up fixing it using isReference() anyway + leaving in the custom check for assignment usage. Thanks for reporting this error.

PR #2110, opened 2018-08-03
Squiz.WhiteSpace.FunctionSpacing is removing indents from the start of functions when fixing

jrfnl:
When too many blank lines are found before a function, the `Before/BeforeFirst` fixer would unnecessarily also remove the indentation of the `$nextContent` line, meaning that this sniff would **always** need to be accompanied by one or more sniffs to fix the indentation again.

As this sniff is about blank lines between functions, not about indentation, I consider this a bug.

This minor change fixes it.

Includes unit tests demonstrating the issue.


jrfnl:
Just realized that the same problem existed when there would be too few blank lines before a function.
The fixer would remove the indentation for the `$nextContent` line and leave trailing indentation on the first added "blank" line. I've added a new fix for this in a separate commit, including unit tests.

This new fix also documents how this sniff handles PHPCS annotations found before a function - which may not be as desired -, but I haven't changed the behaviour regarding this.

gsherwood:
100% agree. Thanks for fixing this.

PR #2116, opened 2018-08-10
Fixes suggested variable type with single-line comment

### LINKED ISSUE #2115: Squiz.Commenting.VariableComment not checking var types when the @var line contains a comment
Code
------
```php
/**
 * @var int Type
 */
public $type;
```

Actual Result
--------------
Produces no error

Expected Result
-----------------
Expected "integer" but found "int" for @var tag in member variable comment

josephzidell:
Fixes #2115

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/VariableCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/VariableCommentSniff.php
@@ -98,7 +98,7 @@ class VariableCommentSniff extends AbstractVariableSniff
             return;
         }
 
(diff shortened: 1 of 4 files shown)


gsherwood:
Thanks a lot for submitting this fix. I'm not sure why the tests didn't fail when the PR was submitted, but I've noticed that the auto fixing for this sniff now removes the comment after the var type, so a few tests are failing. I can fix up that bit though.

gsherwood:
Actually, I've just noticed that this fix doesn't work for complex array types like `array<string, string|null>` as it splits on space. I'll have to revert this and come up with a new fix.

PR #1219, opened 2016-11-23
added return type declaration sniff

akvankorlaar:
Created according to: https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md

For return type declarations, this sniff checks the spacing between the functions closing parenthesis, and the colon. It also checks the spacing between the colon and the return type. The spacings are still configurable

It does not check if the colon and return type are on the same line as the closing parenthesis. I can also add that if required.. but maybe that has to be placed in a separate sniff?

Edit: Removed requirement that return type declaration should be lowercase, because object names can also be in the return type declaration

Update: Added fixer for return type declaration sniff

PR #2118, opened 2018-08-12
[PSR12] Add return type sniff

duncan3dc:
Following on from the amazing work done by @akvankorlaar in #1219 this PR:
* Updates it to handle the breaking change from 1f55125d76385c48a36801628d40be1f482cf7ac
* Changes the auto-fixer to prevent it from affecting inline comments (pointed out by @jrfnl and @gsherwood)
* Add some tests for inline comments
* Defines it as part of PSR-12

### DIFF
--- a/src/Standards/Generic/Docs/Functions/ReturnTypeDeclarationStandard.xml
+++ b/src/Standards/Generic/Docs/Functions/ReturnTypeDeclarationStandard.xml
@@ -1,0 +1,27 @@
+<documentation title="Return Type Declaration">
+    <standard>
+        <![CDATA[
+        Return type declarations are optional, and have a specified amount of spaces between the function's closing parenthesis and colon, and a
+        specified amount of spaces between the colon and the return type.
+        The default is zero spaces between closing parenthesis and colon and one space between colon and return type.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Zero spaces between closing parenthesis and colon, one space between colon and return type.">
+            <![CDATA[
+function fooFunction()<em>: string</em>
+{
+...
+}
+        ]]>
+        </code>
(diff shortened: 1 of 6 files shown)


gmponos:
Hello, I hope this gets merged soon. Or the other PR that you mentioned on the comment.. what ever.

Personally neither on this PR or on the other I see tests related about nullable return types. I hope I didn't missed them. I believe related tests should exist.

gsherwood:
Just a note that the auto-fixer for this does not correctly fix nullable return types:
```diff
$ cat temp.php
<?php
function this():  ?string
{
    return null;
}
$ phpcs temp.php --standard=PSR12 --report=diff
--- temp.php
+++ PHP_CodeSniffer
@@ -1,5 +1,5 @@
 <?php
-function this():  ?string
+function this():?string
 {
     return null;
 }
```

The error messages are also a bit strange as they try to show what the content should be rather than the standard way of saying how much spacing is required. I'll have a look at changing both these things.

gsherwood:
I've had a deeper look into this code and I'd basically need to write the sniff to make it work the way I would expect. I appreciate the work people have done on this, but the code style, method of checking spacing, var names, unit tests all need changing if I'm going to accept it into the core and maintain it from now on. I started to make those changes and realised it would be faster to rewrite the sniff and tests, but I don't have the time to do that so I've removed it from the 3.4.0 milestone.

lkraav:
Hi @gsherwood since this PR was closed, but no activity on #1219 either since 2018, is there another PR where this is being worked on?

gsherwood:
> Hi @gsherwood since this PR was closed, but no activity on #1219 either since 2018, is there another PR where this is being worked on?

Not that I know of.

lkraav:
OK, I went with Slevomat for time being so I think that solves it for most cases.

PR #2103, opened 2018-07-27
Report Info: various fixes to improve readability of multi-line metrics

jrfnl:
As discussed in #2102, this PR makes the following changes to improve the readability of the `Info` report for multi-line metrics:

1. Start the first metric value on a new line.
2. Align the arrows, and by extension the counts, of the metrics.
3. Move the totals to a separate line underneath the list of metric values.
4. Right align the counts.
5. Add thousands separators to the counts.
6. Right align the percentages.
7. Stop it reporting 0% for metrics and report <0.01% instead.

Fixes #2102

### Screenshot before (on Windows):
report-info-before

### Screenshot after (on Windows):
report-info-after

PR #2119, opened 2018-08-12
Report info: minor sorting improvement

jrfnl:
Just realized that I hadn't added this to the previous PR #2103 yet.

As the metric keys may be numeric, the `SORT_NATURAL` flag (introduced in PHP 5.4) will ensure that numbers with equal counts will be sorted properly, i.e.:
Without this these would be sorted like:
```
1  => 10 (... %)
12 => 10 (... %)
2  => 10 (... %)
```
Now they will be sorted like so:
```
1  => 10 (... %)
2  => 10 (... %)
12 => 10 (... %)
```

### DIFF
--- a/src/Reports/Info.php
+++ b/src/Reports/Info.php
@@ -123,7 +123,7 @@ class Info implements Report
 
                 echo "$metric:".PHP_EOL;
 
-                ksort($values);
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks

PR #2121, opened 2018-08-15
Squiz.PHP.DisallowMultipleAssignments false positive in while loop conditions

jrfnl:
A variable assignment in the condition of a `while` loop may well be enclosed in additional parenthesis to prevent issues with operator precedence when comparing the assigned value within the same condition.

Example:
```php
while ( ( $csvdata = fgetcsv( $handle, 2000, $separator ) ) !== false ) {}
```

Previously only the lowest level `parenthesis_owner` would be examined to determine whether or not to report this as an error. Now, for `while` loops only, the parenthesis owners of all nested parenthesis will be examined.

Examining them early on in the sniff as the rest of the logic is superfluous if the assignment should be ignored anyway.

Includes unit test.


gsherwood:
Thanks!

PR #2127, opened 2018-08-19
File::findExtendedClassName() doesn't support nested classes

jrfnl:
The method used the class scope closer to end searches, which meant that a non-extended class which contained a nested extended class would return the extended class name of the nested class, instead of `false`.

Switching this to the scope opener, fixes this.

Included unit tests.

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -2377,12 +2377,12 @@ class File
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


gsherwood:
That's a strange one. I wonder why I would have picked the scope closer over the opener when looking for an extends keyword 🤷‍♂️ 

Thanks for fixing it.

jrfnl:
You're welcome.

PR #2129, opened 2018-08-20
FindImplementedInterfaceNamesTest: Remove redundant line

jrfnl:
Looks like this line was copied/pasted at some point, but has no function in this test file, so removing it.


gsherwood:
I think you're right. Thanks.

PR #2131, opened 2018-08-20
Adds InconsistentReturnSniff

donatj:
I am looking to see if there is any interest in this Sniff I wrote and have been using in my personal code.  I am quite open to any feedback.

Adds a sniff to detect when not all code paths within a function return / throw.

For example:

```php
function foo($bar) {
    if($bar) {
        return true;
    }
}
```

It correctly handles return/throws from Switch `default`'s as well as `catch`'s

If there is interest in my sniff I will **very happily** add tests to back it up.

### DIFF
--- a/src/Standards/Generic/Sniffs/Functions/InconsistentReturnSniff.php
+++ b/src/Standards/Generic/Sniffs/Functions/InconsistentReturnSniff.php
@@ -16,6 +16,7 @@ use PHP_CodeSniffer\Sniffs\Sniff;
 class InconsistentReturnSniff implements Sniff
 {
 
+
     /**
      * Returns an array of tokens this test wants to listen for.
      *
@@ -23,10 +24,10 @@ class InconsistentReturnSniff implements Sniff
      */
     public function register()
     {
-        return [
-            T_FUNCTION
-        ];
-    }
+        return [T_FUNCTION];
+
+    }//end register()
+
 
     /**
      * Processes this test, when one of its tokens is encountered.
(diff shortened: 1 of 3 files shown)


gsherwood:
This is an interesting sniff. I ran it over PHPCS itself so see how it goes and it picked up a few functions that are in this sort of form:
```php
function foo()
{
    if ($foo === true) {
        return $foo;
    } else {
        if ($bar !== false) {
            if ($baz !== 0) {
                return $baz;
            } else {
                return $bar;
            }
        } else {
            return false;
        }
    }

}
```
I obviously tend to write code like this sometimes, where the function itself wont actually return void due to the conditional structure, but I don't put the final return outside the conditions. If I rewrite the code like this:
```php
function foo()
{
    if ($foo === true) {
        return $foo;
    } else {
        if ($bar !== false) {
            if ($baz !== 0) {
                return $baz;
            } else {
                return $bar;
            }
        }
    }

    return false;

}
```
Then everything is fine.

I actually don't mind changing my functions to use the second style, but obviously not everyone is going to want to do the same. They have the option of not using the sniff, but I'm wondering if there is anything that can be done to make it useful for everyone. I'm trying to stop the inevitable bug reports where a sniff doesn't work exactly like someone wants it to, but I realise this might be hard as you'd need to inspect the conditions more and track the ELSE statements. Still, it might be something you'd like to try?

My only other suggestion is that the error message and error token make it hard to figure out which paths are missing the return statement as you need to track your way through the function and locate them by hand. I know that might be asking a bit much given paths have intentional fall-throughs, but I thought I'd put it to you anyway.

As for accepting it into the core, it all depends on how well tested it is. Once I merge it, I need to fix all the bugs people find with it, and changes required with new PHP versions.  So having a good test suite that lays out what it can and cannot detect is critical. So if you want to give that a go, I'd be happy to do a bit more testing. I'd suggest running the sniff over a few projects with large code bases just to make sure you've got a few different styles of coding covered.

donatj:
I'm glad to hear there is interest 😄 

I think fixing the else logic should be relatively simple. I'll work on that first.

I'm sure it simply never came up in my own testing because I have php-cs-fixer's `no_useless_else` fix enabled in our CI suite, which would have prevented that from ever cropping up.

I will go about running this on some large code bases as you suggested, as well as getting a fair number of tests prepared.

donatj:
@gsherwood Is there a built in way with the phpunit Test loader to run just a specific test? 

The entire suite takes a couple minutes on my machine, and when I try to limit what I run I receive:

```
Undefined index: PHP_CODESNIFFER_TEST_DIRS
 /Users/jdonat/Projects/donatj/PHP_CodeSniffer/tests/Standards/AbstractSniffUnitTest.php:59
```

donatj:
Testing has already proven fruitful. I've found a number of cases that lead to false "Good" results. I'll report back on getting these resolved.

donatj:
I thought I closed this a long time ago. When I started poking edge cases it ended up way more complex than I could handle.

PR #2134, opened 2018-08-22
FindImplementedInterfaceNamesTest: Girlscouting & two additional tests

jrfnl:
* Replace all the individual tests which are basically all duplicates of each other, with one test method and a data provider.
* Add two more tests
    Redundancy - just making sure for now and later that the method will not get confused over class declarations which both `extend` as well as `implement`.


gsherwood:
Thanks for this one too :)

PR #2137, opened 2018-08-22
Filters: New `GitStaged` filter

jrfnl:
With the new file added in this commit, a new filter becomes available which can be used in a `pre-commit` git hook, as in:
```bash
phpcs --filter=gitstaged
```
... to only examine the files which you are about to commit.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -127,6 +127,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
      <file baseinstalldir="PHP/CodeSniffer" name="ExactMatch.php" role="php" />
      <file baseinstalldir="PHP/CodeSniffer" name="Filter.php" role="php" />
      <file baseinstalldir="PHP/CodeSniffer" name="GitModified.php" role="php" />
+     <file baseinstalldir="PHP/CodeSniffer" name="GitStaged.php" role="php" />
     </dir>
     <dir name="Generators">
      <file baseinstalldir="PHP/CodeSniffer" name="Generator.php" role="php" />
--- a/src/Filters/GitStaged.php
+++ b/src/Filters/GitStaged.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * A filter to only include files that have been staged for commit in a Git repository.
+ *
+ * This filter is the ideal companion for your pre-commit git hook.
+ *
+ * @author    Juliette Reinders Folmer <phpcs_nospam@adviesenzo.nl>
+ * @copyright 2018 Juliette Reinders Folmer. All rights reserved.
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ */
+
+namespace PHP_CodeSniffer\Filters;
+
+use PHP_CodeSniffer\Util;
+
+class GitStaged extends ExactMatch
+{
+
+
+    /**
+     * Get a list of blacklisted file paths.
+     *
+     * @return array
+     */
+    protected function getBlacklist()
+    {
+        return [];
+
+    }//end getBlacklist()
+
+
+    /**
+     * Get a list of whitelisted file paths.
+     *
+     * @return array
+     */
+    protected function getWhitelist()
+    {
+        $modified = [];
+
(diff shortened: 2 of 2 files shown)


gsherwood:
@jrfnl Any thoughts on the above comment?

jrfnl:
> However, the current implementation will not work properly, as it only looks at the filenames.
>
> What you need is a way to run it over the actual staged content of the files.
> 
> If you have a name only, you will get "the current disk-based content of files that contain staged changes", but the actual staged changes might be different than the disk-based content.

@gsherwood Sorry, yes, been meaning to respond to this, but hadn't got round to it so far.

@schlessera Thank you for your feedback.

While I understand your objections, in my opinion, this is still the best solution.

1. Making a copy of the file to be examined _as staged_ to a temporary location and sniffing that version of a file would break any custom ruleset which uses `<file>`, `<exclude-pattern>` or `<include-pattern>` directives which include (partial) file paths.
    The same applies for a detailed command set in the precommit hook using `--ignore`, `--file-list` or a specific `<file>` directives.
2. Not staging additional changes made after the initial staging of a file, is a work-flow problem, not a problem with this filter as such.
    This is no different to running `phpcs` from the command-line before committing your staged files, fixing them and then forgetting to stage the additional changes.
    This filter just allows you the convenience of automating that step of having to run phpcs manually before committing a change and allows PHPCS to be a lot quicker about it as it only examines the relevant files.
3. I would expect most projects which use PHPCS to also include it in their build testing via Travis or other CI tools.
    If you _would_ forget to stage the fixes you made after the commit was initially rejected, the CI tool would (should) reject the change anyway, which should prevent the "dirty" changes from getting into your codebase.

The only solution I can think of to solve the problem you point out, is to:
1. Get a list of the staged files.
2. Make a copy of the original file(s) as they are to some temporary location.
3. Adjust the *real* file(s) to only contain the staged changes.
4. Run `phpcs` over the *real* (adjusted) file(s) as they are on disk in their real location and report back.
5. Remove the adjusted file(s) and move the original file(s) back.

The above could possibly be done using git worktrees.

With this flow, the file would be examined as staged and the ruleset would still be respected. No need to unstage/stage in between as the list of staged files from (1) would be used and after the file resets, the staging should still be the same. 

However... this is not something PHPCS can (currently) do using filters, as filter are set up to (yes) _filter_ the list of files to be examined at the start of a run and there is nowhere (I know of) to hook into PHPCS at the end of a run to revert the changes made.
_Quite apart from the question of whether it within the purview of PHPCS to actually make those kind of changes to the file system in the first place, no matter whether temporary or not._

So, while I understand where you are coming from, this is the best I can currently come up with, though I'm open to thought on how to improve the filter (without rewriting PHPCS completely).

---

As a side-note: the exact same observations made about this filter, also apply to the existing `GitModified` filter which also examines the whole file as it is on disk (in that case, for files containing unstaged + staged changes, while only unstaged "should" be examined).

jrfnl:
Another thought which just occurred to me: what about if the staged file list contains a (changed) custom ruleset, i.e. the `[.]phpcs.xml[.dist]` file ?

Should the filter in that case be disabled and let PHPCS run over all files in a project ?

gsherwood:
> Quite apart from the question of whether it within the purview of PHPCS to actually make those kind of changes to the file system in the first place, no matter whether temporary or not.

Even if filters could do this, I don't think they should. It sounds like you agree.

I actually don't think there is anything wrong with this filter as-is. Yes, there are cases where it could be tripped up, but it's just there to filter the file list, not add any additional functionality to PHPCS.

jrfnl:
@schlessera Alain, just checking: did you have a chance to have a look at my reply yet ?

jrfnl:
FYI: I've added a second commit to this PR to apply the same fix as applied in #2393 to the `GitModified` filter to the new `GitStaged` filter.

Props: @lmanzke

gsherwood:
Merged now. Thanks for the filter - works well for me.

morozov:
> Having a way to only run PHPCS against the staged changes would be awesome.

FWIW, https://github.com/diff-sniffer/diff-sniffer does this. It doesn't copy the staged changes anywhere (it analyzes the output of `git show :/path/to/file`) and respects all include/exclude rules.

PR #2139, opened 2018-08-25
Squiz/CommentedOutCode: fix false positive + false negative

jrfnl:
### Squiz/CommentedOutCode: fix false positive for PHPMD suppression annotation

Updated the regex used to detect annotations from other tools, to also allow for PHPMD suppression annotations.

This gets rid of the false positive on unit test case file line 138.

Depending on end-user reports, the regex can be improved further.

### Squiz/CommentedOutCode: fix false negative

After the changes I made in PR #1839, this code now added to the unit test case file was incorrectly no longer recognized as commented out code.

This PR further iterates on the previous changes and implements the following:
* Treat a line comment followed by a block comment (and visa versa) as two separate comments.
* Allow for blank lines within block comments, but not within line comment blocks.


gsherwood:
Thanks

PR #2141, opened 2018-08-26
To line up with line 37

alnutile:
This way, either install choice, it ends up in dev

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ php phpcbf.phar -h
 ### Composer
 If you use Composer, you can install PHP_CodeSniffer system-wide with the following command:
 
-    composer global require "squizlabs/php_codesniffer=*"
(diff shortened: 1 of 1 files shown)


gsherwood:
I'm not actually sure what `--dev` does here. The goal of this command is just to have PHPCS always available on your machine, whereas the composer.json section is about using it as a project dependency on this and other machines. Can you explain why `--dev` is needed and what it does?

gsherwood:
> This should probably be a separate guideline on how to to install on a per-project basis.

The README contains two sections for composer. For a global install, it doesn't use a dev dependency because I'm not really sure what that means in the global context given you aren't installing a new copy for every project. The second section include PHPCS in the dev dependancy list as it is describing how to install it per-project. I'm not sure what else I need here.

PR #2143, opened 2018-08-28
PSR2.Namespaces.UseDeclaration does not properly fix "use function" and "use const" statements

thewilkybarkid:
Currently works correctly with curly braces, but turns function and constant uses into classes.


gsherwood:
Clear report. Unit tests. Good fix. Perfect PR - thanks so much for reporting and fixing. 👍 👍

PR #2148, opened 2018-08-31
Fix abstract/interface methods in PEAR.Functions.FunctionDeclaration

thewilkybarkid:
Currently methods without a body fail to have the whitespace before the opening parenthesis and before the semicolon removed.


src/Standards/PEAR/Sniffs/Functions/FunctionDeclarationSniff.php
```diff
@@ -105,7 +105,7 @@ public function process(File $phpcsFile, $stackPtr)
         // Unfinished closures are tokenized as T_FUNCTION however, and can be excluded
         // by checking for the scope_opener.
         if ($tokens[$stackPtr]['code'] === T_FUNCTION
-            && isset($tokens[$stackPtr]['scope_opener']) === true
+            && (isset($tokens[$stackPtr]['scope_opener']) === true || $tokens[$stackPtr]['level'] === 1)
```
gsherwood: I don't think that checking the token level is the best way to do this. It wouldn't work, for example, if you had the class wrapped in a condition (like an autoloader maybe) and while I can't think of why you'd define an abstract method in an anon class, it wouldn't catch that either. It's probably best to call File::getMethodProperties() and check the isAbstract array index to be sure.

thewilkybarkid: Added a failing test in eeb758af9b599d910304032110e26a8fd4cd30b6, will have another look at the implementation.

thewilkybarkid:
@gsherwood Couldn't use `is_abstract` as it doesn't cover methods on interfaces, but used the same technique to determine if there's a body or not.

gsherwood:
Thanks a lot for making those changes and including tests.

Issue #2153, opened 2018-09-05
Support property-level extending for array properties


Majkl578:
Adding new property entries for array properties (i.e. ForbiddenFunctions and more in user-land) is cumbersome as one has to copy&paste all previous values and add new.

My proposal is to allow extending previously declared property values by adding new entries instead of replacing them:
base ruleset:
```xml
<rule ref="Generic.PHP.ForbiddenFunctions">
    <properties>
        <property name="forbiddenFunctionNames" type="array">
            <element key="foo" value="fixedFoo"/>
        </property>
    </properties>
</rule>
```

extending (project-level) ruleset:
```xml
<rule ref="Generic.PHP.ForbiddenFunctions">
    <properties>
        <property name="forbiddenFunctionNames" type="array" extend="true">
            <element key="bar" value="fixedBar"/>
        </property>
    </properties>
</rule>
```

This would result in Generic.PHP.ForbiddenFunctions->forbiddenFunctionNames having two entries:
```
foo => fixedFoo
bar => fixedBar
```

This would also overwrite previously declared `foo` if extending ruleset defined it. For lists, it would just append the new entries.

gsherwood:
I think this is a good idea and the implementation seems fine to me. I'll put this on the roadmap. Thanks.

gsherwood:
Added as part of PR #2154

PR #2154, opened 2018-09-05
Support property-level extending for array properties

Majkl578:
Implementation for #2153.

Since Ruleset is untested, I didn't add any tests (please advice if you need some), only tested this in our consuming packages.


gsherwood:
Thanks a lot for this PR.

PR #2156, opened 2018-09-07
Fix typos in code comments and documentation

TysonAndre:


src/Standards/Generic/Docs/CodeAnalysis/JumbledIncrementerStandard.xml
```diff
@@ -1,7 +1,7 @@
-<documentation title="Jumbled Incrementers">
+<documentation title="Jumbled Increments">
```
jrfnl: I don't think this one should be changed as it refers to the name of the sniff (and renaming the sniff would break BC).

TysonAndre: Fixed - I reverted other fixes that wouldn't work before submitting this but missed that

gsherwood:
Thanks for these corrections.

PR #2158, opened 2018-09-15
Support PHP 7.3

NickWilde1990:
As per https://wiki.php.net/rfc/continue_on_switch_deprecation the use of `continue` is deprecate in switch statements and should use either break or continue 2.

### DIFF
--- a/CodeSniffer/File.php
+++ b/CodeSniffer/File.php
@@ -1763,7 +1763,7 @@ class PHP_CodeSniffer_File
                 }
                 break;
             default:
-                continue;
+                continue 2;
             }//end switch
         }//end for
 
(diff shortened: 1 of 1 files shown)


jrfnl:
FYI: In the PHPCS 3.x `master`, this was fixed in PHCPS 3.3.1 via PR #2086.

NickWilde1990:
Yeah I saw it was fixed in 3x, but I didn't check for when, thanks for the link/history @jrfnl 
Re Failing check, that is on PHP 5.4 and looks like it is *not* related to this (composer errors before tests start)

gsherwood:
Thanks. I've merged this in and fixed the two additional ones in the CSS tokenizer.

PR #2162, opened 2018-09-19
Report/Summary: minor usability improvement

jrfnl:
The "Totals" line at the bottom of the `Summary` report shows the number of files scanned:
```
A TOTAL OF 76 ERRORS AND 0 WARNINGS WERE FOUND IN 188 FILES
```

However, the actual report only lists the files in which errors or warnings were found, which is often a lot less than the total number of files scanned, causing a disconnect between the content of the report and the "Totals" line.

This PR changes the "Totals" line to include the number of files in which errors/warnings were found, like so:
```
A TOTAL OF 76 ERRORS AND 0 WARNINGS WERE FOUND IN 10 OUT OF 188 FILES
```

### DIFF
--- a/src/Reports/Summary.php
+++ b/src/Reports/Summary.php
@@ -159,9 +159,10 @@ class Summary implements Report
             echo 'S';
         }
 
-        echo ' WERE FOUND IN '.$totalFiles.' FILE';
-        if ($totalFiles !== 1) {
-            echo 'S';
+        if ($totalFiles === 1) {
+            echo ' WERE FOUND IN '.$totalFiles.' FILE';
+        } else {
+            echo ' WERE FOUND IN '.count($reportFiles).' OUT OF '.$totalFiles.' FILES';
         }
 
         echo "\033[0m";


gsherwood:
In `generateFileReport()`, the summary report returns `false` if there are no errors in a file. This means that the reporter class doesn't increment `$totalFiles` for this report type. This is true for almost all the reports.

To see a number that is out of sync with the files listed, you need to combine the summary report with one that processes every file, like the info report. So using `---report=summary,info` will show the wrong number at the bottom.

What probably needs to happen here is a bigger change where the reporter actually passes in the total number of files processed (as is documented - although incorrectly) rather than the number of files the report showed errors for. Then all the reports need to be changed to not use this number (if they are) and possibly all show the "IN x OUT OF n FILES" that this PR is adding.

I can't merge it this as-is because using the summary report alone (or with most of the others) now just shows "IN x OUT OF x FILES".

jrfnl:
@gsherwood Thanks for that feedback. I hadn't realized that.

I'll close this PR for now while I have a think about this.

jrfnl:
Reminder to self for when I do look at this again: when the `summary` report is used in combination with the `-v` option, it will list **all** files, not just the files with warnings/errors and the summary line will have the total count of all files too.

PR #2174, opened 2018-10-01
New Generic.WhiteSpace.IncrementDecrementSpacing sniff

### LINKED ISSUE #2172: Feature suggestion: sniff to check whitespace around incrementors/decrementors
While working on something else, I suddenly realized that there is no sniff to check the whitespace before/after incrementors/decrementors.

```php
$i = 10;
--$i; // This is what would be valid.
-- $i; // This should be flagged and fixed.
-- /*comment*/ $i; // This should just be flagged.
$i++; // This is what would be valid.
$i ++; // This should be flagged and fixed.
$i /*comment*/ ++; // This should just be flagged.
```

I realize it is rare to encounter such code, but it is perfectly valid PHP code and PHP will just ignore the whitespace and comments and function as expected.

I haven't (yet) checked if there are any other sniffs which would get confused by this, but running the above code samples against all the included standards in PHPCS itself, doesn't throw any unexpected false positive/negatives.

@gsherwood Do you think its a good idea to add a new sniff for this ? I'm thinking `Generic.WhiteSpace.IncrementDecrementSpacing`.
And if so, should it be added to any of the existing rulesets ?

I'd be happy to create the sniff.

jrfnl:
New sniff to check that there are no spaces (or comments) between an incrementor/decrementor and the variable it applies to.

The sniff can examine both PHP as well as JS files for this, though I suspect there may be some false positive reports for JS at some point as `T_STRING` is quite a generic token to check for.

Includes unit tests.
Includes auto-fixer.
Includes code-consistency metrics.

I've made a conscious decision to not make the number of spaces accepted between the in/decrementor and the variable configurable as it is a widely accepted convention that there should be no spaces between them.
If it would be so desired, I can add a public property to make this configurable.

Fixes #2172

### DIFF
--- a/src/Standards/Generic/Sniffs/WhiteSpace/IncrementDecrementSpacingSniff.php
+++ b/src/Standards/Generic/Sniffs/WhiteSpace/IncrementDecrementSpacingSniff.php
@@ -1,0 +1,171 @@
+<?php
+/**
+ * Verifies spacing between variables and increment/decrement operators.
+ *
+ * @author    Juliette Reinders Folmer <phpcs_nospam@adviesenzo.nl>
+ * @copyright 2018 Juliette Reinders Folmer. All rights reserved.
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ */
+
+namespace PHP_CodeSniffer\Standards\Generic\Sniffs\WhiteSpace;
+
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Files\File;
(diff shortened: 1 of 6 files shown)


jrfnl:
This sniff should probably be added to the PSR12 ruleset as - in part inspired by this sniff -, a rule to that effect has now been added to the draft PSR12: https://github.com/php-fig/fig-standards/pull/1089/commits/4f57067179ffed7dae9d8462cca598b9f6f7f28e

gsherwood:
Thanks a lot for this. I've added it to the Squiz and PHPCS standards as well.

PR #2180, opened 2018-10-06
Added a sniff about constants visibility

gmponos:
Hello,

According to PSR12 all constants must have their visibility declared. 
I added a new sniff about it. The code is copied from @slevomat (ping @kukulich as he is the main contributor as I can see). 

The sniff is altered in order be able to be used inside the current package (function calls that do not exist here) and also removed the autofix. IMHO adding visibility can cause some issues.

If slevomat and you are fine with adding this here I will proceed with adding tests.

### DIFF
--- a/src/Standards/PSR12/Sniffs/Classes/ConstantsVisibilitySniff.php
+++ b/src/Standards/PSR12/Sniffs/Classes/ConstantsVisibilitySniff.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * Verifies that constants have their visibility declared.
+ *
+ * @author    Greg Sherwood <gsherwood@squiz.net>
+ * @copyright 2006-2015 Squiz Pty Ltd (ABN 77 084 670 600)
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ */
+
+namespace PHP_CodeSniffer\Standards\PSR12\Sniffs\Classes;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+
+/**
+ * All constants must have their visibility declared.
+ *
+ * Mainly this sniff is copied from slevomat/coding-standard
+ *
+ * @see https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/Classes/ClassConstantVisibilitySniff.php
+ */
+class ConstantsVisibilitySniff implements Sniff
+{
+
+
+    /**
+     * Returns an array of tokens this test wants to listen for.
+     *
+     * @return array
+     */
+    public function register()
+    {
+        return [T_CONST];
+
+    }//end register()
+
+
+    /**
+     * Processes this test, when one of its tokens is encountered.
+     *
+     * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+     * @param int                         $stackPtr  The position of the current token in the
+     *                                               stack passed in $tokens.
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        if (PHP_VERSION_ID < 70000) {
+            return;
+        }
+
+        $tokens = $phpcsFile->getTokens();
+
+        if (count($tokens[$stackPtr]['conditions']) === 0) {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


jrfnl:
@gmponos I seem to recall having a similar conversation with you before, but unless *you* actually wrote the original sniff, you are violation copyright by pulling this.

gmponos:
It has been done before by asking the permition of the author:

https://github.com/squizlabs/PHP_CodeSniffer/pull/1106
https://github.com/squizlabs/PHP_CodeSniffer/pull/2118/files#diff-7da014129f54d9f8b4da0b2165218b74R2

So I don't see what I am doing different?

Majkl578:
This does violate the license and copyright.
Slevomat CS is licensed under MIT, but you are NOT retaining the copyright nor retaining the authorship.
Also it's sad to see that you made yourself the author here.

> IMHO adding visibility can cause some issues.

Mind to elaborate what issues? No visibility and public visibility is absolutely equivalent code.

gmponos:
closing this then.

Majkl578:
I'm not the maintainer of this package so such decision is not up to me.
If I did this PR I would:
* Keep authors and contributors as the commit authors (GitHub supports multiple authors using "Co-Authored-By") - https://github.com/slevomat/coding-standard/commits/master/SlevomatCodingStandard/Sniffs/Classes/ClassConstantVisibilitySniff.php
* Keep Slevomat asi the `@author` (as class level tag) if `@author is required in this repo)
* Keep the original copyright notice as per MIT license requirement: https://github.com/slevomat/coding-standard/blob/master/LICENSE.md

gmponos:
> Keep the original copyright notice as per MIT license requirement:

Anyway. I see you as one of the authors of that sniffs. My intention as I did was to ask from all the authors there if they are fine to have it under this repo with the license of the repo.. 

Since it's not I rather not be in the middle of this.

Majkl578:
Btw I am still interested in hearing why the fixer is dangerous (since I am the author of the fixer part). 😊

Majkl578:
Strictly speaking you _don't have to_ ask all contributors for code under MIT, you just need to include MIT's copyright notice (imho quoting Slevomat + original license should be sufficient).

gmponos:
> @gmponos I seem to recall having a similar conversation with you before, but unless you actually wrote the original sniff, you are violation copyright by pulling this.

from this tone I had a feeling that I am doing something terrible. I am little bit afraid of @jrfnl  😄 

So @jrfnl and @gsherwood and @slevomat (or @kukulich) are we good?

Majkl578:
The only terrible part is not following original license. Otherwise MIT is one of the most permissive licenses out there, you can do basically anything with such code, as long as you keep the reference to the origin. :)

gmponos:
> Btw I am still interested in hearing why the fixer is dangerous (since I am the author of the fixer part). 

The intention of the developer might have been to have it as private... I have worked with a lot of developers that they do not see the changes caused by fixer.. and the just commit the code after the fix. 

For me it's better to notify the developer and tell him to do the change manually.. I think the same mentallity is kept in other sniffs of this package.. like visiblity in properties.

gmponos:
Also I am not sure what is happening in case you have Class A and class B is extending Class A, Class A has constant with no visibility and class B has the constant with visibility as private. Havent't tested this case to know... So PHP might be fine.. not sure...

Majkl578:
I understand this reasoning, but due to the BC policies I think having it fixable by default is more convenient. But this controversy is why fixability was made opt-in in Slevomat CS (and Doctrine CS opts in as we applied CS changes in stable branches too).

kukulich:
My point of view: It does not make sense to me copy the sniff if you can very easy use the original sniff

The licence issues are already explained by Majkl578

Majkl578:
Also, making fixability opt-in satisfies both lands while no fixer is just annoying for one.

Inheritance is irrelevant - `public const` and `const` are identical.

gmponos:
> It does not make sense to me copy the sniff if you can very easy use the original sniff

At some point having a sniff about const visibility will be needed inside this repo because of PSR12.

Since at some point of my life I have seen the code of your repo even if I tried to start it from scratch the chances of creating a different one or not having in my mind your sniff would be minimum. So I thought that referencing it here as source would be a good thing.

Anyway.. I have closed this.. I understood the points of licenses etc etc but never the less I will leave this for someone else to create it.

Issue #2039, opened 2018-06-03
Yoda or NoYoda


gmponos:
There are those who prefer to have their condition in Yoda style, and there are those who do not.

It reminds me a lot of the issue tabs vs spaces. This issue is not about starting a war again ☮️ .

Would you be interested in including in `General` two sniffs one enforcing Yoda and the other one not to enforce Yoda?

gsherwood:
> Would you be interested in including in General two sniffs one enforcing Yoda and the other one not to enforce Yoda?

I sure would

VincentLanglet:
@gmponos Do you plan to add the EnforceYodaConditionSniff too ?

PR #2181, opened 2018-10-06
Added DisallowYodaCondition sniff

gmponos:
related to #2039 

@dereuromark can confirm that I asked about adding this here. Can you also confirm that you are fine for relicensing this under BSD?

@gsherwood can you confirm if you want this in one sniff or two separate sniffs? Also do you believe an autofix should also be included?


dereuromark:
Sure, no problem. But bear in mind that this was a very basic draft.
There is a more sophisticated one in the slevomatik standard afaik.

gmponos:
> There is a more sophisticated one in the slevomatik standard afaik.

I prefer not to... https://github.com/squizlabs/PHP_CodeSniffer/pull/2180

dereuromark:
You can easily reuse MIT code inside this package.
The file itself just needs to keep the license part/info etc.

gsherwood:
I think there would need to be 2 sniffs - one for forcing yoda conditions and one for banning them. It's how a few other sniffs are done as well.

I prefer all sniffs to be licensed under the same BSD licence as PHPCS, especially given I have to take on the job of maintaining them. Sniffs typically aren't complicated enough for me to worry about licensing and making sure the listed authors are correct and contacted.

If someone has written a sniff themselves and wants to contribute it to PHPCS, I am more than happy to take a look. If you are submitting sniffs that other people have written, I'll probably reject the PR unless the original author is part of the submission process as well.

gmponos:
> Sniffs typically aren't complicated enough for me to worry about licensing

👍 

> I'll probably reject the PR unless the original author is part of the submission process as well

Anyway I believe the same argument about simplicity can apply also here. The current sniffs are pretty simple. Never the less I would be happy if @dereuromark could have a look also and share his opinion.

gmponos:
OK, I have a question to whoever may see this and I need to be sure about this. 

Is this  `if (($value + 1) === $value)` a Yoda condition or not.

gmponos:
Hello there,

I believe I finished the work on this sniff.

Some things that we need to notice

1. Please take a very close look at the tests about what is and what is not a Yoda condition.
2. I have tested against a very known framework and it found valid errors. I will check it against more packages to make sure it works as expected.
3. I will create the EnforceYodaConditionsSniff on another PR.

sertand:
Thanks for the sniff.

PR #2183, opened 2018-10-06
Ignore BOM in Generic.Files.InlineHTML and Generic.PHP.CharacterBeforePHPOpeningTag

thewilkybarkid:
Having a BOM currently triggers `Generic.Files.InlineHTML` and `Generic.PHP.CharacterBeforePHPOpeningTag`, duplicating `Generic.Files.ByteOrderMark`. This causes them to ignore it.


gsherwood:
Thanks for making those changes. Make sense to me.

PR #2185, opened 2018-10-13
Squiz.Classes.SelfMemberReference: remove any number of newlines

ostrolucky:
Stop relying on global while loop with limit to 50 iterations.
Doing this by sniff itself increases performance, reliability
and decouples it from default implementation of Sniff Runner


src/Standards/Squiz/Sniffs/Classes/SelfMemberReferenceSniff.php
```diff
@@ -103,9 +103,10 @@ protected function processTokenWithinScope(File $phpcsFile, $stackPtr, $currScop
             $error = 'Expected 0 spaces before double colon; %s found';
             $data  = [$found];
             $fix   = $phpcsFile->addFixableError($error, $calledClassName, 'SpaceBefore', $data);
+            $i     = $stackPtr;
 
-            if ($fix === true) {
-                $phpcsFile->fixer->replaceToken(($stackPtr - 1), '');
+            while ($fix === true && $tokens[--$i]['code'] === T_WHITESPACE) {
```
jrfnl: You should probably wrap this in a changeset to make sure it gets applied in one go.
```php
// Various fixes
$phpcsFile->fixer->endChangeset();
```

ostrolucky: done

jrfnl: @ostrolucky Sorry, I wasn't clear enough. It should just wrap the fixer parts (for each of the code blocks were you made changes), like so:
```php
        if ($fix === true) {
            while ($fix === true && $tokens[--$i]['code'] === T_WHITESPACE) {
                $phpcsFile->fixer->replaceToken($i, '');
            }
            $phpcsFile->fixer->endChangeset();
        }
```

ostrolucky: Why?

jrfnl: Because PHPCS will check whether other sniffs try to change the same tokens within a loop and if so, needs to be able to revert the whole set as one. The changeset will then be offered again in the next loop and hopefully be applied without issue.

The way you have set up the changesets now:
* Disregards whether the fixer is even active.
* Combines the changes for two significantly different errors into one changeset.

Does that explain it well enough ?

ostrolucky: got it

gsherwood:
For consistency with the rest of the code base, I'd would use a standard class method in place of the closure, or I'd just repeat the code given it is quite small in this specific case. If you don't want to make that change, I can use your test case and do that myself.

I get that you're trying to put in a test case that hits the 50 loop limit, but I don't think this is necessary for the unit tests in general, so I would shorten that up as well. I may, for example, choose to have the unit tests loop a max of 5 times for speed, so trying to test the standard run loop isn't needed here.

ostrolucky:
Alright. What kind of test should I write then? Standard test with less than 50 newlines won't test this change, it would pass without it.

ostrolucky:
removed closure, used duplication

jrfnl:
Just a question, doesn't necessarily have to be solved in this PR: what about if there are comments or annotations found in the tokens ? AFAICS, at this moment, the sniff does not handle that situation well.

Think:
```php
self
/* some comment */

::

// phpcs:ignore Standard.Category.SniffName -- for reasons.
function_name();
```

I also wonder if this sniff for the `NotUsed` error is susceptible to the same sort of bug as reported in #2178 and fixed by #2182, i.e. the `processTokensWithinScope()` method gets called for each matching scope, meaning that a token may get examined multiple times with slightly different contexts.

ostrolucky:
better to create issue about that

jrfnl:
@ostrolucky Nah, in that case I would test & solve it straight away, but I didn't want to get wires crossed with this PR. Figured you may want to take a look as you were working on this sniff anyway.

ostrolucky:
Not really, it is huge edge case. Never seen comments in between like that and don't see why would somebody do that either. Furthermore, as this is my first contribution,  in order for it to be successful I want to minimize things that can be picked apart in reviews.

@gsherwood let me know if removing test case is acceptable for you, or if you would like me to write mechanism which can change number of passes in testcases

jrfnl:
> Furthermore, as this is my first contribution, in order for it to be successful I want to minimize things that can be picked apart in reviews.

Sorry, I hadn't realized. Totally understand. I'll have a look at it once this PR has been merged.

gsherwood:
@ostrolucky It's on the roadmap, which I am slowly working through. I'm pretty sure if any changes are required, they'll be very minor and I will do them as I merge and test.

PR #2187, opened 2018-10-16
Build/Travis: validate rulesets against XSD and check code style consistency

jrfnl:
### Build: lint the XML rulesets and check for consistent code style

Notes:
* Runs only once per build as the results are independent of the PHP version.
* Indent is set at four spaces as per commit 232f7da .
* I've elected to add this check to the smallest (fastest) build, so the build time will be spread more evenly.
* Not linting the PHPCS native ruleset as it's being run anyway, so is automatically checked.

### Clean the XML ruleset files/consistent code style

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,11 @@ matrix:
     - php: 5.6
     - php: 7.0
     - php: 7.0
-      env: CUSTOM_INI=1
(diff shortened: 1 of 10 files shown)


gsherwood:
Thanks. You're a 💎 !

PR #2190, opened 2018-10-18
PEAR.Functions.FunctionCallSignature incorrect error when encountering trailing PHPCS annotation

jrfnl:
When `$allowMultipleArguments` is set to false and the sniff would encounter a trailing PHPCS annotation in a function call, it would incorrectly throw a _"Only one argument is allowed per line in a multi-line function call"_ error.

The sniff did handle normal trailing comments correctly, so only needed a minimal adjustment to allow for the PHPCS annotations.

Additional notes:
* This commit also adds a few unit test cases using trailing comma's in function calls - as allowed per PHP 7.3 - to confirm that the sniff handles these correctly and safeguard this for the future.
* Also includes a minor unit test case documentation fix.


gsherwood:
Thanks. The extra test cases were a good idea too.

PR #2192, opened 2018-10-18
Bugfixes for 6 sniffs / `processTokensWithinScope()` vs nested classes

jrfnl:
These fixes are all loosely related to and inspired by the test case provided in #2178 and fixed via #2182.

There were a number of other sniffs using the `AbstractScopeSniff` as a basis which would throw the same error twice for functions nested in anonymous classes.

I've reviewed most of these now and have fixed them in separate commits.

All fixes include unit tests.

The reasoning and solution used is the same as previously applied in #2182 for the `Generic.NamingConventions.ConstructorName` sniff.

> Turns out the `processTokensWithinScope()` method gets called for each whitelisted scope a `T_FUNCTION` token exists in.
>
> In effect, this means that the first time it gets called for the `nested_function()` method, it is because it is within the scope of the `Nested` class, the second time the function gets called, it is for the method being in the scope of the anonymous class.
>
> The fix I've now applied checks that the scope for which the method is called is actually the deepest scope first and bows out if it's not.

### `Generic.NamingConventions.CamelCapsFunctionName`
* Fixed: Sniff would throw the same error twice for functions in nested classes.
* UX fix: For methods in anonymous classes, the error message would read `Public method name "::nested_function" is not in camel caps format` which could be considered confusing.
    I've changed this to now read: `Public method name "[Anonymous Class]::nested_function" is not in camel caps format`
* Class and method names in PHP are case _in_sensitive. While the sniff took that into account for the magic methods, it did not when checking for PHP4 style constructors/destructors.
     This has now been fixed by lowercasing both the class name as well as the function name before comparing them.

### `PEAR.NamingConventions.ValidFunctionName`
* Fixed: Sniff would throw the same error twice for functions in nested classes.
* UX fix: For methods in anonymous classes, the error message would read `Method name "::nested_function" ....` which could be considered confusing.
    I've changed this to now read: `Method name "[Anonymous Class]::nested_function" ....`
* Class and method names in PHP are case _in_sensitive. While the sniff took that into account for the magic methods, it did not when checking for PHP4 style constructors/destructors.
     This has now been fixed by lowercasing both the class name as well as the function name before comparing them.

### `PSR1.Methods.CamelCapsMethodName`
* Fixed: Sniff would throw the same error twice for functions in nested classes.
* UX fix: For methods in anonymous classes, the error message would read `Method name "::nested_function" ...` which could be considered confusing.
    I've changed this to now read: `Method name "[Anonymous Class]::nested_function" ...`

### `PSR2.Methods.MethodDeclaration`
* Enhancement: added support for anonymous classes while preventing the double error issue.

### `Squiz.Scope.MethodScope`
* Enhancement: added support for anonymous classes.
* Fixed: bug where nested structures would not be examined.

### `Squiz.Scope.StaticThisUsage`
* Enhancement: added support for anonymous classes and traits while preventing the double error issue.

### DIFF
--- a/src/Standards/Generic/Sniffs/NamingConventions/CamelCapsFunctionNameSniff.php
+++ b/src/Standards/Generic/Sniffs/NamingConventions/CamelCapsFunctionNameSniff.php
@@ -98,6 +98,16 @@ class CamelCapsFunctionNameSniff extends AbstractScopeSniff
      */
     protected function processTokenWithinScope(File $phpcsFile, $stackPtr, $currScope)
     {
+        $tokens = $phpcsFile->getTokens();
+
+        // Determine if this is a function which needs to be examined.
+        $conditions = $tokens[$stackPtr]['conditions'];
+        end($conditions);
+        $deepestScope = key($conditions);
+        if ($deepestScope !== $currScope) {
+            return;
+        }
+
         $methodName = $phpcsFile->getDeclarationName($stackPtr);
         if ($methodName === null) {
             // Ignore closures.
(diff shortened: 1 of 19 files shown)


jrfnl:
Oh and just in case anyone is wondering: the reason I didn't add this fix to the abstract sniff, but to the individual sniffs instead, is that child sniffs in external standards may be relying on receiving the token for each scope, so changing that now would be a BC-break.

gsherwood:
Thanks for these fixes

jrfnl:
You're welcome.

PR #2195, opened 2018-10-19
Generic/LanguageConstructSpacing: bug fix / don't fix spacing namespace operator

### LINKED ISSUE #2194: Generic.Whitespace.LanguageConstructSpacing should not be checking namespace operators
The `Generic.Whitespace.LanguageConstructSpacing` sniff demands one space after PHP language construct keywords.

The `namespace` keyword however, is both a construct as well as an operator, depending on the context.

```php
<?php
// Here it is a language construct:
namespace MyProject;

// Here it is an operator or as the documentation states:
// "It is the namespace equivalent of the self operator for classes. "
namespace\func(); // calls function MyProject\func()
$a = new namespace\sub\cname(); // instantiates object of class MyProject\sub\cname
```
See:  http://php.net/manual/en/language.namespaces.nsconstants.php#example-256

Currently, the sniff will trigger errors for the use of the keyword as an operator and demand one space after the keyword.

This, to me, is unexpected.

Is this intended behaviour or a bug ? /cc @gmponos

jrfnl:
The sniff should only address the spacing after the `namespace` keyword when used as a language construct, not when it is used as an operator.

For more details, see issue #2194.

Fixes #2194

### DIFF
--- a/src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
+++ b/src/Standards/Generic/Sniffs/WhiteSpace/LanguageConstructSpacingSniff.php
@@ -11,7 +11,8 @@ namespace PHP_CodeSniffer\Standards\Generic\Sniffs\WhiteSpace;
 
 use PHP_CodeSniffer\Sniffs\Sniff;
 use PHP_CodeSniffer\Files\File;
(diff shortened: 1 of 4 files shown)


jrfnl:
You're welcome.

PR #2196, opened 2018-10-19
File::getTokensAsString(): prevent endless loop

### LINKED ISSUE #2042: Potential endless loop in getTokensAsString when used wrong
Hey,

while writing a custom sniff I made a mistake in not checking the result of findPrevious and found an endless loop in `getTokensAsString` that I'm not sure you want to fix. But I thought I'll write and see.

A sniff containing:
```php
$commentStart = $phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, $stackPointer);
$commentEnd = $phpcsFile->findPrevious(T_DOC_COMMENT_CLOSE_TAG, $stackPointer);

$content = $phpcsFile->getTokensAsString($commentStart, $commentEnd - $commentStart + 1);
```

Min reproduce of the core issue: 

https://3v4l.org/LuUS9

For reference: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Files/File.php#L1905

```php
public function getTokensAsString($start, $length, $origContent=false)
    {
        $str = '';
        $end = ($start + $length);
        if ($end > $this->numTokens) {
            $end = $this->numTokens;
        }
        for ($i = $start; $i < $end; $i++) {
            // If tabs are being converted to spaces by the tokeniser, the
            // original content should be used instead of the converted content.
            if ($origContent === true && isset($this->tokens[$i]['orig_content']) === true) {
                $str .= $this->tokens[$i]['orig_content'];
            } else {
                $str .= $this->tokens[$i]['content'];
            }
        }
        return $str;
    }//end getTokensAsString()
```

When calling this with `(false, false)` $end will be set to numTokens but $i will always be false as `false++` is also `false` so it keeps looping until $str exceed the memory limit.

Of course it doesn't make any sense to call this with (false, false) at all and after debugging I've fixed the issue in the code using the phpcs API :) 

Hope this is useful and understandable!
Cheers,
Volker

jrfnl:
As discussed in issue #2042, this adds an exception to the `File::getTokensAsString()` method for when the value of the `$start` parameter is invalid.

As for the `$length` parameter, I've chosen to just bow out early in that case.

Note: I've tested adding an exception for the `$length` parameter as well, but that would actually give issues with at least one existing PHPCS native sniff, so this is the more elegant solution.

The existing issue is with the function call on line 309 of the `AbstractPatternSniff` in combination with the `PEAR.ControlStructures.ControlSignature` sniff. I haven't looked into that issue any deeper though.

Fixes #2042

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -1911,6 +1911,14 @@ class File
      */
     public function getTokensAsString($start, $length, $origContent=false)
     {
+        if (is_int($start) === false || isset($this->tokens[$start]) === false) {
+            throw new RuntimeException('The $start position for getTokensAsString() must exist in the token stack.');
+        }
+
+        if (is_int($length) === false || $length <= 0) {
+            return '';
+        }
+
         $str = '';
         $end = ($start + $length);
         if ($end > $this->numTokens) {


gsherwood:
Thanks for fixing that.

PR #2202, opened 2018-10-23
Squiz.WhiteSpace.OperatorSpacing throws error for negative index when using curly braces for string access

anbuc:
Adding T_OPEN_CURLY_BRACKET to the list of tokens that indicate that the following token does not
belong to an arithmetic operation.
e.g. `$myString{-1}`

See https://secure.php.net/manual/en/language.types.string.php#language.types.string.substr


gsherwood:
Thanks for this. I didn't even know curly braces could be used as a replacement for the square braces in string access.

I fixed the same problem in `Squiz.Formatting.OperatorBracket`, but didn't see it anywhere else.

anbuc:
>  I didn't even know curly braces could be used as a replacement for the square braces in string access.

Yes, I love it. Helps me to differentiate between string and array access

PR #2210, opened 2018-11-01
Generic.NamingConventions.CamelCapsFunctionName not ignoring SoapClient __getCookies() method

jrfnl:
The list of `$methodsDoubleUnderscore` from the SoapClient extension was missing the `__getCookies()` method.

Ref: http://php.net/manual/en/class.soapclient.php

Includes unit test.

Includes ordering the list in alphabetic order.


gsherwood:
Thanks

PR #2211, opened 2018-11-01
PSR2.Methods.MethodDeclaration gets confused over comments between modifier keywords

jrfnl:
The `PSR2.Methods.MethodDeclaration` sniff would get unnecessarily confused over comments between the modifier keywords.

This minor changes fixes that.

Includes (adjusted) unit tests demonstrating the issue.


gsherwood:
Thanks for fixing this

PR #2212, opened 2018-11-01
FUNCTION and CONST in use groups being tokenised as T_FUNCTION and T_CONST

thewilkybarkid:
Added example to the test for `Squiz.WhiteSpace.FunctionSpacing` which has a problem with this (refs https://github.com/libero/php-coding-standard/pull/30).


gsherwood:
Thanks for fixing those token

PR #2214, opened 2018-11-01
File::getMemberProperties() is recognizing method params as properties

jrfnl:
Method parameters were recognized as if they were class properties.

Looks like I already included an initial unit test case for this in #1715, but that case was not being tested.
The change in this PR should fix the issue, activates the existing unit test and adds some extra as well.


gsherwood:
Thanks

Issue #1574, opened 2017-07-24
(binary) not tokenized as T_BINARY_CAST


jrfnl:
There are two ways to cast to binary as introduced in PHP 5.2.1:
* `$binary = b"binary string";`
* `$binary = (binary) $string;`

The first one is identified by PHPCS as `T_BINARY_CAST`, while the second is not.

By the looks of it, PHP itself incorrectly tokenizes this as `T_STRING_CAST`. See: https://3v4l.org/n0DJj

Ref: http://php.net/manual/en/language.types.type-juggling.php#language.types.typecasting

jrfnl:
On second look: neither are currently tokenized correctly.

It does look like there is some code in the PHP Tokenizer which is supported to take care of the first case, but by the looks of it, it's not working correctly.

See: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Tokenizers/PHP.php#L567-L575

gsherwood:
The binary cast code only gets called if PHP has tokenized the contents of the double quoted string. So it needs some variable in it, or some other reason for PHP to break it up, such as: `$binary = b"binary string $foo";`

I can fix that, and convert the `(binary)` cast into the T_BINARY_CAST token I created for the `b` type cast despite PHP calling it a T_STRING_CAST, which it presumably does because it is casting to a "binary string". Feels like it makes more sense.

jrfnl:
> The binary cast code only gets called if PHP has tokenized the contents of the double quoted string. So it needs some variable in it, or some other reason for PHP to break it up, such as: $binary = b"binary string $foo";

I didn't know that and the example in the PHP documentation doesn't show that either. Is the variable a PHPCS requirement for the token parsing or a PHP requirement for binary casting ?

> I can fix that, and convert the (binary) cast into the T_BINARY_CAST token I created for the b type cast despite PHP calling it a T_STRING_CAST, which it presumably does because it is casting to a "binary string". Feels like it makes more sense.

Well, they both make sense in a way, but as all the casts have their own token, it felt a bit odd that this one doesn't.

For the sniff where I noticed the issue, I've worked around it for now, but I can imagine this may be useful/more intuitive to work with for people anyway

gsherwood:
> I didn't know that and the example in the PHP documentation doesn't show that either. Is the variable a PHPCS requirement for the token parsing or a PHP requirement for binary casting ?

Sorry, I forgot to reply to this.

[...]

PR #2215, opened 2018-11-02
Tokenizer: fix binary cast recognition with inside parentheses whitespace

jrfnl:
As reported in [my last comment](https://github.com/squizlabs/PHP_CodeSniffer/issues/1574#issuecomment-435278428) on the original issue #1574, the fix to the tokenizer does not take whitespace inside of the binary cast parentheses into account.

This PR fixes this.

It also adds a unit test which will guard against regressions to the `Squiz.WhiteSpace.CastSpacing` sniff.

As that sniff did not have a `fixed` file, while it _does_ in actual fact contain a fixer, the `.fixed` file has also been added.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1567,6 +1567,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
        </dir>
        <dir name="WhiteSpace">
         <file baseinstalldir="PHP/CodeSniffer" name="CastSpacingUnitTest.inc" role="test" />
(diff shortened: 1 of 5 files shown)


gsherwood:
Thanks a lot. I guess I've coded my type casts the same for so long that I forgot you could have whitespace in there.

jrfnl:
No worries, I just came across it when I was checking the PHPCS-cross-version compatibility with the token for the `PHPCompatibility.TypeCasts.NewTypeCasts` sniff. (and then had to find a sniff in PHPCS which would show the same issue)

PR #2217, opened 2018-11-02
New `Generic.Formatting.SpaceBeforeCast` sniff

jrfnl:
... which verifies that there is exactly one space before a type cast.

Sister-sniff to the `Generic.Formatting.SpaceAfterCast` sniff.

Depending on the coding-standard in which this sniff is included, this sniff may throw "duplicate" messages, i.e. given `$a =(bool) $b;`, a lot of coding standards will throw an error already demanding a space after the assignment operator.
All the same, type casts can be used in all sorts of code constructs, so there may well be situations in which the preceding token would not trigger an error, in which case, this sniff becomes relevant.

Notes:
* If the preceding whitespace is indentation, the sniff bows out.
* If the type cast is at the start of line, the sniff bows out.

Includes unit tests.

### DIFF
--- a/src/Standards/Generic/Sniffs/Formatting/SpaceBeforeCastSniff.php
+++ b/src/Standards/Generic/Sniffs/Formatting/SpaceBeforeCastSniff.php
@@ -1,0 +1,73 @@
+<?php
+/**
+ * Ensures there is a single space before cast tokens.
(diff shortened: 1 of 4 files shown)


gsherwood:
Thanks

PR #2218, opened 2018-11-03
PSR2/PropertyDeclaration: add modifier order error + fixing

### LINKED ISSUE #2198: Check visibility before static for properties in PropertyDeclarationSniff
MethodDeclarationSniff checks that visibility (public, private, protected) is before the static keyword, but I have not found a sniff to do this check for properties.

It is possible to add the check from MethodDeclarationSniff to PropertyDeclarationSniff to check for visibility before static? The final and abstract keyword is not needed.

It seems there is no Tokens::$propertyPrefixes similar to Tokens::$methodPrefixes for this purpose.

Thanks.

jrfnl:
IMO the feature request made in #2198 is unequivocally covered by PSR2 (emphasis in the quote is mine):

> Visibility MUST be declared on all **_properties_** and methods; `abstract` and `final` MUST be declared before the visibility; `static` MUST be declared after the visibility.
Ref: https://www.php-fig.org/psr/psr-2/#1-overview

With that in mind, I've adjusted the `PSR2.Classes.PropertyDeclaration` sniff to include a check for the modifier keyword order.

Includes unit tests.

Notes:
* As the sniff didn't have any fixers up to now, I've added a `.fixed` file.
* The fixer tries to gracefully handle any comments found between the keywords and the property variable.
* The sniff now defers to the `File::getMemberProperties()` method for the information on whether scope (and static) is specified.
    This also fixes a bug where the visibility was incorrectly not recognized as present when it was declared on another line. See unit test on line 22.
* Added a unit test to cover PHP4 style static properties and documented an intentional parse error in the test case file.
* Additionally, this PR contains some minor defensive coding improvements to the sniff.

Fixes #2198

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1066,6 +1066,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="ClassDeclarationUnitTest.inc.fixed" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="ClassDeclarationUnitTest.php" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="PropertyDeclarationUnitTest.inc" role="test" />
+        <file baseinstalldir="PHP/CodeSniffer" name="PropertyDeclarationUnitTest.inc.fixed" role="test" />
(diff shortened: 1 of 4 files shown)


gsherwood:
Thank you very much for doing this.

jrfnl:
You're very welcome ;-)

PR #2220, opened 2018-11-06
PSR2/PropertyDeclaration: update docs

jrfnl:
In PR #2218, I updated the `PSR2.Classes.PropertyDeclaration` sniff and added a new check, but I forgot to update the associated documentation.

Sorry about that. This should fix it.


gsherwood:
Thanks for doing this

PR #2224, opened 2018-11-12
Add support for multiline indenting to ObjectOperatorIndentSniff

marcospassos:
This PR adds supports for multilevel indenting to the amazing sniff `ObjectOperatorIndentSniff` created by @gsherwood.

Multilevel indenting contributes to improve code readability in long chained method calls. This convention is [widely used by Symfony in configuration definitions](https://symfony.com/doc/current/components/config/definition.html#variable-nodes):

```php
$rootNode
    ->children()
        ->booleanNode('auto_connect')
            ->defaultTrue()
        ->end()
        ->scalarNode('default_connection')
            ->defaultValue('default')
        ->end()
    ->end()
;
```

The option `multilevel` allows enabling/disabled the new behavior, which is disabled by default to be backward compatible with previous versions.

The sniff can automatically fix violations by adjusting the number of spaces based on the value of property `indent` and the previous indentation:

```php
// Before fix
$rootNode
    ->children()
		        ->booleanNode('auto_connect')
        ->end();

// After fix
$rootNode
    ->children()
        ->booleanNode('auto_connect')
    ->end();
```
Tests are included to confirm the correctness.

### DIFF
--- a/src/Standards/PEAR/Sniffs/WhiteSpace/ObjectOperatorIndentSniff.php
+++ b/src/Standards/PEAR/Sniffs/WhiteSpace/ObjectOperatorIndentSniff.php
@@ -22,6 +22,13 @@ class ObjectOperatorIndentSniff implements Sniff
      */
     public $indent = 4;
 
+    /**
+     * Indicates whether multilevel indenting is allowed.
+     *
+     * @var boolean
+     */
+    public $multilevel = false;
+
 
     /**
      * Returns an array of tokens this test wants to listen for.
@@ -71,12 +78,12 @@ class ObjectOperatorIndentSniff implements Sniff
             }
         }
 
-        $requiredIndent = 0;
-        if ($i >= 0 && $tokens[$i]['code'] === T_WHITESPACE) {
-            $requiredIndent = strlen($tokens[$i]['content']);
+        $baseIndent = 0;
+        if ($i >= 0 && $tokens[$i]['code'] === \T_WHITESPACE) {
+            $baseIndent = \strlen($tokens[$i]['content']);
         }
 
-        $requiredIndent += $this->indent;
+        $baseIndent += $this->indent;
 
         // Determine the scope of the original object operator.
         $origBrackets = null;
@@ -95,6 +102,8 @@ class ObjectOperatorIndentSniff implements Sniff
         if ($tokens[$stackPtr]['line'] > $tokens[$start]['line']) {
             $next = $stackPtr;
         }
+
+        $previousIndent = $baseIndent;
 
         while ($next !== false) {
             // Make sure it is in the same scope, otherwise don't check indent.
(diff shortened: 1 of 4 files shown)


src/Standards/PEAR/Sniffs/WhiteSpace/ObjectOperatorIndentSniff.php
```diff
@@ -121,22 +130,34 @@ public function process(File $phpcsFile, $stackPtr)
...
                         $fix = $phpcsFile->addFixableError($error, $next, 'Incorrect', $data);
                         if ($fix === true) {
-                            $spaces = str_repeat(' ', $requiredIndent);
+                            $spaces = str_repeat(' ', $expectedIndent);
                             if ($foundIndent === 0) {
                                 $phpcsFile->fixer->addContentBefore($next, $spaces);
                             } else {
                                 $phpcsFile->fixer->replaceToken(($next - 1), $spaces);
                             }
                         }
```
jrfnl: Just wondering: should `$previousIndent` be (re)set in this leaf of the condition as well ? Maybe to `$expectedIndent` ?

marcospassos: This is one of the possible heuristics. 

Notice that it affects error reporting only:

```php
// Keeping the last valid
$builder
    ->start()
            ->do() // Excpected 8, found 12
                ->end(); // Expected 8, found 16

// Keeping the last adjusted
$builder
    ->start()
            ->do() // Excpected 8, found 12
                ->end(); // Expected 12, found 16
```
Let's go with your suggestion as it may lead to more correct results most of the time, assuming that if there are extraneous spaces, it is more likely that there should be indentation than not.

marcospassos:
@jrfnl any chances of merging it anytime soon, once the PR includes tests?

marcospassos:
Sorry, I didn't realize you don't have such privilege.

> Other than that, you may want to reset the custom property to the default after the tests covering the property, to prevent confusion if more unit tests are added later, but that's minor.

Nice catch, I'll do it.

marcospassos:
Done!

marcospassos:
@gsherwood, did you have time to take a look into this?

marcospassos:
@gsherwood thanks for reviewing it.

> If I had to document this (which I will need to), would I say that the multilevel option allows for a line in a chained call to either be indented 1 level more or less than the previous line, with the only exception being that lines must be indented at least 1 level from the first line?

Exactly, this describes precisely the current implementation.

I believe the current implementation covers the most common use cases. The widespread use case is balanced method calls, such as `open()` and `close()`, `start()` and `end()`, etc. Another common use case is nested builders that requires calling `parent()` to return to the parent builder, which is also covered in the current implementation. The last case you mentioned is valid to support cases where the chained calls are not necessarily balanced as shown in the following snippet:

```php
$writer
    ->writeln("First line")
    ->indent()
        ->writeln("Indent level 1")
        ->indent()
            ->writeln("Indent level 2")
```

As a side note, I've been using it for two months in different projects with no issues so far.

marcospassos:
@gsherwood are there any issues still pending in this PR?

gsherwood:
> are there any issues still pending in this PR?

No, sorry. Just haven't had time to look at this yet.

gsherwood:
Sorry this took so long to merge, but it is in now and will be released in 3.5.0. Thanks a lot for this PR.

PR #2057, opened 2018-06-10
Create NoSpaceAfterNotSniff

carusogabriel:
Hello there :wave: 

Recently, while trying to recreate the Symfony CS with PHPCS, I couldn't found a Sniff to forbid whitespaces after the not operator `!`.

Ref: https://symfony.com/doc/current/contributing/code/standards.html#structure

As my first Sniff PR, here we have it :smile:

PR #2233, opened 2018-11-20
Generic/SpaceAfterNot: make the sniff configurable

jrfnl:
This PR makes the `Generic.Formatting.SpaceAfterNot` sniff configurable as discussed in PR https://github.com/squizlabs/PHP_CodeSniffer/pull/2057#issuecomment-396093803.

Notes:
* Adds a public `spacing` property. Defaults to `1` to maintain BC.
* Adds a public `ignoreNewLines` property. Defaults to `false` to maintain BC.
* Adds a new non-fixable `CommentFound` error for when non-whitespace tokens are found between the NOT operator and the next non-empty token.
* Adjusts the error message for the `Incorrect` error to allow for the flexibility now needed, what with the configurable `spacing` property.
    Note: the error **code** has not been changed to prevent a BC-break.
* Makes the fixer more efficient compared to before.
    Previously, if a whitespace + new line + whitespace would be found with `spacing` set to `1`, the fixer would need three loops to apply all the fixes, now this is fixed in one go.

Includes extensive additional unit tests.
Unit tests have been added to the PHP file, but the changes should work just as well for JS.

Includes basic xml documentation for the sniff.

### DIFF
--- a/src/Standards/Generic/Docs/Formatting/SpaceAfterNotStandard.xml
+++ b/src/Standards/Generic/Docs/Formatting/SpaceAfterNotStandard.xml
@@ -1,0 +1,25 @@
+<documentation title="Space After NOT operator">
+    <standard>
+    <![CDATA[
+    Exactly one space is allowed after the NOT operator.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: A NOT operator followed by one space.">
+        <![CDATA[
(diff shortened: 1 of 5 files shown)


gsherwood:
Thanks for the PR. New props documented here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#genericformattingspaceafternot

PR #2234, opened 2018-11-20
Generic/SpaceAfterCast: make the sniff configurable & deprecates NoSpaceAfterCast

jrfnl:
This PR makes the `Generic.Formatting.SpaceAfterCast` sniff configurable as discussed in PR https://github.com/squizlabs/PHP_CodeSniffer/pull/2057#issuecomment-396103995 and deprecates the `Generic.Formatting.NoSpaceAfterCast` sniff.

Notes:
* Adds a public `spacing` property. Defaults to `1` to maintain BC.
* Adds a public `ignoreNewLines` property. Defaults to `false` to maintain BC.
* Adds a new non-fixable `CommentFound` error for when non-whitespace tokens are found between the type cast and the next non-empty token.
* Adds a new fixable `TooLittleSpace` error code for when `$spacing` is set to more than `1` and there is whitespace, but not enough.
* Adjusts the error message for the `NoSpace` and `TooMuchSpace` errors to allow for the flexibility now needed, what with the configurable `spacing` property.
    Note: the error **codes** have not been changed to prevent a BC-break.
* Makes the fixer more efficient compared to before.
    Previously, if a whitespace + new line + whitespace would be found with `spacing` set to `1`, the fixer would need three loops to apply all the fixes, now this is fixed in one go.

### DIFF
--- a/src/Standards/Generic/Sniffs/Formatting/NoSpaceAfterCastSniff.php
+++ b/src/Standards/Generic/Sniffs/Formatting/NoSpaceAfterCastSniff.php
@@ -5,6 +5,9 @@
  * @author    Greg Sherwood <gsherwood@squiz.net>
  * @copyright 2006-2015 Squiz Pty Ltd (ABN 77 084 670 600)
  * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ *
+ * @deprecated 3.4.0 Use the Generic.Formatting.SpaceAfterCast sniff with
+ *                   the $spacing property set to 0 instead.
  */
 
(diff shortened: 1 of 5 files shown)


gsherwood:
Thanks for this. The props are documented here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#genericformattingspaceaftercast

Forgot to post in the other PR about the message changes. I removed the passing of "s" as a data value for consistency but also because it's not really useful when translating the message due to how plurals work in some languages (Polish comes to mind). PHPCS could use a better translation system, probably based on gettext, but it's never been requested.

PR #2235, opened 2018-11-20
Squiz/ObjectOperatorSpacing: fix whitespace in one loop

jrfnl:
Efficiency fix. Prevents the fixer having to be triggered multiple times when, for instance, whitespace + newline + whitespace or several new lines are found between the object operator and the token before/after.

Existing unit tests already cover this change.

The effect of the change can be seen by running `phpcbf -v --standard=Squiz ./src/Standards/Squiz.Tests/WhiteSpace/ObjectOperatorSpacingUnitTest.inc --sniffs=Squiz.WhiteSpaceObjectOperatorSpacing` first on `master` (3 passes needed), then on this branch (2 passes needed).

Includes minor improvement to an existing unit test checking the handling of whitespace in combination with comments.


gsherwood:
Thanks

Issue #2150, opened 2018-09-03
Sniff to check whitespace around namespace separators


jrfnl:
While working on something else, I suddenly realized that there is no sniff to check the whitespace around namespace separator tokens `\` and that some other sniffs get terribly confused and throw false positives/incorrect error messages when they encounter whitespace and/or comments within a namespace name.

```php
namespace MyNS \  WhiteSpace\OK\ NotOK;

class ABC extends \ MyNamespace \ /* comment */ Something \    Other \
    Deeper \ ClassName
{}
```

I realize it is rare to encounter such code, but it is perfectly valid PHP code and PHP will just ignore the whitespace and comments and function as expected. Also see: https://3v4l.org/EFgCp

While not explicit in the textual guidelines, using whitespace and comments within a namespace name does appear to be forbidden in [PSR-2 based on all the example code](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md).

The sniffs which currently throw false positives when they encounter such code - based on a cursory analysis - are `PSR2.Classes.ClassDeclaration` and `Squiz.Classes.ClassDeclaration`.

@gsherwood Do you think its a good idea to add a new sniff for this ? I'm thinking `Generic.WhiteSpace.NamespaceSeparatorSpacing` with a public `$ignoreNewlines` property which defaults to `false`.
And if so, should it be added to any of the existing rulesets ?

I'd be happy to create the sniff and fix the other two I mentioned above (I actually already started to do so), but figured this warranted an issue to discuss before I opened a PR.

gsherwood:
A similar issue was fixed in PSR2.Namespaces.NamespaceDeclaration as part of bug #2095

A sniff sounds like a good idea. Fixes would be great.

jrfnl:
That might well have been the issue which inspired me to look into the handling of this. Expect some PRs soon.

jrfnl:
I'm going to close this issue.

I did work on this at the time, but it was more complicated than I anticipated.

All the same, as whitespace and comments within namespaced names will now become a [parse error in PHP 8.0](https://wiki.php.net/rfc/namespaced_names_as_token#backward_incompatible_changes), a sniff for this was needed for the [PHPCompatibility PHPCS standard](https://github.com/PHPCompatibility/PHPCompatibility) anyway and with the benefit of two more years of sniff writing, I have finished the sniff and will pull it to PHPCompatibility in the near future.

[...]

PR #2237, opened 2018-11-21
Squiz/SelfMemberReference: various bug fixes

### LINKED ISSUE #2232: SelfMemberReference.NotUsed reports a violation inside anonymous classes
This sniff reports a violation inside anonymous classes where it is not applicable:

```
class Foo {
	public const BAR = 1;

	public function example() : object
	{
		return new class {
			public function __toString() : string
			{
				// Must use "self::" for local static member reference
				return Foo::BAR;
			}
		}
	}
}
```

jrfnl:
This PR takes care of the issues identified in https://github.com/squizlabs/PHP_CodeSniffer/pull/2185#issuecomment-430027868 as well as some other bugs.

**Note**: As the sniff is now more directly targeted at the use of `self` in the right places, I would strongly recommend for the whitespace related errors - `SpaceBefore` and `SpaceAfter` - to be removed from this sniff.
The `Squiz.WhiteSpace.ObjectOperatorSpacing` sniff already handles this anyway and handles this better.

### Squiz/SelfMemberReference: prevent false positives icw anonymous classes

Fixes #2232

### Squiz/SelfMemberReference: prevent false negatives for NotUsed error

... when the double colon is preceded by whitespace/comments.

### Squiz/SelfMemberReference: don't hide one error behind another

Prevent the sniff from not throwing the whitespace errors when the `NotUsed` error is also found.

### Squiz/SelfMemberReference: allow for namespace declarations interspersed with whitespace/comments

Loosely related to #2150

### Squiz/SelfMemberReference: only strip leading namespace separator 

... to prevent the name comparison breaking.

The PR includes unit tests for all fixes.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Classes/SelfMemberReferenceSniff.php
+++ b/src/Standards/Squiz/Sniffs/Classes/SelfMemberReferenceSniff.php
@@ -108,9 +108,10 @@ class SelfMemberReferenceSniff extends AbstractScopeSniff
 
                     $phpcsFile->fixer->replaceToken($stackPtr, 'self::');
                     $phpcsFile->fixer->endChangeset();
-                }
 
-                return;
+                    // Fix potential whitespace issues in the next loop.
+                    return;
+                }
             }//end if
         }//end if
 
--- a/src/Standards/Squiz/Sniffs/Classes/SelfMemberReferenceSniff.php
+++ b/src/Standards/Squiz/Sniffs/Classes/SelfMemberReferenceSniff.php
@@ -78,7 +78,8 @@ class SelfMemberReferenceSniff extends AbstractScopeSniff
         } else if ($tokens[$calledClassName]['code'] === T_STRING) {
             // If the class is called with a namespace prefix, build fully qualified
             // namespace calls for both current scope class and requested class.
-            if ($tokens[($calledClassName - 1)]['code'] === T_NS_SEPARATOR) {
+            $prevNonEmpty = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($calledClassName - 1), null, true);
+            if ($prevNonEmpty !== false && $tokens[$prevNonEmpty]['code'] === T_NS_SEPARATOR) {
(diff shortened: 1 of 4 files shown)


marcospassos:
I can confirm this PR fixes #2232.

jrfnl:
@marcospassos Thanks for testing!

gsherwood:
Thanks a lot for doing this.

I agree that removing the whitespace checks in favour of the other sniff is a good idea. It is technically a BC break, but there is a direct replacement and this change would make the sniff more targeted.

But the main fixer in there assumes it can remove whitespace and re-position things. If the sniff doesn't check whitespace, it also can't alter it. I went to unwind the whitespace checks and realised I didn't have enough time to fix the fixers that are in there, so I've left it for now.

If anyone wants to give it a go, please do as I think it's a worthwhile change.

jrfnl:
You're welcome.

> But the main fixer in there assumes it can remove whitespace and re-position things. If the sniff doesn't check whitespace, it also can't alter it. I went to unwind the whitespace checks and realised I didn't have enough time to fix the fixers that are in there, so I've left it for now.

Can you clarify the above statement a bit more ? I'm not sure what you mean.
And is it this sniff or the other one for which the fixer would need adjusting ?

gsherwood:
> Can you clarify the above statement a bit more ? I'm not sure what you mean.

After removing the two blocks looking at whitespace before/after, I noticed another fixer was also removing whitespace while fixing the non-self reference. Because the sniff would no longer be enforcing no whitespace around the double colon, it would need to also leave that whitespace there while fixing so it doesn't conflict with other rules.

It may not be a complicated change, or I was seeing things, but I had to move on. I felt like I'd need a bunch more text cases to make sure comment and whitespace was retained by the sniff during all fixing.

jrfnl:
@gsherwood I'll look into this in more detail later (as part of the conflict checks), but a quick look at this tells me that this is not so much a bug in this sniff itself, but that the newly added unit tests expose existing bugs in other sniffs.

PR #2239, opened 2018-11-23
[WIP] Squiz.WhiteSpace.FunctionSpacing and nested functions

thewilkybarkid:
Currently nested functions still have extra lines appended after. Since this applies to the last one in the list, I wouldn't expect it (refs https://github.com/libero/php-coding-standard/pull/30#discussion_r230062042). I've left this as WIP, however, because I'm not sure that this would apply globally (maybe should be a new config option?).


gsherwood:
> Maybe the specs of what needs to be solved, if it should be solved in PHPCS and if so, how ? should be discussed in an issue first ?

This is a good idea. I'd be open to the idea of changing the Squiz sniff as it was written to apply to class methods. We don't often use global functions or nested functions, so the rules could be changed for those - either in a different sniff or this one.

gsherwood:
Closing as this needs discussion before a PR is proposed. A new issue would be good, if this is still relevant.

Issue #314, opened 2014-11-07
Ability to customise `$checkedTokens` in `Generic.CodeAnalysis.EmptyStatement` from within a custom rule


u01jmg3:
- At [Generic.CodeAnalysis.EmptyStatement#L41-50](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Sniffs/CodeAnalysis/EmptyStatementSniff.php#L41-50) it would be useful to be able to customise `$checkedTokens` from within a custom rule
  - Currently I have had to edit [Generic.CodeAnalysis.EmptyStatement](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Sniffs/CodeAnalysis/EmptyStatementSniff.php#L47) to exclude `T_IF` but if I could do this within my ruleset it would be much better. Unless I am missing a way of doing this without editing the file?

e.g.

``` xml
<rule ref="Generic.CodeAnalysis.EmptyStatement">
    <properties>
        <property name="checkedTokens" type="array" value="T_CATCH,T_DO,T_ELSE,T_ELSEIF,T_FOR,T_FOREACH,T_SWITCH,T_TRY,T_WHILE" />
    </properties>
</rule>
```

PR #2240, opened 2018-11-23
Generic.CodeAnalysis.EmptyStatement Add option to allow only comments in statements

thewilkybarkid:
There are cases where it's valid to only have a comment in a statement, this adds an option to allow it (refs https://github.com/libero/php-coding-standard/pull/34#discussion_r233818322).

### DIFF
--- a/src/Standards/Generic/Sniffs/CodeAnalysis/EmptyStatementSniff.php
+++ b/src/Standards/Generic/Sniffs/CodeAnalysis/EmptyStatementSniff.php
@@ -30,6 +30,13 @@ use PHP_CodeSniffer\Util\Tokens;
 class EmptyStatementSniff implements Sniff
 {
 
+    /**
+     * Whether to allow statements that contain only comments.
+     *
+     * @var boolean
+     */
+    public $allowComments = false;
+
 
     /**
      * Registers the tokens that this sniff wants to listen for.
@@ -74,10 +81,16 @@ class EmptyStatementSniff implements Sniff
             return;
         }
 
+        if ($this->allowComments === true) {
+            $emptyTokens = ([T_WHITESPACE => T_WHITESPACE] + Tokens::$phpcsCommentTokens);
+        } else {
+            $emptyTokens = Tokens::$emptyTokens;
+        }
+
         $next = $phpcsFile->findNext(
-            Tokens::$emptyTokens,
+            $emptyTokens,
             ($token['scope_opener'] + 1),
-            ($token['scope_closer'] - 1),
+            $token['scope_closer'],
             true
         );
 
--- a/src/Standards/Generic/Tests/CodeAnalysis/EmptyStatementUnitTest.inc
+++ b/src/Standards/Generic/Tests/CodeAnalysis/EmptyStatementUnitTest.inc
@@ -70,3 +70,17 @@ try {
 }
 
 if (true) {} elseif (false) {}
+
+// phpcs:set Generic.CodeAnalysis.EmptyStatement allowComments true
+
+if ($foo) {
+    // Just a comment
+} elseif ($bar) {
+    /*
+     * Yet another comment
+     */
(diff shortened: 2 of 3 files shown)


src/Standards/Generic/Sniffs/CodeAnalysis/EmptyStatementSniff.php
```diff
@@ -74,10 +81,16 @@ public function process(File $phpcsFile, $stackPtr)
...
+        } else {
+            $emptyTokens = Tokens::$emptyTokens;
+        }
+
         $next = $phpcsFile->findNext(
-            Tokens::$emptyTokens,
+            $emptyTokens,
             ($token['scope_opener'] + 1),
-            ($token['scope_closer'] - 1),
+            $token['scope_closer'],
```
thewilkybarkid: Wasn't expecting to have to change this, but skips the last token otherwise.

jrfnl: This is a known issue, see https://github.com/squizlabs/PHP_CodeSniffer/issues/1319

src/Standards/Generic/Tests/CodeAnalysis/EmptyStatementUnitTest.inc
```diff
@@ -69,4 +69,18 @@ try {
...
+// phpcs:set Generic.CodeAnalysis.EmptyStatement allowComments true
+
+if ($foo) {
+    // Just a comment
+} elseif ($bar) {
+    /*
+     * Yet another comment
+     */
+} elseif ($baz) {
+    // phpcs:set Generic.CodeAnalysis.EmptyStatement somethingMadeUp true
```
jrfnl: To prevent confusion between real test related PHPCS annotations and annotations which are part of the test, it's better to use something like `//phpcs:ignore Standard.Category.SniffName -- for reasons` here.

src/Standards/Generic/Tests/CodeAnalysis/EmptyStatementUnitTest.inc
```diff
@@ -69,4 +69,18 @@ try {
...
+} else {
+
+}
```
jrfnl: You should probably reset the changed property to the default at the end of this test to prevent issues with future test cases being added.

jrfnl:
Just thinking out loud: is a property the way to go here or should this be solved via error codes, i.e. having a different error code for a completely empty statement vs an empty statement only containing a comment ?

That way, you could, for instance, exclude `DetectedCatchComment`, while still throwing errors for `DetectedCatch` and `DetectedIfComment`.

thewilkybarkid:
I like the idea of more granular errors. Might need to go for something like `DetectedCatch.ContainsComment` to maintain BC? (eg to not change the behaviour of https://github.com/doctrine/coding-standard/blob/0247a0d440e196d93611bd0f3c561ee12bd82789/lib/Doctrine/ruleset.xml#L24-L27)

jrfnl:
Error codes are not supposed to contain `.`s, which is why I suggested `DetectedCatchComment`.

Adding additional error codes will be BC-break, but only a quite minor one in this case, so I suspect it would be an acceptable one, though @gsherwood can give you a more definite answer.

gsherwood:
I prefer the new error codes over the sniff property as it is more flexible, but this BC break will silently stop existing include and exclude rules from working.

This sniff was changed to use new error codes (issues #314) so that it could be customised more easily, so I assume people are doing that today.

What PHPCS probably needs is the ability to provide aliases for message codes to allow for these sort of deprecations to occur. For example, we throw a `EmptyCatchWithComment` but we alias it as `EmptyCatch` so it still works with old rulesets. These aliases can then be removed during a major release.

jrfnl:
PR #3409 was marked as a duplicate of this.

PR #2244, opened 2018-11-24
[Minor] Move isSpecialMethod

gmponos:
Minor improvement by moving $isSpecialMethod to the next lines. This way in case you return to the if statement before you save a not needed comparison.


gsherwood:
Nice one. Thanks.

PR #2245, opened 2018-11-24
Added an autofix for Squiz.Commenting.FunctionComment.ParamCommentNotCapital

gmponos:
😄


gsherwood:
The things that worry me about this change are:
- it's doing a lot of spacing fixes for an error that is only supposed to be checking and fixing the first letter of a comment
- if the comment does not start with a letter, the error is not actually fixable and so shouldn't say it is
- I don't know enough about languages to know if `ucfirst()` will actually work on everything, and it will cause the fixer to fail if it doesn't

For the last 2 reasons, I generally don't want to play around with developer comments, so I'm hesitant to merge this in.

gmponos:
Sounds reasonable... closing

PR #2249, opened 2018-11-27
Squiz.CSS.ClassDefinitionClosingBrace: add fixed file and various improvements

jrfnl:
[Fixer conflict series PR] + [Missing fixed files series PR]

The sniff contains fixers, but didn't have a `fixed` file.

* The sniff did not handle multiple statements on the same line correctly.
* The fixer would inadvertently remove line indentation. Indentation is not the concern of this sniff, so should be left alone.
* The sniff did not take trailing comments or PHPCS annotations into account, which would cause a fixer conflict with itself.
    - The fixer has been adjusted to handle trailing comments correctly.


gsherwood:
Thanks for this PR, and the other CSS ones I just merged in.

PR #2252, opened 2018-11-27
Squiz/LowercasePHPFunctions: add fixed file and various improvements

jrfnl:

The sniff contains a fixer, but didn't have a `fixed` file.

* The sniff will now bow out early when the `T_STRING` content does not match a PHP native function.
* The sniff did not detect non-lowercase PHP native functions when these were being imported via a `use function` statement.
    Detection of this has been added with a distinct, separate error code.
* The sniff would incorrectly not throw an error when the namespace separator was used to indicate global namespace, i.e. `$a = \str_repeat();`
* The sniff did not take randomly interspersed comments or PHPCS annotations into account.
* Minor code de-duplication fixes.
* More tests to document that some other `T_STRING`s, such as class and namespace declarations are not incorrectly recognized as function calls. This is mostly to guard against breaking this in the future.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1600,6 +1600,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="InnerFunctionsUnitTest.inc" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="InnerFunctionsUnitTest.php" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="LowercasePHPFunctionsUnitTest.inc" role="test" />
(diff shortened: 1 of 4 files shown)


jrfnl:
Just realized some of the fixes in this PR overlap with PR #1546.
for those parts which overlap, this is a more simplistic compared to what is being done in #1546. /cc @webimpress

gsherwood:
Thanks for this. I did make some changes though.

I really didn't like the idea of calling `addError` when there may or may not be an error, or passing in a partial error message string (e.g., `Calls to`), so I reverted those changes. Then I realised that the check to see if the function name was lowercase should probably be the first thing done, so I moved that check up the top to compliment your other shortcutting change. I left the code repeated just because there wasn't much left to share after that.

jrfnl:
@gsherwood I totally understand. Makes sense to me (and I'm left wondering why I didn't do that myself) :+1:

PR #2253, opened 2018-11-27
Squiz.Objects.ObjectMemberComma: add fixed file and minor fix

jrfnl:

The sniff contains fixers, but didn't have a `fixed` file.

Notes:
* While this constitutes a BC-break, the error code used by this sniff was incorrect as the error is thrown when a comma is _found_, not when it's _missing_.
* Added an additional unit test documenting that the sniff correctly skips over PHPCS annotations and comments.
* Removed some redundant code from the unit test file.


gsherwood:
Thanks. The error code BC issue can be resolved by referencing both the old and new code in a ruleset as PHPCS will just never use the new code in old PHPCS versions, or the old code in new PHPCS versions. So I think that change is ok, and obviously makes sense.

PR #2254, opened 2018-11-27
Squiz/LogicalOperatorSpacing: add fixed file

jrfnl:


talaseth:
Is it possible you could tell it to warn if there are no spaces? Currently ```if ($a&&$b)``` generates no warning, nor does ```if ($a==$b)``` and so on.

jrfnl:
> Is it possible you could tell it to warn if there are no spaces? Currently `if ($a&&$b)` generates no warning, nor does `if ($a==$b)` and so on.

@talaseth While you pose a valid question, it is unrelated to this PR, so you may want to open a new issue for your question next time.

All the same, the sniff will actually throw errors for the `if ($a&&$b)` example you gave. Just make sure you include this sniff in your ruleset.

As for the `if ($a==$b)` code, a `==` is not a logical operator, so that code is irrelevant to this sniff.
If you want the spacing around the `==` operator checked, use the `Squiz.WhiteSpace.OperatorSpacing` sniff instead or in addition to this one.

PR #2261, opened 2018-11-27
PSR2/EndFileNewline: add missing fixed files + minor improvement

jrfnl:

Additionally, I've hard-coded the end of line character to be used by the fixer to `\n` as PSR2 demands that anyway.
This should give a minor efficiency improvement for when a file has, for instance, Windows line endings which would be recorded as such in `$phpcsFile->eolChar`. This way the `Generic.Files.LineEndings` sniff doesn't have to go over the file to correct it (again) in a second loop.


gsherwood:
I don't hard-code the line endings on purpose because it is not the job of this sniff to change line endings. It makes it easier to mix and match sniffs into custom standards, in my mind anyway.

I'll change that back, but thanks for the fixed files (and all the others you've submitted).

jrfnl:
Fair enough, though for mixing and matching I would expect people to use the `Generic` version of this sniff PSR2 is quite clear after all about the line endings.

PR #2262, opened 2018-11-27
Squiz/LowercaseClassKeywords: add fixed file and minor improvements

jrfnl:

The sniff contain a fixer, but didn't have a `fixed` file.

* The sniff will now also examine the `class` keyword for anonymous classes.
* Minor code de-duplication fix.


gsherwood:
Thanks for this

PR #2263, opened 2018-11-27
Generic/LineEndings: add fixed files + minor bug fix

jrfnl:

The sniff contain a fixer, but didn't have a `fixed` file for the CSS and JS test case files.

While creating the `fixed` files, I noticed that the `T_CLOSE_TAG`, which for CSS and JS files is empty, would also be replaced with a new line, resulting in an extra, superfluous new line at the end of a file.
This has now been fixed.


gsherwood:
Thanks

PR #2265, opened 2018-11-27
Generic/EndFileNoNewline: add missing fixed files + minor improvements

jrfnl:

While I was add it, I've reviewed the sniff and applied any improvements I could think of:
* Fix the file in one go instead of removing one blank line per loop.
* Added additional unit test to make sure a line containing non-line-ending whitespace at the end of a file will not get removed by this sniff.
* Remove some redundant code from the unit test file.


gsherwood:
Thanks

PR #2273, opened 2018-11-29
Generic/ScopeIndent: lower memory usage

jrfnl:
Based on an analysis of some runs with [BlackFire](https://blackfire.io/), this small tweak will lower the memory usage of any run which includes the `Generic.WhiteSpace.ScopeIndent` - and by extension, the `PEAR.WhiteSpace.ScopeIndent` - sniff significantly.

A comparison between a run on `master` and a run on this branch can be found here: https://blackfire.io/profiles/compare/7adc96c7-09a4-4ec8-8821-bbf9d5906b2f/graph
**Note**: this comparison is only available for 24 hours.

The benchmark was created by running PHPCS over itself with the native ruleset and this command:
`phpcs -p --report-source --no-cache`.

For when the comparison isn't available anymore, this is the significant part:
phpcs-generic-scopeindent-tweak

### DIFF
--- a/src/Standards/Generic/Sniffs/WhiteSpace/ScopeIndentSniff.php
+++ b/src/Standards/Generic/Sniffs/WhiteSpace/ScopeIndentSniff.php
@@ -238,8 +238,9 @@ class ScopeIndentSniff implements Sniff
                     if (isset($tokens[$parenCloser]['nested_parenthesis']) === true
                         && empty($tokens[$parenCloser]['nested_parenthesis']) === false
                     ) {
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for this. Where did you discover this trick? Or did you just play around with it for a while?

jrfnl:
Can't remember specifically, but learned it somewhere along the way over the years.

jrfnl:
[ClearPHP](https://github.com/dseguy/clearPHP) does come to mind as a good resource for things like this though.

PR #2274, opened 2018-11-30
Generic/DisallowMultipleStatements: add fixed file + minor bug fix

jrfnl:
The sniff contain a fixer, but didn't have a `fixed` file.

The fixer did not yet take the new PHPCS annotations into account. The only one which is problematic if it's moved, is the `T_PHPCS_IGNORE` token, so if that token would be encountered, the error will still be thrown, but will no longer be fixable.

Includes additional unit test covering this.


gsherwood:
Thanks

Issue #2222, opened 2018-11-12
Feature suggestion: AbstractDocblockSniff


jrfnl:
Currently, the PEAR and the Squiz standard both contain a number of docblock related sniffs which are not configurable and are opinionated on what tags should be present in the various docblocks.
Additionally, the various sniffs contain duplicate code.

External standards can, of course, create their own sniffs for these type of checks, however, that would involve yet more code duplication.

@gsherwood Would there be interest in an `AbstractDocblockSniff` ?

I imagine that sniff could roughly look like:
```php
abstract class AbstractDocblockSniff implements Sniff
{
    /**
     * Format similar to the format used in the PEAR.Commenting.FileComment sniff.
     *
     * I.e.:
     * `tagname` => array(
     *     'tag_required'     => true/false,
     *     'allow_multiple'   => true/false,
     *     'content_required' => true/false,
     * )
     */
    public $tags = array();

    /**
     * Whether the tag order is important and should be checked.
     *
     * @var bool
     */
    public $checkTagOrder = false;

    abstract public function register();

    public function processToken()
    {
        $docblockOpener = $this->findTargetDocblock();
        if ($docblockOpener === false) {
            $this->MissingDocBlock();
            return;
        }

        $this->examineDocBlockFormat();
        $this->examineTags();
    }

    /**
     * Find the target docblock based on whatever token(s) was registered in the register() method.
     */
    abstract public function findTargetDocblock();

    public function missingDocBlock()
    {
        // Throw error for missing docblock.
    }

    public function examineDocblockFormat()
    {
        // Check for docblock vs comment format.
        // Superfluous empty comment lines etc.
    }

    public function examineTags()
    {
        // Check based on tag_required status
        // Check based on allow_multiple status
        // Check based on content_required status

        // Allow for additional tag specific checks in child sniffs.
        if (method_exists($this, 'examine'.$tagName)) {
            call_user_func(array($this, 'examine'.$tagName), $args);
        }

        // Maybe check tag order.
    }

    // Possibly some examineTagName() methods for common tags.
    // Those can, of course, be overloaded by child sniffs with either a completely different
    // check of with a call to the parent in combination with additional checks.

    // Can be used by child sniffs from the tag specific check methods.
    public function checkPunctuation( $string )
    {
        // Check for capital at start of phrase.
        // Check for valid end of sentence character.
    }
}
```

[...]

PR #2275, opened 2018-11-30
Squiz/FileComment: add fixed file

jrfnl:

While I was at it, I've done a cursory review of the sniff and have added some additional unit test files for those errors which were, so far, untested.


gsherwood:
The fixer uses the current year to generate the copyright statement, which has since broken these tests now that we are in 2019. I'll update them to use 2019 so at least the fixers are tested for 12 months, but a better solution would be nice. I can't think of one off the top of my head though.

gsherwood:
Thanks for the test files

gsherwood:
@jrfnl Yeah, option 2 is probably the way to go there. Ultimately, this sniff needs to be removed. It's a long forgotten sniff from the early days of PHPCS and it should be gone. So maybe it's not worth making this change at all.

jrfnl:
@gsherwood Well, in that case, maybe we should leave it until the `AbstractDocblockSniff` is in place (#2222) ?
Once that abstract is available it will be so much easier for external standards to create their own `FileComment` sniff that removing this sniff should give less friction.

gsherwood:
> Well, in that case, maybe we should leave it until the `AbstractDocblockSniff` is in place

Done

PR #2276, opened 2018-11-30
Squiz/FunctionOpeningBraceSpace: add missing fixed files + some extra unit tests

jrfnl:

While I was at it, I've done a cursory review of the sniff and have added some additional unit test files for situations, so far, untested.

Includes removing some redundant code from the unit test file.


gsherwood:
Thanks

PR #2277, opened 2018-11-30
Squiz/Functionspacing: fix fixer conflict + a fixer bug

jrfnl:
[Fixer conflicts series PR]

`phpcbf -p -s ./src/Standards/Squiz/Tests/WhiteSpace/FunctionOpeningBraceSpaceUnitTest.inc --standard=Squiz`

Introduced by me via #2110, which was included in the `3.3.2` release, and didn't take global functions straight after a PHP open tag into account. Mea Culpa.

Basically, when a function is on the first line after a PHP open tag at the start of the file, the whitespace would be added before the open tag, not before the function.
This in turn causes a conflict with the `SuperfluousWhiteSpace` sniff which will try to remove whitespace before the open tag at the start of file.

While adding unit tests for this, I also discovered a second, more serious bug, where, when there were too _many_ blank lines between a PHP open tag at the start of a file and the first function, the PHP open tag would be removed altogether, which would cause the file to be regarded as inline HTML.

These situations were not covered by unit tests yet and all need separate files to be tested properly. These tests have now been added to prevent this issue from reoccurring in the future.

Additionally, I've added:
* A unit test documenting how the sniff handles blank lines around functions when the function is not on its own line.
* Unit tests documenting that the sniff also handles functions in traits and anonymous classes correctly.
* Some minor other code simplifications and documentation improvements.


jrfnl:
Rebased to get passed imaginary merge conflict.

gsherwood:
Thanks

PR #2278, opened 2018-11-30
Squiz.Operators.ComparisonOperatorUsage false positive when inline IF contained in parentheses

aboks:
This bug caused false positives from Sniff.Operators.ComparisonOperatorUsage for inline if statements that were completely surrounded by a pair of parentheses, e.g. from a function call.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Operators/ComparisonOperatorUsageSniff.php
+++ b/src/Standards/Squiz/Sniffs/Operators/ComparisonOperatorUsageSniff.php
@@ -125,7 +125,7 @@ class ComparisonOperatorUsageSniff implements Sniff
                         ) {
                             break;
                         }
-                    }
+                    }//end if
                 }//end for
 
                 $start = $phpcsFile->findNext(Tokens::$emptyTokens, ($i + 1), null, true);
--- a/src/Standards/Squiz/Sniffs/Operators/ComparisonOperatorUsageSniff.php
+++ b/src/Standards/Squiz/Sniffs/Operators/ComparisonOperatorUsageSniff.php
@@ -116,6 +116,15 @@ class ComparisonOperatorUsageSniff implements Sniff
                         if (isset($tokens[$i]['scope_closer']) === true) {
                             break;
                         }
+                    } else if ($tokens[$i]['code'] === T_OPEN_PARENTHESIS) {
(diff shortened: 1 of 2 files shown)


src/Standards/Squiz/Sniffs/Operators/ComparisonOperatorUsageSniff.php
```diff
@@ -116,7 +116,16 @@ public function process(File $phpcsFile, $stackPtr)
                         if (isset($tokens[$i]['scope_closer']) === true) {
                             break;
                         }
-                    }
+                    } else if ($tokens[$i]['code'] === T_OPEN_PARENTHESIS) {
+                        // Stop if this is the start of a pair of
+                        // parentheses that surrounds the inline
+                        // IF statement.
+                        if (isset($tokens[$i]['parenthesis_opener']) === true
```
jrfnl: Just checking: did you mean for this to be `'parenthesis_closer'` ?

aboks: Hmm, as far as I understand (but I may be mistaken), `parenthesis_opener` and `parenthesis_closer` always occur in pairs, so both conditions would be equivalent. I meant this condition as "if this parenthesis opens a parenthesis scope and it's corresponding closer is beyond the inline if...". For me, the check on `parenthesis_opener` was a natural translation of the first part of that condition. If that makes the code more confusing, I'm happy to change it to `parenthesis_closer` however.

jrfnl: Well, what parse errors or live coding ? The open parenthesis may exist, but the closer may not yet have been entered.

As you're comparing against the `parenthesis_closer` on the next line, using the same in the `isset()` should make the code more stable against potential future tokenizer changes and will prevent potential `undefined index` notices.

aboks: Ah, in that case I've probably misunderstood how `parenthesis_opener` and `parenthesis_closer` work. I've changed the `isset` to check for `parenthesis_closer`. Thanks for pointing this out!

gsherwood:
Thanks a lot for fixing this.

PR #2279, opened 2018-11-30
Squiz/SuperfluousWhitespace: fix fixer conflict with nested anonymous functions

jrfnl:
[Fixer conflicts series PR]

Similar to the issues which were fixed in #2182 and #2192, the `Squiz.WhiteSpace.SuperfluousWhitespace` sniff also did not yet take nested anonymous classes into account.

Spacing around properties and functions within nested anonymous classes have their own rules and the "multiple empty lines in a row" error should not be triggered for these.

This fixes a fixer conflict between the `FunctionSpacing` sniff and this one.

Includes unit test.


gsherwood:
Thanks

Issue #2146, opened 2018-08-30
Zend.Files.ClosingTag removes closing tag from end of file without inserting a semicolon


TysonAndre:
Version: I'm able to reproduce this in the latest commit.
Expected: If `<?=` is used and a semicolon is needed in the fixed file, a semicolon should be inserted at an appropriate place (e.g. not as part of a line comment). (Or skip warning about it and fixing it)
Observed: No semicolon is inserted

The input file

```php
text
<?php include($this->add('arg'))?>
```

The observed result (results in a parse error):

```php
text
<?php include($this->add('arg'))
```

An example ruleset.xml:

```
<?xml version="1.0"?>
<ruleset name="ifwe">
    <description>Example coding standard</description>

    <rule ref="Zend">
    </rule>

</ruleset>
```

Command: `path/to/PHP_CodeSniffer/bin/phpcbf  --standard=ruleset.xml src/bug.php`

----

A workaround is to exclude Zend.Files.ClosingTag.NotAllowed and include PSR2.Files.ClosingTag.NotAllowed or PSR2.

gsherwood:
Thanks for reporting this. The fix will be in 3.3.2.

You may also consider using PSR2.Files.ClosingTag instead of the Zend sniff. It already accounted for this and only enforces no closing tag when the file is PHP-only. In the mixed case (like in your example) the closing tag would be allowed. That's obviously personal preference, but thought I'd point it out in case you wanted the tag kept in these cases.

PR #2280, opened 2018-11-30
Zend/ClosingTag: bug fix /fixer conflict

jrfnl:
[Fixer conflicts series PR]

`phpcbf -p -s ./src/Standards/Generic/Tests/PHP/CharacterBeforePHPOpeningTagUnitTest.1.inc --standard=Squiz`

Introduced in commit e64a980f7bc5a68a4d720e3e74f68d3997bf919e in response to issue #2146.

If there is nothing between the PHP open/close tag or the only thing between them is a comment, a stray semi-colon is introduced by the fixer.
This then causes a conflict with the `Squiz.WhiteSpace.SemicolonSpacing` sniff, quite apart from it being invalid code.

Includes new unit tests covering the issue.


gsherwood:
Thanks for fixing that

PR #2283, opened 2018-12-01
PSR1/PSR2 rulesets: improve inline documentation

jrfnl:
Use complete sniff name instead of partial file name reference to avoid confusion what with there being sniffs with the same name in different standards.


gsherwood:
Having the path there indicated to me that the sniff was part of this standard and would be found in a particular place, but the message codes are fine for that as well, and probably reduce confusion when you've got a few rulesets open. Thanks.

PR #2288, opened 2018-12-04
Few more minor memory management tweaks

jrfnl:
Related to #2273 - I've done a quick search of rest of the code base and adjusted any usage of the `end()` / `key()` functions in combination with the `$tokens` array.

All relevant instances of this in the existing code base have now been fixed for the memory leakage.


gsherwood:
Thanks

PR #2289, opened 2018-12-04
Unit Tests: report on number of sniff test case files tested

jrfnl:
This commit adds an additional line below a unit test run, listing how many sniff test case files were examined. This information can be used as input to check and debug unit test runs.

### DIFF
--- a/tests/Standards/AbstractSniffUnitTest.php
+++ b/tests/Standards/AbstractSniffUnitTest.php
@@ -127,6 +127,7 @@ abstract class AbstractSniffUnitTest extends TestCase
 
         // Get a list of all test files to check.
         $testFiles = $this->getTestFiles($testFileBase);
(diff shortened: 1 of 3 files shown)


gsherwood:
Good idea, although the extra lines were getting to me, so I just compressed it to one line. Looks like this now:
```
$ phpunit
PHPUnit 4.8.35 by Sebastian Bergmann and contributors.

...............................................................  63 / 366 ( 17%)
............................................................... 126 / 366 ( 34%)
.......................................S....................... 189 / 366 ( 51%)
............................................................... 252 / 366 ( 68%)
............................................................... 315 / 366 ( 86%)
.........................................SS........

371 sniff test files generated 643 unique error codes; 374 were fixable (58.16%)

Time: 16.98 seconds, Memory: 36.00MB

OK, but incomplete, skipped, or risky tests!
```

jrfnl:
LGTM :+1:

PR #2290, opened 2018-12-04
Remove unnecessary strlen() function calls

### LINKED ISSUE #2285: FunctionDeclarationArgumentSpacingSniff uses strlen on content not spaces
The `FunctionDeclarationArgumentSpacingSniff` uses `strlen` on token content, so technically it could be `\t` not the space, and the error is not going to be reported. But configuration says about number of spaces and fixer uses spaces.

jrfnl:
The `$tokens` array already contains the `length` of each and every token, so there is no need to (re-)calculate this in individual sniffs, unless we're examining `T_COMMENT` or `T_INLINE_HTML` tokens.

Using the `length` from the tokenizer array should also be more accurate when tabs have been replaced with spaces or when dealing with Unicode texts.

Fixes #2285


gsherwood:
Thanks a lot of finding all those

PR #2291, opened 2018-12-04
Generic/InlineControlStructure: fix fixer conflict / parse error resilience

jrfnl:
[Fixer conflict series PR]

The `Generic.ControlStructures.InlineControlStructure` sniff did not handle live coding situations/parse errors and would try to fix the file anyway.

This causes a fixer conflict. This is now fixed by adding a number of checks against parse errors/live coding.

The fixer conflict can be seen by checking out any of the new test case files and running the following command:
`phpcbf -ps ./src/Standards/Generic/Tests/ControlStructures/InlineControlStructureUnitTest.3.inc --standard=squiz -vv`

At the start of the fixer loop you can see that the sniff adds `{;}` which then causes other sniffs to kick in and conflict.
Some of these other sniffs also still need fixes for this, but this PR addresses the part which is in the `Generic.ControlStructures.InlineControlStructure` sniff.

Includes unit tests in dedicated unit test files to prevent these tests interfering with each other and/or the main test file.


gsherwood:
Thanks

PR #2294, opened 2018-12-04
Squiz/FunctionCommentThrowTag: remove some redundant code

jrfnl:
This simplifies the check on whether or not the sniff should run.


gsherwood:
Nice one, thanks.

Issue #2189, opened 2018-10-17
Refactor suggestion / move utility functions from File to static classes


jrfnl:
Since PHPCS 3.x now has a minimum PHP requirement of PHP 5.4, I was wondering whether it might be an idea to refactor a number of groups of methods in the `File` class to traits.

The reasoning behind this is two-fold:
1. Only a limited number of sniffs use the methods I have in mind for this, so having them in traits rather than in the `File` class makes the re-useable code more modular and easier to maintain.
2. Over the years I have written a number of additional token walking utility methods which I think belong in PHPCS rather than in the WPCS/PHPCompatibility standards in which they now live.
    Traits would allow for these methods to be added to PHPCS more easily without making the `File` class even larger than it already is.

### Timing

The transition could be made in PHPCS 3.x.
In PHPCS 3.x, the `File` class would in that case `use` the new traits and for the transitioned methods just call the method in the trait under the hood.
The methods in the `File` class would be deprecated and could be removed in PHPCS 4.x.
This would allow external standards some time to implement the new way of doing things and make the BC break in PHPCS 4.x smaller as it can already be mitigated beforehand.

### Proposal details

To be concrete, I'm currently thinking of the following Traits:
* `DeclarationName`, containing:
    - `getName()`, previously called `getDeclarationName()`
* `FunctionDeclaration`, containing:
    - `getParameters()`, previously called `getMethodParameters()`
    - `getProperties()`, previously called `getMethodProperties()`
    - and using `DeclarationName` for the `getName()` method
* `PropertyDeclaration`, containing:
    - `getMemberProperties()`
* `ObjectDeclaration` containing:
    - `getClassProperties()`
    - `findExtendedClassName()`
    - `findImplementedInterfaceName()`
    - `findExtendedInterfaceNames()` - open PR #2128
    - and using `DeclarationName` for the `getName()` method
* `Generic` containing:
    - `isReference()`
* `ScopeConditions`, containing:
    - `hasCondition()`
    - `getCondition()`

Additionally, I'd be happy to pull the following based on battle-tested code I've written for WPCS/PHPCompatibility:
* `FunctionCall`, containing:
    - `hasParameters()`
    - `getParameterCount()`
    - `getParameters()`
    - `getParameter()` (to get the details for the parameter at a specific position)
    These methods all work on both function calls as well as for splitting an array into the individual items and could be extended to cover `isset()`, `unset()` and `list()` as well.
[...]

PR #2295, opened 2018-12-04
Core tests: set up abstract class and remove code duplication

jrfnl:
Note: this PR will be easiest to review by looking at the individual commits.

I've created this PR as a precursor to #2189 (move utility methods to static classes) as most of the new utility functions which are mentioned in that issue will be accompanied by unit tests when they are pulled.

Improving the setup of the utility function testing, as done in this PR, will make that so much simpler.

---

### New `Tests\Core\AbstractMethodUnitTest` class

Refactor to remove duplicate code.

This new class abstracts out the `setUp()` and `tearDown()` methods which were included in nearly all the utility method test classes.

The methods have been changed to static `setUpBeforeClass()` and `tearDownAfterClass()` with a static property as the initial processed `File` doesn't change between tests, so there is no need to tokenize and process the unit test case file for each test individually.

This makes the utility method unit tests significantly faster as well as lower the memory usage.

Additionally, the _standard_ against which the files are processed has been changed from `Generic` (75 sniffs) to `PSR1` (7 sniffs), which, again, makes the unit test suite for the Core utility functions significantly faster.

The current implementation allows for only one test case file per utility method test file, but as the utility method tests are not sniff specific, adding additional tests files for a method if needed, should not be a problem.

By default test case files are expected to be PHP `inc` files. Child classes can overload the static `$fileExtension` property when the test case file is a JS or CSS file.

### Implement use of the new AbstractMethodUnitTest class

... in the various utility method unit test classes.

### Tests\Core\AbstractMethodUnitTest: add new getTargetToken() method

This adds a new `getTargetToken()` method which can retrieve the target token for testing with higher precision than the (duplicate) code which was so far used in the individual test methods.

The improvements this method offers are:
* Avoid test leaking/contamination.
    If/when the token to start the test from was retrieved by doing a `findNext()` from the delimiter comment, a typo could cause a token from the *next* test to be used for the testing instead of the target token.
    If the expected results would be the same for both tests, this would go completely unnoticed.
    This is now no longer possible.
    The only requirement is that the delimiter comments start with `/* test...`.
* No more token counting when setting up the unit tests, just pass the target token type constant to this method and it will get you the correct token.
    This also allows for not having to jump through hoops when deciding where to place the delimiter comment for the test.
* If the token a test looks for is a `T_STRING` or text based token, the optional `$tokenContent` allows for selecting the correct token, even when there are several of the same type in the test case line.

### Implement use of the new AbstractMethodUnitTest::getTargetToken() method

... in the various utility method unit test classes.

### GetMethodPropertiesTest: remove even more duplicate code

Implement use of the new `AbstractMethodUnitTest::getTargetToken()` method as well as abstract out the actual testing to a helper method as the logic was the same in all methods.

I've chosen not to implement this with a data provider as the separate test methods creating the `$expected` arrays make the tests more readable and easier to understand.

### Tests\Core\AllTests: simplify creation of the test suite

As the refactor will add a lot of new unit test files, let's automate the creation of the test suite some more by automatically adding all Test files within the `Test\Core` directory.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -206,6 +206,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
        <file baseinstalldir="" name="AcceptTest.php" role="test" />
       </dir>
      </dir>
+     <file baseinstalldir="" name="AbstractMethodUnitTest.php" role="test" />
      <file baseinstalldir="" name="AllTests.php" role="test" />
      <file baseinstalldir="" name="ErrorSuppressionTest.php" role="test" />
      <file baseinstalldir="" name="IsCamelCapsTest.php" role="test" />
@@ -1959,6 +1960,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
    <install as="TestSuite.php" name="tests/TestSuite.php" />
    <install as="TestSuite7.php" name="tests/TestSuite7.php" />
    <install as="tests/bootstrap.php" name="tests/bootstrap.php" />
+   <install as="CodeSniffer/Core/AbstractMethodUnitTest.php" name="tests/Core/AbstractMethodUnitTest.php" />
    <install as="CodeSniffer/Core/AllTests.php" name="tests/Core/AllTests.php" />
    <install as="CodeSniffer/Core/IsCamelCapsTest.php" name="tests/Core/IsCamelCapsTest.php" />
    <install as="CodeSniffer/Core/ErrorSuppressionTest.php" name="tests/Core/ErrorSuppressionTest.php" />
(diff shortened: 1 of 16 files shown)


jrfnl:
Rebased for imaginary merge conflict.

jrfnl:
Could this PR be milestoned to `3.5.0` please ? It's a prerequisite for the utility method refactor.

jrfnl:
I've added an additional commit to this PR which simplifies the creation of the `Core`/utility method test suite as new tests don't need to be added manually anymore, but will automatically be included.

jrfnl:
As discussed in #2189, this commit is included within #2456.

jrfnl:
I've rebased this PR, fixed the merge conflicts and added two additional commits and minor changes in the existing commits.

The significant differences with the original PR:
* Moved the `FileList` class to the `tests` directory as per feedback in #2456.
    Includes making the necessary changes to the `ValidatePEARPackageXML` functionality to allow for this.
* Actually use the `FileList` Iterator in the `Core\AllTests` class instead of using an interim array.
* Removed the `unset()` for the token positions from the `GetMethodParametersTest`, `GetMethodPropertiesTest` and `GetMemberPropertiesTest` and switched the `assertSame()` to a strict `assertArraySubset()`.
    Ref: https://phpunit.de/manual/4.8/en/appendixes.assertions.html#appendixes.assertions.assertArraySubset

gsherwood:
Thanks a lot for updating this PR.

The core test speed improvement was so noticeable that I had to go and break things just to make sure they were still running :)

PR #2300, opened 2018-12-09
Require one space before name - class declaration

### LINKED ISSUE #2298: PSR2.Classes.ClassDeclaration - one space before name
```php
<?php

class X extends
 Foo
{
}
```

the code is valid, but I think it shouldn't. "Foo" should be in the same line as `extends`.

If we have two spaces before "Foo" then we getting error:

```
[x] Expected 1 space before "Foo"; 2 found (PSR2.Classes.ClassDeclaration.SpaceBeforeName)
```

webimpress:
We require just one space before parent class name (after extends keyword).

Fixes #2298

### DIFF
--- a/src/Standards/PSR2/Sniffs/Classes/ClassDeclarationSniff.php
+++ b/src/Standards/PSR2/Sniffs/Classes/ClassDeclarationSniff.php
@@ -344,19 +344,32 @@ class ClassDeclarationSniff extends PEARClassDeclarationSniff
                         $prev = ($className - 1);
                     }
 
-                    $spaceBefore = strlen($tokens[$prev]['content']);
-                    if ($spaceBefore !== 1) {
+                    $last    = $phpcsFile->findPrevious(T_WHITESPACE, $prev, null, true);
+                    $content = $phpcsFile->getTokensAsString(($last + 1), ($prev - $last));
+                    if ($content !== ' ') {
+                        $found = strlen($content);
+
                         $error = 'Expected 1 space before "%s"; %s found';
                         $data  = [
                             $tokens[$className]['content'],
-                            $spaceBefore,
+                            $found,
                         ];
 
                         $fix = $phpcsFile->addFixableError($error, $className, 'SpaceBeforeName', $data);
                         if ($fix === true) {
-                            $phpcsFile->fixer->replaceToken($prev, ' ');
+                            if ($tokens[$prev]['code'] === T_WHITESPACE) {
+                                $phpcsFile->fixer->beginChangeset();
+                                if ($tokens[($prev - 1)]['code'] === T_WHITESPACE) {
+                                    $phpcsFile->fixer->replaceToken(($prev - 1), ' ');
+                                }
+
+                                $phpcsFile->fixer->replaceToken($prev, ' ');
+                                $phpcsFile->fixer->endChangeset();
+                            } else {
+                                $phpcsFile->fixer->addContent($prev, ' ');
+                            }
                         }
-                    }
+                    }//end if
                 }//end if
             }//end if
 
--- a/src/Standards/PSR2/Tests/Classes/ClassDeclarationUnitTest.inc
+++ b/src/Standards/PSR2/Tests/Classes/ClassDeclarationUnitTest.inc
@@ -163,3 +163,8 @@ class C2
 {
 
 } // phpcs:ignore Standard.Category.Sniff
+
+class C3 extends
+ C2
+{
+}
--- a/src/Standards/PSR2/Tests/Classes/ClassDeclarationUnitTest.inc.fixed
+++ b/src/Standards/PSR2/Tests/Classes/ClassDeclarationUnitTest.inc.fixed
@@ -161,3 +161,7 @@ class C2
 {
 
 } // phpcs:ignore Standard.Category.Sniff
+
+class C3 extends C2
+{
+}
--- a/src/Standards/PSR2/Tests/Classes/ClassDeclarationUnitTest.php
+++ b/src/Standards/PSR2/Tests/Classes/ClassDeclarationUnitTest.php
@@ -50,6 +50,7 @@ class ClassDeclarationUnitTest extends AbstractSniffUnitTest
             130 => 2,
             131 => 1,
             158 => 1,
+            168 => 1,
         ];
 
     }//end getErrorList()


src/Standards/PSR2/Sniffs/Classes/ClassDeclarationSniff.php
```diff
@@ -344,19 +344,32 @@ public function processOpen(File $phpcsFile, $stackPtr)
...
+                            if ($tokens[$prev]['code'] === T_WHITESPACE) {
+                                $phpcsFile->fixer->beginChangeset();
+                                if ($tokens[($prev - 1)]['code'] === T_WHITESPACE) {
+                                    $phpcsFile->fixer->replaceToken(($prev - 1), ' ');
+                                }
+
+                                $phpcsFile->fixer->replaceToken($prev, ' ');
+                                $phpcsFile->fixer->endChangeset();
+                            } else {
+                                $phpcsFile->fixer->addContent($prev, ' ');
```
jrfnl: Not sure this is needed. `extendsClassName` would be a parse error, though maybe for `extends/*comment*/ClassName` ? in which case, it would be good to add that as a unit test.

webimpress: `extendsClassName` is invalid, but `extends\ClassName` is valid. I can add test case.

src/Standards/PSR2/Sniffs/Classes/ClassDeclarationSniff.php
```diff
@@ -344,19 +344,32 @@ public function processOpen(File $phpcsFile, $stackPtr)
...
-                            $spaceBefore,
+                            $found,
                         ];
 
                         $fix = $phpcsFile->addFixableError($error, $className, 'SpaceBeforeName', $data);
                         if ($fix === true) {
-                            $phpcsFile->fixer->replaceToken($prev, ' ');
+                            if ($tokens[$prev]['code'] === T_WHITESPACE) {
+                                $phpcsFile->fixer->beginChangeset();
+                                if ($tokens[($prev - 1)]['code'] === T_WHITESPACE) {
```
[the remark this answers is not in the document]
webimpress: Good point, I am going to change it to the loop.

src/Standards/PSR2/Sniffs/Classes/ClassDeclarationSniff.php
webimpress: I've added more test cases, it seems to be working fine in case with comment and no space after `extends`.

gsherwood:
This PR doesn't actually enforce the PSR-2 rule that:
```
Lists of extends/implements MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one interface per line.
```

Example code that doesn't generate any errors:
```php
class X extends //comment
 Foo
{
}
```

Removing the space before `Foo` causes an error, but it's just about the missing space and not that the value must be indented. If you change `extends` to `implements` then the errors all look good.

So this is fixing something that doesn't actually conform to the PSR-2 standard because the sniff was already missing code to conform the standard.

I won't merge it in, but the PSR2 sniff still needs fixing.

jrfnl:
> I won't merge it in, but the PSR2 sniff still needs fixing.

@gsherwood Changed your mind ?

gsherwood:
> Changed your mind ?

Thanks, made a mistake and forgot to revert.

webimpress:
@gsherwood @jrfnl I think PSR-2 enforce it, because we have said:

> The extends and implements keywords MUST be declared on the same line as the class name.

In case when we have multiple extends class names (interface) then it is possible to define them in each line, but then we do not have `implements` keyword.

So I think it should be like that - extends, class name and implements MUST be in the same line for classes/traits, and for interfaces it could be each in separate, but I think it works as expected.

webimpress:
Hm... actually it doesn't work for interfaces correctly :( Example:

```
interface I1 extends
 I2,
 I3,
 I4,
 I5
{
}
```

Above code is not showing any errors, but it's wrong.

So what would be the plan? Allow class name (after `extends`/`implements`) in new line but with correct indent and if there is `extends` and `implements` then enforce both of them to be in the same line as class?

webimpress:
@gsherwood 

>> Lists of extends/implements MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one interface per line.


Can you provide link to PSR-2 rules you are quoting?

I have found this:

> Lists of implements MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one interface per line.

https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#41-extends-and-implements

I am confused ...

gsherwood:
> Can you provide link to PSR-2 rules you are quoting?

It's in the errata: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide-meta.md

They put clarifications and changes there instead of releasing a new version of PSR-2, which makes life a little hard when you only read the main doc.

webimpress:
@gsherwood 

> It's in the errata: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide-meta.md
> 
> They put clarifications and changes there instead of releasing a new version of PSR-2, which makes life a little hard when you only read the main doc.

And to quote, it is said there:

> When extending multiple interfaces, the list of extends should be treated the same as a list of implements, as declared in Section 4.1.

So this is the case when we have __interface__ which extends __multiple__ interfaces.
Still I think, PSR-2 enforces class name after `extends` to be in the same line as class declaration and only we can have them in separate lines for `implements` in classes/traits and `extends` in interfaces.

What do you think?

I can prepare the fix, but I need to know how it should really works :)

gsherwood:
Yes, sorry, my code sample used a class when it should have used an interface, and I confused myself.

The sniff does show errors for this code:
```php
interface I1 extends
 I2,
 I3,
 I4,
 I5
{
}
```
I get:
```
$ phpcs temp.php --standard=PSR2

FILE: temp.php
-------------------------------------------------------------------------------------------------------
FOUND 5 ERRORS AFFECTING 5 LINES
-------------------------------------------------------------------------------------------------------
 2 | ERROR | [ ] Each interface must be in a namespace of at least one level (a top-level vendor name)
 3 | ERROR | [x] Expected 4 spaces before interface name; 1 found
 4 | ERROR | [x] Expected 4 spaces before interface name; 1 found
 5 | ERROR | [x] Expected 4 spaces before interface name; 1 found
 6 | ERROR | [x] Expected 4 spaces before interface name; 1 found
-------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 4 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-------------------------------------------------------------------------------------------------------
```
So I think the fixes for this sniff need to focus on ensuring that the `extends` keyword, and the extended class name, appear on the same line as the object type keyword, unless an interface is extending multiple interfaces.

The sniff has the same problem with `implements`, as this should show an error because only one interface is being implemented:
```php
<?php
class X implements
    A
{
}
```
Anyway, the sniff needs work :(

Issue #2304, opened 2018-12-11
RuntimeException: Undefined variable: childOutput in squizlabs/php_codesniffer/src/Runner.php on line 705


yoeunes:
- phpcs version : 3.3.2
- php version: 7.2 and 7.3
- operating system: osx

### Description
I get this exception when i run phpcs using a parallel value greater than 1 

```bash
Fatal error: Uncaught PHP_CodeSniffer\Exceptions\RuntimeException: Undefined variable: childOutput in /Users/yoeunes/.composer/vendor/squizlabs/php_codesniffer/src/Runner.php on line 705 in /Users/yoeunes/.composer/vendor/squizlabs/php_codesniffer/src/Runner.php:562
Stack trace:
#0 /Users/yoeunes/.composer/vendor/squizlabs/php_codesniffer/src/Runner.php(705): PHP_CodeSniffer\Runner->handleErrors(8, 'Undefined varia...', '/Users/yoeunes/...', 705, Array)
#1 /Users/yoeunes/.composer/vendor/squizlabs/php_codesniffer/src/Runner.php(502): PHP_CodeSniffer\Runner->processChildProcs(Array)
#2 /Users/yoeunes/.composer/vendor/squizlabs/php_codesniffer/src/Runner.php(114): PHP_CodeSniffer\Runner->run()
#3 /Users/yoeunes/.composer/vendor/squizlabs/php_codesniffer/bin/phpcs(18): PHP_CodeSniffer\Runner->runPHPCS()
#4 {main}
  thrown in /Users/yoeunes/.composer/vendor/squizlabs/php_codesniffer/src/Runner.php on line 562
```

this is my phpcs.xml file:

```xml
<?xml version="1.0"?>
<ruleset name="coding standard">
    <description>coding standard</description>

    <arg name="extensions" value="php"/>
    
    <arg name="parallel" value="2"/>

    <file>src</file>

    <rule ref="PSR2" />

</ruleset>
```

gsherwood:
I can't replicate this (php 7.3, mac os), but it would mean that something pretty bad has gone wrong. If you are using a parallel value of 2, it means at least one of the two forks has exited but not written any data to the output file. I assume this means it died early, but I'm not sure why. It could be that the temp output file couldn't be written for some reason.

Without adding debug information into the runner, I'm not really sure how to figure out where it is dying. But there are some things to try:
- Run phpcs with the `-v` command line argument to see debug output and ensure it is at least checking files
[...]

PR #2305, opened 2018-12-12
Fix Undefined variable: childOutput in src/Runner.php where using more than one parallel process

yoeunes:
This fix the issue in : https://github.com/squizlabs/PHP_CodeSniffer/issues/2304


gsherwood:
I'm closing this as it has been confirmed to mask a real PHP fatal error that is stopping PHPCS errors/warnings from being displayed.

PR #2307, opened 2018-12-12
Fix for PSR2 ClassDeclarationSniff - position of class name - follow up #2300

webimpress:
Follow up #2300
As the previous hotfix was merged and unmerged this hotfix provides better check for class names after implements/extends keywords in class/interface/trait definition.

According to PSR-2 multiline definition is allowed only if we implements more than one interface (or interface extends more than one interface). Everything else should be always in the same line as class name.


webimpress:
I bet it worked yday when I was submitting the PR... 🤣 I will do it later on :)

webimpress:
@gsherwood Now it's 💚 . I have no idea how I broke the condition before push yday.

Is that what you've expected? Would you like me to add any more tests?

webimpress:
@gsherwood Is it going to be included in 3.4.1 release?

webimpress:
Rebased to resolve conflicts.

gsherwood:
> so probably not

Well, maybe. I'll add it to 3.4.1 and see if I get to it.

src/Standards/PSR2/Sniffs/Classes/ClassDeclarationSniff.php
```diff
@@ -231,10 +248,11 @@ public function processOpen(File $phpcsFile, $stackPtr)
         $classCount         = count($classNames);
         $checkingImplements = false;
         $implementsToken    = null;
-        foreach ($classNames as $i => $className) {
+        foreach ($classNames as $n => $className) {
```
webimpress: The change here is not because of my preference, but `$i` is used in another loop inside and because of that in some case we had invalid check in line 398.

webimpress:
@gsherwood sorry, I've missed your comment. PR is now updated. Thanks!

gsherwood:
Thanks a lot for making that change.

PR #2322, opened 2018-12-20
Squiz/FunctionCommentThrowTag: two bug fixes

jrfnl:
### Squiz/FunctionCommentThrowTag: fix incorrect `WrongNumber` error

When it isn't clear what type of Exception a `throw` will generate, but there are more than one `throw`s in a function, the "unknown" exceptions wouldn't be taken into account when determining whether the correct number of exceptions were being thrown.

Includes unit test.

(In case anyone is wondering: the `Prefix_Invalid_Argument_Exception::invalid_string_parameter()` method returns an instance of `self` in the code where I encountered this issue.)

### Squiz/FunctionCommentThrowTag: fix incorrect `Missing` errors with FQN Exception names

The code did not take into account that Fully Qualified names could be used in the `@throws` tag while a relative Exception name may be used in the `throw` statement.

Includes unit tests (probably a few too many, but I wanted to be sure that I understood how the sniff was dealing with this).

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
@@ -69,6 +69,7 @@ class FunctionCommentThrowTagSniff implements Sniff
         $thrownExceptions = [];
         $currPos          = $stackPtr;
         $foundThrows      = false;
+        $unknownCount     = 0;
         do {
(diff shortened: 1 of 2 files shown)


jrfnl:
FYI: one of the Travis builds needs to be retriggered. Error on their side, not an issue with the PR.

gsherwood:
Thanks for these fixes

jrfnl:
You're welcome ;-)

PR #2326, opened 2018-12-23
Changelog 3.4.0: fix incorrect entry

jrfnl:
See #2148

This should probably also be adjusted in https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.4.0


gsherwood:
Thanks for fixing that up. I've changed the github release notes, although can't ever update the PEAR one.

PR #2331, opened 2018-12-25
PSR12/ShortFormTypeKeywords: bug fix / false positive

jrfnl:
The PSR 12 standard (currently) says nothing about whether or not spaces within type casts are allowed.

As it was, the `PSR12.Keywords.ShortFormTypeKeywords` sniff would throw a false positive for a short form type cast like `( bool )` and "fix" it by removing the spaces, which is not the concern of this sniff.

If at some point the PSR12 standard would take a stand about spaces within type casts, the `Squiz.WhiteSpace.CastSpacing` sniff could be added to check & fix these issues.

In the mean time, the `ShortFormTypeKeywords` sniff should be space-agnostic.

This PR takes care of that.

It also makes the error message slightly more descriptive by being explicit about the type cast found in the error message.

Includes unit tests.

### DIFF
--- a/src/Standards/PSR12/Sniffs/Keywords/ShortFormTypeKeywordsSniff.php
+++ b/src/Standards/PSR12/Sniffs/Keywords/ShortFormTypeKeywordsSniff.php
@@ -42,24 +42,31 @@ class ShortFormTypeKeywordsSniff implements Sniff
      */
     public function process(File $phpcsFile, $stackPtr)
     {
-        $tokens = $phpcsFile->getTokens();
+        $tokens     = $phpcsFile->getTokens();
+        $typecast   = str_replace(' ', '', $tokens[$stackPtr]['content']);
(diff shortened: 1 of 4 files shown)


gmponos:
> If at some point the PSR12 standard would take a stand about spaces within type casts, the Squiz.WhiteSpace.CastSpacing sniff could be added to check & fix these issues.

Would you mind if I raise this to @phpfig with a PR or do you want to do it?

jrfnl:
> Would you mind if I raise this to @php-fig with a PR or do you want to do it?

Wouldn't mind at all, though if I remember the process correctly, you may want raise the subject in the mailinglist first.

gmponos:
FYI 

https://github.com/php-fig/fig-standards/pull/1141
https://groups.google.com/forum/#!topic/php-fig/Qdc7XLygeO8

gsherwood:
Thanks for fixing that up.

PR #2332, opened 2018-12-25
Efficiency fixes

jrfnl:
Array union `+` is much faster than `array_merge()`, though the behaviour is different.

The differences are:

-- | Array union | Array merge
-- | -- | --
Numeric keys | Will keep the keys which already exist | Will renumber & add keys for all values
String keys | Will keep the keys which already exist | Will keep the keys which already exist
Values | Will *not overwrite* existing value | Will *overwrite* existing value

As the arrays which are defined in the `Tokens` class have the same keys as well as values, using array union on those token arrays will be:
* more efficient than using a function call to `array_merge()`;
* will auto-deduplicate the arrays, so no need for a call to `array_unique()`;
* and the result will be the same.

Similarly, using straight assignments to an array is also more efficient than using `array_merge()` when it's a small number of assignments.

### DIFF
--- a/src/Sniffs/AbstractPatternSniff.php
+++ b/src/Sniffs/AbstractPatternSniff.php
@@ -260,10 +260,9 @@ abstract class AbstractPatternSniff implements Sniff
         $errors      = [];
         $found       = '';
 
-        $ignoreTokens = [T_WHITESPACE];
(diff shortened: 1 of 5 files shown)


gsherwood:
Nice one. Thanks.

PR #2340, opened 2019-01-02
Hotfix for PEAR install

jrfnl:
In the WordPressCS repo we received a bug report from someone using PHPCS 3.4.0 but missing the `Generic.WhiteSpace.LanguageConstructSpacing` sniff.
Ref: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1614

Turned out that that sniff is missing from the `package.xml` file which is why it's not available when someone is using a PEAR install of PHPCS.

I've now done a comprehensive check of all relevant files in the repo versus the files listed in the `<contents>` tag of the `package.xml` file and found some more discrepancies with this commit as a result.

As this basically means that the PEAR install is currently "broken" with regards to the `Code` report (PHPCS 3.0), the `Generic.Formatting.SpaceBeforeCast` sniff (PHPCS 3.4.0) as well as the `Generic.WhiteSpace.LanguageConstructSpacing` sniff (PHPCS 3.3.0), this should be considered a hotfix.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -37,7 +37,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
   <dir name="/">
    <file baseinstalldir="PHP/CodeSniffer" name="autoload.php" role="php">
     <tasks:replace from="@test_dir@" to="test_dir" type="pear-config" />
-  </file>
+   </file>
    <file baseinstalldir="PHP/CodeSniffer" name="CodeSniffer.conf.dist" role="data" />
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for finding this. I `pear package-validate` all the time but forget it doesn't tell me about missing entries.

Unfortunately, there is no way to go back and fix previous PEAR releases, so this will just have to go into the next release.

jrfnl:
It is what it is, but at least it's fixed now and with the additional validation, won't be able to happen again.

PR #2341, opened 2019-01-02
Travis: add additional validation for the PEAR package.xml file

jrfnl:
This PR adds a custom script which will validate the `<contents>` and `<phprelease>`  tags in the PEAR-targeted `package.xml` file.

The script executes the following checks to verify the validity of the tags:
* Check that every file mentioned in the `<contents>` tag exists.
* Check that the value of the `role` attribute for a `<file>` tag is valid.
* Check that the value of the `baseinstalldir` attribute for a `<file>` tag is valid.
* Checks that all files in the `src` and `tests` directories are listed in the `<contents>` tag.
* Check that every file mentioned in the `<phprelease>` tags exists.
* Check that the value of the `as` attribute for a `<install>` tag is valid.
* Checks that all files in the `tests` directory are listed in the `<phprelease>` tags.

More/improved checks can be added if/when needed, but this should prevent the most common issues.

This check only needs to be run on one build as the results won't change between PHP versions.
For that reason, I've elected to add it to one of the `CUSTOM_INI` runs to balance out the runs some more.

As the PHP version makes no difference to the `xmllint` check, I've moved that to a lower PHP version, so we can use a higher (faster) PHP version for this additional check.

**NOTE**: The build for this PR won't be able to pass until PR #2340 has been merged, however, once this PR is merged, issues such as identified and fixed by #2340 won't be possible anymore as the Travis build would fail in those cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,15 +8,15 @@ matrix:
     - php: 5.4
     - php: 5.5
     - php: 5.5
-      env: CUSTOM_INI=1
-    - php: 5.6
-    - php: 7.0
-    - php: 7.0
       env: CUSTOM_INI=1 XMLLINT=1
       addons:
         apt:
           packages:
             - libxml2-utils
+    - php: 5.6
+    - php: 7.0
+    - php: 7.0
+      env: CUSTOM_INI=1 PEAR_VALIDATE=1
     - php: 7.1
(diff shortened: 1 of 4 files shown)


gsherwood:
You're amazing. Thank you for this.

PR #2344, opened 2019-01-04
Squiz/BlockComment: don't require a blank line between PHP open tag and block comment

jrfnl:
Just like with scope openers, if a block comment is the first thing after a PHP open tag, no blank line should be required.

Includes unit tests.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/BlockCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/BlockCommentSniff.php
@@ -360,8 +360,9 @@ class BlockCommentSniff implements Sniff
 
         // Check that the lines before and after this comment are blank.
         $contentBefore = $phpcsFile->findPrevious(T_WHITESPACE, ($stackPtr - 1), null, true);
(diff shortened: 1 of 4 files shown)


gsherwood:
This is a change in behaviour of the sniff, so I'm going to remove this from the 3.4.1 release unless there is a reason behind the change (fixer conflict maybe?).

If it's just a matter of preference, then I need to think about how this impacts the code we write within Squiz and if we'd want this behaviour change as part of our standard. I suspect we never put block comments directly after PHP tags, so it may not have any impact.

jrfnl:
@gsherwood I'm fine with removing this one from the 3.4.1 milestone, just hoping that the "thinking about it" will happen at some point in the not too far future.

I don't recall there being a fixer conflict with this, though I haven't run any specific tests to see if there is either.

It is more that this felt inconsistent with how other sniffs behave after a PHP open tag.

gsherwood:
This doesn't look like it will have an impact, so merged in now. Thanks.

jrfnl:
Thank you!

Issue #1926, opened 2018-03-07
Improve PHPCS documentation


Basje:
In response to [this message](https://www.reddit.com/r/PHP/comments/824p11/serious_alternatives_to_squizlabs_phpcs/dv8ijn9/) on Reddit.

I haven't got loads of time to help, but if more people join in then we could make some steps to proper documentation!

@gsherwood: you mentioned that you write documentation professionally. Could you set up some guidelines for us laymen on how to write quality documentation? Links are fine too.

I will post this on Reddit as well, see if we can get some people to join this effort.

jrfnl:
Just a thought: if all sniffs had proper XML docs associated with them ([example](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Generic/Docs/Arrays/DisallowLongArraySyntaxStandard.xml)), would it not be possible to auto-generate a documentation site based on those ?

This will require the docs to be improved or even added for a lot of sniffs.

It may also require the docs XML standard to be extended and formalized.

One of the things I've been missing, for instance, would be to have a block per error code and to have those error codes as nodes or attributes in the XML.
Another thing would be to have the ability to add links for each sniff/error code, which could link to the official standard, like PSR2 could link to the individual paragraphs in the [document on the FIG website](https://www.php-fig.org/psr/psr-2/).

gsherwood:
The XML format is certainly one place to start.

My intention is to eventually bring up a PHPCS site with hand-written documentation for each error message, with code samples. I could copy/paste from the XML and then re-write anything a bit old or informal.

> you mentioned that you write documentation professionally

I do write a lot documentation as part of my job, but I'm not a professional writer. Part of what I do is write from scratch. Another part is taking existing rough notes and turning it into a more complete document. That's really all I think is needed in this specific case. It's a bit of grunt work to start with, followed by a lot of polish.

I thought that the first thing I'd need to do is actually collect a list of all the error messages that PHPCS produces by going through the code of all the sniffs. For each, I wanted to gather:
- the error code
- the error message
- the severity (error or warning)
- the error message data (e.g., a class name, a method name, whatever helps customise messages)

And if I had time, I'd also produce a snippet of good and bad sample code for each one showing the condition that could cause the error message to be generated.

My next step would be to take a look at that data and try to figure out an easy-to-read format for displaying it all, before re-formatting and adding custom descriptions for each message.

[...]

Issue #1603, opened 2017-08-11
Feature request: Establish a pattern of providing a link to explain a lint failure


andrewhowdencom:
Invariably, lints are used to communicate some piece of knowledge from one developer to another. However, a lint is not a good medium to communicate the nuance of a piece of knowledge (for example, "Try not to use unserialize"). Instead, projects like shellcheck provide a key that can be looked up in a wiki that explains the lint, as well as why it exists and another way to solve the issue the lint violates.

Example: https://github.com/koalaman/shellcheck/wiki/SC1003

It would be good to establish such a pattern here. There are already keys that are provided (something like PSR2.Whitespace.Foo.Bar), which can be used as wiki pages. So, this request would consist of simply establishing a wiki that contained the necessary justification.

The GitHub wiki is a good place to host such knowledge, as in addition to the standards here, there are other project specific standards -- most of which seem to be in some way related to GitHub.

For bonus, a link could be built as first class support -- embedded in the error message shown about the link (perhaps with a shortner; l.squiz.io/PSR2.Whitespace.Foo.Bar)

Such a link could be provided optionally in the standard declaration:

```
<?xml version="1.0"?>
<ruleset name="PSR2">
 <description>The PSR-2 coding standard.</description>
 <arg name="tab-width" value="4"/>
 <!-- the new bit -->
 <help_url>https://l.squiz.labs/</help_url>
```

(Edit: 2019-01-03): I use mkdocs for this sort of documentation with other projects. Suuper simple, can be setup and pushed to netlify trivially, comes with a pretty material skin.

igorsantos07:
I'm getting here after some doc-frustration at #1752. This is a great idea indeed, but I would only vote against using the GitHub Wiki for that. It's a private wiki only open for project members, and it would make it harder for contributions and updates to land - which is one of the key-points against documentation at the mentioned issue.

This would be better placed into a GitHub Pages branch maybe, as it's just as navigable as the wiki, but it's also PR-able and possible to be automated.

gsherwood:
Closing to ensure all this is tracked on  #1926

PR #2345, opened 2019-01-04
Move documentation in repo for consumption by mkodcs, establish sniff documentation pattern.

andrewhowdencom:
See commit for design notes et. al. 

This effectively deprecates the wiki. 

Demo: https://phpcs.netlify.com/

 (just a build that runs `pip install mkdocs && mkdocs build`, and serves the generated `site` directory)

### DIFF
--- a/docs/Documentation/Advanced-Usage.md
+++ b/docs/Documentation/Advanced-Usage.md
@@ -1,0 +1,398 @@
+## Table of contents
+  * [Specifying Valid File Extensions](#specifying-valid-file-extensions)
+  * [Ignoring Files and Folders](#ignoring-files-and-folders)
+  * [Ignoring Parts of a File](#ignoring-parts-of-a-file)
+  * [Limiting Results to Specific Sniffs](#limiting-results-to-specific-sniffs)
+  * [Filtering Errors and Warnings Based on Severity](#filtering-errors-and-warnings-based-on-severity)
+  * [Replacing Tabs with Spaces](#replacing-tabs-with-spaces)
+  * [Specifying an Encoding](#specifying-an-encoding)
+  * [Using a Bootstrap File](#using-a-bootstrap-file)
+  * [Using a Default Configuration File](#using-a-default-configuration-file)
+  * [Specifying php.ini Settings](#specifying-phpini-settings)
+  * [Setting Configuration Options](#setting-configuration-options)
+  * [Deleting Configuration Options](#deleting-configuration-options)
+  * [Viewing Configuration Options](#viewing-configuration-options)
+  * [Printing Verbose Tokeniser Output](#printing-verbose-tokeniser-output)
+  * [Printing Verbose Token Processing Output](#printing-verbose-token-processing-output)
+  * [Quieting Output](#quieting-output)
+
+***
+
+## Specifying Valid File Extensions
+By default, PHP_CodeSniffer will check any file it finds with a `.inc`, `.php`, `.js` or `.css` extension, although not all standards will actually check all these file types. Sometimes, this means that PHP_CodeSniffer is not checking enough of your files. Sometimes, the opposite is true. PHP_CodeSniffer allows you to specify a list of valid file extensions using the `--extensions` command line argument. Extensions are separated by commas.
+
+To only check .php files:
+
+    $ phpcs --extensions=php /path/to/code
+
+To check .php, .inc and .lib files:
+
+    $ phpcs --extensions=php,inc,lib /path/to/code
+
+## Ignoring Files and Folders
+Sometimes you want PHP_CodeSniffer to run over a very large number of files, but you want some files and folders to be skipped. The `--ignore` command line argument can be used to tell PHP_CodeSniffer to skip files and folders that match one or more patterns.
+
+In the following example, PHP_CodeSniffer will skip all files inside the package's tests and data directories. This is useful if you are checking a PEAR package but don't want your test or data files to conform to your coding standard.
+
+    $ phpcs --ignore=*/tests/*,*/data/* /path/to/code
+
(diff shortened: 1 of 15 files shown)


jrfnl:
@andrewhowdencom Just some generic questions:

* Where do you envision the docs be hosted as a site ? Keep in mind that the `gh-pages` branch is already in use for the metrics.
* If they are not intended to be put online in the form of a website, why would the `mkdocs` setup be needed ?
* Does `mkdocs` work cross-platform ?
* Should the docs building be tested via Travis to prevent a failing docs site ?

andrewhowdencom:
Holaa, 

> Where do you envision the docs be hosted as a site ? Keep in mind that the gh-pages branch is already in use for the metrics.

Netlify. That's where the demo is running, and it's free and extremely quick. You can see the configuration here:

https://app.netlify.com/sites/trusting-aryabhata-8abf17/deploys

Happy to transfer ownership of that to a maintainer, but it took about 10 minutes to setup.

> If they are not intended to be put online in the form of a website, why would the mkdocs setup be needed ?

This is the intention. I would probably stick it at `docs.phpcs.io` or smth like this. 

> Does mkdocs work cross-platform ?

I think so, Documentation gives some specific steps for Windows, and it's a python program so it's probably fine on Mac.

It's fairly old / widely used; I imagine it should be super stable.

https://www.mkdocs.org/

> Should the docs building be tested via Travis to prevent a failing docs site ?

Negative; netlify will fail the build directly if the build fails. Travis would be double work.

---

I additionally added an example sniff documentation: 

https://trusting-aryabhata-8abf17.netlify.com/sniffs/psr2/classes/classdeclaration/openbracenewline/

The true purpose of this PR was to solve:

https://github.com/squizlabs/PHP_CodeSniffer/issues/1603

Which I think this would allow. Moving the other docs into a website is more just to acknowledge there is already documentation, but this PR would allow community maintenance of it, as well as set a pettern for other providers in the future (see that issue for details, as well as the last commit notes)

luckydonald:
In one way or another, this is looking like a good way.
Markdown in a docs folder can also be viewed in github directly, so that's a nice fallback.

jcrben:
I've noticed that vscode has a repo for their github wiki (https://github.com/Microsoft/vscode-wiki) - each page has a link to fork and make a PR (https://github.com/Microsoft/vscode/wiki). Not sure how they do that, however.

jrfnl:
@jcrben All repos which are publicly editable have that nowadays.

gsherwood:
> All repos which are publicly editable have that nowadays.

If you have any more info about this, please let me know because I'm not aware of any such feature. But the general idea of maintaining a wiki project that I can then accept PRs on and push to the main wiki repo actually sounds pretty good.

gsherwood:
Just a note here that I really do appreciate the effort, but I don't intend to move away from the Github wiki. I will investigate setting up a seperate wiki repo that can be synced with the main wiki repo, when I get some time.

jrfnl:
> If you have any more info about this, please let me know because I'm not aware of any such feature.

As far as I'm aware it won't work with PRs, but you can get a local copy of the wiki via that link and I presume - if the wiki is open or if you have edit rights - push any edited markdown files directly to it. (I haven't tried that yet, but noticed the clone option a few weeks ago on WordPressCS)

Also see:
* https://help.github.com/articles/adding-or-editing-wiki-pages/#adding-or-editing-wiki-pages-locally
* https://grantwinney.com/5-things-you-can-do-with-a-locally-cloned-github-wiki/

Regarding the documentation itself: I've been meaning to add my voice to issue #1926, but hadn't gotten round to it. Short summary of the idea I'd been playing with:
* Create a XSD for the `SniffNameStandard.xml` type files as already used, but with quite a few more options. Think: a node per error code with a description and code sample for each error code, property nodes explaining the public properties, a top level `tags` node to add keyword for searchability etc. Ability to add links to the docs for error codes and/or complete sniffs etc
* Invite everyone to contribute to improving the existing XML docs based on the XSD, as well as to add XML docs for all sniffs which don't currently have them. (I have a list available of what's missing and that could be turned into a checklist in an issue)
* Improve the existing `Generator` files to work with the new XSD and possibly add some additional Generator options.
* The Markdown generated docs could then, for instance, go into the wiki or the HTML generated docs put up on a website.

gsherwood:
> Short summary of the idea I'd been playing with

This is how I did coding standard documentation before PHPCS became a PEAR package. I'd use these files to generate a PDF. Example from the long long ago: [coding_standards.pdf](https://github.com/squizlabs/PHP_CodeSniffer/files/2747167/coding_standards.pdf)

Resurrecting these documentation files is a great idea.

gsherwood:
> As far as I'm aware it won't work with PRs, but you can get a local copy of the wiki via that link and I presume - if the wiki is open or if you have edit rights - push any edited markdown files directly to it.

My intention was to setup a squizlabs/PHP_CodeSniffer-wiki repo and treat that as if it was the wiki. Then I'd clone it locally and do edits on there, accept PRs etc. When pushing, I'd set up my local copy to push to both squizlabs/PHP_CodeSniffer-wiki and squizlabs/PHP_CodeSniffer.wiki. Hopefully this would make it appear as if the wiki repo was available for PRs and collaboration, although it is really just a copy.

jrfnl:
@gsherwood If we take the XSD/XML idea as a base and add a GH Wiki Generator or improve the existing markdown Generator to allow for multi-page generation and more, you could just run the generator, and update the squizlabs/PHP_CodeSniffer.wiki with that (at least for the sniff documentation).

gsherwood:
@jrfnl Totally with you. I was just thinking of the other docs on there as well. Maybe that's a stage 2 thing.

jrfnl:
@gsherwood :+1: I'd be happy to draw up a proposal for the XSD after I've finished the utility method refactor (#2189)

svivian:
If you’re going to set up a new repository anyway to update the wiki, why not make a ‘GitHub Pages’ site instead? You can make a simple site with Jekyll using Markdown files, and host it on e.g. `phpcs.github.io` (if you own the phpcs username), otherwise `squizlabs.github.io` or your own (sub)domain.

I’ve done this with another project and it worked out pretty well.

PR #2347, opened 2019-01-04
Test bootstrap/printPHPCodeSnifferTestOutput(): bug fix

jrfnl:
PR #2289 added a count of the number of test case files to the test output, however did not take into account that when only the unit tests for the utility functions would be run, i.e. via for instance `phpunit --filter FindExtendedClassNameTest`, the `$GLOBALS['PHP_CODESNIFFER_SNIFF_CASE_FILES']` variable would not have any arrays in it.

This adds some minor defensive coding which gets rid of the PHP error message which would be generated in that situation.


gsherwood:
Thanks

PR #2350, opened 2019-01-06
Girlscouting: various minor code fixes

jrfnl:
Minor code fixes for the following:
* Use short form type casts
    These were the only places were long form type casts were being used, so for consistency these have been replaced with short form.
* Remove unused `use` statements
* Add missing `use` statement
* Minor code simplifications for variable concatenations
* Declare variables
    The `$newToken` array variable was not (yet) initialized when `content` was being added.
* Use strict comparisons when using `in_array()`
    Prevent quirks caused by type juggling when using the default of `$strict = false`.
    Ref: http://php.net/manual/en/function.in-array.php

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -895,9 +895,9 @@ class File
         // due to the use of the --sniffs command line argument.
         if ($includeAll === false
             && ((empty($this->configCache['sniffs']) === false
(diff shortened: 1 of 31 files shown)


gsherwood:
Thanks

PR #2351, opened 2019-01-06
Girlscouting: various minor documentation fixes

jrfnl:
* Use FQN for exceptions in `@throws` tags and add missing `@throws` tags
* Use FQN in `@param` tags and add missing type declaration
* One minor documentation correction


gsherwood:
Thanks

PR #2353, opened 2019-01-07
Use Exceptions consistently

jrfnl:
Following up on https://github.com/squizlabs/PHP_CodeSniffer/pull/2351/commits/bd961bbe7277aca4689b8059acfe6cbac647cbfa, a further review of the Exceptions being thrown showed some inconsistencies.

As far as I can see, Exceptions thrown by sniffs/utility methods when, for instance, a wrong token is passed, should be `RuntimeException`s, while Exceptions thrown by the `Tokenizer` classes should be `TokenizerException`s.

In a number of places this was applied inconsistently. This has now been fixed.

Aside from that, for quite a number of methods, the `Exception` being thrown has now been documented.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -584,6 +584,7 @@ class Config
      * @param int    $pos The position of the argument on the command line.
      *
      * @return void
+     * @throws \PHP_CodeSniffer\Exceptions\DeepExitException
      */
     public function processShortArgument($arg, $pos)
     {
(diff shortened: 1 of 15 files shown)


jrfnl:
As discussed in #2189, this commit is included within #2456.

gsherwood:
Thanks a lot for fixing these up.

PR #2358, opened 2019-01-11
Generic/DeprecatedFunctions: add missing unit tests

jrfnl:

Note: PHP internally doesn't always correctly register functions as deprecated, so this sniff will only work when it has (as the information is otherwise not available to the Reflection extension).

Some trial and error showed that the two functions now in the tests are correctly registered, but numerous other functions I tried to add to the test case file failed as PHP didn't provide the expected information.


gsherwood:
> Note: PHP internally doesn't always correctly register functions as deprecated

Well that's not a great, and also good to know. Thanks for adding the tests.

jrfnl:
@gsherwood I have the feeling it's better in more recent PHP versions and that functions which are deprecated because the whole extension is deprecated may not have the annotation with the individual function, but I haven't looked at the C code to verify that.

For this PR, I did a spot check with a number of the functions listed in the [PHPCompatibility RemovedFunctions](https://github.com/PHPCompatibility/PHPCompatibility/blob/master/PHPCompatibility/Sniffs/FunctionUse/RemovedFunctionsSniff.php) sniff and didn't get the expected results for the first five tests I ran, even though I ran the tests against PHP versions in which the tested functions were supposed to be deprecated and not yet removed.

PR #2361, opened 2019-01-11
Generic/ESLint: add missing unit tests

jrfnl:

Note: it may be worth it to check if `eslint` is available on Travis by default or do a `npm i eslint` in the Travis `before_script` section to install it, so the sniff will actually be tested properly.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -437,6 +437,10 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="InlineControlStructureUnitTest.js.fixed" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="InlineControlStructureUnitTest.php" role="test" />
        </dir>
+       <dir name="Debug">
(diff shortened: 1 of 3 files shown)


jrfnl:
@gsherwood In that case, my two pennies as input for your internal debat:
While the CSS/JS linters available out there are great at checking against parse errors and sorting out all the whitespace related things, when it comes down to checking the logic of things or cross-checking things, I've always gotten the impression that it is nowhere near as accessible/easy or even possible to write custom checks for those tools, so I, for one, definitely see the value of having CSS/JS tokenizing in PHPCS.

As an example: for the WP Theme Review board I wrote a sniff a while back which checks that the WordPress admin bar is not disabled/hidden/made invisible through CSS/JS by themes.

In their [PHPCS standard repo](https://github.com/WPTRT/WPThemeReview/issues/) there are a lot of open issues which are about checking CSS/JS related things which would be a lot more difficult to implement without PHPCS's ability to handle CSS/JS files.

(_Context: All themes published on wordpress.org are (manually) reviewed for compliance with the conditions for publication on wordpress.org. The team wants to try and automate a lot of the more code related review checks to allow reviewers to focus on the theme itself_)

PR #2363, opened 2019-01-11
Generic/ClosureLinter: add missing unit tests

jrfnl:

It may be worth it to add `sudo apt-get install python-pip && sudo pip install https://github.com/google/closure-linter/zipball/master` in the Travis `before_script` section to install it, so the sniff will actually be tested properly, though trying to get this running locally was quite a struggle as the tool is deprecated and has issues when run on Python 3.


jrfnl:
Rebased for merge conflict.

PR #2359, opened 2019-01-11
Generic/SubversionProperties: add perfunctory unit test files

jrfnl:
[Sniff completeness series PR]

This sniff cannot be tested as it needs to read out information from files in a `.svn` directory which cannot easily be simulated.

The files now added just document why there are no unit tests.

PR #2364, opened 2019-01-11
Travis: add check to make sure that sniffs are "feature complete"

jrfnl:
This PR adds a custom script which will check whether each and every sniff is accompanied by a documentation XML file (warning) as well as unit test files (error).

Notes:
* To only show errors, the script can be run in `quiet` mode by passing the `-q` flag on the command line.
* The script has been set up in such a way that it is not only suitable for use by PHPCS itself, but also for use by external standards which use the PHPCS native unit test suite.
    For external standards to use the tool, run it from the root of the external standards repo like so:
    `php -f "path/to/PHP_CodeSniffer/scripts/check-sniff-completeness.php"`
* The list of sniffs to check completeness for, is determined by getting a list of all files in `Sniffs` directories, where the files have a name ending with `Sniff.php` and not prefixed with `Abstract`.
* The "has tests" check verifies that a `UnitTest.php` file is present in the standard's `Tests` directory and _at least_ one test case file.
* The "has docs" check verifies that a `Standard.xml` file is present in the standard's `Docs` directory.

This check only needs to be run on one build as the results won't change between PHP versions.
For that reason, I've elected to add it to the fastest PHP version run to balance out the runs some more.

As quite a large number of sniffs at this moment do not have documentation files, `quiet` mode is currently enabled for the check in Travis.
If/when the missing documentation files have been added, the `-q` flag should be removed and having unit tests as well as documentation should be enforced for all sniffs.

**Note: the build for this won't pass until all sniffs are accompanied by unit tests.**

Currently, the [following sniffs do not have unit tests](https://travis-ci.org/jrfnl/PHP_CodeSniffer/jobs/478410647#L1194-L1202):
- [x] `Generic.Debug.ClosureLinter` - see PR #2363
- [x] `Generic.Debug.CSSLint` - see PR #2360
- [x] `Generic.Debug.ESLint` - see PR #2361
- [x] `Generic.Debug.JSHint` - see PR #2362
- [x] `Generic.PHP.DeprecatedFunctions` - see PR #2358
- [x] `Generic.VersionControl.SubversionProperties` - see PR #2359
- [ ] `MySource.Channels.IncludeOwnSystem` - to be addressed, possibly in the same way as #2359.
    As it isn't completely clear to me without access to the proprietary system for which this sniff was written, what and how it should be tested, this is for someone else to sort out.

Related: https://github.com/squizlabs/PHP_CodeSniffer/pull/2345#issuecomment-453264115

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
     - php: 7.2
       env: PHPUNIT_INCOMPAT=1
     - php: 7.3
-      env: PHPUNIT_INCOMPAT=1
+      env: PHPUNIT_INCOMPAT=1 SNIFF_COMPLETE=1
     - php: "7.4snapshot"
       env: PHPSTAN=1 PHPUNIT_INCOMPAT=1
       addons:
@@ -62,6 +62,7 @@ script:
   - if [[ $CUSTOM_INI != "1" ]]; then composer validate --no-check-all --strict; fi
   - if [[ $CUSTOM_INI != "1" ]]; then php scripts/build-phar.php; fi
   - if [[ $CUSTOM_INI != "1" ]]; then php phpcs.phar; fi
+  - if [[ $SNIFF_COMPLETE == "1" ]]; then php scripts/check-sniff-completeness.php -q; fi
   # Validate the xml ruleset files.
   # @link http://xmlsoft.org/xmllint.html
   - if [[ $XMLLINT == "1" ]]; then xmllint --noout --schema phpcs.xsd ./src/Standards/*/ruleset.xml; fi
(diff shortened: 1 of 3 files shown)


jrfnl:
I've rebased this PR now all the PRs for missing unit test files have been merged.

The only sniff still missing unit tests is `MySource.Channels.IncludeOwnSystem`. @gsherwood How would you like to address this ?

jrfnl:
Rebased to fix the merge conflict state.

jrfnl:
@gsherwood Any chance this can be added to the 3.5.0 milestone ? This tool would be very useful for external standards as well.

jrfnl:
Closing in favour of an much improved version of this script now available via https://github.com/PHPCSStandards/PHPCSDevTools

If PHPCS still has any interest in using a script like this, I suggest adding the above mentioned repo as a `dev` dependency and using the script from `PHPCSDevTools`.

For maintainers of external PHPCS standards: see the [readme](https://github.com/PHPCSStandards/PHPCSDevTools/blob/develop/README.md) to learn how to add this check to your CI-toolchain.

PR #2365, opened 2019-01-11
Declare strict types

floverdevel:
Hello !
Since #1771 seems not to be be going forward, i forked it, rebased on master and added a commit to remove the public members of the class that were mentioned by @gsherwood here : https://github.com/squizlabs/PHP_CodeSniffer/pull/1771#issuecomment-452150153


webimpress:
Hi, I do not understand why we can't have them public? It is configurable, as it was requested before... see: https://github.com/squizlabs/PHP_CodeSniffer/pull/1771#issuecomment-375499597

floverdevel:
> Hi, I do not understand why we can't have them public? It is configurable, as it was requested before... see: [#1771 (comment)](https://github.com/squizlabs/PHP_CodeSniffer/pull/1771#issuecomment-375499597)

see this comment : https://github.com/squizlabs/PHP_CodeSniffer/pull/1771#issuecomment-452150153

webimpress:
@floverdevel For me it is unclear. I don't know if @gsherwood changed his mind and he doesn't want to have this sniff configurable anymore?

webimpress:
@gsherwood ok, so basically you want make `$format` protected and two other properties keep as public, or you don't like them as well to be configurable? I think these two are consistent with other sniffs.

gsherwood:
> ok, so basically you want make `$format` protected and two other properties keep as public

That's not what I said at all. Please read my comment here: https://github.com/squizlabs/PHP_CodeSniffer/pull/1771#issuecomment-446446890

webimpress:
@gsherwood sorry, I am so lost... I just understand you don't like `$format`, and I can understand it. If you don't want any configuration right now, it's fine, we can always add them later if requested, that's right. But I can't see reason to make them protected as these are already implemented and works, I mean:

- `spacesBefore` - number of blank lines before the declaration (default 1)
- `spacesAfter` - number of lines after the declaration and before the script (default 1)

and I think these are quite useful.

webimpress:
Even more maybe we should be able to allow strict type declaration in the same line as php open tag, I've seen some people/project are doing it, no it's not possible.

vladyslavstartsev:
Is this a thing that really needs to be done? 
Why not just use https://github.com/slevomat/coding-standard#slevomatcodingstandardtypehintsdeclarestricttypes-

gsherwood:
> sorry, I am so lost...

My comments related *only* to the `format` property. The format property was being used to decide how many spaces are around the equals sign and parenthesis (for example). I said that it would be better to make these additional properties rather than matching a fixed format property.

I never said that the `spacingBefore` of `spacingAfter` properties should be removed. I have no idea why you think that. I specifically asked for them to be renamed and you did that.

webimpress:
@gsherwood Ok, all clear now. So I've changed `$format` visibility to `private`. Please see original PR #1771. Thanks and sorry for confusion.

gsherwood:
After reviewing this sniff again, the use of the property to check the format of the declare statement made it impossible to properly report and fix errors. Instead, the sniff needed to be rewritten to enforce each of the individual rules of the PSR-12 standard so each message could have a different error code and fix. This also lets it account for PHPCS annotations much more easily.

So I've decided not to merge this PR, but it was very helpful in figuring out what needed to be done. Thank you to those who contributed. I know this took a long time to get done, but hopefully the sniff that goes into 3.5.0 properly enforces declare formatting and has been written to enable customisation in the future.

jrfnl:
@gsherwood I've just had a look at the sniffs which were added to PHPCS 3.5.0, but - in contrast to what was discussed here -, the formatting sniff now merged does not allow for _any_ customization in spaces required on the inside of parenthesis, around the `=` etc.
Is this intentional ? What happened to everything discussed above ?

gsherwood:
> Is this intentional ? What happened to everything discussed above ?

Yes, it was an intentional decision to code this for PSR-12's rules only so it could be done in the time allowed.

jrfnl:
@gsherwood Would you be open to a PR at a later time to add the properties as discussed above ?

gsherwood:
> Would you be open to a PR at a later time to add the properties as discussed above ?

Yes

PR #2368, opened 2019-01-17
MySource.PHP.AjaxNullComparison throws error when first function has no doc comment

MariusLab:
A RuntimeException is thrown in File::getTokensAsString() if the first parameter is not an int or if the supplied int does not exist as a position in the token stack.

AjaxNullComparisonSniff ends up throwing this exception on every file that contains functions with no comment blocks.


gsherwood:
Thanks for this fix. If you'd like me to credit you in the changelog, please let me know what name you'd like me to use.

PR #2377, opened 2019-01-22
Finer forbidden function token

### LINKED ISSUE #2376: Using __halt_compiler() breaks Generic.PHP.ForbiddenFunctions unless it's last in the function list
Currently, the `ForbiddenFunctionsSniff` get all tokens for forbidden functions by concatenating their names together to a string, then tokenize it using `token_get_all`.
However, `token_get_all` does not work properly if `__halt_compiler` is one of the forbidden function and is not last element of `forbiddenFunctions` array.

This would cause bug in the following situation:
Consider file:
```
<?php
foo();
```
and `forbiddenFunctions` is set to `['__halt_compiler' => null, 'foo' => 'bar']`.
When `ForbiddenFunctionsSniff` is trying to get tokens for each forbidden function by tokenize the joined string, any function after `__halt_compiler` will not be considered. Therefore usage of `foo();` won't be detected in this case.

Billz95:
Fixes #2376


src/Standards/Generic/Sniffs/PHP/ForbiddenFunctionsSniff.php
```diff
@@ -74,15 +74,15 @@ public function register()
 
         // If we are not pattern matching, we need to work out what
         // tokens to listen for.
-        $string = '<?php ';
+        $tokens = [];
         foreach ($this->forbiddenFunctionNames as $name) {
-            $string .= $name.'();';
+            $token = token_get_all(sprintf('<?php %s();', $name));
+            array_shift($token);
+            $tokens = array_merge($tokens, $token);
```
jrfnl: Using `array_merge()` within a loop is highly inefficient.

Looking at your report in #2376, why not just look at the array key/value and skip (`continue`) if it's `__halt_compiler` ? and if found, add it to the end of the string before the tokenization ?

Billz95: Yes that would be a valid fix as well, and would have much better performance if this `Sniff` is instantiated many times.
My main reason for using this rather inefficient method is to avoid considering any edge/special cases (just in case some other functions would break `token_get_all` in the future).
If performance is an important concern here, I'll change it accordingly.

jrfnl: I don't know of any other PHP function (nor of RFCs to add any) which would halt the compiler and therefore the tokenizer, so IMO it should be safe to change this.

gsherwood:
Thanks for fixing that issue. If you'd like a credit in the changelog, please let me know your name and I'll add it in. Otherwise, thanks again.

Billz95:
> Thanks for fixing that issue. If you'd like a credit in the changelog, please let me know your name and I'll add it in. Otherwise, thanks again.

No worries. Though I'm not a show-off person, this is my first PR so please have my name included: Sijun Zhu :smiley:

gsherwood:
> No worries. Though I'm not a show-off person, this is my first PR so please have my name included: Sijun Zhu 😃

Very happy to. Thanks again.

PR #2378, opened 2019-01-22
Fixed docblock

kukulich:
Because custom tokens are strings.


gsherwood:
Very true. They once were all ints, but I changed them to strings to stop conflicts with other tools. Thanks for this.

PR #2380, opened 2019-01-23
allow disabling camel case property variable sniff separately

neanton:


gsherwood:
After reviewing the sniff, I've decided not to merge this PR.

The next error code down from this one (the one that would remain as `NotCamelCaps`) is also used for referencing static properties, so differentiating between `->$var` and `::$var` using different error codes does not seem correct.

There is also an existing `MemberNotCamelCaps` error code here, which is then confusing if there is a new `PropertyNotCamelCaps` because they are referencing the same concept.

The sniff itself uses error codes to indicate how the variable is being used and not the type of the variable. So making this change doesn't make this sniff any clearer. The sniff could possibly use a rewrite, but that would be a job for version 4 as it would be a complete BC break.

PR #2385, opened 2019-01-24
Generic/DocComment: add missing fixed file and minor other improvements

jrfnl:

Includes:
* Minor usability improvements for two error messages.
    These error message refer to tags without naming them. While there can't be much confusion anyway as there is normally only one tag per line, mentioning the name of the tag in the error message makes the message more descriptive.
* Defer to the `length` key in the token array instead of recalculating lengths.
    This is the same change as was made for other sniffs in #2290, which I'd kept back as I was afraid it would conflict with this PR otherwise.
* Adding extra unit test documenting the behaviour of the fixers with indented docblocks.
    The fact that the alignment/indentation of the fixes is off is not addressed in this PR and IMO doesn't need to be addressed by this sniff in the first place as the `Squiz.Commenting.DocCommentAlignment` sniff can handle that.
* Adding extra unit tests documenting the sniff and fixer behaviour when tabs are used to align tag values.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -463,7 +463,9 @@ http://pear.php.net/dtd/package-2.0.xsd">
        </dir>
        <dir name="Commenting">
         <file baseinstalldir="PHP/CodeSniffer" name="DocCommentUnitTest.inc" role="test" />
+        <file baseinstalldir="PHP/CodeSniffer" name="DocCommentUnitTest.inc.fixed" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="DocCommentUnitTest.js" role="test" />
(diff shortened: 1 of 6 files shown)


jrfnl:
@gsherwood You're welcome.

Twentyfive down, three more to go.

Once those are done as well, I can pull a PR (ready & waiting) which will start marking tests with missing `.fixed` files as incomplete (and if so desired, could fail the build on them).

The ones which still need to be addressed are:
- [ ] `PEAR.WhiteSpace.ScopeIndent`
- [ ] `Squiz.ControlStructures.SwitchDeclaration` (php + js)
- [ ] `Squiz.CSS.LowercaseStyleDefinition`

PR #2388, opened 2019-01-26
Squiz+Zend/ValidVariableName: improve support for anonymous classes

jrfnl:
Underscore prefixes were ignored for plain variables in classes, interfaces and traits, but not for anonymous classes.

This commit makes the behaviour consistent for all OO structures by ignoring underscore prefixes for variables in anonymous classes too.

Includes unit tests.


gsherwood:
Thanks

Issue #2391, opened 2019-01-28
Sniff-specific ignore rules inside rulesets are filtering out too many files


wvega:
Consider the following custom standard (`phpcs.xml`):

```xml
<?xml version="1.0"?>
<ruleset name="Custom Standard">
    <rule ref="WordPress">
        <exclude-pattern>/the/content/of/the/pattern/is/irrelevant/</exclude-pattern>
    </rule>
</ruleset>
```

If the path to the project using that standard file includes `wordpress` (the name of the referenced rule) at any level, PHPCS will ignore all files on that project. The list of files to check is empty, even if other rules are being referenced in the standard without using an exclude pattern.

The way I understand it, on v3.4.0 and up to https://github.com/squizlabs/PHP_CodeSniffer/commit/8c94cf9231dc70e8cb45c52a8f3a34501e271da0 (at least), when exclude-patterns are defined for specific standards, categories or sniffs, the `Ruleset::getIgnorePatterns()` method returns an array with a structure similar to:

```php
$ignorePatterns = [
    'WordPress' => [
        '/the/content/of/the/pattern/is/irrelevant/' => 'absolute'
    ],
    ...
    'WordPress.PHP.YodaConditions.NotYoda' => [
        '/the/content/of/the/pattern/is/irrelevant/' => 'absolute'
    ],
    ...
]
```

That array is later used on the [shouldIgnorePath() method](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Filters/Filter.php#L231-L263), as follows:

```php
foreach ($ignorePatterns as $pattern => $type) {
    // ...
    if (preg_match($pattern, $testPath) === 1) {
        return true;
    }
}
```

So that the name of the standard, category or sniff ends up being used as the pattern to ignore. 

When the exclude-pattern is specific to a category of sniffs or a single sniff, then is very unlikely that the name of that category or sniff matches the path of any file on the project, but when the exclude-pattern is specific to a standard with a single word name (like `WordPress` above) the problem is more likely to appear.

---

I'll submit a PR with a test for this issue in a moment.

---

Related to https://github.com/squizlabs/PHP_CodeSniffer/issues/686.

jrfnl:
@wvega Thanks for setting up that unit test, I'm looking into this now.

jrfnl:
[...]

PR #2390, opened 2019-01-27
ValidatePEARPackageXML: minor tweaks

jrfnl:
* Add a new line between the different checks.
* Allow for utility method tests using a JS or CSS test case file.
* Allow for utility method tests using a XML ruleset test file. (see PR #2392 / https://github.com/wvega/PHP_CodeSniffer/pull/1)

PR #2392, opened 2019-01-28
Check paths are accepted if no exclude-pattern matches

wvega:
This PR includes a test to check that paths that include the name of a standard with associated exclude-patterns are still accepted for the list of files to be checked.

The test reproduces https://github.com/squizlabs/PHP_CodeSniffer/issues/2391. 

I followed the coding standards (`php bin/phpcs` reports no problems) but I'm not sure if I added the test to the right place or used the correct approach by using `AcceptTest.inc` as the XML file with the custom standard that reproduces the problem.

I'll be happy to make modifications if the test is considered valid and additional instructions are provided.

As for the solution, I could try to provide one, but I'm not sure what's the right approach in this case. It seems to me that the ignore patterns that are specific to a standard, category or sniff shouldn't be considered on `Filter::shouldIgnorePath()` because that could cause some paths to be excluded entirely only because the custom standard says a subset of the sniffs shouldn't be used on that path.

What do you think?

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -138,6 +138,12 @@ http://pear.php.net/dtd/package-2.0.xsd">
       <file baseinstalldir="" name="IsReferenceTest.inc" role="test" />
       <file baseinstalldir="" name="IsReferenceTest.php" role="test" />
      </dir>
+     <dir name="Filters">
+      <dir name="Filter">
+       <file baseinstalldir="" name="AcceptTest.xml" role="test" />
+       <file baseinstalldir="" name="AcceptTest.php" role="test" />
+      </dir>
+     </dir>
      <file baseinstalldir="" name="AllTests.php" role="test" />
      <file baseinstalldir="" name="ErrorSuppressionTest.php" role="test" />
      <file baseinstalldir="" name="IsCamelCapsTest.php" role="test" />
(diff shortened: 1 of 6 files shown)


jrfnl:
Note: the build failure is not directly related to this PR and will be fixed by #2390.

peternolland:
Thanks for the patch and for providing the update to package.xml

I moved is_array check earlier in shouldIgnorePath() as:
    $this->config->ignored was an array of pattern => string,
where as
    $this->ruleset->getIgnorePatterns() was an array of ref => [pattern => string]

Which meant the original array_merge was not merging the arrays correctly, which caused all files to be excluded as when the rule ref name appeared in the projects path.

jrfnl:
You're welcome ?

wvega:
@peternolland thank you for merging this into master.

I think the [credit](https://github.com/squizlabs/PHP_CodeSniffer/commit/8b713e14f230c238580a2b669b76460c3a611017#diff-34ae68f4adad56c25c5bc05dcb64794e) should go to @jrfnl as well. I reported the bug and submitted an initial version of a failing test, but [she provided the patch, the update to package.xml and greatly improved the tests](https://github.com/wvega/PHP_CodeSniffer/pull/1) :)

gsherwood:
@jrfnl my bad - didn't notice the commit history. Thanks for the fix.

PR #2393, opened 2019-01-29
The gitmodified filter will infinitely loop when encountering deleted file paths

lmanzke:
Deleted files cause realpath($path) to return false. When this value
enters the while loop, it will run indefinitely as the condition will
never be matched.


gsherwood:
Thanks a lot for finding and fixing this.

lmanzke:
You're welcome :)

PR #2399, opened 2019-02-05
Generic/FunctionCallArgumentSpacing: remove assignment operator spacing checks

jrfnl:
As discussed in #2387, there are dedicated sniffs for checking the spacing around assignment operators:
* `PSR12.Operators.OperatorSpacing`
* `Squiz.WhiteSpace.OperatorSpacing`

Both of these are more comprehensive and more appropriate to use than the limited assignment operator spacing check in this sniff.

Removing the assignment operator spacing checks from this sniff, lowers the risk of fixer conflicts regarding operator spacing and gets rid of duplicate notices in the `PSR12` and `Squiz` standard and inappropriate notices for the `PEAR`, `PSR2` and `Zend` standard.

Includes adjusted unit tests.

### DIFF
--- a/src/Standards/Generic/Sniffs/Functions/FunctionCallArgumentSpacingSniff.php
+++ b/src/Standards/Generic/Sniffs/Functions/FunctionCallArgumentSpacingSniff.php
@@ -87,7 +87,6 @@ class FunctionCallArgumentSpacingSniff implements Sniff
 
         $find = [
             T_COMMA,
-            T_VARIABLE,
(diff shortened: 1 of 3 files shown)


sertand:
Thanks for the patch.

gsherwood:
An unintended benefit of this change was that the PEAR, PSR2, and PSR12 standard are no longer checking assignment operators inside function calls. They should never have been doing this, but the Squiz standard was, and so they inherited it.

PR #2401, opened 2019-02-05
Squiz/IncrementDecrementUsage: suggest pre-in/decrement

### LINKED ISSUE #2354: Wrong suggestion for increment/decrement sniff
Hello,

First, thanks for this great tool :)

I believe I spotted an issue.

If you have:

~~~php
$a = 3;
echo $a += 1; // 4
~~~

You'll get an error (Squiz.Operators.IncrementDecrementUsage.Found), and the suggestion is to use `$a++`. But this is wrong in some cases, the correct suggestion imho would be to have `++$a`.

See: https://3v4l.org/KFHi9

What do you think?

Regards,
~Nico

Note: link to the code:
https://github.com/squizlabs/PHP_CodeSniffer/blob/957a2b63c084cc8b34cdb1a2c77ae12257c79a3d/src/Standards/Squiz/Sniffs/Operators/IncrementDecrementUsageSniff.php#L208

jrfnl:
... rather than post in/decrement.

Fixes #2354

### DIFF
--- a/src/Standards/Squiz/Sniffs/Operators/IncrementDecrementUsageSniff.php
+++ b/src/Standards/Squiz/Sniffs/Operators/IncrementDecrementUsageSniff.php
@@ -205,7 +205,7 @@ class IncrementDecrementUsageSniff implements Sniff
                 }
             }
 
-            $expected = $tokens[$assignedVar]['content'].$operator.$operator;
+            $expected = $operator.$operator.$tokens[$assignedVar]['content'];
             $found    = $phpcsFile->getTokensAsString($assignedVar, ($statementEnd - $assignedVar + 1));
 
             if ($operator === '+') {


gsherwood:
This is a pretty big change to the Squiz coding standard, and potentially a huge BC break for existing custom standards. I'll have to discuss it internally before a change like this could be made and consider if it needs to be held to 4.0 as well. An alternative would be to provide sniffs for both formats as this wouldn't impact existing standards.

jrfnl:
@gsherwood Well, the change is smaller than it may feel. The only difference is in the error message displayed.
* The sniff does not contain any auto-fixers, so code will not be fixed differently than before.
* The sniff will not be triggered on post-increment, nor suggest that existing post-in/decrements be changed to pre-in/decrements.

In other words, people can still (manualy) fix code with _post_-in/decrement and the sniff will accept that as perfectly valid.

Anyways, have a think about it. I do believe the op had a valid point, but can totally understand it if this is a change which would be punted to 4.0 or moved to a separate sniff.

gsherwood:
Thanks a lot for doing this. You were right - the change isn't as dramatic as I made it out to be in my head, and the people I've spoken to understand and agree with the change.

PR #2402, opened 2019-02-05
Squiz/DisallowMultipleAssignments: split errorcode

### LINKED ISSUE #1679: Squiz.PHP.DisallowMultipleAssignments error on code with comparison and assignment
I updated PHPCS to 3.1.0 and I’m now getting a lot of `Assignments must be the first block of code on a line` errors.

They’re on code like this:

    if ( ( $key = array_search( 'tag', $classes, true ) ) !== false ) {

Are you intending to prohibit these sorts of expressions? If so, it would be good to use a different type of error message. If not, can you fix this?

### LINKED ISSUE #2083: Split Squiz.PHP.DisallowMultipleAssignments.Found into two different error codes
Code:
```php
if ($app->runningInConsole() && ($input = new ArgvInput())->hasParameterOption('--env')) {
}
```

jrfnl:
As per the feature request in #2083, the errorcode of the sniff has now been split into two:
* `FoundInControlStructure` when the assignment is within a control structure (with the exception of `while` statements and the first leaf of a `for` statement as those won't be reported anyway).
* `Found` for all other cases.

Fixes #2083
Fixes #1679

### DIFF
--- a/src/Standards/Squiz/Sniffs/PHP/DisallowMultipleAssignmentsSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/DisallowMultipleAssignmentsSniff.php
@@ -153,8 +153,28 @@ class DisallowMultipleAssignmentsSniff implements Sniff
             return;
         }
 
-        $error = 'Assignments must be the first block of code on a line';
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for making this change

PR #2407, opened 2019-02-07
Remove deprecated each() function from test examples

phil-davis:
``each()`` is deprecated from PHP 7.2 http://php.net/manual/en/function.each.php
My IDE was complaining about this in an example test file.
Even though the test examples are not required to be strictly-valid PHP (they are more for testing format parsing), it seems reasonable to remove this sort of deprecated stuff from examples.


gsherwood:
Good idea - each() is not core to this sniff so it can be changed. Thanks.

PR #2418, opened 2019-02-11
Added Generic.PHP.DisallowAlternativePHPTags to PSR1 ruleset

### LINKED ISSUE #2415: PSR1 not reporting or fixing alternative tags
In [PSR1 documentation](https://www.php-fig.org/psr/psr-1/) says:

> PHP code MUST use the long <?php ?> tags or the short-echo <?= ?> tags; it MUST NOT use the other tag variations.

I noticed that when running the sniffer (PSR1) against a file with alternative (ASP or script) tags, that it does not find the problem:
```bash
$ ./bin/phpcs -s -vvv --standard=PSR1 testfile.php
```
In my opinion the PSR1 ruleset.xml should include also the Generic.PHP.DisallowAlternativePHPTags:
```xml
<rule ref="Generic.PHP.DisallowAlternativePHPTags"/>
```
I'm wrong?

elplaza:
Close #2415

### DIFF
--- a/src/Standards/PSR1/ruleset.xml
+++ b/src/Standards/PSR1/ruleset.xml
@@ -7,6 +7,7 @@
     <!-- 2.1. PHP Tags -->
 
     <!-- PHP code MUST use the long <?php ?> tags or the short-echo <?= ?> tags; it MUST NOT use the other tag variations. -->
+    <rule ref="Generic.PHP.DisallowAlternativePHPTags"/>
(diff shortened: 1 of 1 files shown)


gsherwood:
Awaiting feedback on #2415

jrfnl:
Suggest: close as this has been fixed via https://github.com/squizlabs/PHP_CodeSniffer/commit/55d563ec5b997a217478f702d9316e4f47e017a8

PR #2432, opened 2019-02-23
File::findStartOfStatement/findEndOfStatement(): minor tweak

jrfnl:
`unset()` will silently fail if the array key/variable to unset doesn't exist, so there is no need to wrap this in an `isset()`.


gsherwood:
Nice find, thanks.

PR #2434, opened 2019-02-24
Native PHPCS ruleset: ignore Utility method test case files

jrfnl:
.. and make the pattern for ignoring the Sniff test case files slightly more specific.

Includes removing existing `phpcs:ignoreFile` comments from utility test case files.


gsherwood:
Good idea. Thanks.

PR #2445, opened 2019-03-16
Tokenizer: assign a parenthesis_owner for list() parenthesis

jrfnl:
`list()`, just like `array()` is a language construct.

While the parentheses for `array()` would receive the `array` token as the `parenthesis_owner`, the parentheses for `list()` did not.

This commit fixes that inconsistency.

Tested by removing the `T_LIST` token from the "additional tokens indicating that parenthesis are not arbitrary" list in the `Generic.WhiteSpace.ArbitraryParenthesesSpacing` sniff.


gsherwood:
Thanks for this

PR #2446, opened 2019-03-18
Refer to semantic versioning

xuanquynh:
I think we can give the others more information about versioning by adding a reference to semantic versioning home page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for information.
 
 ## Versioning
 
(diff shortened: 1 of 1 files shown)


gsherwood:
Following that semver doc is more than just the version number format. The rules listed on that page are not followed by this project, which is why I list the project version rules in the README and also why I don't mention semver. I need to define terms like "bug" and "compatible" myself, as they apply to PHPCS users (developers using the tool, developers using the XML format, and developers writing sniffs).

For that reason, I'm going to close this PR. I hope I've explained that well enough.

PR #2451, opened 2019-03-19
SonarQube report format

### LINKED ISSUE #2404: [Feature request] Add a Sonar reporter
SonarQube supports external analyzer but needs a specific format: https://docs.sonarqube.org/latest/analysis/generic-issue/

A Sonar reporter will be helpful in order to track CodeSniffer rules history in SonarQube.

freezy-sk:
SonarQube report format as described in https://docs.sonarqube.org/latest/analysis/generic-issue/
Closes #2404

### DIFF
--- a/src/Reports/Sonar.php
+++ b/src/Reports/Sonar.php
@@ -1,0 +1,128 @@
+<?php
+/**
+ * SonarQube report for PHP_CodeSniffer.
+ *
+ * @author  Marek Víger <marek.viger@gmail.com>
+ * @license https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ */
+
+namespace PHP_CodeSniffer\Reports;
+
+use PHP_CodeSniffer\Files\File;
+
+class Sonar implements Report
+{
+
+
+    /**
+     * Generate a partial report for a single processed file.
+     *
+     * Function should return TRUE if it printed or stored data about the file
+     * and FALSE if it ignored the file. Returning TRUE indicates that the file and
+     * its data should be counted in the grand totals.
+     *
+     * @param array $report      Prepared report data.
+     * @param File  $phpcsFile   The file being reported on.
+     * @param bool  $showSources Show sources?
+     * @param int   $width       Maximum allowed line width.
+     *
+     * @return bool
+     */
+    public function generateFileReport($report, File $phpcsFile, $showSources=false, $width=80)
+    {
+        foreach ($report['messages'] as $line => $lineErrors) {
+            foreach ($lineErrors as $column => $colErrors) {
+                foreach ($colErrors as $error) {
+                    $error['message'] = strtr(
+                        $error['message'],
+                        [
+                            "\n" => '\n',
+                            "\r" => '\r',
+                            "\t" => '\t',
+                        ]
+                    );
+
+                    $issue = [
+                        'engineId'        => 'PHP_CodeSniffer',
+                        'ruleId'          => $error['source'],
+                        'type'            => 'CODE_SMELL',
+                        'severity'        => $this->convertErrorTypeToSonarSeverity($error['type']),
+                        'primaryLocation' => [
+                            'message'   => $error['message'],
+                            'filePath'  => $report['filename'],
+                            'textRange' => [
+                                'startLine'   => $line,
+                                'endLine'     => $line,
+                                'startColumn' => $column,
+                                'endColumn'   => $column,
+                            ],
+                        ],
+                    ];
+
+                    echo json_encode($issue).',';
+                }//end foreach
+            }//end foreach
+        }//end foreach
+
+        return true;
+
+    }//end generateFileReport()
+
+
+    /**
+     * Generates a SonarQube report.
+     *
+     * @param string $cachedData    Any partial report data that was returned from
+     *                              generateFileReport during the run.
+     * @param int    $totalFiles    Total number of files processed during the run.
+     * @param int    $totalErrors   Total number of errors found during the run.
+     * @param int    $totalWarnings Total number of warnings found during the run.
+     * @param int    $totalFixable  Total number of problems that can be fixed.
+     * @param bool   $showSources   Show sources?
+     * @param int    $width         Maximum allowed line width.
+     * @param bool   $interactive   Are we running in interactive mode?
+     * @param bool   $toScreen      Is the report being printed to screen?
+     *
+     * @return void
+     *
+     * @see https://docs.sonarqube.org/latest/analysis/generic-issue/
+     */
+    public function generate(
+        $cachedData,
+        $totalFiles,
+        $totalErrors,
+        $totalWarnings,
+        $totalFixable,
+        $showSources=false,
+        $width=80,
+        $interactive=false,
+        $toScreen=true
+    ) {
+        printf('{"issues":[%s]}', rtrim($cachedData, ','));
+
+    }//end generate()
+
+
+    /**
+     * Convert error type to SonarQube severity key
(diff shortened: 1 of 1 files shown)


gsherwood:
My hesitation with this is that I can't quickly test that this format works (unless there is a validator somewhere) and I'm not keen on maintaining this report format unless there is a significant number of users.

It's a very well done PR though, and people could use this report format with the `--report` command line argument right now if they want to test it out. Maybe those who are interested could try this report class and confirm it works.

freezy-sk:
@gsherwood I could write some test cases for this report format. Can you give me some ideas how  would you like to test report formats? I don't see any report format test cases yet

gsherwood:
> I could write some test cases for this report format

Sorry, I wasn't clear enough.

The testing I was talking about is testing of the report inside SonarQube itself, by something who is using SonarQube. I don't use it, but as soon as I accept this PR, I'm taking responsibility for maintaining this report format.

So not only do I need to make sure it actually works, I also need to make sure it works on all future versions of SonarQube. I have no experience with that platform and I don't know if the report format is likely to change.

This is always my hesitation with accepting new report formats for platform I don't use and can't easily confirm if the report is working.

freezy-sk:
@gsherwood I understand. I can't guarantee you, that SonarQube don't change this format in some new major release. But what do you think about possibility to install external report formats like it can be done with standards?

gsherwood:
> But what do you think about possibility to install external report formats like it can be done with standards?

You can already use external reports by specifying the path on the command line, which is what I'm hoping people do with your report format to confirm it's all working.

I don't think I'd add support for actually installing report formats though, unless there was a growing collection of them.

freezy-sk:
I think that collection will not grow without possibility to install external report format. Specifying path to report installed to global or local vendor folder is little bit impractical.

jrfnl:
@freezy-sk Have a look at PR #1948 (merged in PHPCS 3.3.0), slightly different approach, but it allows end-users to use reports shipped with an external standard by name.

You can find an example implementation here: https://github.com/jrfnl/QA-WP-Projects/

dingo-d:
How does this works with the native SQ scanning @freezy-sk? It owerwrites it in favor of phpcs or?

mdeboer:
> How does this works with the native SQ scanning @freezy-sk? It owerwrites it in favor of phpcs or?

No, in those cases you would see a duplicate issue. But it's really easy to disable rules in SQ, it's much much harder to add custom rules, and why would you if there are reliable tools ;) So if you get duplicates, create a new quality profile and set a parent profile or clone your current. Then disable the rules you see duplicates.

@gsherwood I would be really pleased to see this merged as I also use [Psalm](https://github.com/vimeo/psalm) with SQ this way. I don't expect the format to change much or often at all and it's pretty straight forward (simple JSON).

The output of PHP_CodeSniffer could be tested against a [JSON schema](https://json-schema.org/understanding-json-schema/index.html) to make sure it works with SQ. I'd be happy to provide you with one.

src/Reports/Sonar.php
```diff
@@ -0,0 +1,128 @@
...
+                        'ruleId'          => $error['source'],
+                        'type'            => 'CODE_SMELL',
+                        'severity'        => $this->convertErrorTypeToSonarSeverity($error['type']),
+                        'primaryLocation' => [
+                            'message'   => $error['message'],
+                            'filePath'  => $report['filename'],
+                            'textRange' => [
+                                'startLine'   => $line,
+                                'endLine'     => $line,
+                                'startColumn' => $column,
```
mdeboer: The start column should be less than the end column (not the same or more), this causes the sonar-scanner to fail. As far as I know there is no quick and easy way to determine the real start of the line (excluding whitespace) and imho I see no reason why we should. Simply set the `startColumn` to `0` and make sure `endColumn` is always more than `0` (or `startColumn` for that matter). I suggest checking that beforehand and incrementing with 1 if needed.

I tested this with SQ 7.9.1.27448 and with that modification it works fine.

mdeboer:
Here is a JSON schema I wrote according to the specs at https://docs.sonarqube.org/latest/analysis/generic-issue/

```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "type": "object",
    "required": [
        "issues"
    ],
    "properties": {
        "issues": {
            "type": "array",
            "items": {
                "type": "object",
                "required": [
                    "engineId",
                    "ruleId",
                    "severity",
                    "type",
                    "primaryLocation"
                ],
                "properties": {
                    "engineId": {
                        "type": "string",
                        "minLength": 1
                    },
                    "ruleId": {
                        "type": "string",
                        "minLength": 1
                    },
                    "severity": {
                        "type": "string",
                        "enum": [
                            "BLOCKER",
                            "CRITICAL",
                            "MAJOR",
                            "MINOR",
                            "INFO"
                        ]
                    },
                    "type": {
                        "type": "string",
                        "enum": [
                            "BUG",
                            "VULNERABILITY",
                            "CODE_SMELL"
                        ]
                    },
                    "primaryLocation": {
                        "type": "object",
                        "required": [
                            "message",
                            "filePath",
                            "textRange"
                        ],
                        "properties": {
                            "message": {
                                "type": "string",
                                "minLength": 1
                            },
                            "filePath": {
                                "type": "string",
                                "minLength": 1
                            },
                            "textRange": {
                                "type": "object",
                                "required": [
                                    "startLine",
                                    "endLine",
                                    "startColumn",
                                    "endColumn"
                                ],
                                "properties": {
                                    "startLine": {
                                        "type": "integer",
                                        "description": "Starting line of issue.",
                                        "minimum": 1
                                    },
                                    "endLine": {
                                        "type": "integer",
                                        "description": "Ending line of issue.",
                                        "minimum": 1
                                    },
                                    "startColumn": {
                                        "type": "integer",
                                        "minimum": 0
                                    },
                                    "endColumn": {
                                        "type": "integer",
                                        "description": "Ending column of issue. This value must be higher than startColumn when issue is on one line.",
                                        "minimum": 1
                                    }
                                }
                            }
                        }
                    },
                    "effortMinutes": {
                        "type": "integer",
                        "default": 0,
                        "minimum": 0
                    },
                    "secondaryLocations": {
                        "type": "array",
                        "items": {
                            "type": "object",
                            "required": [
                                "message",
                                "filePath"
                            ],
                            "properties": {
                                "message": {
                                    "type": "string",
                                    "minLength": 1
                                },
                                "filePath": {
                                    "type": "string",
                                    "minLength": 1
                                },
                                "textRange": {
                                    "type": "object",
                                    "required": [
                                        "startLine",
                                        "endLine",
                                        "startColumn",
                                        "endColumn"
                                    ],
                                    "properties": {
                                        "startLine": {
                                            "type": "integer",
                                            "description": "Starting line of issue.",
                                            "minimum": 1
                                        },
                                        "endLine": {
                                            "type": "integer",
                                            "description": "Ending line of issue.",
                                            "minimum": 1
                                        },
                                        "startColumn": {
                                            "type": "integer",
                                            "minimum": 0
                                        },
                                        "endColumn": {
                                            "type": "integer",
                                            "description": "Ending column of issue. This value must be higher than startColumn when issue is on one line.",
                                            "minimum": 1
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

src/Reports/Sonar.php
freezy-sk: thanks @mdeboer I didn't noticed crashes of sonar-scanner.
What about making `endColumn` optional as it's written in documentation?

mdeboer: Ah yes, I missed that. I will try that as that'd be a better solution!

PR #2457, opened 2019-03-24
Mark recent tests as such

DavidPrevot:
It avoids installing them with the PEAR package.


jrfnl:
@DavidPrevot Thanks! #2458 should prevent this from ever happening again ;-)

gsherwood:
Thanks for finding and fixing this

PR #2458, opened 2019-03-24
Minor improvement to ValidatePEARPackageXML script

jrfnl:
This minor change means that the `role` attributes for typical test and sniff files will now be checked to prevent the issue reported and fixed in #2457.

Note: the build won't pass until #2457 has been merged.


DavidPrevot:
Good idea, thanks!

gsherwood:
Thanks a lot

PR #2460, opened 2019-03-25
PSR-12: update the ruleset to reflect current version as under review

jrfnl:
As [PSR-12](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md) is now [in review](https://www.php-fig.org/psr/#review) again, an update of the ruleset to reflect the current version was needed.

* Updated all inline documentation texts with the texts as currently in PSR-12.
* Added any new statements which were added to PSR-12.
* Added the `Generic.WhiteSpace.IncrementDecrementSpacing` sniff to satisfy the `There MUST NOT be any whitespace between the increment/decrement operators and the variable being incremented/decremented.` rule.
* Added the `Squiz.WhiteSpace.CastSpacing` sniff to satisfy the `There MUST NOT be any spaces inside the type casting parentheses.` rule.

Note: The `5.3.1`/`5.3.2` split is my doing as it is otherwise completely unclear what the difference is between the two statements.

While updating the inline documentation, I noticed a few issues in PSR-12 itself and have opened PRs to fix those. If and when those would be accepted, the inline documentation in the ruleset will need to be updated (again).

Refs:
* [PSR 12: minor grammatical fix / declare statements](https://github.com/php-fig/fig-standards/pull/1154)
* [PSR-12: minor clarification / object declarations](https://github.com/php-fig/fig-standards/pull/1155)
* [PSR-12: minor correction / multi-line do-while](https://github.com/php-fig/fig-standards/pull/1156)

### DIFF
--- a/src/Standards/PSR12/ruleset.xml
+++ b/src/Standards/PSR12/ruleset.xml
@@ -10,6 +10,8 @@
     <!-- Code MUST follow all rules outlined in PSR-1. -->
     <rule ref="PSR1"/>
 
+    <!-- The term 'StudlyCaps' in PSR-1 MUST be interpreted as PascalCase where the first letter of each word is capitalized including the very first letter. -->
+
     <!-- 2.2 Files -->
 
     <!-- All PHP files MUST use the Unix LF (linefeed) line ending only. -->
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for doing this

jrfnl:
You're welcome ;-)

jrfnl:
@gsherwood FYI: This will need a changelog entry for the two new sniffs added to the PSR-12 ruleset.

PR #2461, opened 2019-03-25
Help: add link to wiki at the bottom of the messages

### LINKED ISSUE #2412: The "phpcs" doesn't use error exit code when invalid standard is given
I was trying to use PHP_CodeSniffer 2.x standard on PHP_CodeSniffer 3.x binary. Since `ruleset.xml` format wasn't changed I wasn't getting any exception due wrongly named sniff classes.

I however did get this output:

```
ERROR: Referenced sniff "Squiz.PHP.DisallowObEndFlush" does not exist

Run "phpcs --help" for usage information
```

Unfortunately exit code from that operation was 0 (= success) and therefore any build system using PHP_CodeSniffer wouldn't fail.

jrfnl:
Fixes #2412

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -1421,6 +1421,10 @@ class Config
         echo ' <standard>     The name or path of the coding standard to use'.PHP_EOL;
         echo ' <stdinPath>    If processing STDIN, the file path that STDIN will be processed as'.PHP_EOL;
         echo ' <tabWidth>     The number of spaces each tab represents'.PHP_EOL;
+        echo PHP_EOL;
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks, but this isn't something I want to include in the CLI help text as it forces this URL to exist forever.

jrfnl:
@gsherwood I understand the objection, though the URL could be updated if it would change. Alternatively the URL could be changed to a permalink for which the underlying redirect could be changed if the end-point would change.

I'd be happy to adjust the PR.

PR #2465, opened 2019-03-28
Excluding a sniff by path is not working

lmanzke:
This is one of the places where the strict comparison breaks things. If you exclude a sniff in the XML via this line, the strict = true causes in_array to return false as it is iterating over XMLElements.

There _might_ be other places where this is also wrong, so I suggest doublechecking every place affected by https://github.com/squizlabs/PHP_CodeSniffer/commit/c280c14885a73c9f0dc18036f13caa79e8b583e6 if the behavior is still the desired one.

### DIFF
--- a/src/Ruleset.php
+++ b/src/Ruleset.php
@@ -588,7 +588,7 @@ class Ruleset
         // sniff list, but filter out any excluded sniffs.
         $files = [];
         foreach ($includedSniffs as $sniff) {
-            if (in_array($sniff, $excludedSniffs, true) === true) {
+            if (in_array($sniff, $excludedSniffs) === true) {
                 continue;
             } else {
                 $files[] = Util\Common::realpath($sniff);


lmanzke:
To still use strict comparison, one could also just use forEach to determine if the element is in the array.

lmanzke:
Or probably not, as the SimpleXMLElement class really behaves weirdly ...

jrfnl:
@lmanzke Could you provide an example so I can reproduce what was broken by this ?

gsherwood:
I think this might be caused by excluding sniffs by path and not code. But the required change here is not to remove the strict check - I need to figure out why a SimpleXMLElement object is being included in the array at all.

gsherwood:
The ruleset logic for including sniffs already cast these objects to strings, but the excluding logic didn't. I've fixed this now, so the strict comparison will now pass. Thanks for reporting this.

lmanzke:
Thanks for fixing - sorry I didn't respond ealier @jrfnl - was too busy...

PR #2474, opened 2019-04-08
PHP_CodeSniffer\Exceptions\RuntimeException extends \RuntimeException

donatj:
I'm not sure if this is a controversial change or not. 

I can fully understand it being rejected outright, however it is a change I would personally appreciate.

I develop a lot of sniffs in my work, and PhpStorm barks about `PHP_CodeSniffer\Exceptions\RuntimeException` being unhandled. I can add it to the list of handled exceptions but I end up doing this on every project, and I touch hundreds. 

PhpStorm however specifically does _not bark_ about things that extend \RuntimeException out of the box - and it my book it makes sense for `PHP_CodeSniffer\Exceptions\RuntimeException` to extend `\RuntimeException`

Screen Shot 2019-04-08 at 2 53 44 PM

### DIFF
--- a/src/Exceptions/RuntimeException.php
+++ b/src/Exceptions/RuntimeException.php
@@ -9,7 +9,7 @@
 
 namespace PHP_CodeSniffer\Exceptions;
 
-class RuntimeException extends \Exception
+class RuntimeException extends \RuntimeException
 {
 
 }//end class


jrfnl:
@donatj Please correct me if I'm wrong (I don't use PHPStorm much ATM), but I don't think this is what is meant by that notice and I expect your change will not have the desired effect.

When I see that notice, I think PHPStorm expects you to write your code like this:
```php
try {
    $funcName = $phpcsFile->getDeclarationName($stackPtr);
} catch ( RuntimeException $e ) {
    // Handle it.
}
```
... as the `getDeclarationName()` method can potentially throw a `PHP_CodeSniffer\Exceptions\RuntimeException`.

donatj:
@jrfnl Correct, it's expecting it to be wrapped in a try. As `getDeclarationName` is used throughout the code base of PHP_CS however it's not a try'd exception in general, and in writing a similar style of sniff it's a continual complaint.

https://github.com/squizlabs/PHP_CodeSniffer/blob/16e89838a491c06af60acffd2acfc6a56c505cf8/src/Standards/PSR1/Sniffs/Methods/CamelCapsMethodNameSniff.php#L58-L64

https://github.com/squizlabs/PHP_CodeSniffer/blob/16e89838a491c06af60acffd2acfc6a56c505cf8/src/Standards/PEAR/Sniffs/NamingConventions/ValidFunctionNameSniff.php#L79-L87

PhpStorm specifically excludes \RuntimeException from these complaints because it figures it is something you would not be catching. 

Anyway **regardless** of PhpStorm, I think making I think making `PHP_CodeSniffer\Exceptions\RuntimeException` a `\RuntimeException` is quite logical.

jrfnl:
> As getDeclarationName is used throughout the code base of PHP_CS however it's not a try'd exception in general

That's because the `Exception` is thrown when the method receives an unexpected type of token and the PHPCS native sniffs using the method wouldn't call the method in that case (either because the sniff isn't active for other types of tokens or because it has a preliminary check).

> Anyway regardless of PhpStorm, I think making `PHP_CodeSniffer\Exceptions\RuntimeException` a `\RuntimeException` is quite logical.

Makes sense to me :+1:

gsherwood:
Makes sense to me. Thanks.

PR #2483, opened 2019-04-22
Allows indexes containing hyphens

josephzidell:
Fixes this code:
```js
foo['bar-baz']
```


jrfnl:
@josephzidell Could you please add a unit test for your change ?

josephzidell:
@jrfnl Is that what you had in mind?

jrfnl:
> @jrfnl Is that what you had in mind?

@josephzidell Sort of, though I see no reason why all the other lines in the test file were changed.
The new test could have just been added at the end of the test case file. That way, the diff is cleaner and it's a lot clearer what your actual change was.

josephzidell:
Updated

gsherwood:
Missed this one - but closing now as the JS tokenizer will be removed in version 4, so no work is being done to improve it. Sorry this was missed when you submitted it.

PR #2491, opened 2019-05-06
XSD schema: allow for phpcs-only and phpcbf-only attributes

jrfnl:
While validating a custom ruleset against the PHPCS XSD schema, I realized that the `phpcs-only` and the `phpcbf-only` attributes were not accounted for in the schema.

Ref: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-Ruleset#selectively-applying-rules

This PR fixes that.

Notes:
* I have added the `attributeGroup` to nearly all XML elements/nodes, with the exception of:
    - The toplevel `ruleset` and the `description` element as per the documentation.
    - The `autoload` element as not using autoload for either `phpcs` or `phpcbf` would most likely break the standard using the element.
    - The `message` element of the `ruleType` as this only applies to `phpcs` anyway.
    - The individual `element` nodes within an array `<property>` group.
* Includes adjusting the `extend` attribute for the `properties` node from type `string` (restricted to `true|false` via enumeration) to type `boolean`.

Suggestion: it may not be a bad idea to validate the XSD schema during the Travis build test using something along the lines of:
```
curl -O https://www.w3.org/2012/04/XMLSchema.xsd
xmllint.exe --noout --schema XMLSchema.xsd ./phpcs.xsd
```

### DIFF
--- a/phpcs.xsd
+++ b/phpcs.xsd
@@ -9,20 +9,31 @@
                     <xs:complexType>
                         <xs:attribute name="name" type="xs:string" use="required"></xs:attribute>
                         <xs:attribute name="value" type="xs:string" use="required"></xs:attribute>
+                        <xs:attributeGroup ref="applySelectively"/>
                     </xs:complexType>
                 </xs:element>
-                <xs:element name="file" type="xs:string" maxOccurs="unbounded" minOccurs="0"></xs:element>
+                <xs:element name="file" maxOccurs="unbounded" minOccurs="0">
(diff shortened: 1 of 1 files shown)


gsherwood:
I'm very glad you did this because XSDs drive me crazy.

Adding the validation of phpcs.xsd was a good idea, so I've done that too. Thanks.

jrfnl:
Thanks for merging this.

PR #2501, opened 2019-05-11
Ruleset::processRule(): fix handling of rules included via path

### LINKED ISSUE #2497: Sniff properties not set when referencing a sniff using relative paths or non-native slashes
When using this syntax:

```xml
<rule ref="./path/to/sniff/class.php">
    <properties>
        <property name="foo" value="bar" />
    </properties>
</rule>
```

The sniff will the value from the class, not the one configured.

jrfnl:
This fixes a bug where properties set for rules included via a path to the sniff file or to an included standards file, would be disregarded if the slashes used in the path did not match the slashes expected for the OS on which the ruleset is being run.

Includes adding initial integration tests for the Ruleset class.

Fixes #2497

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -239,6 +239,15 @@ http://pear.php.net/dtd/package-2.0.xsd">
        <file baseinstalldir="" name="AcceptTest.php" role="test" />
       </dir>
      </dir>
+     <dir name="Ruleset">
+      <file baseinstalldir="" name="RuleInclusionTest.php" role="test" />
+      <file baseinstalldir="" name="RuleInclusionTest.xml" role="test" />
+      <file baseinstalldir="" name="RuleInclusionTest-include.xml" role="test" />
+      <file baseinstalldir="" name="RuleInclusionAbsoluteLinuxTest.php" role="test" />
+      <file baseinstalldir="" name="RuleInclusionAbsoluteLinuxTest.xml" role="test" />
+      <file baseinstalldir="" name="RuleInclusionAbsoluteWindowsTest.php" role="test" />
+      <file baseinstalldir="" name="RuleInclusionAbsoluteWindowsTest.xml" role="test" />
+     </dir>
      <dir name="Tokenizer">
       <file baseinstalldir="" name="AnonClassParenthesisOwnerTest.inc" role="test" />
       <file baseinstalldir="" name="AnonClassParenthesisOwnerTest.php" role="test" />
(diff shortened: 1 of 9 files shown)


jrfnl:
Rebased for merge conflicts.

jrfnl:
And rebased _again_ for merge conflicts.

jrfnl:
Rebased _yet again_

gsherwood:
Sorry, have finally got to this.

I'm a bit worried about modifying committed test files during a test run (much more so for me running them locally). Would it instead be possible to have different XML files that are included conditionally based on the OS the test is run on?

jrfnl:
> Would it instead be possible to have different XML files that are included conditionally based on the OS the test is run on?

No, it wouldn't, or rather: there are already different test files for different OSes.

The problem is that I cannot hard-code an absolute file-path in the XML file as the path will be different depending on where the tests are being run from, while an _absolute_ file path is _exactly_ the situation which _needs_ testing.

As I documented in the test file:
```
// On-the-fly adjust the ruleset test file to be able to test sniffs included with absolute paths.
```


> I'm a bit worried about modifying committed test files during a test run (much more so for me running them locally).

I understand the question, but I actually anticipated that.
The `tearDown()` method brings the XML files back to their original state after the test, so after running the tests, there should be no changes.

gsherwood:
> The problem is that I cannot hard-code an absolute file-path in the XML file as the path will be different depending on where the tests are being run from

Yep, this is the bit I missed - I thought it was just slashes.

> The `tearDown()` method brings the XML files back to their original state after the test

I saw that, but I'm still nervous about modifying committed files. Ideally, I'd refactor the class to allow the contents of the XML file to be passed instead of only read from disk, but I'm not sure I have the time for that at the moment, so may need to leave as is.

Thanks for clarifying for me.

jrfnl:
> I saw that, but I'm still nervous about modifying committed files

I understand, but there just isn't any other way to test this and not having tests for a change like this, which is fundamental in the ruleset processing, seemed irresponsible.

While working on this, I've ran the tests a significant number of times and didn't run into any issues with the logic now in place.

The only other thing I can suggest for now, would be to have a `git reset` for each of these files in the `tearDownAfterClass()` as an additional safeguard.

gsherwood:
Thanks a lot for this fix.

> The only other thing I can suggest for now, would be to have a `git reset` for each of these files in the `tearDownAfterClass()` as an additional safeguard.

They run fine, so I'll just make sure I don't commit something if they ever fail in the future. Thanks a lot for writing these.

jrfnl:
@gsherwood Thanks for the merge, one more issue down :tada:

PR #2502, opened 2019-05-12
Generic.WhiteSpace.ScopeIndent false positives with nested switch indentation and case fall-through

klausi:
When you have a nested switch statement without break statements then the scope indent sniff gets confused. Here is a testcase to demonstrate the problem, I have not found a solution yet.

### DIFF
--- a/src/Standards/Generic/Tests/WhiteSpace/ScopeIndentUnitTest.5.inc
+++ b/src/Standards/Generic/Tests/WhiteSpace/ScopeIndentUnitTest.5.inc
@@ -1,0 +1,16 @@
+<?php
+
+function entity_permissions($entity_type_id, $operation) {
+    switch ($entity_type_id) {
+        case 'entity_test':
+            switch ($operation) {
+                case 'delete':
+                    return ['administer entity_test content'];
+            }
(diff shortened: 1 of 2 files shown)


gsherwood:
The problem is that the return statements are being run twice, closing the CASE statement (as they should) but also the containing SWITCH. That leaves the indent in a bit of a mess.

I think I've found the problem and I'm testing a fix now.

gsherwood:
The sniff was checking the indent in front of the return statement in one loop, then checking the return statement itself in the next loop. This is fine, but it ended up closing the scope of the CASE statement twice, which actually closed the scope of the SWITCH statement the second time.

I've added a test case and a fix here: https://github.com/squizlabs/PHP_CodeSniffer/commit/802e6efe3fb26a34efda7ee344357ac963a3db4c

Your test case is now passing for me, although it would still be great if you were able to test the fix as well.

Thanks for reporting this.

klausi:
Thanks a lot Greg! I was trying to get into this sniff myself and thought closing the scope twice was correct, but your fix is working as intended!

klausi:
And found another bug with this sniff, opened #2504

PR #2504, opened 2019-05-13
Generic.WhiteSpace.ScopeIndent false positives with nested arrays and nowdoc string

klausi:
Using nowdoc syntax in a nested array confuses the ScopeIndent sniff. Here is a test case that should not throw any indentation errors, but it does.


gsherwood:
Thanks for reporting this. I've fixed this one up now - the here/nowdoc processing was too aggressive for when they are included as array values.

klausi:
Thanks a lot yet again, I can confirm that this fixes my problem!

PR #2515, opened 2019-05-22
Add option to allow checking assignment operators

grongor:
I would like to enforce checking of spaces around assignments as I'm not allowing statements alignment. This simple change (no BC-break) allows me to do so.


gsherwood:
I like this idea.

I think the public var should be renamed - maybe to `ignoreSpacingBeforeAssignments` - to keep it consistent. It would then be `TRUE` by default.

gsherwood:
Thanks for the PR. I ended up changing the name of the sniff property and have documented it here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#squizwhitespaceoperatorspacing

PR #2522, opened 2019-05-30
Text generator throws error when code sample line is too long

jrfnl:
If an individual line of the code sample in the code comparison would be longer than 47/48 characters, a `Warning: str_repeat(): Second argument has to be greater than or equal to 0` would be thrown.

This fixes the error.

Note: this fix does not introduce code-wrapping as that would probably reduce the readability of the code too much.
Instead, the line will just be longer than other lines.

:bulb: **Idea**: should this class respect a `report_width` if set by the user either in their `CodeSniffer.conf` or via the command-line ?

### DIFF
--- a/src/Generators/Text.php
+++ b/src/Generators/Text.php
@@ -231,9 +231,9 @@ class Text extends Generator
             }
 
             echo '| ';
-            echo $firstLineText.str_repeat(' ', (47 - strlen($firstLineText)));
+            echo $firstLineText.str_repeat(' ', max(0, (47 - strlen($firstLineText))));
             echo '| ';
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for fixing this.

I'm not keen on trying to implement a report width setting for the docs at this stage as I'm still trying to sort out a proper doc strategy, which may be easier to manage if the docs aren't sitting as XML blobs being output the screen. Ideally, the generator feature eventually goes away in favour of using a tool designed for documentation. But I'm not sure yet.

jrfnl:
Thanks for merging this. Even though the future direction may not be clear, at least the current implementation doesn't throw the error anymore.

Let me know if you'd like to brainstorm about the future documentation generation :wink:

PR #2525, opened 2019-06-05
Add support to skip LineLengthSniff on Use lines

lukasvrany:
We have in our project long names of namespaces. These names overlap line length limit on use statement lines. We won't increase limit for all lines so great solution for use is skip use statement lines for line length check like comment lines.


lukasvrany:
> I haven't run any tests on this (yet), but is this intended to only ignore import `use` statements ? If so, please make sure there are tests that check that trait `use` statements and closure `use` statements are not affected.
> 
> FYI: The unit tests are failing on test 6 as the `phpcs:set` directives are not file specific and you are not resetting the property values to their defaults at the end of each unit test case file.
> 
> Also: please undo the `break` indentation changes in the unit test file.

Thanks for feedback.
Yes, this only ignore `use` statements. I forget for traits and closures so I add a test for it and fix it.

merlindiavova:
Anything I can do to help this along? this will be really useful for us

gsherwood:
Sorry I missed this when submitted. Closing as this isn't a feature I want to add - it opens the door to even more exceptions, for which a different solution probably needs to be found (if required).

PR #2527, opened 2019-06-13
Report/XML: bug fix for output on Windows

### LINKED ISSUE #2526: XML report format has bad syntax on Windows
Exactly the same source code of version 3.4.2 with default config with xml output format produces expected results on Linux, but bad results on Windows:

Windows output example:
```
<phpcs version="3.4.2">
xml version="1.0" encoding="UTF-8"?>
<file ...removed... file>
xml version="1.0" encoding="UTF-8"?>
<file ...removed... file>
xml version="1.0" encoding="UTF-8"?>
<file ...removed...
```

Expected/Linux output:
```
<phpcs version="3.4.2">
<file ...removed... file>
<file ...removed... file>
<file ...removed...
```

jrfnl:
It looks like the `XMLWriter` extension does not respect the OS-specific `PHP_EOL` character, but always uses `\n`, though I have not been able to test this on various OS-es to confirm.
This could be regarded as a bug in PHP itself which may or may not be fixed at some point.

With the above in mind, I have created a fix which contains some redundancy, but should be stable on all OS-es, including if/when the PHP native issue would be fixed.

Alternatively, the whole snippet could be replaced by the code on line 81 of the patch.

Fixes #2526

### DIFF
--- a/src/Reports/Xml.php
+++ b/src/Reports/Xml.php
@@ -75,7 +75,12 @@ class Xml implements Report
         // add that manually later. We only have it in here to
         // properly set the encoding.
         $content = $out->flush();
-        $content = substr($content, (strpos($content, PHP_EOL) + strlen(PHP_EOL)));
+        if (strpos($content, PHP_EOL) !== false) {
+            $content = substr($content, (strpos($content, PHP_EOL) + strlen(PHP_EOL)));
+        } else if (strpos($content, "\n") !== false) {
+            $content = substr($content, (strpos($content, "\n") + 1));
+        }
+
         echo $content;
 
         return true;


mvorisek:
As discussed here https://github.com/squizlabs/PHP_CodeSniffer/issues/2526#issuecomment-501745007 I would recommend more robust solution, hovewer I confirm that the issue is here and both ways to fix it seems to be working on Linux/Debian as well as Windows.

gsherwood:
Can confirm no impact on macOS with this change

gsherwood:
Thanks a lot for fixing this

PR #2532, opened 2019-06-16
Squiz.Commenting.FunctionComment.IncorrectTypeHint does not work with class aliases

tstoeckler:
If you are using an aliased class in a typehint, `Squiz.Commenting.FunctionComment.IncorrectTypeHint` will complain even if the correct typehint is actually being used.

### DIFF
--- a/src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
+++ b/src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
@@ -252,6 +252,18 @@ public function typeHint(MyClass $a1, $a2, myclass $a3, $a4)
 }//end typeHint()
 
 
+use MyOtherClass as MyAliasedClass;
+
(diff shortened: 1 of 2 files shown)


tstoeckler:
The PR contains a little addition to the the test suite to prove that this is currently a problem. There is no fix yet, not sure if that's not above my head...

gsherwood:
Class aliases are not something PHPCS tracks, so none of the sniff are going to support them without significant effort.

tstoeckler:
Thanks @gsherwood for the quick response! That's unfortunate, but it's good to know.

What is particular annoying about this is that it's not really easy to disable this particular sniff via `// phpcs:disable` because (depending on the particular ruleset) there is no way to put the `//`-style comment. It would be helpful if I could add something like `phpcs:disable` to the `/**` style function comment, but I think that's not possible, right?

In any case thanks for the info!

PR #2534, opened 2019-06-17
Unresolvable installed_paths can lead to open_basedir errors

ndm2:
Install paths defined via the `installed_paths` configuration option do sometimes contain relative paths that may not exist depending on the current environment.

Resolving such non-existent paths causes `false` to be included in the results, which when used for looking up files/folders, can cause wrong locations to be looked up, and/or errors being thrown. [**`\PHP_CodeSniffer\Util\Standards::getInstalledStandardDetails()`**](https://github.com/squizlabs/PHP_CodeSniffer/blob/3.4.2/src/Util/Standards.php#L85-L88) would for example look for `/ruleset.xml`, ie in the root of the current drive, which could trigger an open basedir restriction error.

### DIFF
--- a/src/Util/Standards.php
+++ b/src/Util/Standards.php
@@ -16,7 +16,9 @@ class Standards
 
 
     /**
-     * Get a list paths where standards are installed.
+     * Get a list of paths where standards are installed.
+     *
+     * Unresolvable relative paths will be excluded from the results.
      *
      * @return array
      */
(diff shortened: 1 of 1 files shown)


jrfnl:
IMO this is a valid fix.

I wonder: should a notice be shown to the user about unresolved paths to help debugging ?

ndm2:
@jrfnl In my experience, I've mostly seen possibly non-existent paths when additional locations are being defined as a fallback. So I'm not sure, looking at for example `getInstalledStandardDetails()`, non-existent/accessible standard paths are deliberately silently stepped over, introducing notices would break with that... or are you maybe referring to output in verbose mode?

As far as I can tell, non existent paths would always finally trigger an error when a non existent sniff is referenced, one would then receive a message like:

> ERROR: Referenced sniff "XYZ" does not exist

gsherwood:
Fix looked good to me too. Thanks a lot for finding and fixing this.

PR #2548, opened 2019-07-10
New Generic.WhiteSpace.SpreadOperatorSpacingAfter sniff (PSR12 related)

jrfnl:
This introduces a new sniff which checks the whitespace between the spread operator and the variable/function call it applies to.

This new sniff has two `public` properties:
* `$spacing` which defaults to `0`, i.e. no space.
* `$ignoreNewlines` which defaults to `false`.

Includes unit tests.
Includes fixers.
Includes documentation.

~~Includes adding this sniff to the PSR12 ruleset.~~

Note: PSR12 only specifies _There MUST NOT be a space between the variadic three dot operator and the argument name_ for **function declarations**.

This sniff, however, checks the spacing for the spread operator in all circumstances in which it can be used, i.e.:
* In function declarations (PHP 5.6).
* In function calls (PHP 5.6).
* In array declarations (PHP 7.4).

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -407,6 +407,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="DisallowSpaceIndentStandard.xml" role="php" />
         <file baseinstalldir="PHP/CodeSniffer" name="DisallowTabIndentStandard.xml" role="php" />
         <file baseinstalldir="PHP/CodeSniffer" name="ScopeIndentStandard.xml" role="php" />
+        <file baseinstalldir="PHP/CodeSniffer" name="SpreadOperatorSpacingAfterStandard.xml" role="php" />
        </dir>
       </dir>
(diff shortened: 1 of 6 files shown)


gsherwood:
I'm not sure I can actually include this in PSR12 because people tend to get angry when PHPCS enforces rules that the standard doesn't explicitly describe. But it's a very useful sniff regardless.

jrfnl:
Yes, I noticed you'd added a different sniff for PSR12. Shall I remove the PSR12 ruleset change from the PR so this can be merged for anyone who does want a more extensive check for the spread operator ?

gsherwood:
> Shall I remove the PSR12 ruleset change from the PR so this can be merged for anyone who does want a more extensive check for the spread operator ?

Yes please, if you have time.

jrfnl:
Easy-peasy. Done.

gsherwood:
Thanks a lot for making that change.

PR #2559, opened 2019-07-20
Squiz/OperatorSpacing: fix potential fixer conflict

jrfnl:
[Fixer conflict series]

The sniff had a potential fixer conflict with the `SuperfluousWhiteSpace` and the `EndFileNewline` sniffs.

This minor change fixes this by bowing out from analysing the space after the operator if the operator is the last non-whitespace token in a file.

Fixer conflict can be reproduced by running the below command over the unit test added by this patch (without the included fix):
`phpcbf ./squiz/tests/whitespace/operatorspacingunittest.inc --standard=Squiz -vv`

Includes unit test.


gsherwood:
Thanks

PR #2560, opened 2019-07-21
Replace usage of curly braces for indexing with square brackets

### LINKED ISSUE #2558: PHP 7.4 throwing offset syntax with curly braces is deprecated message
Hello!

https://github.com/squizlabs/PHP_CodeSniffer/blob/f6fd84859659ab25d74646e7eff7397f009e258e/src/Config.php#L420

Current version of PHP 7.4 (`PHP 7.4.0-dev (cli) (built: Jul 19 2019 10:22:49) ( ZTS )`) considers practice of using curly braces to address offset of a string/array as deprecated. PHPCS still works okay now but emits warnings: 
```
Deprecated: Array and string offset access syntax with curly braces is deprecated in /my-project-path/vendor/squizlabs/php_codesniffer/src/Config.php on line 420
```
Changing brackets to square ones in `src/Config.php` should solve the problem.

MasterOdin:
Closes #2558 

As stated in the issue, PHP 7.4 has deprecated curly braces for string/array access (Accepted [RFC Proposal](https://wiki.php.net/rfc/deprecate_curly_braces_array_access)). This replaces all usages in the codebase with the standard square brackets.


gsherwood:
Thanks for this, but the removal of curly braces from test files is actually removing test cases. Some of these tests ensure the sniffs (and tokenizer) don't get confused by curly braces being used for array access, so they need to remain in most of the cases. It's probably safer to just revert all changes to test files and leave the PR for the core changes, which I will very happily accept.

MasterOdin:
Ah, I did not realize that with the tests. Reverted all changes to the test files and left only the source code changes.

src/Tokenizers/Tokenizer.php
```diff
@@ -1255,7 +1255,7 @@ private function recurseScopeMap($stackPtr, $depth=1, &$ignore=0)
                         $ignore++;
                     } else {
                         // Make sure this is actually an opener and not a
-                        // string offset (e.g., $var{0}).
+                        // string offset (e.g., $var[0]).
```
jrfnl: This comment should be left unchanged.

MasterOdin: Reverted this change.

MasterOdin:
Thanks for pointing out those additional cases I missed with my original regex search. Fixed these as well.

gsherwood:
Thanks a lot for making those changes, and for the PR itself. I appreciate it.

PR #2561, opened 2019-07-21
PHP 7.4 compatibility fix / implode argument order

jrfnl:
`implode()` takes two parameters, `$glue` and `$pieces`.
For historical reasons, `implode()` accepted these parameters in either order, though it was recommended to use the documented argument order of `implode( $glue, $pieces )`.

PHP 7.4 is slated to deprecate the tolerance for passing the parameters for `implode()` in reverse order.
PHP 8.0 is expected to remove the tolerance for this completely.

Refs:
* https://wiki.php.net/rfc/deprecations_php_7_4#implode_parameter_order_mix
* https://php.net/manual/en/function.implode.php

These instances were discovered while testing a new PHPCompatibility sniff to detect this issue.

### DIFF
--- a/src/Reports/Gitblame.php
+++ b/src/Reports/Gitblame.php
@@ -51,7 +51,7 @@ class Gitblame extends VersionControl
         }
 
         $parts  = array_slice($parts, 0, (count($parts) - 2));
-        $author = preg_replace('|\(|', '', implode($parts, ' '));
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks a lot for fixing these.

jrfnl:
You're welcome ;-)

PR #2567, opened 2019-07-24
Generic/InlineControlStructure: bug fix

### LINKED ISSUE #2565: Generic.ControlStructures.InlineControlStructure confused by mixed short/long tags
Example:
```php
<?php if (true) : ?>
<? endif ?>
<?php if (true) { ?>
<? } ?>

<? if (false) : ?>
<? endif ?>
<?php if (false) { ?>
<?php } ?>
```

CLI output (reduced for brevity):
```
$ phpcs -n --standard=PSR2 phpcs-test.php
  1 | ERROR | [x] Inline control structures are not allowed
  3 | ERROR | [x] Inline control structures are not allowed
$ phpcbf -n --standard=PSR2 phpcs-test.php
  /home/me/phpcs-test.php                            FAILED TO FIX
```

Resulting "fixed" file:
```php
<?php if (true) {
    :;
} ?>
<? endif ?>
<?php if (true) { { { { { { { { { { { { { { { { { { { { { { { { { { { {;}}}}}}}}}}}}}}}}}}}}}}}}}} } ?>
<? } ?>

<? if (false) : ?>
<? endif ?>
<?php if (false) { ?>
<?php } ?>
```

As can be seen in the `false` if's, using `<?php` or `<?` on both lines do not trigger the phpcs error. I'm not so sure what `InlineControlStructure` means in those cases (maybe the humanized message doesn't make much sense for those cases?), but we can say for sure that `phpcbf` got lost with the alt-syntax and nuts with the bracketed one :joy: 

That's for versions 3.3.2/3.4.2 at PHP 7.2.20.

jrfnl:
Prevent creating incorrect fixes and getting into a fixer conflict when both short and long open tags are used in combination with alternative control structures and the short open tags are not recognized because the system on which the scan is run has `short_open_tag=Off`.

Solved by bowing out early when an alternative control structure is detected which was not recognized as such (`scope_opener` not set).

Fixes #2565

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -503,6 +503,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="InlineControlStructureUnitTest.4.inc" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="InlineControlStructureUnitTest.5.inc" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="InlineControlStructureUnitTest.6.inc" role="test" />
+        <file baseinstalldir="PHP/CodeSniffer" name="InlineControlStructureUnitTest.7.inc" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="InlineControlStructureUnitTest.js" role="test" />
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks for looking into this, and fixing it.

jrfnl:
You're welcome. Sorry for the forgotten `var_dump()` :fearful:

PR #2580, opened 2019-08-11
Fix capitalization of getFilename function

MasterOdin:
s/getFileName/getFilename

I assume these two are trying to reference this function (like the other instances of the correct spelling/capitalization):
https://github.com/squizlabs/PHP_CodeSniffer/blob/a7403540490022691559380cfc8bfc1342890998/src/Files/File.php#L1212-L1216


MasterOdin:
While true, my editor (VS Code) does a lot better with Intellisense when the spelling/case match the declaration. 😄

gsherwood:
Although it technically doesn't matter, I like clean code :) Thanks.

Issue #2581, opened 2019-08-11
README.md: A lot of good installation instructions, and no instructions for actual use.


tchalvak:
I would recommend adding a section to the README for instructions on a good example command usage.  For instance, a trivial usage of `phpcs ./` just fatal errors on memory size on a clean wordpress install.  Without a clear, simple starting usage suggestion, it's much more difficult to determine whether individual configuration is the problem or otherwise debug.

dingo-d:
There is a [wiki](https://github.com/squizlabs/PHP_CodeSniffer/wiki) filled with examples and explanations 🙂

gsherwood:
@tchalvak is the stuff in the wiki what you were looking for in terms of a basic example of usage?

There isn't going to be any documentation about how to diagnose a fatal memory error in the docs though.

tchalvak:
Yeah, the usage page there was pretty much what I was looking for.  I think there could be some major value in clarifying the expected usage of the command (very very briefly) in the README before recommending going to a second site for that.  Would y'all be open to a trivial PR to that effect?

gsherwood:
> I think there could be some major value in clarifying the expected usage of the command (very very briefly) in the README before recommending going to a second site for that

Yeah, I'd be interested in seeing what you recommend here. Something basic that doesn't need to change much between PHPCS releases would probably be quite helpful. Then a link off to more documentation.

tchalvak:
@gsherwood Created a PR for an example of the trivial change to the README that I mean.
https://github.com/squizlabs/PHP_CodeSniffer/pull/2583

PR #2583, opened 2019-08-13
Add simple usage section to readme.

tchalvak:
To simplify running through the README install instructions and give a slight recommendation for use immediately after install.

Trivial readme change based on discussion in: https://github.com/squizlabs/PHP_CodeSniffer/issues/2581


gsherwood:
Thanks for submitting this. I decided to copy some of the info from the wiki into there but still keep it really short.  I'd love to know what you think.

gsherwood:
Thanks for reviewing the changes.

PR #2585, opened 2019-08-14
Squiz/FunctionSpacing + MemberVarSpacing: add metrics

jrfnl:
For repos not yet using these sniffs, having metrics available will be useful to find out what the prevalent number of blank lines between properties/functions is.

This will make it more easy to determine what to set for the value of the various properties available.


gsherwood:
Thanks for these metrics. Helped me notice that I forget the member var spacing rule in the project ruleset as well.

PR #2589, opened 2019-08-21
Allow newlines in control structures (PSR12)

### LINKED ISSUE #2113: PSR12: Support multiple lines control structure spacing sniffs
Following the PSR12 :

> Expressions in parentheses MAY be split across multiple lines, where each subsequent line is indented at least once. When doing so, the first condition MUST be on the next line. The closing parenthesis and opening brace MUST be placed together on their own line with one space between them. Boolean operators between conditions MUST always be at the beginning or at the end of the line, not a mix of both.

```php
<?php

if (
    $expr1
    && $expr2
) {
    // if body
} elseif (
    $expr3
    && $expr4
) {
    // elseif body
}
```

The current PSR12 ruleset will show the following error code : `PSR2.ControlStructures.ControlStructureSpacing.SpacingAfterOpenBrace`

The expected behavior is that no error is shown for the example code.

scheb:
I'd like to suggest this solution to fix #2113.

Adds properties to configure sniffs `Squiz.ControlStructures.ForLoopDeclaration` and `PSR2.ControlStructures.ControlStructureSpacing` to allow newlines after/before parentheses according to PSR12:

> Expressions in parentheses MAY be split across multiple lines, where each subsequent line is indented at least once. When doing so, the first condition MUST be on the next line. The closing parenthesis and opening brace MUST be placed together on their own line with one space between them. Boolean operators between conditions MUST always be at the beginning or at the end of the line, not a mix of both.

These properties are enabled in the PSR12 `ruleset.xml`.

### DIFF
--- a/src/Standards/PSR12/ruleset.xml
+++ b/src/Standards/PSR12/ruleset.xml
@@ -233,7 +233,11 @@
     <rule ref="Squiz.WhiteSpace.ControlStructureSpacing.SpacingBeforeClose"/>
     <rule ref="Squiz.WhiteSpace.ScopeClosingBrace"/>
     <rule ref="Squiz.ControlStructures.ForEachLoopDeclaration"/>
-    <rule ref="Squiz.ControlStructures.ForLoopDeclaration"/>
+    <rule ref="Squiz.ControlStructures.ForLoopDeclaration">
+        <properties>
+            <property name="ignoreNewline" value="true"/>
+        </properties>
+    </rule>
     <rule ref="Squiz.ControlStructures.ForLoopDeclaration.SpacingAfterOpen">
         <severity>0</severity>
     </rule>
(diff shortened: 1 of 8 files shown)


gsherwood:
Solving the problem with settings might work, but the setting to ignore newlines is normally called `ignoreNewlines`. See an example here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#squizwhitespaceoperatorspacing

Using the new settings in your PR also causes the sniff to `return` when a new line is found, which skips processing the rest of the sniff. This would need to be changed.

I need to take another look at the PSR2 sniffs to see if it is still worth using those for PSR12 or if a new one should be written. It might be possible to refactor and extend the PSR2 sniff as well, instead of trying to use properties to change it.

scheb:
Did some changes according to your feedback.

But I see your point with having a sniff that checks PSR12 standard and nothing more, instead of changing the behavoir of the other sniffs via properties. Your call if we should continue this way or the other.

gsherwood:
I ended up writing a new sniff to check the PSR-12 rules directly instead of trying to modify the PSR-2 sniff. The new PSR-12 sniff uses the PSR-2 sniff when it detects that a single line definition is being used. Otherwise, it implements the set of checks outlined in the PSR-12 standard, except for some checks handled by other sniffs.

Thanks a lot for the PR though. It just turned out that the PSR-12 one need to be more specific anyway, so a new sniff was required even if this feature was added.

PR #2590, opened 2019-08-23
Constant visibility MUST be an error

gmponos:
According to PSR this is a MUST hence it MUST be an error.

### DIFF
--- a/src/Standards/PSR12/Sniffs/Properties/ConstantVisibilitySniff.php
+++ b/src/Standards/PSR12/Sniffs/Properties/ConstantVisibilitySniff.php
@@ -40,6 +40,10 @@ class ConstantVisibilitySniff implements Sniff
      */
     public function process(File $phpcsFile, $stackPtr)
     {
+        if (PHP_VERSION_ID < 70100) {
+            return;
+        }
+
         $tokens = $phpcsFile->getTokens();
         $prev   = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($stackPtr - 1), null, true);
         if (isset(Tokens::$scopeModifiers[$tokens[$prev]['code']]) === true) {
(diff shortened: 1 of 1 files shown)


gmponos:
From PSR-12:

> Visibility MUST be declared on all constants if your project PHP minimum version supports constant visibilities (PHP 7.1 or later).

gsherwood:
From the changelog:
> Uses a warning instead of an error due to this conditionally requiring the project to support PHP 7.1+

I can't make this an error because if a project does not support PHP 7.1 then it can never pass PSR-12 validation.

gmponos:
Sorry haven't read the changelong.. never the less I thought that this might be the reason you had it as an error and  that's why I added the check for the PHP version...

WDYT?

gsherwood:
> that's why I added the check for the PHP version

Sorry, forgot to talk about that bit.

I was going to do the same thing, but the fact that someone is using PHP 7.1 to run PHPCS doesn't mean that the code base itself is for a project supporting PHP 7.1.

Similarly, if someone is running PHPCS using PHP 7, it doesn't mean that the check should be ignored.

There is no way to determine what version of PHP a project supports, so there is nothing the sniff can do besides throw it up as a warning and let the developer decide what to do.

But it is trivial to turn the warning into an error using a custom ruleset, so projects can do this if they want to enforce this as an error.

gmponos:
> But it is trivial to turn the warning into an error using a custom ruleset, so projects can do this if they want to enforce this as an error.

Same thing could be said for ignoring the rule :)

Anyway your points are taken.. if you rethink it let me know.. closing for now :)

gsherwood:
> Same thing could be said for ignoring the rule :)

True. It could go either way. I was just anticipating the issues that would be opened when the default position of PHPCS is to produce an error with a specific version number in it. Hope it makes sense.

PR #2596, opened 2019-08-29
Fix tokenizer scope resolution for inline statements followed by an anonymous block

ht-bs:
A case like the following was not picked up by the InlineControlStructureSniff because the scope resolution in the tokenizer fails to recognize that the curly braces are not related to the foreach.

```
foreach ([] as $a)
echo 'bar';
{
    echo 'baz';
} 
```

Since all tests are working I expect that this is not a breaking change.


gsherwood:
Thanks for reporting this, but the fix needed to be made in the tokenizer itself as the incorrectly assigned curly braces cause issues for other sniffs as well. I've committed a fix for the core issue.

Thanks a lot for reporting the issue. I used your code sample in the unit tests.

PR #2601, opened 2019-09-05
Squiz.WhiteSpace.FunctionSpacing incorrect fix when spacing is 0

JanGalek:
Fix this usecase:

Settings
------------
spacing: 0
spacingBeforeFirst: 0
spacingAfterLast: 0

Code:
```php
<?php
declare(strict_types = 1);

namespace Nette\CodingStandard\Examples;

interface OneBlankLineBeforeFirstFunctionClassInterface
{


    /** @return mixed */
    public function interfaceMethod();
}
```

it was returning this:
```php
<?php
declare(strict_types = 1);

namespace Nette\CodingStandard\Examples;

interface OneBlankLineBeforeFirstFunctionClassInterface
{/** @return mixed */
    public function interfaceMethod();
}
```

it return now correct this:

```php
<?php
declare(strict_types = 1);

namespace Nette\CodingStandard\Examples;

interface OneBlankLineBeforeFirstFunctionClassInterface
{
    /** @return mixed */
    public function interfaceMethod();
}
```

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/FunctionSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/FunctionSpacingSniff.php
@@ -310,7 +310,7 @@ class FunctionSpacingSniff implements Sniff
                 if ($nextSpace === false) {
                     $nextSpace = ($stackPtr - 1);
                 }
-                
+
                 if ($requiredSpacing === 0) {
                     $nextSpace++;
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks a lot for this report and PR. I ended up fixing this is a slightly different way so that it reflects how many of the other sniffs solve this problem, which is just to ignore any content of the same line as prevContent.

Issue #2609, opened 2019-09-19
PSR12.Classes.ClassDeclaration allows extended classname on new line


wietsewarendorff:
Question in somewhat similiar to https://github.com/squizlabs/PHP_CodeSniffer/issues/2298. Somehow this seems valid, while I think it should not be (note the newline between the word `class` and the actual classname):

```
<?php

class
Foo
{
}
```

I think the word `class` and the classname should be on the same line.

gsherwood:
Using the PSR12 standard on that code shows the error `Expected 1 space after class keyword; newline found (PSR2.Classes.ClassDeclaration.SpaceAfterKeyword)`. Note that it comes from PSR2's sniff because PSR12 uses PSR2 sniffs where possible, just as PSR2 uses PSR1 sniffs. They each build on each other rather than duplicating the code.

wietsewarendorff:
Took me some time to figure it out: the issue is indeed fixed in current master-branch, but the check fails on the current tagged version (3.4.2). I can make a pull request to add a test for this if you like (which will passes automatically on the current master branch) because this specific scenario is not yet implemented (looking [here](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/PSR2/Tests/Classes/ClassDeclarationUnitTest.inc)).

gsherwood:
> I can make a pull request to add a test for this if you like

Sure :)

It does look like the case isn't covered by those tests, so they will need updating. Thanks for looking into it.

wietsewarendorff:
No problem, thanks for your efforts in managing this project!

Merge request created: https://github.com/squizlabs/PHP_CodeSniffer/pull/2611

gsherwood:
Thanks for the new test. Closing this now as this case is finally covered.

PR #2611, opened 2019-09-20
Add extra testcase for PSR2.Classes.ClassDeclaration.SpaceAfterKeyword

wietsewarendorff:
MR for an extra testcase, as mentioned in #2609


gsherwood:
Thanks for doing that.

Issue #2616, opened 2019-09-27
PSR12.Files.FileHeader false positive when file only contains docblock


dereuromark:
```
FOUND 1 ERROR AFFECTING 1 LINE
---------------------------------------------------------------------------------------------------------------------
 5 | ERROR | [x] Header blocks must be followed by a single blank line
```
is unfixable.
It doesn't respect a file that only contains a docblock.

code is:
```php
<?php

/**
 * Here go your local configs
 */
```

morrisonlevi:
I think Datadog/dd-trace-php is hitting this same issue for a different reason in [branch levi/phpcs](https://github.com/DataDog/dd-trace-php/tree/3e81033a99b09188472b2128a7ffe6c6c7211710):
```bash
circleci@82b8c67c4988:~/app$ composer lint
> phpcs -s --ignore=src/DDTrace/Integrations/ZendFramework/V1/Ddtrace.php

FILE: /home/circleci/app/tests/Unit/Util/TryCatchFinallyTest.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 226 | ERROR | [x] Header blocks must be followed by a single blank
     |       |     line
     |       |     (PSR12.Files.FileHeader.SpacingAfterBlock)
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------

Time: 39.58 secs; Memory: 20.25MB

Script phpcs -s --ignore=src/DDTrace/Integrations/ZendFramework/V1/Ddtrace.php handling the lint event returned with error code 2
circleci@82b8c67c4988:~/app$ composer fix-lint
> phpcbf

PHPCBF RESULT SUMMARY
----------------------------------------------------------------------
FILE                                                  FIXED  REMAINING
----------------------------------------------------------------------
...cleci/app/tests/Unit/Util/TryCatchFinallyTest.php  FAILED TO FIX
----------------------------------------------------------------------
A TOTAL OF 0 ERRORS WERE FIXED IN 1 FILE
----------------------------------------------------------------------
PHPCBF FAILED TO FIX 1 FILE
----------------------------------------------------------------------

Time: 37.31 secs; Memory: 20.25MB


Script phpcbf handling the fix-lint event returned with error code 2
```

gsherwood:
[...]

PR #2617, opened 2019-09-27
[PSR12] Add failing test for valid code

duncan3dc:
More if an issue than a pull request sorry.

The following code is valid as per PSR-12, however phpcs reports the following issues:
```
------------------------------------------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 2 LINES
------------------------------------------------------------------------------------------------------
 2 | ERROR | [ ] The file header must be the first content in the file
 6 | ERROR | [ ] The file-level docblock must follow the opening PHP tag in the file header
 6 | ERROR | [x] Header blocks must be followed by a single blank line
```

It recognises the inline docblock as a file header, which is incorrect. The fix for this is simple, however that affected the check for leading blank lines in `FileHeaderUnitTest.2.inc`.

The code for the opening PHP tag also doesn't seem to be correct, PSR-12 only talks about mixed HTML and PHP files, it doesn't require the opening tag to be the first thing in the file in all scenarios (eg the hashbang seen here is valid)

### DIFF
--- a/src/Standards/PSR12/Tests/Files/FileHeaderUnitTest.7.inc
+++ b/src/Standards/PSR12/Tests/Files/FileHeaderUnitTest.7.inc
@@ -1,0 +1,7 @@
+#!/usr/bin/env php
+<?php
+
+use Psr\Container\ContainerInterface;
+
+/** @var ContainerInterface $container */
+$container = require __DIR__ . "/../bootstrap.php";


dereuromark:
Refs https://github.com/squizlabs/PHP_CodeSniffer/issues/2616

gsherwood:
> The code for the opening PHP tag also doesn't seem to be correct, PSR-12 only talks about mixed HTML and PHP files, it doesn't require the opening tag to be the first thing in the file in all scenarios (eg the hashbang seen here is valid)

To PHP, everything outside PHP tags is T_INLINE_HTML. I can't really detect that the content is actually HTML without trying to parse it. It could very well be XML, JSON, YAML, whatever. PSR-12 still wants the file header up the top in these cases.

What I will try and do is just allow hashbang lines as these are very obviously not HTML and have to be up top.

The other issues in this PR is being addressed by issue #2626.

gsherwood:
I've allowed the hashbang line and also fixed the issue with the @var comment. I'm closing this as your sample code now passes, and I've included it in a test even though the PR wouldn't merge due to other changes.

Thanks a lot for this report.

PR #2618, opened 2019-09-27
Allow files from stdin AND file list (config files or commandline args)

samifruit514:
Everything is in the title :)

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -356,6 +356,7 @@ class Runner
             include $bootstrap;
         }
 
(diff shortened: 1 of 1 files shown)


gsherwood:
Out of interest, what did you need this change for?

The reason I ask is that using STDIN disables the cache, so it is potentially faster to do 2 PHPCS runs rather than try and process both together.

floverdevel:
@samifruit514 I think that it could be interesting to tell us what use case you want to accomplish that needs this change.

PR #2622, opened 2019-09-30
Allow trailing comma in multi line function calls

### LINKED ISSUE #2621: PSR12.Classes.AnonClassDeclaration.CloseBraceSameLine false positive for anon class passed as function argument
When using anonymous class the following should be allowed, currently it is reported as error.

```
Closing class brace must be on a line by itself
(PSR12.Classes.AnonClassDeclaration.CloseBraceSameLine)
```

```php
$foo->bar(
    new class implements Bar {
        // ...
    },
);
```

martinssipenko:
Fixes #2621


gsherwood:
Thanks a lot for the PR - worked well

Issue #1657, opened 2017-09-15
"Uninitialized string offset: 0" when sniffing CSS


h3xx:
## Steps to reproduce

Invoke phpcs on the test code:

```
phpcs --standard=Squiz -s --sniffs=Squiz.CSS.DuplicateClassDefinition [FILE]
```

## Test code

```
.foo /* any comment */
{ color: red; }
```
## Produces output

```
FILE: test.css
---------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
---------------------------------------------------------------------------------
 1 | ERROR | An error occurred during processing; checking has been aborted. The
   |       | error message was: Uninitialized string offset: 0 in
   |       | /.../git/PHP_CodeSniffer/src/Standards/Squiz/Sniffs/CSS/DuplicateClassDefinitionSniff.php
   |       | on line 98 (Internal.Exception)
---------------------------------------------------------------------------------

Time: 77ms; Memory: 8Mb
```
## Expected output

No output. No duplicate class definitions should be found.

Found in phpcs version 276f561764e13d869a14b0093111548bb8b00b62 (2017-09-15)

Note: The amount of whitespace surrounding the comment doesn't seem to affect whether this happens. Can be on the same line.

h3xx:
Note: The amount of whitespace surrounding the comment doesn't seem to affect whether this happens. Can be on the same line.

gsherwood:
I'm not sure why the sniff was looking for comments (there were no tests explaining it) but removing that check makes this work. Thanks for the report.

h3xx:
Confirmed fixed in `master`. Thanks!

PR #2639, opened 2019-10-06
Update Squiz.CSS.DuplicateClassDefinitionSniff

### LINKED ISSUE #2638: Squiz.CSS.DuplicateClassDefinitionSniff sees comments as part of the class name
This sniff threw false-positives in my code. After #985 was fixed, they disappeared. After the fix of #1657 they rose again.

The difference between my code and the same in #1657 is, that I have the comments before the class name:
`/* any comment */
.foo
{ color: red; }`

rhorber:
Remove comments from class names, before searching for `@media` and checking for duplicates.

This closes #2638.
The fix of #1657 is still in place.


gsherwood:
The test cases in this PR pass without any modification to the sniff itself, so they are not replicating the problem that the PR aims to solve. I will take a look into them a bit more and see if I can figure out what's going on.

gsherwood:
Ah, I see what went wrong. The comments in the test file are all the same, so they always look like they match. I will change them to ensure the current code sees them as different classes.

gsherwood:
Thanks for the PR. Fix worked well after the tests were changed.

PR #2640, opened 2019-10-07
Squiz.WhiteSpace.OperatorSpacing false positives for some negation operators

grongor:
Not all tokens were handled, so I added them + refactored the code a bit.

I also added a bunch of test code to test all the possible (hopefully) variants.

On a side note: I think that the "fix tests" are somewhat broken. I would expect that the code I added won't be altered at all during the fixing process (because the sniffs should ignore it completely) but it seems that some other sniffs "fixes it" - is that right? Is that expected behaviour?

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
@@ -9,8 +9,8 @@
 
 namespace PHP_CodeSniffer\Standards\Squiz\Sniffs\WhiteSpace;
 
-use PHP_CodeSniffer\Sniffs\Sniff;
 use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
 use PHP_CodeSniffer\Util\Tokens;
 
 class OperatorSpacingSniff implements Sniff
@@ -42,6 +42,13 @@ class OperatorSpacingSniff implements Sniff
      */
     public $ignoreSpacingBeforeAssignments = true;
 
+    /**
+     * A list of tokens that aren't considered as operands.
+     *
+     * @var string[]
+     */
+    private $nonOperandTokens = [];
+
 
     /**
      * Returns an array of tokens this test wants to listen for.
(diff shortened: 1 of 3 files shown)


src/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
```diff
@@ -57,6 +64,45 @@ public function register()
...
+        $this->nonOperandTokens = array_merge(
+            // Trying to operate on a negative value; eg. ($var * -1).
+            Tokens::$operators,
+            // Trying to compare a negative value; eg. ($var === -1).
+            Tokens::$comparisonTokens,
+            // Trying to compare a negative value; eg. ($var || -1 === $b).
+            Tokens::$booleanOperators,
+            // Trying to assign a negative value; eg. ($var = -1).
+            Tokens::$assignmentTokens,
+            [
```
jrfnl: Please set this up with the tokens as the array keys, so you can use `isset()` instead of `in_array()` below.

jrfnl:
> On a side note: I think that the "fix tests" are somewhat broken. I would expect that the code I added won't be altered at all during the fixing process (because the sniffs should ignore it completely) but it seems that some other sniffs "fixes it" - is that right? Is that expected behaviour?

The tests only consider this sniff and this sniff alone. If the fixer is being invoked on code where it shouldn't be, I suspect there's still something wrong with this PR.

src/Standards/Squiz/Tests/WhiteSpace/OperatorSpacingUnitTest.inc
```diff
@@ -129,11 +129,11 @@ $a = $x?$y:$z;
...
 
 $y = 1
-   + 2  
+   + 2
    - 3;
 
-$y = 1 + 
-     2 - 
+$y = 1 +
+     2 -
```
jrfnl: Please revert the whitespace changes here. IIRC, the trailing spaces are testing a specific situation with the fixer.

src/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
grongor: I don't think that this is necessary, but sure, I'll use `T_TOKEN => T_TOKEN` style, as used in the `Tokens` class.

src/Standards/Squiz/Tests/WhiteSpace/OperatorSpacingUnitTest.inc
grongor: Sure, sorry, Phpstorm was too proactive :)

grongor:
> The tests only consider this sniff and this sniff alone. If the fixer is being invoked on code where it shouldn't be, I suspect there's still something wrong with this PR.

Ok, I'll look into it in more detail ...

grongor:
@jrfnl you were, there was a bug :) thx ... now it should be alright :)

gsherwood:
Thanks a lot for these changes. This fix will be in 3.5.3.

PR #2651, opened 2019-10-16
'Undefined offset: -1' if findPrevious '$end = false'

babeuloula:
Hi,

If you try to find a previous token with `$end = false` you've got this error :

```
----------------------------------------------------------------------------------------------------------
----------------------------------------------------------------------------------------------------------
 1 | ERROR | An error occurred during processing; checking has been aborted. The error message was:
   |       | Undefined offset: -1 in
   |       | /home/babeuloula/***/vendor/squizlabs/php_codesniffer/src/Files/File.php on line 2123
----------------------------------------------------------------------------------------------------------
```

My code :

```php
$findPreviousPtr = $phpcsFile->findPrevious(
    [T_USE, T_OPEN_USE_GROUP],
    $stackPtr,
    $phpcsFile->findPrevious([T_OPEN_CURLY_BRACKET], $stackPtr)
);
```

Because `$phpcsFile->findPrevious([T_OPEN_CURLY_BRACKET], $stackPtr)` return `false` if he did not find any token.

I have not write a unit test because I don't understand how you write your tests. And I have find anything on your documentation.

Have a nice day.

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -2113,7 +2113,7 @@ class File
     ) {
         $types = (array) $types;
 
-        if ($end === null) {
+        if ($end === null || \is_bool($end) === true) {
             $end = 0;
         }
 
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -2100,6 +2100,8 @@ class File
      *                                  will not be checked. IE. checking will stop
      *                                  at the previous semi-colon found.
      *
+     * @thorw \InvalidArgumentException
+     *
      * @return int|bool
      * @see    findNext()
      */
@@ -2113,8 +2115,11 @@ class File
     ) {
         $types = (array) $types;
 
-        if ($end === null || \is_bool($end) === true) {
+        if ($end === null) {
             $end = 0;
+        } elseif (\is_bool($end) === true) {
+            throw new \InvalidArgumentException('$end must be a boolean. 
+                Maybe you\'ve used findPrevious and he did not found anything.');
         }
 
         for ($i = $start; $i >= $end; $i--) {


jrfnl:
The `false` response is documented and passing `false` to the function is not an allowed variable type for `$end`. So nesting `find...()` calls is just not a good idea.

Try this:
```
$openCurly = $phpcsFile->findPrevious([T_OPEN_CURLY_BRACKET], $stackPtr);
if ($openCurly === false) {
    return;
}
$findPreviousPtr = $phpcsFile->findPrevious(
    [T_USE, T_OPEN_USE_GROUP],
    $stackPtr,
    $openCurly
);
```

babeuloula:
@jrfnl That what I've do, but without strict type on the method we can make the mistake. It would be nice to prevent this kind of error.

jrfnl:
@babeuloula As it's a mistake, I'd say it's a good thing you're getting an error to alert you to your mistake...

gsherwood:
I don't really mind this change, but merging it in would mean that passing FALSE to $end now needs to be a documented behaviour of the method. This make it a lot harder to be strict with the passed types in the future.

Once PHP 7 is the min required PHP version for PHPCS (probably from v4 I'd say) then I can type hint this with a nullable integer. For now, I'd be happy to accept a PR that enforced that manually, but not a PR that silently allows booleans.

VincentLanglet:
I don't think this kind of check should be added. Waiting for php 7 seems the best.

Why throwing an error if `$end` is a bool, but not if `$end` is a string, a function, or something else ? 
Why doing this check for `$end` but not for `$start`, `$exclude`, ... ?
Why for the `findPrevious` method but not the `find` method or every other methods ?

PR #2656, opened 2019-10-17
Squiz.WhiteSpace.MemberVarSpacing removes comments before first member var during auto fixing

jrfnl:
Squiz/MemberVarSpacing: bug fix - don't remove comments before first member var

If there would be comments (or commented out code) before the first member variable, the fixer as it was, would remove those comments, including potential PHPCS annotations.

By allowing for other comments between the class opener/trait import `use` statement and a property, this is prevented.

The fix now implemented means that if there is anything between the class opener/trait import `use` and the first property, other than a property docblock/comment, the property will be treated as any property and won't get the special "first property" treatment.

Includes unit test.


gsherwood:
Thanks a lot for this fix.

PR #2657, opened 2019-10-17
Squiz.WhiteSpace.FunctionSpacing can remove spaces between comment and first/last method during auto-fixing

jrfnl:
Similar to #2656

If there would be comments (or commented out code) before the first method or after the last method in an OO structure, the fixer as it was, would - depending on the various `$spacing` settings, potentially remove the space between the method and the comment which would easily conflict with other comment rules.

By allowing for comments before the first method/after the last method, this is prevented.

The fix now implemented means that if there is anything between the class opener/trait import `use` and the first method, other than a function docblock/comment, the method will be treated as any method and won't get the special "first method" treatment.
Similarly, when there is anything between the last method and the class closer, other than a trailing comment, the method will be treated as any method and won't get the special "last" method treatment.

Includes unit test.


gsherwood:
Thanks

jrfnl:
You're welcome

PR #2660, opened 2019-10-21
Clarify that blocks should be separated rather than followed by a blank line

duncan3dc:
When dealing with the following code:
```php
<?php
namespace Example\Project;
```
The error message `Header blocks must be followed by a single blank line` was confusing to me.
I realise that PSR-12 calls `<?php` a block, but intuitively it's not, so this PR aims to make the message easier to understand in this situation


gsherwood:
Really good change - especially given the message now uses the wording from the standard itself. Thanks for the PR.

PR #2661, opened 2019-10-21
Correct the return type of process() when jumping to the end of the tokens

duncan3dc:
While working on #2660 I noticed that the declared `@return` for the `process()` method was not correct, I did a quick search for find any other instances where `$phpcsFile->numTokens` is returned and this PR corrects them all


gsherwood:
Thanks for spotting these and sending the PR.

PR #2689, opened 2019-11-13
Add "insteadof" and "as" to Squiz\WhiteSpace\OperatorSpacing

michalbundyra:
As we have added `instanceof` we should probably have also `insteadof` and `as` here.

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
@@ -118,6 +118,8 @@ class OperatorSpacingSniff implements Sniff
         $targets[] = T_INLINE_THEN;
         $targets[] = T_INLINE_ELSE;
         $targets[] = T_INSTANCEOF;
+        $targets[] = T_INSTEADOF;
+        $targets[] = T_AS;
 
         return $targets;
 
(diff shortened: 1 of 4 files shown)


jrfnl:
@michalbundyra Just my two pennies: We'd been contemplating this in the WordPress Coding Standards not too long ago and decided this needs a separate sniff, as - while the sniff functionality needed is the same -, `insteadof` and `as` are PHP keywords, not [operators](https://www.php.net/manual/en/language.operators.precedence.php).

michalbundyra:
@jrfnl 
> are PHP keywords, not operators

that's actually very good point. So you think it should be closed here?

jrfnl:
> that's actually very good point. So you think it should be closed here?

@michalbundyra That's not for me to say. Just wanted to leave the input to consider.

gsherwood:
> while the sniff functionality needed is the same -, `insteadof` and `as` are PHP keywords, not [operators](https://www.php.net/manual/en/language.operators.precedence.php).

Agreed. Even thought the rule is the same, I know that someone will take issue with non-operators being included in the operator spacing sniff.

I have thought about this before and the only other option I came up with was to include the token type in the error code (so it could be excluded) or make a sniff property where operators can be excluded (or included I guess, but that would be a very long list). So it's probably better to make new sniffs for `insteadOf` and `as` spacing.

PR #2693, opened 2019-11-13
Feature: fixable LeadingSlash error in PSR12.Files.ImportStatement

michalbundyra:
This can be easily fixable, just be removing the slash.
Added .fixed file with expected result of the fixer.


gsherwood:
Thanks a lot for the PR

PR #2695, opened 2019-11-14
Hotfix: correct array indicies when using ternary or coalesce ops

### LINKED ISSUE #2694: AbstractArraySniff produces invalid indices when using ternary operator
Example:

```
<?php

$array = [
    'foo' => 'foo',
    'bar' => $baz ? ['abc'] : ['def'],
    'baz' => 'baz',
];
```

Indices:
```
Array
(
    [0] => Array
        (
            [index_start] => 9
            [index_end] => 9
            [arrow] => 11
            [value_start] => 13
        )

    [1] => Array
        (
            [index_start] => 17
            [index_end] => 17
            [arrow] => 19
            [value_start] => 21
        )

    [2] => Array
        (
            [value_start] => 31
        )

    [3] => Array
        (
            [index_start] => 37
            [index_end] => 37
            [arrow] => 39
            [value_start] => 41
        )
)
```

Above is dump of variable `$indices` in method `AbstractArraySniff::processMultiLineArray`.

It is wrong, because array contains 3 elements, all 3 with keys.
The problem is because of ternary operator.

michalbundyra:
Fixes #2694

I have find out that the issue is also when using coalesce operator.
Example, which is valid code, but has no sens would be:

```php
$array = [
    'foo' => 'foo',
    'bar' => $baz ? ['abc'] : ['def'],
    'hey' => $baz ?? ['one'] ?? ['two'],
];
```

Without fix we have 5 elements in `$indices` but we should have 3.

Added tests to cover changes.

### DIFF
--- a/src/Sniffs/AbstractArraySniff.php
+++ b/src/Sniffs/AbstractArraySniff.php
@@ -89,6 +89,13 @@ abstract class AbstractArraySniff implements Sniff
                 continue;
             }
 
+            if ($tokens[$checkToken]['code'] === T_INLINE_THEN
+                || $tokens[$checkToken]['code'] === T_COALESCE
+            ) {
+                $checkToken = $phpcsFile->findEndOfStatement($checkToken);
+                continue;
+            }
+
             if ($tokens[$checkToken]['code'] === T_ARRAY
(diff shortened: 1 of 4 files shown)


gsherwood:
Thanks for fixing this.

PR #2697, opened 2019-11-14
PSR12/OperatorSpacing: hot fix

### LINKED ISSUE #2696: PSR12.Operators.OperatorSpacing: false positives
Looks like PR #2640 has introduced some new issues in `PSR12.Operators.OperatorSpacing` which extends `Squiz.WhiteSpace.OperatorSpacing`.

Example code:
```php
if (strpos($tokenContent, 'b"') === 0 && substr($tokenContent, -1) === '"') {}

$oldConstructorPos   = -1;

$content = -$content;
```

Previously, no errors would be thrown, now you get the following false positives:

```
-------------------------------------------------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
-------------------------------------------------------------------------------------------------------------
 3 | ERROR | [x] Expected at least 1 space after "-"; 0 found
 5 | ERROR | [x] Expected at least 1 space after "-"; 0 found
 7 | ERROR | [x] Expected at least 1 space after "-"; 0 found
-------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 3 MARKED SNIFF VIOLATIONS AUTOMATICALLY
------------------------------------------------------------------------------------------------------------- 
```

jrfnl:
PR #2640 made it so the `Squiz.WhiteSpace.OperatorSpacing` sniff now sets a property with the `$nonOperandTokens` from the `register()` method.

However, the PSR12 sniff for the same extends the Squiz sniff, but overloads the `register()` method to set different targets.

This means that the `$nonOperandTokens` array was empty and therefore the `isOperator()` method would nearly always return `true`, which was often incorrect.

Fixed by calling the `parent::register()` method before setting the `$targets` for the PSR12 sniff.

Includes adding unit tests to the PSR12 sniff related to the exclusions handled via the `isOperator()` method, to prevent breakage like this from being able to be merged in the future.

Fixes #2696

### DIFF
--- a/src/Standards/PSR12/Sniffs/Operators/OperatorSpacingSniff.php
+++ b/src/Standards/PSR12/Sniffs/Operators/OperatorSpacingSniff.php
@@ -24,6 +24,8 @@ class OperatorSpacingSniff extends SquizOperatorSpacingSniff
      */
     public function register()
     {
+        parent::register();
+
         $targets   = Tokens::$comparisonTokens;
         $targets  += Tokens::$operators;
         $targets  += Tokens::$assignmentTokens;
--- a/src/Standards/PSR12/Tests/Operators/OperatorSpacingUnitTest.inc
+++ b/src/Standards/PSR12/Tests/Operators/OperatorSpacingUnitTest.inc
@@ -46,3 +46,13 @@ $foo = $foo?:'bar';
 try {
 } catch (ExceptionType1|ExceptionType2 $e) {
 }
+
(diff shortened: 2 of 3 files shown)


gsherwood:
Thanks so much for finding and fixing this before it was released.

jrfnl:
> Thanks so much for finding and fixing this before it was released.

A failing external standards build kind of helps with that ;-)

Issue #943, opened 2016-04-04
Squiz.PHP.InnerFunctions.NotAllowed reported in anonymous classes


ondrejmirtes:
I think this sniff rule is supposed to prevent old  and discouraged PHP feature, but it also detects methods inside anonymous classes.

gsherwood:
Thanks for reporting this. It has been fixed.

Th3Mouk:
Hello !

The problem still occurs with

```php
new class {
    public function valueObject(): object
    {
         return new class {
            public function string(): string {
                 return 'string';
            }
        };
    }
};
```

A workaround :

```php
new class {
    public function valueObject(): object
    {
         return $x = new class {
            public function string(): string {
                 return 'string';
            }
        };
    }
};
```

PR #2701, opened 2019-11-18
Fix 🐛 #943

Th3Mouk:
Hello ! Just found this bug in my CI


gsherwood:
This fix is not complete as it will miss code like this:
```php
new class {
    public function outer()
    {
        if (!function_exists('inner')) {
            function inner() {
            }
        }
    }
};
```
This is actually a regression as this code passes with the current sniff. I'm adding this as a test case to ensure it is picked up.

I think the solution is relatively straight-forward. Instead of using the first anon class in the conditions array, it needs to use the closest anon class in order to determine if the functions is actually a method. I'll make this change and commit the fix.

gsherwood:
I've committed a fix for this new case. Thanks a lot for reporting it.

PR #2704, opened 2019-11-19
Hotfix: adds T_FN_ARROW to assignmentTokens list

michalbundyra:
We have there also T_DOUBLE_ARROW so I believe it is right place.


gsherwood:
> I don't think this is right.

Agreed.

Issue #2523, opened 2019-06-02
PHP 7.4: arrow functions


jrfnl:
PHP 7.4 will contain the ability to declare [arrow functions](https://wiki.php.net/rfc/arrow_functions_v2) (short, one statement closures). This RFC has already been [implemented in the PHP 7.4](https://github.com/php/php-src/commit/f3e5bbe6f37ce52a9ecd42812389e6aaf3aa2892) branch.

Arrow functions will make code sniffing more complicated as the `T_DOUBLE_ARROW` token is now also used for those, quite apart from these functions being declared without "scope"/curly braces, making it _interesting_ to determine the end of a statement, particularly when these would be used within arrays, ternaries or even in nested arrow functions.

I imagine, it may even be desirable to assign a different token to the `T_DOUBLE_ARROW` when used for arrow functions.

Based on the RFC and the implementation details, I foresee that - at least - the following changes will need to be made:
- [x] Back-fill the new `T_FN` token for PHP < 7.4.
- [ ] Add `T_FN` to the [condition related to anonymous structures](https://github.com/squizlabs/PHP_CodeSniffer/blob/200eae562f8c449107d2378a7f886586e77c6c8e/src/Files/File.php#L1235) in the `File::getDeclarationName()` method.
- [x] Allow for `T_FN` in the `File::getMethodParameters()` method.
- [x] Allow for `T_FN` in the `File::getMethodProperties()` method.
- [ ] Register `T_FN` with various sniffs which sniff function declarations.
- [ ] Any sniff/function relying on the `T_DOUBLE_ARROW` only being used for array key/value separation will also need adjusting (depending on whether or not a different token will be used or not).

Userland external standards will also need to be aware of this and will need to make similar changes in their sniffs for the last two action points.

There are a lot of code examples available in the [RFC](https://wiki.php.net/rfc/arrow_functions_v2), but IMO, more will be needed to test the above identified issues around arrow functions in arrays, in ternaries and nested arrow functions.

gsherwood:
From #2570, Generic.WhiteSpace.ScopeIndent would need support:
```php
<?php

declare(strict_types=1);

$numbers = [1, 2, 3];

$result = array_map(
    static fn(int $number) : int => $number + 1,
    $numbers
);
```

From #2571, PSR2.Methods.FunctionCallSignature would need support:
```php
<?php

declare(strict_types=1);

$container
    ->add(
        'identifier',
[...]

PR #2712, opened 2019-11-19
Hotfix: PHP 7.4 fn closure Return type hint tokens

michalbundyra:
For type hints and return type we usually have T_STRING, but we have couple
different tokens, and these are:

- T_CALLABLE
- T_PARENT
- T_SELF

We need to skip them here.

Related to #2523


michalbundyra:
Requires unit tests, not sure yet where to add them.

```php
fn (self $a) : self => $a;
fn (parent $a) : parent => $a;
fn (callable $a) : callable => $a;
```

michalbundyra:
I am not sure about T_CALLABLE, I've just checked couple examples and I am getting T_STRING always.

gsherwood:
Thanks for this PR. I added some tests for these cases as well, which were all failing before your fix.

PR #2714, opened 2019-11-19
QA: optimize imports - remove unused and order alphabetically

michalbundyra:
Honestly I am not sure what standard the library is using in terms of
sorting global imports but I find it quite confusing when these are not
ordered alphabetically and when non used imports are there.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -12,8 +12,8 @@
 
 namespace PHP_CodeSniffer;
 
-use PHP_CodeSniffer\Exceptions\RuntimeException;
 use PHP_CodeSniffer\Exceptions\DeepExitException;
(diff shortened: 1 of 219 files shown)


jrfnl:
I've mostly had the impression they were ordered by the order in which the classes were used in the file, i.e. any `implements`/`extends` classes first, then generally `File` second as it's used as the type declaration for `process()` and then whatever else is used.

gsherwood:
> Honestly I am not sure what standard the library is using in terms of sorting global imports

There is no defined standard. The order doesn't both me at all, but given it bothers you, and you're doing so much work on arrow functions, I'm happy to merge this in.

PR #2717, opened 2019-11-20
QA: fix doc return type

michalbundyra:


src/Standards/PSR2/Sniffs/ControlStructures/SwitchDeclarationSniff.php
```diff
@@ -208,7 +208,7 @@ public function process(File $phpcsFile, $stackPtr)
      * @param int                         $stackPtr  The position to start looking at.
      * @param int                         $end       The position to stop looking at.
      *
-     * @return int | bool
+     * @return int|bool
```
jrfnl: This could potentially be made more precise using `int|false`

gsherwood:
Thanks

PR #2725, opened 2019-11-26
Hotfix: fn arrow closure - Squiz.Scope.StaticThisUsage

michalbundyra:
Related #2523

I am not sure about this change as this is actually invalid code.
We cannot use `$this` even inside closure, as then closure is not gonna
be usable. The same in case of normal closure.

Maybe here we should have another fix then?

### DIFF
--- a/src/Standards/Squiz/Sniffs/Scope/StaticThisUsageSniff.php
+++ b/src/Standards/Squiz/Sniffs/Scope/StaticThisUsageSniff.php
@@ -69,13 +69,10 @@ class StaticThisUsageSniff extends AbstractScopeSniff
         $end  = $tokens[$stackPtr]['scope_closer'];
 
         do {
-            $next = $phpcsFile->findNext([T_VARIABLE, T_CLOSURE, T_FN, T_ANON_CLASS], ($next + 1), $end);
+            $next = $phpcsFile->findNext([T_VARIABLE, T_ANON_CLASS], ($next + 1), $end);
             if ($next === false) {
                 continue;
-            } else if ($tokens[$next]['code'] === T_CLOSURE
-                || $tokens[$next]['code'] === T_FN
-                || $tokens[$next]['code'] === T_ANON_CLASS
-            ) {
+            } else if ($tokens[$next]['code'] === T_ANON_CLASS) {
                 $next = $tokens[$next]['scope_closer'];
                 continue;
             } else if (strtolower($tokens[$next]['content']) !== '$this') {
--- a/src/Standards/Squiz/Sniffs/Scope/StaticThisUsageSniff.php
+++ b/src/Standards/Squiz/Sniffs/Scope/StaticThisUsageSniff.php
@@ -68,14 +68,37 @@ class StaticThisUsageSniff extends AbstractScopeSniff
         $next = $stackPtr;
         $end  = $tokens[$stackPtr]['scope_closer'];
 
+        $this->checkThisUsage($phpcsFile, $next, $end);
+
+    }//end processTokenWithinScope()
+
+
+    /**
+     * Check for $this variable usage between $next and $end tokens.
+     *
+     * @param File $phpcsFile The current file being scanned.
+     * @param int  $next      The position of the next token to check.
+     * @param int  $end       The position of the last token to check.
+     *
+     * @return void
+     */
+    private function checkThisUsage(File $phpcsFile, $next, $end)
+    {
+        $tokens = $phpcsFile->getTokens();
+
         do {
             $next = $phpcsFile->findNext([T_VARIABLE, T_ANON_CLASS], ($next + 1), $end);
             if ($next === false) {
                 continue;
-            } else if ($tokens[$next]['code'] === T_ANON_CLASS) {
+            }
+
+            if ($tokens[$next]['code'] === T_ANON_CLASS) {
+                $this->checkThisUsage($phpcsFile, $next, $tokens[$next]['scope_opener']);
                 $next = $tokens[$next]['scope_closer'];
                 continue;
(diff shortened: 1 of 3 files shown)


gsherwood:
> I am not sure about this change as this is actually invalid code.

That's the point of this sniff. Because it's a runtime error, the sniff tries to determine that the code is invalid during linting.

So I think this change is a good one.

src/Standards/Squiz/Sniffs/Scope/StaticThisUsageSniff.php
```diff
@@ -69,10 +69,11 @@ public function processTokenWithinScope(File $phpcsFile, $stackPtr, $currScope)
...
         do {
-            $next = $phpcsFile->findNext([T_VARIABLE, T_CLOSURE, T_ANON_CLASS], ($next + 1), $end);
+            $next = $phpcsFile->findNext([T_VARIABLE, T_CLOSURE, T_FN, T_ANON_CLASS], ($next + 1), $end);
             if ($next === false) {
                 continue;
             } else if ($tokens[$next]['code'] === T_CLOSURE
+                || $tokens[$next]['code'] === T_FN
                 || $tokens[$next]['code'] === T_ANON_CLASS
             ) {
                 $next = $tokens[$next]['scope_closer'];
```
gsherwood: T_FN tokens don't always have a scope closer, unlike closures and anon classes. This is because the T_FN token  is a real token that exists, and then scope information is added later if the syntax looks valid. But T_CLOSURE and T_ANON_CLASS are custom tokens that only get applied by PHPCS if the syntax looks good. Otherwise they are normal T_FUNCTION and T_CLASS tokens that don't have scope information.

michalbundyra:
@gsherwood 

> So I think this change is a good one.

I think the error should be reported if method is static and `$this` is used inside closure

```php
class A {
    public $name = 'NAME';
    
    public static function m1() { // cannot be static as we are using $this in closure
        return function () {
            return $this->name;
        };
    }
    
    public static function m2() { // cannot be static as we are using $this in closure
        return fn () => $this->name;
    }
}
```
https://3v4l.org/XgXbX

We are not getting any errors on that code, but if we try to call method and execute the result of the method - callback, then we get the error.

Currently we are skipping that case for normal closure (T_CLOSURE), that's why I've added also the same for T_FN, but I believe we should report error in that case.

I'll submit another commit soon and you'll decide what to do.

src/Standards/Squiz/Sniffs/Scope/StaticThisUsageSniff.php
```diff
@@ -69,15 +69,13 @@ public function processTokenWithinScope(File $phpcsFile, $stackPtr, $currScope)
...
             if ($next === false) {
                 continue;
-            } else if ($tokens[$next]['code'] === T_CLOSURE
-                || $tokens[$next]['code'] === T_ANON_CLASS
-            ) {
+            } else if ($tokens[$next]['code'] === T_ANON_CLASS) {
                 $next = $tokens[$next]['scope_closer'];
                 continue;
-            } else if (strtolower($tokens[$next]['content']) !== '$this') {
+            } else if ($tokens[$next]['content'] !== '$this') {
```
michalbundyra: It was wrong, see:
https://3v4l.org/cvN2L

src/Standards/Squiz/Tests/Scope/StaticThisUsageUnitTest.inc
```diff
@@ -57,22 +57,61 @@ class MyClass
...
     }
 }
 
 trait MyTrait {
-	public static function myFunc() {
-		$this->doSomething();
-	}
+    public static function myFunc() {
+        $this->doSomething();
+    }
```
michalbundyra: changes tabs to spaces

src/Standards/Squiz/Tests/Scope/StaticThisUsageUnitTest.inc
```diff
@@ -57,22 +57,61 @@ class MyClass
...
+
+    public static function other() {
+        return fn () => $this->name;
+    }
+
+    public static function anonClassUseThis() {
+        return new class($this) {
```
michalbundyra: Currently, when we find anonymous class we are skipping till scope_closer, but we can use $this in params, so we really shouldn't skip it 'yet'.

The check should be more complicated here - please see also example in lines 98-110.
Errors should be reported only in line 84 and 99, not in 93 or 106.

I've implemented "similar" sniff here: 
https://github.com/webimpress/coding-standard/blob/develop/src/WebimpressCodingStandard/Sniffs/PHP/StaticCallbackSniff.php
which checks if closure can be defined as static or not.

src/Standards/Squiz/Tests/Scope/StaticThisUsageUnitTest.php
```diff
@@ -31,9 +31,13 @@ public function getErrorList()
             9  => 1,
             14 => 1,
             20 => 1,
+            41 => 1,
```
michalbundyra: in my opinion it was invalid implementation before, so the error was not reported on that line - as noted in one of the previous comments - we shouldn't skip closures.

gsherwood:
I'm going to remove this from the 3.5.4 milestone as it contains failing test cases that should be resolved before merging.

michalbundyra:
@gsherwood I've rebased and updated the code of detecting usage of `$this` variable. We need to use recursion, as I pointed before, because `$this` can be used in instantiation of new anonymous class.

Can it be added  back to 3.5.4 release, please? Thanks!

gsherwood:
Thanks a lot for this. Sorry it's taken so long.

PR #2727, opened 2019-11-30
Add fixer for valid logical operator

VincentLanglet:


jrfnl:
@VincentLanglet I don't think this can be auto-fixed without additional checks.

The operators being replaced have a different [operator precendence](https://www.php.net/manual/en/language.operators.precedence.php) than their replacements, so it would need to be verified that the replacing doesn't change the results.

VincentLanglet:
@jrfnl Oh ! I didn't know this.
Indeed
```
<?php
$bool = true && false;
var_dump($bool); // false, that's expected

$bool = true and false;
var_dump($bool); // true, ouch!
?>
```

Does the additional checks looks easy to you or is it simplier to close the PR ?

jrfnl:
> Does the additional checks looks easy to you or is it simplier to close the PR ?

I suspect you'd need a _lot_ of extra code to verify the operator precedence, as well as lots and lots of extra unit tests, so it would definitely be simpler not to have this auto-fixable.

VincentLanglet:
@jrfnl OK !

PR #2731, opened 2019-12-01
Travis: don't allow PHP 7.4 build to fail

jrfnl:
As [PHP 7.4 has been released](https://www.php.net/archive/2019.php#2019-11-28-1), the build against PHP 7.4 should no longer be allowed to fail.

As that build was combined with the PHPStan check which - for now - should be allowed to fail, I've split the build into two:
1. a "normal" PHP 7.4 build which is not allowed to fail.
2. a PHPStan specific build which is allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ matrix:
     - php: 7.3
       env: PHPUNIT_INCOMPAT=1
     - php: "7.4snapshot"
+      env: PHPUNIT_INCOMPAT=1
+    - php: "7.4snapshot"
       env: PHPSTAN=1 PHPUNIT_INCOMPAT=1
       addons:
         apt:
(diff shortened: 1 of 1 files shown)


michalbundyra:
@jrfnl 

> a PHPStan specific build which is allowed to fail.

why it is allowed to fail? Can you explain, please?

jrfnl:
> 
> > a PHPStan specific build which is allowed to fail.
> 
> why it is allowed to fail? Can you explain, please?

You'll need to ask @gsherwood that question as that's what he decided when PHPStan was added.

See: https://github.com/squizlabs/PHP_CodeSniffer/pull/2201#issuecomment-455358323 and the follow up comments for the original discussion.

gsherwood:
Thanks a lot for this.

PR #2733, opened 2019-12-03
Fix warning determining shell width on system without shell_exec

### LINKED ISSUE #2698: PHPCS throws errors determining auto report width when shell_exec is disabled
If a server has shell_exec() disabled for security reasons, the error logs fill up with warnings like this each time a PHP CodeSniffer command is run:
`
PHP Warning:  shell_exec() has been disabled for security reasons in /.composer/vendor/squizlabs/php_codesniffer/src/Config.php on line 222
`
Disabling shell_exec() doesn't interfere with PHP CodeSniffer's functioning, but a simple check to see if shell_exec() is disabled with function_exists() would prevent the warning on servers that disable shell_exec().

MasterOdin:
Closes #2698

Checks for the existence of the `shell_exec` function before attempting to use it to determine shell width, falling into the `else` clause where the width will be set to `0`, similar to what would happen if `stty` program did not exist or failed to run if `shell_exec` did exist.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -219,7 +219,10 @@ class Config
         switch ($name) {
         case 'reportWidth' :
             // Support auto terminal width.
-            if ($value === 'auto' && preg_match('|\d+ (\d+)|', shell_exec('stty size 2>&1'), $matches) === 1) {
+            if ($value === 'auto'
+                && function_exists('shell_exec') === true
+                && preg_match('|\d+ (\d+)|', shell_exec('stty size 2>&1'), $matches) === 1
+            ) {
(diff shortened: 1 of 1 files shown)


MasterOdin:
Improved the title to reflect that other usages of shell_exec are not covered.

The other usages are:
1. Determining if stdin is tty
1. Generating diff report (uses commandline `diff` utility, could be replaced with https://github.com/sebastianbergmann/diff?)
1. Using the `Generic.PHP.SyntaxSniff` sniff

gsherwood:
Thanks a lot for fixing this up

Issue #2736, opened 2019-12-03
Question: Multiple reports in phpcs.xml with output file


nick-zh:
Hey guys

I was unable to find how to do this, so i want to have multiple reporting, which is possile if
i add this in my phpcs.xml:
```
<arg name="report" value="full,gitblame,junit"/>
```
is there a way to put the junit output in a file though?
Thanks in advance for clarifying this.
If this is somewhere in the doc please let me know, if not maybe i could contribute this afterward as an example.

Cheers,
Nick

nick-zh:
i had a quick glance at the source, it's doable like this:
```
    <arg name="report-gitblame"/>
    <arg name="report-full"/>
    <arg name="report-junit" value="./junit.xml"/>
```

PR #2737, opened 2019-12-03
improve config example

nick-zh:
Because of #2736

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -15,6 +15,9 @@
     <arg name="colors"/>
     <arg name="parallel" value="75"/>
     <arg value="np"/>
+    <arg name="report-gitblame"/>
+    <arg name="report-full"/>
+    <arg name="report-junit" value="./build/logs/phpcs/junit.xml"/>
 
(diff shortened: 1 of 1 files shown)


jrfnl:
1. This is not a "config example". This is the actual ruleset for the code of PHPCS itself you are adjusting.
2. All the info you were looking for in #2736 is already documented in the wiki:
    - https://github.com/squizlabs/PHP_CodeSniffer/wiki/Reporting#printing-multiple-reports
    - https://github.com/squizlabs/PHP_CodeSniffer/wiki/Reporting#writing-a-report-to-a-file

Suggest: close.

nick-zh:
@jrfnl thanks, yeah i saw but for me it was not immediate apparent how to use this in a config file, the examples are mostly for cli use. Since i saw this:

> An example phpcs.xml file can be found in the PHP_CodeSniffer repository: phpcs.xml.dist

from here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-default-configuration-file
I thought this was the appropriate place to add it. Do you think maybe this can be improved in the doc or do you think it is not needed?

PR #2746, opened 2019-12-06
Hotfix: correct array indices

### LINKED ISSUE #2745: AbstractArraySniff wrong indices when mixed coalesce and ternary values
Code:

```php
<?php

return [

    $a => static function () {
        return [1, 2, 3];
    },

    $b ?? $c,
    $d ? [$e] : [$f],

];
```

Indicies:

```
Array
(
    [0] => Array
        (
            [index_start] => 8
            [index_end] => 8
            [arrow] => 10
            [value_start] => 12
        )

)
```

The problem is here:

https://github.com/squizlabs/PHP_CodeSniffer/blob/557a1fc7ac702c66b0bbfe16ab3d55839ef724cb/src/Sniffs/AbstractArraySniff.php#L92-L97

as we do not have arrows for two last elements in array, above finds end of the statement - comma - and then skip to next element.

michalbundyra:
I've changed the approach completely. In my opinion it much clearer.
Added tests to cover the issue described in #2745.

Fixes #2745

It solves also the problem when we had `["value_start" => false]` for an empty array.

### DIFF
--- a/src/Sniffs/AbstractArraySniff.php
+++ b/src/Sniffs/AbstractArraySniff.php
@@ -64,139 +64,33 @@ abstract class AbstractArraySniff implements Sniff
         if ($tokens[$lastContent]['code'] === T_COMMA) {
             // Last array item ends with a comma.
             $phpcsFile->recordMetric($stackPtr, 'Array end comma', 'yes');
-            $lastArrayToken = $lastContent;
         } else {
             $phpcsFile->recordMetric($stackPtr, 'Array end comma', 'no');
-            $lastArrayToken = $arrayEnd;
         }
 
-        if ($tokens[$stackPtr]['code'] === T_ARRAY) {
-            $lastToken = $tokens[$stackPtr]['parenthesis_opener'];
-        } else {
-            $lastToken = $stackPtr;
-        }
-
-        $keyUsed = false;
         $indices = [];
 
-        for ($checkToken = ($stackPtr + 1); $checkToken <= $lastArrayToken; $checkToken++) {
-            // Skip bracketed statements, like function calls.
-            if ($tokens[$checkToken]['code'] === T_OPEN_PARENTHESIS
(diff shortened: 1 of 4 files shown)


gsherwood:
This is quite a large refactor that has removed a fair bit of code. Can you provide a bit of an explanation about what changes have been made and why?

michalbundyra:
@gsherwood I just wrote it from scratch instead of "fixing" the old code.

There are two functions now: one which is finding all indices in the array and the other (`getNext`) which returns the next "stop character" in array - and it is either comma or double arrow.
In this method we skip all scopes - like closures, array functions, anonymous classes, nested arrays, etc... and we return the ptr of next comma/double arrow or the end of the array.

If you read just new code you will see it is all pretty clear, and we don't need all these old code, imo.
As you can also see, all tests are fine :)

VincentLanglet:
@michalbundyra I just wrote a Sniff checking for comma and I added the comma in the indices here https://github.com/squizlabs/PHP_CodeSniffer/pull/2759.

How can we add the commas with your refactor ?

michalbundyra:
@VincentLanglet It would be easy to add it here. I am not doing it as it is out of the scope right now and PR #2759 is going to be changes (as I understand from comments there).

gsherwood:
> @gsherwood I just wrote it from scratch instead of "fixing" the old code.

Refactoring a sniff is fine, but I can't determine what the actual fix is here. I can't accept this PR as a bug fix, so I'll remove it from the 3.5.4 milestone.

gsherwood:
Quick update: I'm working on writing tests for the AbstractArraySniff indices processing to expose the original bug. If I get a comprehensive enough suite of tests for it, I'll come back and review this refactor for inclusion.

gsherwood:
After adding some core tests for this sniff, I've been able to verify that this refactor is working correctly. Thanks a lot for this PR. This sniff is much cleaner now, and also fixes #2745.

PR #2750, opened 2019-12-07
Disallow the use of $_Request

### LINKED ISSUE #2373: Disallow the use of $_REQUEST
Like in VIP, but this makes sense generally:
https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/#using-_request

>$_REQUEST should never be used because it is hard to track where the data is coming from (was it POST, or GET, or a cookie?), which makes reviewing the code more difficult. Additionally, it makes it easy to introduce sneaky and hard to find bugs, as any of the aforementioned locations can supply the data, which is hard to predict.  Much better to be explicit and use either $_POST or $_GET instead.

Morerice:
Create a new sniff which will listen on the variable constant ($). If the $_REQUEST super global is detected, add the following error: 

> The $_REQUEST super global should not be used. Use $_GET, $_POST or $_COOKIE instead

This implements #2373

### DIFF
--- a/src/Standards/Generic/Sniffs/PHP/DisallowRequestSuperGlobalSniff.php
+++ b/src/Standards/Generic/Sniffs/PHP/DisallowRequestSuperGlobalSniff.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * Ensures the $_REQUEST super global is not used
+ *
+ * @author    Jeantwan Teuma <jeant.m24@gmail.com>
+ * @copyright 2006-2019 Squiz Pty Ltd (ABN 77 084 670 600)
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ */
+
+namespace PHP_CodeSniffer\Standards\Generic\Sniffs\PHP;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+
+class DisallowRequestSuperGlobalSniff implements Sniff
+{
+
+
+    /**
+     * Returns an array of tokens this test wants to listen for.
+     *
+     * @return array
+     */
+    public function register()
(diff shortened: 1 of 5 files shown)


src/Standards/Generic/Sniffs/PHP/DisallowRequestSuperGlobalSniff.php
```diff
@@ -0,0 +1,55 @@
...
+        $tokens = $phpcsFile->getTokens();
+
+        $varName = $tokens[$stackPtr]['content'];
+        if ($varName !== '$_REQUEST') {
+            return;
+        }
+
+        $type  = 'RequestSuperGlobalAccessed';
+        $error = 'The $_REQUEST super global should not be used. Use $_GET, $_POST or $_COOKIE instead';
+        $phpcsFile->addError($error, $stackPtr, $type, []);
```
gsherwood: What I would normally do here is use `Found` as the error type string. This typically allows for cleaner message codes. For example, `Generic.PHP.DisallowRequestSuperGlobal.Found` instead of `Generic.PHP.DisallowRequestSuperGlobal.RequestSuperGlobalAccessed`.

I don't think it needs a variable defined for it the type either. And no need to pass an empty array as the 4th argument as this is the default value already.

Morerice: Suggestions implemented

gsherwood:
Thanks a lot for this contribution.

I decided to rename the sniff to `Generic.PHP.DisallowRequestSuperglobal` and change references to "super global" to "superglobal" as per the PHP manual. Other than that - I left it alone. Thanks for the docs as well.

Morerice:
@gsherwood You're welcome and thanks for guiding me on the first pr on this project. I hope I have time to tackle some other issues in the future.

PR #2756, opened 2019-12-09
Remove incorrect apostrophe in UseDeclarationSniff

JoshOldenburg:


gsherwood:

Issue #2546, opened 2019-06-28
PHP 7.4: backfill for tokenizing numeric sequences with an underscore separator


jrfnl:
PHP 7.4 will contain the ability to use [an underscore in numeric literals to visually separate groups of digits](https://wiki.php.net/rfc/numeric_literal_separator). This RFC has already been [implemented in the PHP 7.4](https://github.com/php/php-src/commit/f74109d9a4b1e4fbaeba4f68c8fc89950e19d265) branch and can be tested with the alpha builds available.

Now, from a PHPCS/Tokenizer point of view, there are two things we need to consider:
1. While the PHP 7.4 tokenizer will correctly tokenize these sequences as `T_LNUMBER` or `T_DNUMBER`, the tokenizer in any PHP version below PHP 7.4 will not.
    IMO, this is something which can and should be back-filled in PHPCS.
2. The token `content` value for any such token - including the backfilled ones - would contain the `_`(s). This is problematic when a sniff would examine the `content` of a `T_LNUMBER` or `T_DNUMBER`, for instance to check that the parameter value passed to a certain function call is higher or lower than a set value.
    If that sniff would run on PHP 7.4, all would be fine, but on any older PHP version the sniff would break as soon as it would try to do a numeric comparison with the `content` of the token.
    To prevent such breakage of existing (userland) sniffs, I would like to suggest for the tokenizer - after backfilling the token parts to one token -, to save the token value as received from the tokenizer to `orig_content` and to strip out the underscores for the `content` value of the token.

Does that make sense ?

---
<details>
  <summary>Code example based on the RFC:</summary>

  ```php
<?php

// OK: ordinary numeric sequences.
$a = 1000000000;
$b = 107925284.88;
$discount = 13500;

// OK: Constant name starting with an underscore is perfectly valid.
const _100 = 100;

// PHP 7.4 numeric sequences with a numeric literal separator.
$threshold = 1_000_000_000;  // a billion!
$testValue = 107_925_284.88; // scale is hundreds of millions
$discount  = 135_00;         // $135, stored as cents

$a = 6.674_083e-11; // float
$a = 299_792_458;   // decimal
$a = 0xCAFE_F00D;   // hexadecimal
$a = 0b0101_1111;   // binary
$a = 0137_041;      // octal

// More test cases.
$a = 1_2.3_4e1_23;

// Invalid use of underscores in numeric sequences.
// Each underscore in a numeric literal must be directly between two digits.
[...]

PR #2771, opened 2019-12-17
[Hotfix] Tokenizer / numeric separators backfill: fix all known issues

jrfnl:
### Test case file:
1. Remove a stray invisible whitespace character/null byte from the test case file.
2. Add a large number of additional test cases.

### Test file `testBackfill()`:
1. Fixed reversed order of the parameters in the `assertSame()`. The first parameter is `$expected`, the second `$result`.
    Having the parameters in the correct order makes debugging the tests easier as the messages send by PHPUnit will be correct.
2. Test that the final token has the correct code **and** the correct type.
3. Correct the expected type for two test cases based on how PHP 7.4 tokenizes these.
4. Add datasets for the new test cases which are to be handled by the backfill.

### Test file new `testNoBackfill()`
Add tests to verify that numbers using numeric separators which are considered parse errors and/or which aren't relevant to the backfill, do not incorrectly trigger the backfill anyway.

This test verifies that the final token sequence is in line with the token sequence PHP 7.4 would give.

## Tokenizer / numeric separators backfill: fix incorrect token types

Tokenizer / numeric separators backfill: fix incorrect token types

When the final content of a numeric token is more than PHP_INT_MAX or when a numeric token contains a `.` or an `E/e`, it should be tokenized as `T_DNUMBER`.

This fixes the first set of the unit test failures.

### Tokenizer / numeric separators backfill: fix parse errors being handled by the backfill

The prevents the backfill from kicking in when confronted with invalid use of the numeric literal separator which would be a parse error in PHP 7.4 anyway.

This fixes the remaining unit test failures.

Fixes all currently known issues which remained after #2546 

---

Note: on Windows the unit tests will currently not pass when run against PHP 7.4.0 due to a bug in PHP itself. See: https://bugs.php.net/bug.php?id=78978

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -997,12 +997,6 @@ class PHP extends Tokenizer
                     ) {
                         $newContent .= $tokens[$i][1];
 
-                        // Any T_DNUMBER token needs to make the
-                        // new number a T_DNUMBER as well.
-                        if ($tokens[$i][0] === T_DNUMBER) {
-                            $newType = T_DNUMBER;
-                        }
-
                         // Support floats.
                         if ($tokens[$i][0] === T_STRING
                             && substr(strtolower($tokens[$i][1]), -1) === 'e'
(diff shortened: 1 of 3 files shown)


jrfnl:
@gsherwood Could this PR please be milestoned for `3.5.4` ? This issue is a blocker for PHPCSUtils.

gsherwood:
Thanks for doing all this work to make it more accurate.

jrfnl:
@gsherwood Thanks for the merge! This one was a bit of a thorn in my side ;-)

PR #2778, opened 2019-12-21
Add UnusedUseSniff

VincentLanglet:
@jrfnl @gsherwood I never created a rule, I don't know if there is more file to create.

This is ready-to-review. 
I only need to fix the linter, improve the comment, adding the copyright, ... since I wrote this in another project.

This could close
https://github.com/squizlabs/PHP_CodeSniffer/issues/801
https://github.com/squizlabs/PHP_CodeSniffer/pull/1106


gmponos:
I think this is something that is really missing from codesniffer..

Your work seems good for me.. never the less you might want to add some tests also about 

group use namespaces.. 

Namespace with scope.. 

```
namespace That { the code }
```

Two classes in one file

VincentLanglet:
> Your work seems good for me.. never the less you might want to add some tests also about

I'll be happy adding some tests @gmponos 
 
> group use namespaces..

I tested these: 
```
use My\Project\{TestA, TestB};
use My\{Project\TestC, TestD, Project\TestE};
use SomeNamespace\{ClassD, ClassF as F, ClassE};
use function SomeNamespace\{fn_d, fn_e};
use const SomeNamespace\{ConstD, ConstE};
```

Did you mean something else ?

> Namespace with scope..
> ```
> namespace That { the code }
> ```

I don't know this syntax. Do you have a bigger example ? Where are the use ?

> Two classes in one file

I think it works for two classes of the same namespace.
Is it possible to have different namespace in the same file ?

gmponos:
> I don't know this syntax. Do you have a bigger example ? Where are the use ?

Never used it but I think in PHP you are allowed to do this:

```
namespace Whatever {
    use Test\Whatever;

    class Test
    {

    }
}

namespace Standard {
    use Test\Whatever;

    class MyTest{
        
    }
}
```

gmponos:
> I tested these:

OK I missed this 👍 nice going..

> Is it possible to have different namespace in the same file ?

I think my previous comment answers this :)

jrfnl:
> > Namespace with scope..
> > ```
> > namespace That { the code }
> > ```
> 
> I don't know this syntax. Do you have a bigger example ? Where are the use ?
> 
> > Two classes in one file
> 
> I think it works for two classes of the same namespace.
> Is it possible to have different namespace in the same file ?

This manual page has examples and explanations about both: https://www.php.net/manual/en/language.namespaces.definitionmultiple.php

Other things I'm wondering about whether the sniff handles these correctly based on a first quick glance at the code:
* closure declarations with `use` in the global namespace
* use of the `namespace` operator in the global namespace, i.e `echo namespace\my_function();`
* code interlaced with comments/whitespace/annotations (doesn't happen often, but is perfectly valid syntax, so the sniff should take it into account)
* The `T_STRING` token is used for a lot. Does the sniff handle things correctly like a (global) function being called which just happens to have the same name as a class which was imported ? Or a class constant being declared which just happens to have the same name as a class being imported ? The `use` should not be counted as "used" in that case.
    I have tingling feeling that a lot more checking for code context is needed.
* Comments do not normally necessitate a `use` statement. Why are you counting those as "used" ?

Having a test with an exception in a `catch` would also be a good idea.

Lastly: this sniff will scan the complete file for *every* *single* `use` statement encountered. This is very inefficient and will make the sniff pretty slow.

Anyways, that's just from a first glance. I will probably have missed things or possibly overlooked safeguards you have in place already.

gmponos:
One more case that we need to make sure..

```
namespace App\Entity;

use Doctrine\ORM\Mapping as ORM;
use Doctrine\ORM\Mapping\UniqueConstraint;

/**
 * @ORM\Table(name="whatever", uniqueConstraints={@UniqueConstraint(name="idx_whatever_code_UNIQUE", columns={"code"})})
 * @ORM\Entity
 */
class Whatever {}
```

VincentLanglet:
> * closure declarations with `use` in the global namespace

Indeed, infinite loop;

> * use of the `namespace` operator in the global namespace, i.e `echo namespace\my_function();`

Indeed

> * code interlaced with comments/whitespace/annotations (doesn't happen often, but is perfectly valid syntax, so the sniff should take it into account)

Can you give me an example, I don't see the issue.
Are you talking about the fact the comment is removed with the use ?

> * The `T_STRING` token is used for a lot. Does the sniff handle things correctly like a (global) function being called which just happens to have the same name as a class which was imported ? Or a class constant being declared which just happens to have the same name as a class being imported ? The `use` should not be counted as "used" in that case.
>   I have tingling feeling that a lot more checking for code context is needed.

Indeed, sometimes I miss unusedUse. I actually don't have better idea for the checks.

> * Comments do not normally necessitate a `use` statement. Why are you counting those as "used" ?

If you write
```
/* @param PHP_CodeSniffer\Files\File
```
You don't need a use

But if you write
```
/* @param File
```
You do need `use PHP_CodeSniffer\Files\File`.

And there is tricky comments, like
```
/* @Route 
```
 
> Having a test with an exception in a `catch` would also be a good idea.

Ok

> Lastly: this sniff will scan the complete file for _every_ _single_ `use` statement encountered. This is very inefficient and will make the sniff pretty slow.

I know it's not the best ; but in three year I used this sniff, I never got issue with the execution time.

jrfnl:
> Can you give me an example, I don't see the issue.
Are you talking about the fact the comment is removed with the use ?

As I said, I've only skimmed the code, but I suspect that your code in a few places will fail when it encounters (horrible) code like this:
```php
use My\Name\{// comment
    SubLevel \ /* comment*/ ClassName
}

echo NotTheSame\
    /*comment*/
   ClassName;
```


> But if you write
> ```php
> /* @param File
> ```
> You do need use PHP_CodeSniffer\Files\File.

Actually, you don't. PHP will work just fine without it if the classname is only used in comments.

Could it be you are confusing this with use in a type declaration ? In that case, you _do_ need the `use` statement.

VincentLanglet:
> > You do need use PHP_CodeSniffer\Files\File.
> 
> Actually, you don't. PHP will work just fine without it if the classname is only used in comments.

I didn't know. Or I didn't think about it, it's indeed comment so Php will work normally.

But Phpstorm add automatically the use, and considers this one as used.
This also helps for navigating through files.
So I still don't want to remove this use if a developper added it.

> One more case that we need to make sure..
> 
> ```
> namespace App\Entity;
> 
> use Doctrine\ORM\Mapping as ORM;
> use Doctrine\ORM\Mapping\UniqueConstraint;
> 
> /**
>  * @ORM\Table(name="whatever", uniqueConstraints={@UniqueConstraint(name="idx_whatever_code_UNIQUE", columns={"code"})})
>  * @ORM\Entity
>  */
> class Whatever {}
> ```
> 
> This namespace : `Doctrine\ORM\Mapping\UniqueConstraint` is not unused.

Just added.

michalbundyra:
Hello!

I've put a lot of time and effort to make this sniff working here:
https://github.com/webimpress/coding-standard/blob/master/src/WebimpressCodingStandard/Sniffs/Namespaces/UnusedUseStatementSniff.php

There are a lot of tests and I am using it with success with several projects already.

VincentLanglet:
> Hello!
> 
> I've put a lot of time and effort to make this sniff working here:
> https://github.com/webimpress/coding-standard/blob/master/src/WebimpressCodingStandard/Sniffs/Namespaces/UnusedUseStatementSniff.php
> 
> There are a lot of tests and I am using it with success with several projects already.

What do you think about creating a PR to share this sniff with Php_Codesniffer @michalbundyra ?

gsherwood:
> I didn't know. Or I didn't think about it, it's indeed comment so Php will work normally.
> 
> But Phpstorm add automatically the use, and considers this one as used.
> This also helps for navigating through files.
> So I still don't want to remove this use if a developper added it.

This adds significant complexity to the sniff. It's also not possible to correctly determine if a use statement is actually required by single-file static analysis alone given the dynamic nature of the PHP language.

If I accept this sniff into PHPCS, I then have to take responsibility for fixing all bugs found, which takes significant time. So to accept a sniff like this, it really needs to be both very carefully targeted and incredibly well tested.

I don't think I'd accept this sniff if it is doing auto-fixing either. PHPCBF can auto-format code, but removing code is something I try to never do. I'd be fine with the sniff reporting a non-fixable warning, indicating that a manual code review should be performed to determine if the use statement is actually unused.

VincentLanglet:
> I don't think I'd accept this sniff if it is doing auto-fixing either.

I only use sniff with auto-fixing so I prefer to make the pull request on another repository then.

GlennM:
>  I'd be fine with the sniff reporting a non-fixable warning, indicating that a manual code review should be performed to determine if the use statement is actually unused.

How about this one, can it be implemented or @VincentLanglet can you point me to the right direction please? Also looking for a way to re-order the import statements alphabetically.

VincentLanglet:
@GlennM You can start from my branch and remove the auto-fix.

GlennM:
Thanks for your quick reply @VincentLanglet . I'm used to "FriendsOfPHP/PHP-CS-Fixer", and not that familiar yet with PHP_CodeSniffer, so unfortunately I 'm not sure whether I can manage your suggestion...

PR #2792, opened 2020-01-02
Generic/ExecutableFile: skip tests on Windows

jrfnl:
Windows doesn't have the concept of executable files this sniff targets, so the test would always fail.


jrfnl:
The build failure is addressed in PR #2793 and is unrelated to this PR.

gsherwood:
Thanks

PR #2793, opened 2020-01-02
Squiz/FileComment: update the "fixed" files copyright year

jrfnl:
These tests need an update each time the year changes as the `.fixed` file checks for the hard-coded current year.

Happy new year!

PR #2794, opened 2020-01-02
Tokenizer: improve tokenization of T_NULLABLE vs T_INLINE_THEN

### LINKED ISSUE #2791: PSR12.Functions.NullableTypeDeclaration false positive when ternary operator used with instanceof
This code will return `There must not be a space between the question mark and the type in nullable type declarations`

```php
return $value instanceof static ? '(' . $value . ')' : $value;
```

8E7EEA1A-7EEF-45E7-8612-51C7869F6D9E

Snuffer version: 3.5.3

jrfnl:
This is an attempt to make the differentiation between `T_NULLABLE` and `T_INLINE_THEN` more stable.

There is only a limited set of tokens which can be used for type declarations. If a `?` is not followed by one of these, we can be certain that it is a ternary operator `T_INLINE_THEN`.

The more resource intensive "walking back" check will now only be done when it could be either.
The "walking back" token logic could probably still do with an additional check for `T_INSTANCEOF`, but the changes now made, should prevent the majority of issues where `T_INLINE_THEN` is misidentified as `T_NULLABLE`.

Includes unit tests via the PSR12.Functions.NullableTypeDeclaration sniff.

Fixes #2791

Note: the build failure is unrelated to this PR and will be fixed via #2793

### DIFF
--- a/src/Standards/PSR12/Tests/Functions/NullableTypeDeclarationUnitTest.inc
+++ b/src/Standards/PSR12/Tests/Functions/NullableTypeDeclarationUnitTest.inc
@@ -66,7 +66,19 @@ class TestTokenizingOfNullableVsInlineThen {
     }
 }
 
+// Issue #2641.
 $foo = new static(
     is_null($a) ? foo($a) : $a,
     is_null($b) ? $b : $c
 );
+
+// Issue #2791.
+class testInstanceOf() {
+    function testIt() {
+        $foo = $value instanceof static ? '(' . $value . ')' : $value;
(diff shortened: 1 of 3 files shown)


jrfnl:
Rebased to show the build passes.

gsherwood:
> There is only a limited set of tokens which can be used for type declarations. If a `?` is not followed by one of these, we can be certain that it is a ternary operator `T_INLINE_THEN`.

Thanks a lot. This is a much better idea than how I originally implemented it.

PR #2797, opened 2020-01-05
Return int|false instead of int|bool

glen-84:
I'm integrating [Psalm](https://psalm.dev/) into my custom standard, and code like this:

```php
// Find the position of each EOL sequence.
while (($pos = $phpcsFile->findNext(T_WHITESPACE, $start, null, false, $phpcsFile->eolChar))) {
    // Count the number of EOL sequences directly following the current token.
    $innerStart = ($pos + 1);
```

... is an issue, because `findNext` is documented to return `int|bool`, when in fact it returns `int|false`. The `while` loop removes `false` from the type, so `$pos` is then `int|true` when it should be just `int`, and adding `1` to a variable that is possibly `true` is a problem.

I decided to fix all cases of `int|bool` that I could find where the method only returns `false` and not `true`. There were also two cases where `bool` wasn't even included in the return type, which I've fixed.

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -2115,7 +2115,7 @@ class File
      *                                  will not be checked. IE. checking will stop
      *                                  at the previous semi-colon found.
      *
-     * @return int|bool
+     * @return int|false
      * @see    findNext()
      */
     public function findPrevious(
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks for this change

PR #2798, opened 2020-01-05
Specify multiple inline exceptions in FunctionCommentThrowTagSniff

### LINKED ISSUE #2728: FunctionCommentThrowTagSniff should allow multiple exceptions for the same throws tag
I think this should be valid
```
/**
 * @throws PHP_Exception1|PHP_Exception2
 */
public function allOK()
{
    throw new PHP_Exception1('Error');
    throw new PHP_Exception2('Error');
}
```
Or an option could exist to allow having multiple exception on one throws tag.

Morerice:
This pull request closes #2728

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentThrowTagSniff.php
@@ -177,15 +177,15 @@ class FunctionCommentThrowTagSniff implements Sniff
             }
 
             if ($tokens[($tag + 2)]['code'] === T_DOC_COMMENT_STRING) {
-                $exceptions = $tokens[($tag + 2)]['content'];
-                if (strpos($exceptions, '|') !== false) {
-                    // filter out all spaces and create array containing every exception
-                    $exceptions = explode('|', str_replace(' ', '', $exceptions));
+                $commentContents = $tokens[($tag + 2)]['content'];
+                if (strpos($commentContents, '|') !== false) {
+                    $exceptions = explode('|', $commentContents);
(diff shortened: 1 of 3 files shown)


Morerice:
The build failure is addressed in PR #2793

gsherwood:
I don't think this sniff should allow this syntax. The intention of this Squiz sniff is to ensure each exception is individually documented. Merging multiple exceptions doesn't allow for this type of documentation, and goes against the phpDoc recommendation as well: https://docs.phpdoc.org/references/phpdoc/tags/throws.html

Happy to hear other opinions.

jrfnl:
I concur with @gsherwood. Using union types also goes against the recommendation in the (draft) PSR 19:

> It is also RECOMMENDED that this tag occurs for every occurrence of an exception, even if it has the same type. By documenting every occurrence a detailed view is created and the consumer knows for which errors to check.

Source: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#514-throws

Morerice:
I understand the problem. Should both the issue and the pr be closed?

gsherwood:
> Should both the issue and the pr be closed?

Good idea.

VincentLanglet:
Using union types for the `@throws` tag also goes against the recommendation in the (draft) PSR 19:

> > It is also RECOMMENDED that this tag occurs for every occurrence of an exception, even if it has the same type. By documenting every occurrence a detailed view is created and the consumer knows for which errors to check.
> 
> Source: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#514-throws

I wasn't aware about this recommendation.

Does a sniff checking that a `@throws` type does not use union types would be accepted ?

Issue #2790, opened 2019-12-31
PSR12.Traits.UseDeclaration ignores block comments


maras3000:
/src/Standards/PSR12/Sniffs/Traits/UseDeclarationSniff.php ignores block comments.

For example this will cause the `The first trait import statement must be declared on the first non-comment line after the class opening brace` error to be shown: 
```
class MyClass
{
    /*
    some comment
    */
    use RegistersUsers;
```

while this will not:

```
class MyClass
{
    use RegistersUsers;

    /*
    some comment
    */
```

Synchro:
This issue is triggered in recent Laravel versions: https://github.com/laravel/laravel/blob/master/app/Http/Controllers/Auth/ConfirmPasswordController.php#L22

VincentLanglet:
```
class Foo implements Bar
{
    /**
     * Comment here;
     */
    /**
     * Another Comment here;
     */
    use Baz;
}
```
Is also reported as an error.

Synchro:
> I think the error is correctly reported for recent Laravel versions since there is a blank line.

That's true, but only half the story. phpcbf fixes that by removing the blank line, but it then runs into this exact problem.

VincentLanglet:
> > I think the error is correctly reported for recent Laravel versions since there is a blank line.
> 
> That's true, but only half the story. phpcbf fixes that by removing the blank line, but it then runs into this exact problem.

Indeed. My PR can fix the second half of the story by fixing the false positive error for.
```
class Foo implements Bar
{
    /**
     * Comment here;
     */
    /**
     * Another Comment here;
     */
    use Baz;
}
```

gsherwood:
Can confirm that the PR fixes the original issue, and also sorts out that Laravel issue (the blank line should be removed, but no error should then be reported).

ccapotehrd:
> > I think the error is correctly reported for recent Laravel versions since there is a blank line.
> 
> That's true, but only half the story. phpcbf fixes that by removing the blank line, but it then runs into this exact problem.

Same thing still happening to me... `phpcbf` 'fixed' (removed) the empty line between the comments and the trait declaration but `phpcs` keeps showing an error on the file: 

`ERROR | [x] The first trait import statement must be declared on the first non-comment line after the class opening brace`

PR #2799, opened 2020-01-06
Fix use declaration

VincentLanglet:
Fix https://github.com/squizlabs/PHP_CodeSniffer/issues/2790


gsherwood:
Thanks a lot for fixing this

Issue #2238, opened 2018-11-22
how to disable a single comarisonOperatorUsage


adminfriso:
Hello,
I am working on a custom sniff for my (shared) projects. I would lik to allow the T_BOOLEAN_NOT as valid operator on the Squiz.Operators.ComparisonOperatorUsage.NotAllowed rule.
right now the only way (I could find)  to exclude this is to exclude the complete rule, but then you will also lose the other checks.

    <rule ref="Squiz.Operators.ComparisonOperatorUsage">
        <exclude name="Squiz.Operators.ComparisonOperatorUsage.NotAllowed"/>
    </rule>

is there a way to only exclude this single rule?

dingo-d:
You can silence it, that way you won't be including the entire ruleset

```xml
<rule ref="Squiz.Operators.ComparisonOperatorUsage.NotAllowed">
  <severity>0</severity>
</rule>
```

adminfriso:
But then i do also silence the check for the other valid and invald OPS

adminfriso:
for example
 61 | ERROR | Operator ! prohibited; use === FALSE instead // Do not want to check for this
 66 | ERROR | Operator != prohibited; use !== instead // Do want to check for this

dingo-d:
Oh, I see. You could remove this one, and create your own that is modified to your liking (basically fork the current one and make modifications that fit your use case).

adminfriso:
how do you mean? and or is there a way to do it in the current package?

dingo-d:
I'm assuming that you are just using `phpcs.xml.dist` in your project and cherrypicking the rules you want to (the usual way).

What I did was I [created a standard](https://github.com/infinum/coding-standards-wp) for the company I work for, and I added a custom sniff. You could do something like it, define your own sniff and it will be included in your ruleset.

Or you could fork the phpcs and change the sniff in question and then load your fork in the composer. This is a bit more work as you need to keep track of the changes in the upstream branch.

jrfnl:
@adminfriso Just a heads-up: The simplest way to create your own version of this sniff would be to just extend the PHPCS sniff and overload the `$invalidOps` property. That way you are not infringing on the copyright of the code.

Unfortunately that property appears to be `private`, but a PR to this repo to change the property from `private` to `protected` should fix that.
[...]

PR #2803, opened 2020-01-08
Add strict comparison sniff

VincentLanglet:
WIP for https://github.com/squizlabs/PHP_CodeSniffer/issues/2238

### DIFF
--- a/src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
+++ b/src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * Ensures that constant names are all uppercase.
+ *
+ * @author    Vincent Langlet <vincentlanglet@exemple.com>
+ * @copyright 2006-2015 Squiz Pty Ltd (ABN 77 084 670 600)
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ */
+
+namespace PHP_CodeSniffer\Standards\Generic\Sniffs\Operators;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+
+class StrictComparisonSniff implements Sniff
+{
+
+    /**
+     * Types to replace: key is operator to replace, value is operator to replace with.
+     *
+     * @var array
+     */
+    public $operators = [
+        T_IS_EQUAL     => '===',
+        T_IS_NOT_EQUAL => '!==',
+    ];
+
+
+    /**
+     * Returns an array of tokens this test wants to listen for.
+     *
+     * @return array
+     */
+    public function register()
+    {
+        return [
+            T_IS_EQUAL,
+            T_IS_NOT_EQUAL,
+        ];
+
+    }//end register()
+
+
+    /**
+     * Processes this test, when one of its tokens is encountered.
+     *
+     * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+     * @param int                         $stackPtr  The position of the current token in the
+     *                                               stack passed in $tokens.
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+
+        $phpcsFile->addError(
+            'The %s comparator is forbidden, use %s instead',
+            $stackPtr,
+            'NotStrict',
+            [
+                $tokens[$stackPtr]['content'],
+                $this->operators[$tokens[$stackPtr]['code']],
+            ]
+        );
+
+    }//end process()
+
+
+}//end class
--- a/src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
+++ b/src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
@@ -1,9 +1,9 @@
 <?php
 /**
- * Ensures that constant names are all uppercase.
+ * Discourages the use of not strict operators.
  *
  * @author    Vincent Langlet <vincentlanglet@exemple.com>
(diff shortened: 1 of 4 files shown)


src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Returns an array of tokens this test wants to listen for.
+     *
+     * @return array
+     */
+    public function register()
+    {
+        return [
+            T_IS_EQUAL,
```
jrfnl: What's the point of having the public property if you don't use it in `register()` ? People could now add additional operators via the public property, but they would never be sniffed for.

src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
```diff
@@ -0,0 +1,70 @@
...
+     *                                               stack passed in $tokens.
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+
+        $phpcsFile->addError(
+            'The %s comparator is forbidden, use %s instead',
```
jrfnl: `comparator` is an electronic circuit.

```suggestion
            'Using the %s operator is forbidden, use %s instead',
```

src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
```diff
@@ -0,0 +1,70 @@
+<?php
+/**
+ * Ensures that constant names are all uppercase.
+ *
+ * @author    Vincent Langlet <vincentlanglet@exemple.com>
+ * @copyright 2006-2015 Squiz Pty Ltd (ABN 77 084 670 600)
```
jrfnl:
```suggestion
 * @copyright 2020 Squiz Pty Ltd (ABN 77 084 670 600)
```

src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
```diff
@@ -0,0 +1,70 @@
+<?php
+/**
+ * Ensures that constant names are all uppercase.
```
jrfnl: Copy/pasta.

src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
```diff
@@ -0,0 +1,70 @@
...
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+
+        $phpcsFile->addError(
+            'The %s comparator is forbidden, use %s instead',
+            $stackPtr,
+            'NotStrict',
```
jrfnl:
```suggestion
            'Forbidden',
```

src/Standards/Generic/Tests/Operators/StrictComparisonUnitTest.inc
```diff
@@ -0,0 +1,12 @@
...
+if ($true == true) {
+    echo 'True';
+} elseif ($true != true) {
+    echo 'False';
+} elseif (false === $true && true !== $false) {
+    echo 'False';
+}
+
+$a = $b == $c;
+$a = $b != $c;
```
jrfnl: Please add tests which test the public property as well.

src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
```diff
@@ -0,0 +1,70 @@
...
+class StrictComparisonSniff implements Sniff
+{
+
+    /**
+     * Types to replace: key is operator to replace, value is operator to replace with.
+     *
+     * @var array
+     */
+    public $operators = [
+        T_IS_EQUAL     => '===',
```
jrfnl: Using the constant as the key will make this property nearly impossible to use as the value of the token constants may change between PHP versions.

Suggestion: use the type instead, i.e `'T_IS_EQUAL'` (as a string) to allow for setting this from a ruleset in a reliable manner.

src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
VincentLanglet: Whoah, I'm tired. Thanks.

src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
VincentLanglet: How can I get the code in the register method from the type then ?

jrfnl: You mean this ?

```php
$targets = [];
foreach ($this->operators as $type => $replacement) {
    $targets[] = constant($type);
}

return $targets;
```

VincentLanglet:
@jrfnl I'm not sure about the "generic" `ForbiddenOperators` (or even `ForbiddenTokens`). It does lose context and it forces the user to configure this Sniff. There is already a `ValidLogicalOperatorsSniff` and the `BacktickOperatorSniff` for example and I think there're better than a Generic. But an AbstractForbiddenTokensSniff can be made, and they can extends it.

After some thinking, I feel like the $forbiddenOperator option should not be even needed.
The issue with the `ComparisonOperatorUsageSniff` was the fact this sniff check multiple things in one sniff and some people wanted to remove one or multiple checks (never adding one).
By splitting this Sniff in 3 minimal sniff, 
- One for === and !==.
- One for ! usage.
- One for implicit True.
you can take exactly what you want.

src/Standards/Generic/Sniffs/Operators/StrictComparisonSniff.php
VincentLanglet: Yes, thanks, I didn't know the constant function.

PR #2804, opened 2020-01-09
PHP4-style typed properties not tokenized correctly

jrfnl:
Typed properties are explicitly also supported for PHP 4-style properties using just the `var` or `static` keyword.

However, in that case, the nullable indicator `?` was not converted from `T_INLINE_THEN` to `T_NULLABLE`.

Fixed now.

Includes unit tests via the `File::getMemberProperties()` method.

Ref: https://wiki.php.net/rfc/typed_properties_v2


gsherwood:
Thanks

PR #2805, opened 2020-01-09
Undefined Offset notice during live coding of arrow functions

jrfnl:
Defensive coding for the arrow functions backfill. This fixes a number of "undefined index" PHP notices when arrow functions are being declared during live coding.

```
Notice: Undefined offset: 561 in /path/to/PHP_CodeSniffer/src/Tokenizers/PHP.php on line 1807
Notice: Undefined offset: 561 in /path/to/PHP_CodeSniffer/src/Tokenizers/PHP.php on line 1808
Notice: Undefined offset: 561 in /path/to/PHP_CodeSniffer/src/Tokenizers/PHP.php on line 1815
```

It also fixes a bug where an inner for loop was using the wrong variable for the condition part.

Includes unit test.

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -1801,9 +1801,9 @@ class PHP extends Tokenizer
                 }//end if
 
                 continue;
-            } else if ($this->tokens[$i]['code'] === T_FN) {
+            } else if ($this->tokens[$i]['code'] === T_FN && isset($this->tokens[($i + 1)]) === true) {
                 // Possible arrow function.
-                for ($x = ($i + 1); $i < $numTokens; $x++) {
+                for ($x = ($i + 1); $x < $numTokens; $x++) {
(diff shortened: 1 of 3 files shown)


src/Tokenizers/PHP.php
```diff
@@ -1812,7 +1812,7 @@ protected function processAdditional()
                     }
                 }
 
-                if ($this->tokens[$x]['code'] === T_OPEN_PARENTHESIS) {
+                if (isset($this->tokens[$x]) === true && $this->tokens[$x]['code'] === T_OPEN_PARENTHESIS) {
```
tvanrielwendrich: [isset](https://www.php.net/manual/en/function.isset.php) Always returns `bool`. `=== true` is redundant and should be removed.

jrfnl: @tvanrielwendrich While that is true, the PHPCS native code style rules **demand** the `=== true`, so no, this cannot be removed.

gsherwood:
Thanks

PR #2809, opened 2020-01-10
docs(File): Fix data type of token types that can also be string

klausi:
PHPStan error report in Coder:
```
Parameter #2 $ignore of method PHP_CodeSniffer\Files\File::findStartOfStatement() expects array|int|null, string given.
```
But token types can also be strings, let's fix the data types in the doc blocks here.


gsherwood:
Thanks. There are probably quite a few of these around - from back when even the custom token types were all ints.

Issue #2813, opened 2020-01-11
Implement codeclimate reporter


stephanoff:
It is nice to have Codeclimate reporter plugin to be able to integrate codesniffer with gitlab-ci.

jrfnl:
@stephanoff Could you elaborate a little on what you mean by this ? Maybe add some links too ?

stephanoff:
There is existing tool called Codeclimate used for the GitLab CI AutoDevOps.
There is an example their report implementation https://github.com/codeclimate/codeclimate-phpcodesniffer
But running Codeclimate on the large project is not an option since common case is use PHPCS installed from composer.
I didn't happy about their approach hardcoding report generator inside their package.

jrfnl:
@stephanoff Ah, I understand now and I agree it would be better to have a `report` type for that, but that doesn't necessarily have to be in PHPCS itself and as your PR contains hard-coded sniff names and such which would need regular maintenance and is tied into the `composer.json` for CodeClimate, I'm wondering if it wouldn't be better to pull your report to the CodeClimate repo ?

An external repo can use their own report by name. See https://github.com/squizlabs/PHP_CodeSniffer/issues/1942 and https://github.com/squizlabs/PHP_CodeSniffer/pull/1948 for how to implement this.

stephanoff:
Great, thank you!
Will do it as a separate reporter. I am not sure about integrating this reporter into CodeClimate, will ask them.

PR #2814, opened 2020-01-11
#2813 Add Codeclimate report generator

stephanoff:
Added Codeclimate report generator.

### DIFF
--- a/src/Reports/Codeclimate.php
+++ b/src/Reports/Codeclimate.php
@@ -1,0 +1,340 @@
+<?php
+
+namespace PHP_CodeSniffer\Reports;
+
+use PHP_CodeSniffer\Files\File;
+
+/**
+ * Codeclimate report for PHP_CodeSniffer.
+ */
+class Codeclimate implements Report
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function generateFileReport(
+        $report,
+        File $phpcsFile,
+        $showSources = false,
+        $width = 80) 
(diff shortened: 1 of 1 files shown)


src/Reports/Codeclimate.php
```diff
@@ -0,0 +1,340 @@
...
+        "WordPress.Classes.ValidClassName" => 50000,
+        "WordPress.Files.FileName" => 50000,
+        "WordPress.Functions.FunctionRestrictions" => 50000,
+        "WordPress.NamingConventions.ValidFunctionName" => 50000,
+        "WordPress.NamingConventions.ValidVariableName" => 50000,
+        "WordPress.PHP.DiscouragedFunctions" => 50000,
+        "WordPress.PHP.StrictComparisons" => 50000,
+        "WordPress.PHP.StrictInArray" => 50000,
+        "WordPress.PHP.YodaConditions" => 50000,
+        "WordPress.VIP.AdminBarRemoval" => 50000,
```
jrfnl: This looks like CodeClimate is using an old and outdated version of WPCS. Those `VIP` sniffs don't exist anymore.

jrfnl: Actually, looks like all dependencies are way out of date. Last update of their `composer.json`/`composer.lock` is from sept 2018.

jrfnl:
For anyone coming across this: see #2813 for the discussion on why this was closed.

Issue #2077, opened 2018-06-29
Check doc comment only for public methods.


obstschale:
I am aware of the `Squiz.Commenting.BlockComment` sniff. But is it possible to only check BlockComments for `public` methods and ignore `private` and `protected` ones?

gsherwood:
There is no option to do that with the current sniff. A sniff property would need to be created for this.

obstschale:
Unfortunately, I don't know how to do this. So I have to wait for you or someone to implement this. Thank you for your reply.

VincentLanglet:
> Unfortunately, I don't know how to do this. So I have to wait for you or someone to implement this. Thank you for your reply.

Are you still interested ?

obstschale:
Yes, I am. But still, I don't know how and I currently have no time to dig into it. :)

VincentLanglet:
I can make a try. What is you exact need ?
I have difficulty to understand how/why you want to check only `public` method.

For what I understand from this sniff this is how it works.
```
/** This comment is checked */

/**
 * This comment is not checked
 *
 * @param $a
 */
function foo() {
 /** This comment is checked */
}
```

obstschale:
We have decided at our team to document every public method, cause they act like an public API for other parts of our code / modules. So these methods should be documented. private and protected methods are not mandatory to document, cause they are only used withing that class.

Example:

```php
class Foo {

  /**
   * Mandatory DocBlock.
   * If this is DocBlock is missing
   * CodeSniffer should fail.
   */
  public function one() {}

  /**
   * Not mandatory.
   * Code Sniffer does not care if this exists.
   */
  protected function two() {}


  /**
   * Not mandatory.
   * Code Sniffer does not care if this exists.
   */
  private function two() {}

}
```

VincentLanglet:
@obstschale Ok, I understand.
But you talked about the `Squiz.Commenting.BlockComment` sniff, but I think this sniff does not check there is a `BlockComment function`.

For what I know, this is the `Squiz.Commenting.FunctionComment` sniff, extending the `PEAR.Commenting.FunctionComment` one.

These sniffs are checking :
- Function has doc comment
- `/**` style is used for function comment
- The @return tag
- The @param tags
- The @throws tags.

And then if you also use the `Generic.Commenting.DocComment` sniff, it checks the doc comment will start with a short description.

So it seems more difficult to make something coherent, I need more example.

```
class Foo {

  /**
[...]

PR #2815, opened 2020-01-12
[PEAR.Commenting.FunctionComment] Add option to disable the check for protected/private function

VincentLanglet:
Close https://github.com/squizlabs/PHP_CodeSniffer/issues/2077

### DIFF
--- a/src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
@@ -16,6 +16,14 @@ use PHP_CodeSniffer\Util\Tokens;
 class FunctionCommentSniff implements Sniff
 {
 
+    /**
+     * Disable the check for functions with a lower visibility than the value given.
+     * Allowed values are public, protected and private.
+     *
+     * @var string
+     */
+    public $minimumVisibility = 'private';
+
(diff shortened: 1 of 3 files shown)


src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -41,16 +49,29 @@ public function register()
...
     {
         $tokens = $phpcsFile->getTokens();
-        $find   = Tokens::$methodPrefixes;
-        $find[] = T_WHITESPACE;
 
-        $commentEnd = $phpcsFile->findPrevious($find, ($stackPtr - 1), null, true);
+        $ignore   = Tokens::$emptyTokens;
+        $ignore[] = T_STATIC;
+
+        $scopeModifier = $phpcsFile->findPrevious($ignore, ($stackPtr - 1), null, true);
```
gsherwood: Might be worth using `$phpcsFile->getMethodProperties($stackPtr)` to gather the method visibility + just check the `scope` index of the return value, instead of looking for it yourself. Makes a bit more future-proof.

VincentLanglet: @gsherwood Done

PR #2816, opened 2020-01-12
docs(LineLength): Method does not return anything, so return type is void

klausi:
I'm overriding this method in Coder and then PHPStan complains that the return type is not correct. Should be void.


gsherwood:
Thanks for the correction

Issue #1932, opened 2018-03-12
Generic.ControlStructures.InlineControlStructure fixer moves new PHPCS annotations


jrfnl:
This is one of the last few sniffs in which I *know* there are still issues with the PHPCS annotations.

In this case, the fixer moves trailing PHPCS annotations down to subsequent lines which changes their meaning completely.

Example code for the unit tests:
```php
while (!$this->readLine($tokens, $tag)); //phpcs:ignore Standard.Category.Sniff

foreach ($stringParade as $hit)
    $hitParade[] = $hit + 0; // phpcs:disable Standard.Category.Sniff

if ($bar)
    if ($foo) echo 'hi'; /* @phpcs:ignore Standard.Category.Sniff */
```

Expected fix:
```php
while (!$this->readLine($tokens, $tag)) {} //phpcs:ignore Standard.Category.Sniff
}
foreach ($stringParade as $hit) {
    $hitParade[] = $hit + 0; // phpcs:disable Standard.Category.Sniff
}

if ($bar) {
    if ($foo) { echo 'hi'; /* @phpcs:ignore Standard.Category.Sniff */
    }
}
```

Fix currently being made by the sniff as it is now:
```php
while (!$this->readLine($tokens, $tag)) {} //phpcs:ignore Standard.Category.Sniff

foreach ($stringParade as $hit) {
    $hitParade[] = $hit + 0;
} // phpcs:disable Standard.Category.Sniff

if ($bar) {
    if ($foo) { echo 'hi';
    }
} /* @phpcs:ignore Standard.Category.Sniff */
```

Changing conditions which look for `T_COMMENT`, to ones which also take `Tokens::$phpcsCommentTokens` into account, does not really help all that much (the first one is worse, the second better, the third still just as bad):
```php
while (!$this->readLine($tokens, $tag)) { 
//phpcs:ignore Standard.Category.Sniff
}
foreach ($stringParade as $hit) {
    $hitParade[] = $hit + 0; // phpcs:disable Standard.Category.Sniff
}

if ($bar) {
    if ($foo) { echo 'hi';
    }
} /* @phpcs:ignore Standard.Category.Sniff */
```

I've stared at this one way too long by now and haven't figured out a solution yet.

I'm pretty sure the issue is somewhere in these lines:
[...]

PR #2827, opened 2020-01-16
Generic/InlineControlStructure: ignore while without body

### LINKED ISSUE #2822: Generic/InlineControlStructures: sniff does not allow while without body
A `while` control structure without body is perfectly valid PHP but gets flagged by the `Generic.ControlStructures.InlineControlStructures` sniff.

```
$i = 10;
while ($i > 0 && --$i);

// $i is now 0
```
See: https://3v4l.org/RXNqL

As far as I can see, the sniff is intended to find inline control structures *with* a body, but *without* braces and fix those. An inline control structure without a body should be outside the scope of this sniff.

I've got a fix ready for this, but found that a number of existing unit test would start failing (line 42, 43, 226, 228 and 230).

@gsherwood With that in mind, I'd like second opinion on how the sniff should handle `while` structures without body.

As an alternative to the fix I've got prepared, I could create a fix which would maintain the existing behaviour, but would give `while` structures without a body a different error code.

Opinions ?

jrfnl:
As discussed in issue #2822, this changes the sniff to ignore `while()` and `for`constructs without body.

This - partially - undoes the unit tests added for issue #1932, but there are still some remaining covering that issue.

Fixes #2822

---

**Edit**: I've updated the PR to now also allow for `for` structures without body. See the accompanied unit test.


gsherwood:
Thanks for fixing this up. I actually had no idea you could have WHILEs and FORs with no body. I guess I just never tried.

jrfnl:
I was running some tests and it turns out that even `foreach` and `if` statements can be without body, though I can't imagine a case in which that could ever useful, so unless someone reports a bug for it, I don't think we need to account for it.

PR #2840, opened 2020-01-27
NullableTypeDeclaration: fix CLI documentation

jrfnl:
* The line length of the code samples was too long.
* The second code sample would not be shown as it was not accompanied by a `valid` case, nor wrapped in its own `code_comparison` block.

#### Was:
image

#### Fixed to:
image


gsherwood:
Thanks

PR #2843, opened 2020-01-30
Tokenizer does not support alternative syntax for declare statements

jrfnl:
The `declare` language construct can also use the alternative control structure syntax, but in that case would not get assigned `scope_opener` or `scope_closer` indexes.

This fixes that.

Includes unit test via the `PSR1.Files.SideEffects` sniff, which also needed as small tweak to disregard the semicolon after the `enddeclare`.


gsherwood:
Thanks a lot for this.

jrfnl:
You're welcome ;-)

PR #2846, opened 2020-01-31
Fix path

remicollet:
Consistency with others, (and `PHP_CodeSniffer` directory may have different name, e.g. PHP_CodeSniffer-3.5.4 from pear Archive)

### DIFF
--- a/tests/Core/Ruleset/RuleInclusionTest.xml
+++ b/tests/Core/Ruleset/RuleInclusionTest.xml
@@ -31,7 +31,7 @@
 	    </properties>
 	</rule>
 
-	<rule ref="./../PHP_CodeSniffer/src/Standards/Generic/Sniffs/NamingConventions/CamelCapsFunctionNameSniff.php">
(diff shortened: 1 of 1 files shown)


tests/Core/Ruleset/RuleInclusionTest.xml
```diff
@@ -31,7 +31,7 @@
 	    </properties>
 	</rule>
 
-	<rule ref="./../PHP_CodeSniffer/src/Standards/Generic/Sniffs/NamingConventions/CamelCapsFunctionNameSniff.php">
+	<rule ref="./src/Standards/Generic/Sniffs/NamingConventions/CamelCapsFunctionNameSniff.php">
```
jrfnl: You are breaking the test. The whole point of the test is that path resolution is checked and as this is a test file, it is save to presume the path.

Suggest: close.

remicollet: Sorry, but `PHP_CodeSniffer` only exists if you clone  the git repo under this name... and not from the pear archive

remicollet: btw, pr#2847 will skip it

PR #2847, opened 2020-01-31
Allow to force PHP_CODESNIFFER_PEAR from env

remicollet:
Needed when test suite run from pear archive (not from pear installation tree)

### DIFF
--- a/tests/AllTests.php
+++ b/tests/AllTests.php
@@ -9,7 +9,7 @@
 
 namespace PHP_CodeSniffer\Tests;
 
-$GLOBALS['PHP_CODESNIFFER_PEAR'] = false;
+$GLOBALS['PHP_CODESNIFFER_PEAR'] = (getenv("PHP_CODESNIFFER_PEAR") ? true : false);
 
 if (is_file(__DIR__.'/../autoload.php') === true) {
     include_once 'Core/AllTests.php';


remicollet:
Running test suite from PEAR archive:

```
There was 1 failure:

1) PHP_CodeSniffer\Tests\Core\Ruleset\RuleInclusionAbsoluteLinuxTest::testLinuxStylePathRuleInclusion
Failed asserting that 1 is identical to '10'.

/dev/shm/BUILD/php-pear-PHP-CodeSniffer-3.5.4/PHP_CodeSniffer-3.5.4/tests/Core/Ruleset/RuleInclusionAbsoluteLinuxTest.php:112
/dev/shm/BUILD/php-pear-PHP-CodeSniffer-3.5.4/PHP_CodeSniffer-3.5.4/tests/TestSuite7.php:28

```

jrfnl:
That should already be handled by line 21 in the same file and that test is explicitly skipped on PEAR. 

https://github.com/squizlabs/PHP_CodeSniffer/blob/dceec07328401de6211037abbb18bda423677e26/tests/AllTests.php#L21

https://github.com/squizlabs/PHP_CodeSniffer/blob/dceec07328401de6211037abbb18bda423677e26/tests/Core/Ruleset/RuleInclusionAbsoluteLinuxTest.php#L48-L53

remicollet:
> That should already be handled by line 21 i

No as autoload.php exist when you unpack the pear archive.

jrfnl:
Hmm.. it shouldn't exist in that location as the PEAR directory layout is different with the package files being placed in the `PEAR/PHP/CodeSniffer` directory and the tests being placed in `tests/PHP_CodeSniffer`. The PHPCS autoload file should not be available that way in a PEAR install.

jrfnl:
Oh hang on... you mean you download a `zip` from the PEAR website and unpack that instead of using a `PEAR` install ?
That's not a supported method of installing PHPCS for development.

remicollet:
But that the way used for distribution packages for ages... and it have always works

jrfnl:
Ok, I see what you mean. Looks like this was introduced in https://github.com/squizlabs/PHP_CodeSniffer/commit/68048f3df9defae0d3995237dd5cdd1dc24dc6dc

As it worked previously, I agree it should be fixed.

From a code perspective, may I suggest removing the check from the `AllTests.php` file and adding it as a `define` in the test bootstrap file which contains more test related `define`s ?

I should have suggested that on the above referenced commit, but as that was pushed in without a pull request I never reviewed it.

There are currently a few tests which use the global var to skip the test, those should then be adjusted to use the constant instead.

Would that work for you ?

remicollet:
> Would that work for you ?

Make sense to use the bootstrap file.

remicollet:
Thinking a bit more about this

there is mostly 2 run cases

* single tree, from git or pear archive
* split tree, from pear installation

So, IMHO, proper condition is `is_dir('../src')` is the boostrap file.

jrfnl:
> So, IMHO, proper condition is is_dir('../src') is the boostrap file.

Makes sense to me.

And a limited set of tests should be skipped when on the split tree pear installation, but should run in both the single tree situations.

gsherwood:
I've taken a look into this and I don't see how changing the PEAR check is going to do anything. It doesn't look like the unit test error has anything to do with the archive itself - it's the name of the directory that you extract the archive to.

This code in `tests/Core/Ruleset/RuleInclusionTest.xml` is the problem:
```xml
<rule ref="./../PHP_CodeSniffer/src/Standards/Generic/Sniffs/NamingConventions/CamelCapsFunctionNameSniff.php">
    <properties>
        <property name="strict" value="false" />
    </properties>
</rule>
```
It makes the assumption that the directory above `src` is called `PHP_CodeSniffer`, which is not the case if you untar a PEAR archive without specifying a new directory name. By default, the directory will be named something like `PHP_CodeSniffer-3.5.4`, which causes the sniff to not be found.

If I change the name of the directory, the unit tests work fine without no code modifications.

gsherwood:
I think the original issue is actually caused by a core issue resolving a sniff name. This is how I replicated it anyway.

I extracted a PEAR archive into a directory called `PHP_CodeSniffer-3.5.4` and ran the tests from in there. As part of testing, the RuleInclusionAbsoluteLinux test, PHPCS tries to resolve the following sniff file: `/path/to//PHP_CodeSniffer-3.5.4/src/Standards/Generic/Sniffs/Formatting/SpaceAfterNotSniff.php`

The `Ruleset::processRule()` method splits this on dots, getting back:
```
Array
(
    [0] => /path/to/PHP_CodeSniffer-3
    [1] => 5
    [2] => 4/src/Standards/Generic/Sniffs/Formatting/SpaceAfterNotSniff
    [3] => php
)
```
Because of the extra dots, this is no longer a 2-part string, so the method doesn't attempt to determine the sniff code (it thinks it is one).

gsherwood:
I have a fix for this, but I need to do a bit more testing.

But with the fix to processRule(), and the change to the RuleInclusionTest.xml file to not assume the dirname, I can run the tests after extracting a PEAR archive.
```
$ tar -xzf PHP_CodeSniffer-3.5.5.tgz
$ cd PHP_CodeSniffer-3.5.5
$ phpunit
PHPUnit 4.8.36 by Sebastian Bergmann and contributors.

......................................S........................  63 / 506 ( 12%)
............................................................... 126 / 506 ( 24%)
............................................................... 189 / 506 ( 37%)
............................................................... 252 / 506 ( 49%)
...............................S............................... 315 / 506 ( 62%)
............................S.................................. 378 / 506 ( 74%)
.......SSSS.................................................... 441 / 506 ( 87%)
............................................SS................. 504 / 506 ( 99%)
..

431 sniff test files generated 724 unique error codes; 441 were fixable (60%)

Time: 5.16 seconds, Memory: 36.00MB

OK, but incomplete, skipped, or risky tests!
Tests: 506, Assertions: 1394, Skipped: 9.
```

jrfnl:
@gsherwood I'm presuming your fix will not remove the whole point of that particular test case ? which is having `./../` in the path at the start ;-)

gsherwood:
I've committed a fix for both issues. The tests pass from a git clone, an extracted PEAR archive, and from a PEAR install (where they continue to be skipped). I'm going to close this as I believe the original issue is now fixed and the 3.5.5 release will work fine.

Thanks for raising this issue.  It sounded like a PEAR/tests compatibility problem, but it turned out to be a problem with PHPCS itself.

PR #2853, opened 2020-02-04
Undefined variable error when using Info report

jrfnl:
This partially reverts the change made to this particular file in e9ebf5253a3e6dd3cc4b584ce54b1f4c34fcb1f3

The `floor( $countWidth / 3)` was causing an _undefined variable_ notice as the original definition of the variable was removed.


gsherwood:
Thanks a lot for fixing this

jrfnl:
You're welcome.

PR #2860, opened 2020-02-08
[Hotfix] PHP Tokenizer: improve arrow function backfill consistency

### LINKED ISSUE #2859: Wrong T_FN token when $this->fn
Having `$xxx->fn` in the code is reported as a `T_FN`, which is wrong.

Example:

```php
<?php
class Test {
    public function foo() {
        $this->fn = 'a';
    }
}
```

`$this->fn` gives
```php
^ array:8 [
  "code" => "PHPCS_T_FN"
  "type" => "T_FN"
  "content" => "fn"
  "line" => 7
  "column" => 16
  "length" => 2
  "level" => 2
  "conditions" => array:2 [
    2 => 361
    11 => 346
  ]
]
```

Related issue: https://github.com/slevomat/coding-standard/issues/896

jrfnl:
This commit implements the proposal as outlined in https://github.com/squizlabs/PHP_CodeSniffer/issues/2859#issuecomment-583695917.

The `fn` keyword will now only be tokenized as `T_FN` if it really is an arrow function, in which case it will also have the additional indexes set in the token array.

In all other cases - including when the keyword is used in a function declaration for a named function or method -, it will be tokenized as `T_STRING`.

Includes numerous additional unit tests.

Includes removing the changes made to the `File::getDeclarationName()` function and the `AbstractPatternSniff` in commit 37dda44ed3bf34dbaad6e0dd1e9e10b03d67ed00 as those are no longer necessary.

Fixes #2859

### DIFF
--- a/tests/Core/Tokenizer/BackfillFnTokenTest.inc
+++ b/tests/Core/Tokenizer/BackfillFnTokenTest.inc
@@ -73,6 +73,49 @@ fn(array $a) : array => $a;
 /* testTernary */
 $fn = fn($a) => $a ? /* testTernaryThen */ fn() : string => 'a' : /* testTernaryElse */ fn() : string => 'b';
 
+/* testConstantDeclaration */
+const FN = 'a';
+
+class Foo {
+    /* testStaticMethodName */
+    public static function fn($param) {
+        /* testNestedInMethod */
+        $fn = fn($c) => $callable($factory($c), $c);
+    }
+
+    public function foo() {
+        /* testPropertyAssignment */
+        $this->fn = 'a';
+    }
+}
+
+$anon = new class() {
+    /* testAnonClassMethodName */
+    protected function fn($param) {
+    }
+}
+
+/* testNonArrowStaticMethodCall */
+$a = Foo::fn($param);
+
+/* testNonArrowStaticMethodCallWithChaining */
+$a = Foo::fn($param)->another();
+
+/* testNonArrowStaticConstant */
+$a = MyClass::FN;
+
+/* testNonArrowStaticConstantDeref */
+$a = MyClass::FN[$a];
+
+/* testNonArrowObjectMethodCall */
+$a = $obj->fn($param);
+
+/* testNonArrowNamespacedFunctionCall */
+$a = MyNS\Sub\fn($param);
+
+/* testNonArrowNamespaceOperatorFunctionCall */
+$a = namespace\fn($param);
+
 /* testLiveCoding */
 // Intentional parse error. This has to be the last test in the file.
 $fn = fn
--- a/tests/Core/Tokenizer/BackfillFnTokenTest.php
+++ b/tests/Core/Tokenizer/BackfillFnTokenTest.php
@@ -102,28 +102,6 @@ class BackfillFnTokenTest extends AbstractMethodUnitTest
 
 
     /**
-     * Test a function called fn.
-     *
-     * @covers PHP_CodeSniffer\Tokenizers\PHP::processAdditional
-     *
-     * @return void
-     */
-    public function testFunctionName()
-    {
-        $tokens = self::$phpcsFile->getTokens();
-
-        $token = $this->getTargetToken('/* testFunctionName */', T_FN);
-        $this->assertFalse(array_key_exists('scope_condition', $tokens[$token]), 'Scope condition is set');
-        $this->assertFalse(array_key_exists('scope_opener', $tokens[$token]), 'Scope opener is set');
-        $this->assertFalse(array_key_exists('scope_closer', $tokens[$token]), 'Scope closer is set');
-        $this->assertFalse(array_key_exists('parenthesis_owner', $tokens[$token]), 'Parenthesis owner is set');
-        $this->assertFalse(array_key_exists('parenthesis_opener', $tokens[$token]), 'Parenthesis opener is set');
-        $this->assertFalse(array_key_exists('parenthesis_closer', $tokens[$token]), 'Parenthesis closer is set');
-
-    }//end testFunctionName()
-
-
-    /**
      * Test nested arrow functions.
      *
      * @covers PHP_CodeSniffer\Tokenizers\PHP::processAdditional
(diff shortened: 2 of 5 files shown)


tests/Core/Tokenizer/BackfillFnTokenTest.inc
```diff
@@ -73,6 +73,49 @@ fn(array $a) : array => $a;
 /* testTernary */
 $fn = fn($a) => $a ? /* testTernaryThen */ fn() : string => 'a' : /* testTernaryElse */ fn() : string => 'b';
 
+/* testConstantDeclaration */
```
gsherwood: This new test was added but is not checked anywhere. Should it be added to dataNotAnArrowFunction()?

tests/Core/Tokenizer/BackfillFnTokenTest.inc
```diff
@@ -73,6 +73,49 @@ fn(array $a) : array => $a;
...
+    }
+}
+
+/* testNonArrowStaticMethodCall */
+$a = Foo::fn($param);
+
+/* testNonArrowStaticMethodCallWithChaining */
+$a = Foo::fn($param)->another();
+
+/* testNonArrowStaticConstant */
```
gsherwood: This new test was added but is not checked anywhere. Should it be added to dataNotAnArrowFunction()?

tests/Core/Tokenizer/BackfillFnTokenTest.inc
```diff
@@ -73,6 +73,49 @@ fn(array $a) : array => $a;
...
+$a = MyClass::FN;
+
+/* testNonArrowStaticConstantDeref */
```
gsherwood: This new test was added but is not checked anywhere. Should it be added to dataNotAnArrowFunction()?

tests/Core/Tokenizer/BackfillFnTokenTest.php
```diff
@@ -553,27 +531,85 @@ public function testTernary()
...
+     *
+     * @return void
+     */
+    public function testNotAnArrowFunction($testMarker)
     {
         $tokens = self::$phpcsFile->getTokens();
 
-        $token = $this->getTargetToken('/* testLiveCoding */', [T_STRING, T_FN]);
-        $this->assertSame($tokens[$token]['code'], T_FN, 'Token not tokenized as T_FN');
+        $token      = $this->getTargetToken('/* testFunctionName */', [T_STRING, T_FN], 'fn');
```
gsherwood: The test marker is hard-coded here so it's not actually using the data provider.

jrfnl:
Saw your remarks and was just fixing up the PR, but I see you've now already merged it.

> I added the missing tests to the data provider and changing the method to use it correctly, but I then received failures, which is to be expected given that the token should be T_STRING and not T_FN:

Actually, that wasn't the problem. The test was already looking for both token types. The problem was that the `$tokenContent` was hard-coded as lowercase `fn`, while the tests related to constants were using uppercase,

> This may require a new method to ensure a T_FN token is not found for these text cases.

That is already being tested in the `testNotAnArrowFunction()` method.

Re the:
> This new test was added but is not checked anywhere. Should it be added to dataNotAnArrowFunction()?

Sorry about that. I was working on both PHPCS as well as PHPCSUtils at the same time and must have not kept them well enough in sync.

I'd be happy to send in the fixes I just committed locally, but as you already merged this now, I presume that's no longer needed ?

tests/Core/Tokenizer/BackfillFnTokenTest.php
jrfnl: Thanks for catching that. Working on too many things at the same time :cold_sweat:

gsherwood:
> I'd be happy to send in the fixes I just committed locally, but as you already merged this now, I presume that's no longer needed ?

Sorry, was in the middle of a few things and pushed this without thinking. Have corrected the tests as per your advice and all is working well now. Thanks.

jrfnl:
@gsherwood Looks like we did the same thing. I saw the tests hadn't been edited and just pulled #2863. I suppose that can be closed now ?

gsherwood:
> Looks like we did the same thing

I cheated and just made the code lowercase :)

jrfnl:
Looks like we solved it slightly differently. Before I do anything else - should I update/rebase #2863 ? Or do you want to leave it as-is now ?

gsherwood:
> Or do you want to leave it as-is now ?

I'm happy to leave it if you are

jrfnl:
> 
> 
> > Looks like we did the same thing
> 
> I cheated and just made the code lowercase :)

I just saw. And yes, that's cheating :stuck_out_tongue_closed_eyes:

jrfnl:
> 
> 
> > Or do you want to leave it as-is now ?
> 
> I'm happy to leave it if you are

As long as it's being tested we should be good, though the uppercase did add an extra safeguard to the test of course.

gsherwood:
> As long as it's being tested we should be good, though the uppercase did add an extra safeguard to the test of course.

I guess it technically needs both lower and upper (and mixed?) to account for different ways someone might define a constant, but I don't actually think that's going to impact the code here.

But you're fix is still better and more flexible, so if you take the time to rebase it, I'll merge it in and add additional lower/mixed tests for it.

jrfnl:
@gsherwood Ok, will do.

jrfnl:
@gsherwood PR has been updated.

jrfnl:
@gsherwood Just a thought: the change made in #2860 to how arrow functions are now handled/tokenized, I think it may now be safe to add them to the `Tokens::$scopeOpeners` and the `Tokens::$parenthesisOpeners` arrays. What do you think ?

tests/Core/Tokenizer/BackfillFnTokenTest.inc
jrfnl: Resolved via #2863

tests/Core/Tokenizer/BackfillFnTokenTest.inc
jrfnl: Resolved via #2863

tests/Core/Tokenizer/BackfillFnTokenTest.inc
jrfnl: Resolved via #2863

tests/Core/Tokenizer/BackfillFnTokenTest.php
jrfnl: Resolved via #2863

PR #2863, opened 2020-02-09
Tokenizer arrow functions backfill tests: fix it ;-)

jrfnl:
Follow up on #2860. This fixes the tests.

#### Test case file:
* Adds tests with mixed case `fn` keyword.
* Removes a few redundant tests (weren't testing anything which wasn't covered already).

#### Test file:
* Adds a couple of tests which were missing from the data provider.
* Updates the test itself to use a `$testContent` parameter to allow for the `fn` in the tests being non-lowercase.


jrfnl:
Rebased, added mixed case tests and updated the PR description above. Should be good to go.

jrfnl:
Related to #2523

gsherwood:
Thanks a lot

jrfnl:
@gsherwood You're welcome.

PR #2866, opened 2020-02-11
FindEndOfStatementTest: simplify the tests

jrfnl:
* Use the `AbstractMethodUnitTest::getTargetToken()` method instead of doing a `findNext()` with a calculation.
    This makes the `$start` token retrieval more descriptive.
* Assert based on the token pointers, not the sub-array value.
    As for both the "expected" as well as "found" values, the token sub-array is retrieved from the same `$tokens` variable, if the token pointers are the same, the sub-arrays will be too, so doing the comparison based on the sub-array doesn't add any extra value to the test.
    Also: doing an array assertion is more expensive than doing a simple integer based assertion.
* Remove redundant calls to the `getTokens()` method.

### DIFF
--- a/tests/Core/File/FindEndOfStatementTest.php
+++ b/tests/Core/File/FindEndOfStatementTest.php
@@ -22,11 +22,10 @@ class FindEndOfStatementTest extends AbstractMethodUnitTest
      */
     public function testSimpleAssignment()
     {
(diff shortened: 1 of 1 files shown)


jrfnl:
Rebased and updated for the changes in commit 2914011675cd13f8a7d5eae16ed1d5ee4f0e701d

gsherwood:
Thanks for cleaning this up.

Issue #2742, opened 2019-12-05
Generic.PHP.DeprecatedFunctions return PHP Fatal error:  Uncaught ReflectionException


harrisdt15f:
squizlabs/php_codesniffer   version   3.5.3 

PHP version
```
PHP 7.3.11-1+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Oct 24 2019 18:23:06) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.11, Copyright (c) 1998-2018 Zend Technologies
with Zend OPcache v7.3.11-1+ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2018, by Zend Technologies
with Xdebug v2.7.2, Copyright (c) 2002-2019, by Derick Rethans
```

**Framework    Laravel    6.6.1**

os version
```
NAME="Ubuntu"
VERSION="16.04.6 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.6 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```
xml file
[phpcs.xml.txt](https://github.com/squizlabs/PHP_CodeSniffer/files/3964899/phpcs.xml.txt)

Problem
`<rule ref="Generic.PHP.DeprecatedFunctions"/>`

> PHP Fatal error:  Uncaught ReflectionException: Function SjczTDG4TDK0Tweg() does not exist in /var/www/jianghu_entertain/vendor/squizlabs/php_codesniffer/src/Standards/Generic/Sniffs/PHP/DeprecatedFunctionsSniff.php:37
> Stack trace:
> #0 /var/www/jianghu_entertain/vendor/squizlabs/php_codesniffer/src/Standards/Generic/Sniffs/PHP/DeprecatedFunctionsSniff.php(37): ReflectionFunction->__construct('SjczTDG4TDK0Twe...')
> #1 /var/www/jianghu_entertain/vendor/squizlabs/php_codesniffer/src/Ruleset.php(1198): PHP_CodeSniffer\Standards\Generic\Sniffs\PHP\DeprecatedFunctionsSniff->__construct()
[...]

PR #2873, opened 2020-02-17
Generic/DeprecatedFunctions: remove redundant `method_exists()`

jrfnl:
This code snippet was introduced in ec29df437d7ce83bf63f73cdf6c4583b5ea91d8e to allow for running the sniff on HHVM, but HHVM support has been dropped in PHPCS 3.3.1.

In native PHP, the `ReflectionFunctionAbstract::isDeprecated()` method is available since PHP 5.2, so this check is redundant.

Ref: https://www.php.net/manual/en/reflectionfunctionabstract.isdeprecated.php

Loosely related to #2742


gsherwood:
Nice find - thanks.

Issue #1941, opened 2018-03-19
[feature request] Allow for a difference of one in alignment with concat_equals, plus_equals etc.


Dacaspex:
Currently, PHPCS gives a warning for the following code sample
```
$foo = 'foo';
$foo .= 'bar';
```
```
Equals sign not aligned with surrounding assignments; expected 2 spaces but found 1 space
```
In our company, we allow for this code style. I would like to request a feature to add an option that we can disable this warning for this specific case. 

This is the case for concat equals (.=), plus equals (+=), minus equals (-=), division equals (/=), times equals (*=), etc.

### Example samples ###
```
$foo = 'foo';
$foo .= 'bar';

$foo .= 'foo';
$bar = 'bar';

$foo = 1;
$foo += 2;
$foo += 5;
$foo -= 2;
$bar = 6;
```

gsherwood:
The alignment in your examples would drive me crazy :)

This isn't a feature I would add personally, but I'd consider merging a PR if there is a lot of support for it. Otherwise, maybe the sniff could be refactored to allow you to write a custom sniff that extends it but requires a different number of spaces in these cases.

Dacaspex:
Thanks you for considering. I wrote a hacky solution for the time being. I understand that this isn't exactly what you want, but let's see what other people think of it

gsherwood:
A new sniff property will be in 3.6.0 that makes this possible. It is documented here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#genericformattingmultiplestatementalignment

PR #2896, opened 2020-03-09
Add configuration to MultipleStatementAlignmentSniff for changing alignment reference

johnpbloch:
Hi there! First off, thanks for your work on phpcs!

This is more or less a fix the underlying issue in #1941. This PR adds a configuation option to `\PHP_CodeSniffer\Standards\Generic\Sniffs\Formatting\MultipleStatementAlignmentSniff` which shifts the reference point for alignment from the end of the assignment token to the beginning. The current behavior is, of course, the default. With this change, though, someone could add this to their config:

```xml
<rule ref="Generic.Formatting.MultipleStatementAlignment">
    <properties>
        <property name="alignAtEndOfAssignToken" value="false"/>
    </properties>
</rule>
```

phpcs would then identify this as passing the sniff:

```php
$test = 'Hello';
$test .= ' World';
```

Thanks for considering this pull request!

### DIFF
--- a/src/Standards/Generic/Sniffs/Formatting/MultipleStatementAlignmentSniff.php
+++ b/src/Standards/Generic/Sniffs/Formatting/MultipleStatementAlignmentSniff.php
@@ -47,6 +47,23 @@ class MultipleStatementAlignmentSniff implements Sniff
      */
     public $maxPadding = 1000;
 
+    /**
+     * Controls which side of the assignment token is used for alignment
+     *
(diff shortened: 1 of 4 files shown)


gsherwood:
Thanks a lot for this PR.

I ended up shortening the sniff property name so it fits a bit more nicely with the existing ones, but everything worked really well. The new property is documented here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#genericformattingmultiplestatementalignment

PR #2900, opened 2020-03-11
AbstractSniffUnitTest: bug fix - warnings not counted in total

jrfnl:
At the bottom of a test run a message along the lines of `4 sniff test files generated 2 unique error codes; 0 were fixable (0%)` is shown.

The unique error codes, as well as the fixable count and percentage would only include `error` codes and would totally disregard the codes coming from `warning`s.

Fixed now.


jrfnl:
@gsherwood Could this PR be earmarked for one of the upcoming releases ?

gsherwood:
Nice one. I probably assumed that only errors were fixable when I first added that. Bad assumption as there appears to be 5 of them that have fixes.

PR #2902, opened 2020-03-13
Fix some phpstan issues

vladyslavstartsev:
I've fixed most of phpstan level 1 issues. Also I've set level to max and made baseline, so that we can ignore current errors, but not make them in future


src/Standards/Generic/Sniffs/CodeAnalysis/JumbledIncrementerSniff.php
```diff
@@ -89,7 +89,7 @@ public function process(File $phpcsFile, $stackPtr)
 
             if (count($diff) !== 0) {
                 $error = 'Loop incrementor (%s) jumbling with inner loop';
-                $data  = [join(', ', $diff)];
+                $data  = [implode(', ', $diff)];
```
jrfnl: :+1:

src/Standards/MySource/Sniffs/Channels/DisallowSelfActionsSniff.php
```diff
@@ -108,7 +108,9 @@ public function process(File $phpcsFile, $stackPtr)
                 // Function was not in this class, might have come from the parent.
                 // Either way, we can't really check this.
                 continue;
-            } else if ($foundFunctions[$funcData['name']] === 'public') {
+            }
+
+            if ($foundFunctions[$funcData['name']] === 'public') {
```
jrfnl: You could even choose to add this condition (reversed) to the previous `if` and lower the indent level for the code.

src/Standards/MySource/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -72,7 +72,7 @@ public function process(File $phpcsFile, $stackPtr)
             }//end if
         }//end foreach
 
-        if ($hasApiTag === true && substr($tokens[$tag]['content'], 0, 4) !== '@api') {
+        if ($hasApiTag === true && isset($tag) === true && substr($tokens[$tag]['content'], 0, 4) !== '@api') {
```
jrfnl: Condition unnecessary, it will always be `true`.

`$hasApiTag` will only be `true` if the `foreach` loop has been entered and if so, `$tag` will be set.

src/Standards/Squiz/Sniffs/Classes/SelfMemberReferenceSniff.php
```diff
@@ -54,7 +54,7 @@ protected function processTokenWithinScope(File $phpcsFile, $stackPtr, $currScop
             }
         }
 
-        if ($conditionToken !== $currScope) {
+        if (isset($conditionToken) === false || $conditionToken !== $currScope) {
```
jrfnl: Condition is unnecessary. There will always be a `$conditionToken` set as otherwise the `AbstractScopeSniff` wouldn't have kicked in in the first place.

src/Standards/Squiz/Sniffs/Commenting/BlockCommentSniff.php
```diff
@@ -343,7 +343,8 @@ public function process(File $phpcsFile, $stackPtr)
                 $error = 'Last line of comment aligned incorrectly; expected %s but found %s';
                 $fix   = $phpcsFile->addFixableError($error, $commentLines[$lastIndex], 'LastLineIndent', $data);
                 if ($fix === true) {
-                    if (isset($tokens[$line]['orig_content']) === true
+                    if (isset($line) === true
```
jrfnl: Condition unnecessary. You can't have a block comment without lines.

src/Config.php
```diff
@@ -143,9 +143,10 @@ class Config
      *
      * If FALSE, arguments that are not command line options or file/directory paths
      * will be ignored and execution will continue. These values will be stored in
-     * $this->unknown.
```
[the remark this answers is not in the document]
vladyslavstartsev: I was trying to add phpDoc `@see` tag

src/Fixer.php
```diff
@@ -469,6 +469,7 @@ public function replaceToken($stackPtr, $content)
             return false;
         }
 
+        $indent = "\t";
```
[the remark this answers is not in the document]
vladyslavstartsev: It's not.
If you search for `$indent = ` in the code, you will see that it is ALWAYS `$indent = "\t";` but then it concatenates to some other thing.

src/Config.php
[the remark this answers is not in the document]
vladyslavstartsev: will fix this

src/Config.php
```diff
@@ -1205,11 +1206,11 @@ public function processLongArgument($arg, $pos)
                     $eqPos = strpos($arg, '=');
                     try {
                         if ($eqPos === false) {
-                            $this->values[$arg] = $arg;
+                            $this->unknown[$arg] = $arg;
```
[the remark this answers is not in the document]
vladyslavstartsev: It's a good thing to do generally, but I think this PR will be delayed, since there should be tests for `Config.php` (there is no tests).

Also it will be good idea to have code coverage for whole PHP_CodeSniffer

vladyslavstartsev:
I'll rebase commits in the end

gsherwood:
Wide-ranging PRs like this as very hard for me to devote time to because the changes are not focused. I've taken a quick look through and seen a number of changes I would not make, but I don't have time to go through them all because the value is not clear.

It would be great if you could point out what specific changes you think are absolutely necessary. If it's really just code cleanup, then I'm going to have to prioritise this lower than bug fixes.

vladyslavstartsev:
@gsherwood okay, what if I'll only leave phpstan.neon and phpstan-baseline.neon in this PR?
This way you will only fix errors when they occur

vladyslavstartsev:
Done, only added max level and baseline

vladyslavstartsev:
@gsherwood anything else I can do?

PR #2903, opened 2020-03-14
PHP tokenizer: add missing verbose message

jrfnl:
PR #2860 made a change to the tokenizer which changes the token type of an `fn` keyword (back) to `T_STRING` if it turns out not to be an arrow function.

That change did not include a verbose message for tokenizer debugging. Fixed now.


gsherwood:
Thanks. I was working in this part of the file but didn't notice this was missing.

I ended up changing the message to be similar to the "positive" case. It was also printing out the internal `PHPCS_T_FN` token name, but I've hard coded that as `T_FN` because `$i` is always `T_FN` in this condition, unless I've read that wrong.

jrfnl:
Sounds good. I just realized at some point I'd not added that message and that it would probably be useful, so glad it's in the codebase now.

PR #2904, opened 2020-03-14
PHPCS 4.0: various fixes related to the PHP < 7.0 version drop

jrfnl:
## Update the PHP requirements

Per commit 25324b3, the minimum PHP version for PHPCS 4.x will be PHP 7.0.

To that end, update the PHP requirements in:
* the `composer.json` file;
* the `Runner::checkRequirements()` method.

## Update the PHPUnit requirements

Per commit 25324b3, the minimum PHP version for PHPCS 4.x will be PHP 7.0. This means that PHPUnit 4.x and 5.x [no longer need to be supported](https://phpunit.de/supported-versions.html).

## Tokenizer/PHP: remove allowance for PHP 7.0

Remove code which was only needed when PHPCS would be run on PHP < 7.0.

## Generic/DisallowAlternativePHPTags: remove code and tests for PHP < 7.0

The ASP and script style open tags are only recognized as `T_OPEN_TAG`/`T_OPEN_TAG_WITH_ECHO` in PHP < 7.0.

This means that the sniff can now be simplified to only look for `T_INLINE_HTML` and that the tests related to PHP < 7.0 can also be removed.

## Squiz/FunctionComment: remove allowance for PHP < 7.0

As PHP 7.0 will now be the minimum PHP version, this type declaration related code no longer needs special casing.


gsherwood:
Thanks for these changes, although I only ended up cherry picking a couple of the commits so will close this now.

I was mid-way through removing tokenizer support and fixing the PHPUnit tests for PHP < 7.2, which is what I'm going with for the min PHP version. I hadn't yet gotten to changing the sniffs though, so those changes were very helpful. Thanks.

jrfnl:
@gsherwood Sounds good to me. Looks like the only change which was missed when cherry-picking was the removal of the PHPUnit aliases from the `tests/bootstrap.php` file.
I imagine you'll still want to do that: https://github.com/squizlabs/PHP_CodeSniffer/pull/2904/commits/c35bd8ce5abe516850931f7c205d883694dbc91d

gsherwood:
Thanks for that - I did miss those as I was updating PHPunit. Have merged that in now.

PR #2907, opened 2020-03-17
Generic/ArbitraryParenthesesSpacing: minor clean up

jrfnl:
As `declare()` is recognized as a parentheses owner, there is no need for it to be listed in the "additional tokens indicating that parenthesis are not arbitrary" list.

Tested via an existing test: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Generic/Tests/WhiteSpace/ArbitraryParenthesesSpacingUnitTest.inc#L29


gsherwood:
Nice find. Thanks.

Issue #1818, opened 2017-12-29
Disable report for phpcbf


GaryJones:
Is there a way to disable reporting for `phpcbf`?

Something like: `phpcbf -q --report=none .`?

gsherwood:
No, but which output do you want to suppress? The output showing you which files are being fixed and or the final report showing you what was done (or both)?

You could always redirect STDOUT to /dev/null (or NUL for Windows, I think it is) if you want no output at all.

gsherwood:
Sorry, should have mentioned `-q` works, but only suppresses that progress output. The final report still shows, so you are probably wanting to suppress that, which there is no way to do.

GaryJones:
My use case: generating files with zend-code (via a composer script), which doesn't completely comply with PSR-2. Running `phpcbf` automatically after generation, I have no need for any report to show i.e. silent fixing, since it's just removing redundant blank lines that break PSR-2 compatibility.

`> /dev/null` seems to work in terms of hiding the output from `phpcbf`, though Composer is giving me a notice that the script returned with error code 1. 

```json
"generate-types": "vendor/bin/soap-client generate:types --config ./config/generate.php && phpcbf -q src/SoapTypes > /dev/null && echo \"Types have been generated to src/SoapTypes. (Ignore the error code message.)\"",
```

Thanks :-)

gsherwood:
The exit codes tell you what PHPCBF did, and can not be changed using config settings:
- Exit code 0 is used to indicate that no fixable errors were found, so nothing was fixed
- Exit code 1 is used to indicate that all fixable errors were fixed correctly
- Exit code 2 is used to indicate that PHPCBF failed to fix some of the fixable errors it found
- Exit code 3 is used for general script execution errors

I'm not sure what config options you have, but if you can have the script error only on exit codes > 1 then it would make more sense. Otherwise, you'll need to see if you can ignore the exit code for PHPCBF if you don't actually care what it did.

GaryJones:
The exit code here is `1`: `phpcbf` fixed everything that could be fixed (running `phpcbf` again doesn't give me any exit code message).

[Composer custom scripts](https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands) don't appear to have a way to override the exit code, so even though PHPCBF was 100% successful, it's still reported with a big scary red message.

I have found one workaround though. The relevant part of my script:

```sh
phpcbf -q src/SoapTypes > /dev/null || true
```

[...]

PR #2918, opened 2020-03-28
Add GitHub Actions annotations report

BrianHenryIE:
Adds a report type for use in [GitHub Actions](https://github.com/features/actions) which annotates PRs etc with PHPCS errors.

~~Nomenclature is changed to use "warning" for errors that can be automatically fixed in CI and "error" for those that cannot.~~

example


jrfnl (commented):
Oh nice! Shiny! I like this idea very much indeed.

Two points of feedback:
* I'd suggest staying away from changing errors to warnings as that defies expectations.
* What about adding some information in the file docblock about how to activate this with GitHub Actions ?

P.S.: The build is failing as the new file hasn't been added to the PEAR `package.xml` file, though that's an easy fix to make,

BrianHenryIE:
# Error/Warning Label

I don't know what actually constitutes an error as distinct from a warning in PHP CodeSniffer.

e.g., from wiki:
```
 47 | ERROR | Line not indented correctly; expected 4 spaces but found 1
```
and
```
 21 | WARNING | Equals sign not aligned with surrounding assignments
```
both are automatically fixable indentation problems but designated different.  I searched the documentation and code but didn't find a definition.

From a CI perspective, the real question is _"does this require manual intervention?"_ which I think is communicated correctly here. So what I'm really trying to do is a PHPCBF report that shows what has been fixed and what needs to be fixed. But when I use `phpcbf --report=GitHubActionsAnnotations` the report used is always [Cbf.php](https://github.com/squizlabs/PHP_CodeSniffer/blob/3.5.4/src/Reports/Cbf.php) and https://github.com/squizlabs/PHP_CodeSniffer/issues/1818 suggests there is no way to override PHPCBF's report. 

A compromise might be to continue fixable/not-fixable as warning/error for GitHub Actions, but prefix the message with the PHPCS warning/error label. 

# Docblock

I think this would be more appropriate in the wiki. Here's an example GitHub Actions workflow, to be saved in `.github/workflow/phpcbf.yml`

```
name: Run PHP CodeSniffer

# NB: Pull requests from forks do not have access to repository secrets so cannot commit changes.

on:
  push:
    branches:
      - master

jobs:
  php-codesniffer:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v2

      - name: Run Composer install
        uses: php-actions/composer@v1

      - name: Run PHPCS
        continue-on-error: true
        run: vendor/bin/phpcs --report=GitHubActionsAnnotations

      - name: Run PHPCBF
        continue-on-error: true
        run: vendor/bin/phpcbf

      - name: Commit PHPCBF changes
        uses: stefanzweifel/git-auto-commit-action@v4.1.1
        with:
          commit_message: "PHPCBF"
```

# Package.xml

I've added this and it seems the build is passing!

jrfnl:
Related: https://github.com/chekalsky/phpcs-action

BrianHenryIE:
As I thought about this over the weekend, I think it does make sense to stick with the true error/warning output from PHPCS. I was thinking of it as running the report before PHPCBF so that the errors that CI was about to fix would be reported. When thinking of it as running PHPCBF then PHPCS, the traditional output makes more sense. So what I was really trying to do was a PHPCBF report. 

I noticed there were projects that were full actions, but thought it's more appropriate as part of the main PHPCS project.

chekalsky:
Let me check it in a few days!

jrfnl:
Just as a heads-up, while it may still be nice to have a PHPCS native report for this, it isn't actually needed to have annotations show in GitHub.

The [CS2PR](https://github.com/staabm/annotate-pull-request-from-checkstyle) tool which can be automatically installed via the `setup-php` action runner can handle the PHPCS checkstyle report perfectly and will show annotations inline in PRs and code views.

```yaml
      - name: Install PHP
        uses: shivammathur/setup-php@v2
        with:
          php-version: '8.1'
          coverage: none
          tools: cs2pr
      - name: Install Composer dependencies
        uses: "ramsey/composer-install@v2"
      - name: Run PHPCS
        run: vendor/bin/phpcs --report=checkstyle -q /path/to/code | cs2pr
```

muno92:
I knew cs2pr, and thought it is nice to annotate PR without any other tool for less dependencies.
But as you wrote, it is not necessary.

I use cs2pr.

Thanks.

nicodemuz:
+1 to have this merged!

jrfnl:
@nicodemuz This repo is abandoned and the project has continued in a new repo: https://github.com/PHPCSStandards/PHP_CodeSniffer Nothing will be merged here anymore.

nicodemuz:
@jrfnl Thank you! I have open a similar PR there: https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/1190

PR #2920, opened 2020-03-30
Tokenizer: set `bracket_opener/closer` indexes for T_OPEN/CLOSE_USE_GROUP

jrfnl:
Efficiency tweak to allow sniffs to quickly get the corresponding bracket opener/closer for a use group token without having to do token walking to find it.

Includes unit test.


jrfnl:
Came across this again while trying to fix something else and have now simplified the implementation.

Based on existing code, this was always intended to work, just never did.

jrfnl:
Looks like some existing sniff and utilities may need adjusting if this change would be accepted. Before I spend time on that - @gsherwood is this a change you'd consider (the scope opener/closer setting for T_USE with use groups) or not ?

gsherwood:
I'm worried about a BC break for this one. I think it's a good idea, I'm just not sure if it is worth working on if it's not working easily. I haven't looked into it, so I'm not sure if this is an easy change or not.

jrfnl:
@gsherwood In that case, I'll have a look at what is breaking in more detail to gauge the impact.

I can rebase the PR on the PHPCS 4.x branch to move the BC-break to the next major.

PR #2922, opened 2020-04-01
Removing Waning caused by "continue 2"

gabrielpcruz:
I was installed the codesniffer, then i run the command to validate my classes his warning about this:

PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in /home/gabriel/.composer/vendor/squizlabs/php_codesniffer/CodeSniffer/File.php on line 1763
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in /home/gabriel/.composer/vendor/squizlabs/php_codesniffer/CodeSniffer/File.php on line 2723

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -1464,7 +1464,7 @@ class File
                 // If it's null, then there must be no parameters for this
                 // method.
                 if ($currVar === null) {
-                    continue 2;
+                    break;
                 }
 
                 $vars[$paramCount]            = [];


jrfnl:
Invalid fix. Suggest: close.

Please upgrade the version of PHP_CodeSniffer you have installed. You are using an outdated version.
PHP 7.3 is supported since PHPCS 3.3.1. PHP 7.4 since PHPCS 3.5.0.

The fix you are proposing now is completely unnecessary. `continue 2` is fine as it targets the `for` and not the `switch` and does not throw any warnings on PHP 7.3+.

The old warning you were seeing from the old PHPCS version was fixed in #2086 way back in 2018.

Issue #2856, opened 2020-02-05
Unknown type hint "string" found for $value


MartkCz:
squizlabs/php_codesniffer: 3.5.4

I've following code:

```php
/**
 * @param mixed $value
 */
function fail(string $name, $value): void {
}
```

Unknown type hint "string" found for $value
(Squiz.Commenting.FunctionComment.InvalidTypeHint)

Following codes are ok:
```php
/**
 * @param mixed $name
 */
function ok($name, string $value): void {
}
/**
 * @param mixed|int $value
 */
function ok(string $name, $value): void {
}

function ok(string $name, string $value): void {
}
```

Dump from FunctionCommentSniff (https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php#L455-L462)
**$realParams**
```php
array (
  0 => 
  array (
    'token' => 30,
    'name' => '$name',
    'content' => 'string $name',
    'pass_by_reference' => false,
    'reference_token' => false,
    'variable_length' => false,
    'variadic_token' => false,
    'type_hint' => 'string',
    'type_hint_token' => 28,
    'type_hint_end_token' => 28,
    'nullable_type' => false,
    'comma_token' => 31,
  ),
  1 => 
  array (
    'token' => 33,
    'name' => '$value',
    'content' => '$value',
    'pass_by_reference' => false,
    'reference_token' => false,
    'variable_length' => false,
    'variadic_token' => false,
    'type_hint' => '',
    'type_hint_token' => false,
    'type_hint_end_token' => false,
    'nullable_type' => false,
    'comma_token' => false,
  ),
)
```
**$pos**: 0
**$param**:
```php
array (
  'tag' => 17,
  'type' => 'mixed',
  'var' => '$value',
  'comment' => '',
  'commentLines' => 
  array (
    0 => 
    array (
      'comment' => '',
    ),
  ),
  'type_space' => 1,
  'var_space' => 0,
)
```
**$typeHint**: 'string'

PR #2929, opened 2020-04-03
Squiz.Commenting.FunctionComment failed if realParams is not equal with docParams

MartkCz:
https://github.com/squizlabs/PHP_CodeSniffer/issues/2856


gsherwood:
PHP_CodeSniffer supports PHP 5.4.0+ so `array_column()` cannot be used until PHP_CodeSniffer version 4.

Issue #2877, opened 2020-02-22
PEAR.Functions.FunctionCallSignature false positive for array of functions


VincentLanglet:
The following code
```
$a = ['a' => function ($b) { return $b; }];
$a['a']( 1 );
```
Return no error for the extra space after the opening and before the closing parentheses.

Since the `(` is correctly ignored in the Generic.WhiteSpace.ArbitraryParenthesesSpacing, this should be handle in the FunctionCallSignatureSniff.

PR #2941, opened 2020-04-23
[FunctionCallSignatureSniff] Handle array of function

VincentLanglet:
Close https://github.com/squizlabs/PHP_CodeSniffer/issues/2877


gsherwood:
Thanks for this fix, and for the original bug report as well.

PR #2952, opened 2020-05-03
File::getMethodProperties(): allow for return type "static"

jrfnl:
As of PHP 8.0, `static` will be allowed as a return type.

Ref: https://wiki.php.net/rfc/static_return_type

This commit adjusts the `File::getMethodProperties()` utility method to allow for that.

Includes unit test.


gsherwood:

PR #2964, opened 2020-05-16
Modernize junit report

GhostLyrics:
This PR improves on the existing Junit report by implementing further attributes of the spec. I have attached samples of the output as well as examples how the results are parsed by a consumer, in this case GitLab.

- `classname` and `file` are added to `testcase` elements
- a generic "PHP_CodeSniffer" string is now used instead of the file name if all tests passed
- Indention of the generated XML is now consistent. 
---
- I have run the CS checker and fixed all issues.
- I have **not** run the phpunit test because no tests cover generating Junit reports as far as I saw.
- I have tried implementing this as a custom report before, but I could not find a way to do both print a custom report to a file AND print a builtin report to stdout at the same time, so I sent this PR (The builtin report would also end up in my `--report-file`).
---
examples:

```xml
<!--BEFORE: inconsistently indented fragment of report.xml -->

<testsuites name="PHP_CodeSniffer 3.5.4" errors="0" tests="104" failures="1">
<testsuite name="tests/unit/MiscUtilTest.php" errors="0" tests="1" failures="0">
    <testcase name="tests/unit/MiscUtilTest.php"/>
</testsuite>
[...]
```

```xml
<!-- AFTER: consistently indented fragment of report.xml -->

<testsuites name="PHP_CodeSniffer 3.5.4" errors="0" tests="104" failures="1">
  <testsuite name="tests/unit/MiscUtilTest.php" errors="0" tests="1" failures="0">
    <testcase name="tests/unit/MiscUtilTest.php"/>
  </testsuite>
[...]
```

```xml
<!-- BEFORE: all information is contained only in "name" -->
<!-- line breaks added for readability of PR -->
<testcase name="PSR2.Methods.FunctionCallSignature.SpaceAfterOpenBracket at tests/unit/DeploymentCommandsTest.php (44:35)">
  <failure type="error" 
           message="Space after opening parenthesis of function call prohibited"/>
</testcase>
```

```xml
<!-- AFTER: information goes into their relevant attributes -->
<!-- line breaks added for readability of PR-->
<testcase classname="DeploymentCommandsTest" 
          file="tests/unit/DeploymentCommandsTest.php" 
          name="PSR2.Methods.FunctionCallSignature.SpaceAfterOpenBracket (44:35)">
  <failure type="error" 
           message="Space after opening parenthesis of function call prohibited (line 44, column 35)"/>
</testcase>
```

Before:
01-tests-tab-before
After: 
01-tests-tab-after

### DIFF
--- a/src/Reports/Junit.php
+++ b/src/Reports/Junit.php
@@ -4,7 +4,8 @@
  *
  * @author    Oleg Lobach <oleg@lobach.info>
  * @author    Greg Sherwood <gsherwood@squiz.net>
- * @copyright 2006-2015 Squiz Pty Ltd (ABN 77 084 670 600)
+ * @author    Alexander Skiba <alexander.skiba@timetac.com>
+ * @copyright 2006-2020 Squiz Pty Ltd (ABN 77 084 670 600)
  * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
  */
 
--- a/src/Reports/Junit.php
+++ b/src/Reports/Junit.php
@@ -120,11 +120,30 @@ class Junit implements Report
         }
 
         $failures = ($totalErrors + $totalWarnings);
-        echo '<?xml version="1.0" encoding="UTF-8"?>'.PHP_EOL;
-        echo '<testsuites name="PHP_CodeSniffer '.Config::VERSION.'" errors="0" tests="'.$tests.'" failures="'.$failures.'">'.PHP_EOL;
-        echo $cachedData;
-        echo '</testsuites>'.PHP_EOL;
 
+        $dom = new \DOMDocument();
+        $dom->formatOutput = True;
+        $dom->encoding = "UTF-8";
+        $dom->preserveWhiteSpace = False;
+
+        $testsuites = $dom->createElement("testsuites");
+        $testsuites->setAttribute("name", 'PHP_CodeSniffer '.Config::VERSION);
+        $testsuites->setAttribute("errors", 0);
+        $testsuites->setAttribute("tests", $tests);
+        $testsuites->setAttribute("failures", $failures);
+
+        $fragment = $dom->createDocumentFragment();
+        # using XML that is partially formatted in appendXML() results in
+        # dom->formatOutput ignoring the fragment during formatting
+        $fragment->appendXML($cachedData);
+
+        $testsuites->appendChild($fragment);
+        $dom->appendChild($testsuites);
+
+        # saving and loading the string forces pretty formatting
+        $tmp = $dom->saveXML();
+        $dom->loadXML($tmp);
+        echo $dom->saveXML();
     }//end generate()
 
 
--- a/src/Reports/Junit.php
+++ b/src/Reports/Junit.php
@@ -44,20 +44,19 @@ class Junit implements Report
 
         $classname = pathinfo($report['filename'])['filename'];
 
-        # successful tests
         if (count($report['messages']) === 0) {
+            // Handle successful tests.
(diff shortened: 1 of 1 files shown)


src/Reports/Junit.php
```diff
@@ -120,10 +136,34 @@ public function generate(
         }
 
         $failures = ($totalErrors + $totalWarnings);
-        echo '<?xml version="1.0" encoding="UTF-8"?>'.PHP_EOL;
-        echo '<testsuites name="PHP_CodeSniffer '.Config::VERSION.'" errors="0" tests="'.$tests.'" failures="'.$failures.'">'.PHP_EOL;
-        echo $cachedData;
-        echo '</testsuites>'.PHP_EOL;
+
+        $dom = new \DOMDocument();
```
jrfnl, line 140: Please be aware that the DOM extension [can be disabled](https://www.php.net/manual/en/dom.installation.php)ay not be available and is currently not a requirement for PHPCS.

The [XMLWriter](https://www.php.net/manual/en/book.xmlwriter.php) extension and the [SimpleXML](https://www.php.net/manual/en/book.simplexml.php) extension _are_ requirements, so please consider if it's possible to use those instead.

Refs:
* https://github.com/squizlabs/PHP_CodeSniffer/blob/ce62dee92da3b038a399a99353fc055b39d2853e/composer.json#L27-L31
* https://github.com/squizlabs/PHP_CodeSniffer/blob/ce62dee92da3b038a399a99353fc055b39d2853e/src/Runner.php#L246-L257

GhostLyrics: > Please be aware that the DOM extension [can be disabled](https://www.php.net/manual/en/dom.installation.php), so may not be available and is currently not a requirement for PHPCS.

I see. This is my first time writing PHP, so I assumed everything that's on PHP.net is shipped as default.

> The [XMLWriter](https://www.php.net/manual/en/book.xmlwriter.php) extension and the [SimpleXML](https://www.php.net/manual/en/book.simplexml.php) extension _are_ requirements, so please consider if it's possible to use those instead.

That depends on whether you consider consistent indentation important or not. 
- If indentation doesn't matter, then all the changes in `generate()` can be dropped.
- Otherwise, I looked through the docs for XMLWriter and SimpleXML several times but did not find ways to work with the partial XML that's emitted by `generateFileReport()`. I couldn't figure out if implementing the echo **within** this function is essential or if I could just create a class member of `Junit` holding the XMLWriter, append content to that during calls of `generateFileReport()` and echo all of it at the end of `generate()`. That would also ensure consistent indentation and likely not require DOM.

jrfnl: > This is my first time writing PHP

Oh wow, respect for taking something like this on straight off the bat in that case :+1: 

> if I could just create a class member of Junit holding the XMLWriter, append content to that during calls of generateFileReport() and echo all of it at the end of generate()

If I remember correctly, the `echo` in `generateFileReport()` is needed with how things are set up in PHPCS now, but this may be a legacy from PHP 4 when PHP didn't have a proper object model. (this project is quite old).
The downside of saving it to a class property instead of to a file, would be the memory this would consume, which could be extensive on a large project with lots of violations.

I suppose you could just save the minimum information you need from each file in non-XML format, like is done for the `Summary` report, and then use the `generate()` method to parse the saved information and format it the way you want it to be.
Will be less efficient as the data is processed twice, but could possibly get you closer to the intended result.

GhostLyrics:
Seeing as this has been open without more interaction or time from my side, I'll close the PR and leave the code online in case someone later wants to refer to it.

For the record, we're using it internally because our setup already comes with the discussed optional dependencies. :+1:

PR #2967, opened 2020-05-22
Markdown generator does not output headings correctly

o5:
## Before
before

## After
after


o5:
@gsherwood Could you check it, please?

o5:
@gsherwood any comments?

gsherwood:
I went to merge this and realised there is a hard-coded unix newline here. The PHP_EOL constant is used throughout doc generation to ensure the newline characters match the OS being used, so this should be used for consistency.

But I realise it's been 2 months, so I can make that change if you don't have time.

gsherwood:
Thanks a lot for that change, and for this fix.

PR #2977, opened 2020-05-31
File::isReference() does not detect return by reference for closures

jrfnl:
The `&` in a closure declared to return by reference was not recognized as a reference.

Fixed now.

Includes unit test.


gsherwood:
Nice find. Thanks for fixing this.

PR #2981, opened 2020-06-02
Improve error message of BooleanOperatorPlacementSniff when using allowOnly

VincentLanglet:
When you're using `allowOnly = first`, you dont expect `'Boolean operators between conditions must be at the beginning or end of the line, but not both'` as error message because end of line is not allowed.

So I improved it.

### DIFF
--- a/src/Standards/PSR12/Sniffs/ControlStructures/BooleanOperatorPlacementSniff.php
+++ b/src/Standards/PSR12/Sniffs/ControlStructures/BooleanOperatorPlacementSniff.php
@@ -135,8 +135,18 @@ class BooleanOperatorPlacementSniff implements Sniff
             return;
         }
 
-        $error = 'Boolean operators between conditions must be at the beginning or end of the line, but not both';
-        $fix   = $phpcsFile->addFixableError($error, $stackPtr, 'FoundMixed');
+        switch ($this->allowOnly) {
+        case 'first':
+            $error = 'Boolean operators between conditions must be at the beginning of the line';
+            break;
+        case 'last':
+            $error = 'Boolean operators between conditions must be at the end of the line';
+            break;
+        default:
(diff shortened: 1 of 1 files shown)


jrfnl:
I just ran into this again and started to create a PR only to realize one is already open.

@gsherwood As this is only a message change, is there a particular reason for this change to be milestoned to `3.6.0`, not `3.5.7` ?

The existing tests cover the change, but just to make the case for this change, here is a test scenario which illustrates the problem this PR solves:

When the "allowOnly" option is enabled, the error message for a condition with only two parts is unclear and confusing.

For example, given this ruleset:
```xml
<?xml version="1.0"?>
<ruleset>
    <rule ref="PSR12.ControlStructures.BooleanOperatorPlacement">
        <properties>
            <property name="allowOnly" value="first"/>
        </properties>
    </rule>
</ruleset>
```
... run against the below code:
```php
if (
    $conditionA === true ||
    $conditionB === true
) {
    // Do something.
}
```

Would result in the following error message:
```
 [x] Boolean operators between conditions must be at the beginning or end of the line, but not both
     (PSR12.ControlStructures.BooleanOperatorPlacement.FoundMixed)
```

... which would presume that the end-user knows that the ruleset demands for the boolean operator to be at the beginning of the line as otherwise the error doesn't make any sense.

gsherwood:
> As this is only a message change, is there a particular reason for this change to be milestoned to `3.6.0`, not `3.5.7` ?

It was done at a time when I was trying to clear 3.5.7 and push through just a few bug fixes. But it's turned into a much bigger bug fix release, so makes sense to include this change again. I've adjusted the milestone for it.

gsherwood:
Thanks a lot for this message change. It has been merged in, but I was doing a rebase and the commit hashes got changed over.

PR #2988, opened 2020-06-17
Undefined offset in Squiz.Strings.ConcatenationSpacing during live coding

thiemowmde:
This can happen when PHPCS runs on a file that is currently being worked on, but not yet completed. The file might end with a dot. We can not assume there are always 2 more tokens after a dot.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Strings/ConcatenationSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/Strings/ConcatenationSpacingSniff.php
@@ -79,10 +79,14 @@ class ConcatenationSpacingSniff implements Sniff
             $phpcsFile->recordMetric($stackPtr, 'Spacing before string concat', $before);
         }
 
-        if ($tokens[($stackPtr + 1)]['code'] !== T_WHITESPACE) {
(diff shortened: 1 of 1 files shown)


thiemowmde:
Nice catch. Indeed, my original proposal would create awkward reports that don't make much sense if the line of code is not even complete. I changed it to don't report anything in such a situation.

jrfnl:
> I changed it to don't report anything in such a situation.

That's not what I intended. The "spacing before" could still be safely reported/fixed in that case. It's only the spacing after where the problem lies.

thiemowmde:
There is no separate message for "spacing before". There are only messages that talk about *"Concat operator must [not] be surrounded by …"*. All these messages are confusing when the line of code that ends with something like `$s = '…' .` is not even complete. Sure, we could introduce separate messages that only talk about the "spacing before" in such a situation. But that would be quite some overkill.

jrfnl:
@thiemowmde You're right. Sorry, I was getting confused by the metrics which are recorded separately
Bit surprising nonetheless as most of these types of sniffs would have separate error codes, but as this one doesn't, I agree, no need to change that at this time.

gsherwood:
Thanks a lot for this PR

PR #2989, opened 2020-06-17
Incorrect auto-fixing in Generic.ControlStructures.InlineControlStructure during live coding

thiemowmde:
… a bracket

While live coding, there might not only be an incomplete `if ():` at the end of a file, it might as well be an incomplete `if () {`. The most minimal example I found is `<?php if ( 1 ) { break;`. The auto-fix currently turns this into `<?php if ( 1 ) { { { { { { { { { { { { { { { { { { { { { { { { { { break;}}}}}}}}}}}}}}}}}}}}}}}}}`. 😲

### DIFF
--- a/src/Standards/Generic/Sniffs/ControlStructures/InlineControlStructureSniff.php
+++ b/src/Standards/Generic/Sniffs/ControlStructures/InlineControlStructureSniff.php
@@ -142,8 +142,10 @@ class InlineControlStructureSniff implements Sniff
             return;
         }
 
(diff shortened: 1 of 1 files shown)


thiemowmde:
What's needed to get a tiny bugfix like this reviewed and merged?

gsherwood:
Auto-fixing during live coding seems a bit dangerous. Thanks for fixing this issue, but I suspect there are other lurking in the sniffs. Please let me know if you find more.

PR #2990, opened 2020-06-17
Fix possible index error in NonExecutableCodeSniff

thiemowmde:
I was running into this while working on another sniff. One trivial example I found to demonstrate the bug is `<?php return array_map(`. This can happen when PHPCS runs on a file I'm currently typing in, but the file is not complete yet.

~~I believe it's safe to assume the `parenthesis_closer` array key only exists on T_OPEN_PARENTHESIS tokens, and the `bracket_closer` key only on T_OPEN_CURLY_BRACKET tokens. I had a quick look at the tokenizer and I believe this is indeed the case. But I might be wrong.~~

Sorry for not providing a test. I would love to, but I'm not familiar enough with the test setup here.

### DIFF
--- a/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
@@ -216,7 +216,9 @@ class NonExecutableCodeSniff implements Sniff
                 continue;
             }
 
-            if (isset($tokens[$start]['bracket_closer']) === true) {
+            if (isset($tokens[$start]['bracket_closer']) === true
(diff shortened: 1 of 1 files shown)


jrfnl:
IIRC `bracket_closer` also applies to `T_OPEN_SQUARE_BRACKET` and `T_OPEN_SHORT_ARRAY`

thiemowmde:
I added both checks back, just to be sure. Is this better?

thiemowmde:
This was (checks calendar) 4 months ago. Is there anything I can do to get this bug fixed?

gsherwood:
Thanks for this fix. I've added a unit test and merged this in. It will be released in version 3.6.0.

Issue #2992, opened 2020-06-21
Enabling caching using a ruleset produces invalid cache files when using --sniffs and --exclude CLI args


jrfnl:
While researching how the result caching was implemented and playing around with it, I noticed that the `--sniffs` and `--exclude` arguments are not taken into account properly, leading to misleading results being cached and displayed for runs.

---
<details>
  <summary><b>Steps to reproduce the issue when using the <code>--sniffs</code> CLI argument</b></summary>

Given the following file:
```php
<?php

namespace Foo;

use const PHP_VERSION_ID;
use func in_array;

class my_class {
	function method() {
		$versions = ['10', '20'];
		return ( in_array( PHP_VERSION_ID, $versions ) );
	}
}
```

With this ruleset saved as `phpcs.xml.dist`:
```xml
<?xml version="1.0"?>
<ruleset name="Issue cache invalidation CLI args">
	<file>.</file>
	<arg value="sp"/>
	<arg name="basepath" value="."/>
	<arg name="cache" value="phpcs.cache"/>
	<rule ref="PSR12"/>
</ruleset>
```

0. If necessary, throw away an existing `phpcs.cache` file in the test directory.
1. Run `phpcs --report=source`
    The output will be:
    ```
    PHP CODE SNIFFER VIOLATION SOURCE SUMMARY
    -----------------------------------------------------------------------
        SOURCE                                                        COUNT
    -----------------------------------------------------------------------
    [x] Generic.WhiteSpace.DisallowTabIndent.TabsUsed                 4
    [x] PSR2.Classes.ClassDeclaration.OpenBraceNewLine                1
    [x] PSR2.Methods.FunctionCallSignature.SpaceAfterOpenBracket      1
    [x] PSR2.Methods.FunctionCallSignature.SpaceBeforeCloseBracket    1
    [ ] PSR12.Files.FileHeader.IncorrectOrder                         1
    [x] PSR12.Files.FileHeader.SpacingAfterUseConstBlock              1
    [ ] Squiz.Classes.ValidClassName.NotCamelCaps                     1
    [x] Squiz.Functions.MultiLineFunctionDeclaration.BraceOnSameLine  1
    [ ] Squiz.Scope.MethodScope.Missing                               1
    -----------------------------------------------------------------------
    A TOTAL OF 12 SNIFF VIOLATIONS WERE FOUND IN 9 SOURCES
    -----------------------------------------------------------------------
[...]

PR #2995, opened 2020-06-21
Result caching: minor optimization tweak

jrfnl:
While debugging #2992, I noticed that `.bak` files were being used for the "PHPCS native file hash".

As non-php files in the PHPCS directories have no effect on the actual run output, we may as well skip them from being part of the hash.

### DIFF
--- a/src/Util/Cache.php
+++ b/src/Util/Cache.php
@@ -105,6 +105,11 @@ class Cache
                     return false;
                 }
 
+                // Skip non-php files.
+                if ($file->isFile() === true && substr($filename, -4) !== '.php') {
+                    return false;
+                }
+
                 $filePath = Common::realpath($file->getPathname());
                 if ($filePath === false) {
                     return false;
--- a/src/Util/Cache.php
+++ b/src/Util/Cache.php
@@ -95,27 +95,25 @@ class Cache
         // hash. This ensures that core PHPCS changes will also invalidate the cache.
         // Note that we ignore sniffs here, and any files that don't affect
         // the outcome of the run.
-        $di     = new \RecursiveDirectoryIterator($installDir);
+        $di     = new \RecursiveDirectoryIterator(
+            $installDir,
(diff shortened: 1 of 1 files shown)


gsherwood:
If you're seeing `.bak` files in the PHPCS `src` dir, they are presumably being placed there while you're editing the PHPCS source itself and are not committed (I hope). Adding extra checks for these files might make the hash more accurate while coding PHPCS, but isn't the extra check going to actually slow down the creating of the file hash for all regular uses?

jrfnl:
@gsherwood I'd think that "slow down" would be so small, it wouldn't be noticeable by humans. (And yes, of course I'm not committing these ... ever saw a PR of mine which contained them ? :grin:)

jrfnl:
@gsherwood As you were concerned about performance, I've had another look at the code and found a number of other optimizations which can be made which should easily offset the "slow down" of the additional check.

I've added a second commit to this PR with those changes.

> Optimize use of the iterators:
> 1. Skip dot files at the `RecursiveDirectoryIterator` level by setting the `FilesystemIterator::SKIP_DOTS` flag and remove the code which was doing the same in the callback.
>     Note: the other two flags are the default flags used by the `RecursiveDirectoryIterator` constructor, so are needed to maintain the existing behaviour.
> 2. No need for the `$file->getPathname()` function call. The `$key` already contains that information, as per the default flags.
> 3. No need for a file-system `is_dir()` call. If it's a directory, the iterator will have children.

gsherwood:
> As you were concerned about performance

I think you missed the point I was trying to make. Adding a check for .bak files in the src directory isn't something that should be required given there are none, so why add a check for them?

Thanks for the extra optimisations though - I'll take a look when I get a chance.

jrfnl:
@gsherwood Oh no, I got your point and I agree that those files will only be there for contributing devs, though it effectively means those people who use PHPCS most will rarely, if ever, benefit from caching, which is a shame.

Either way, the extra optimizations should more than offset the additional check.

gsherwood:
Finally merged :) Thanks

Issue #2994, opened 2020-06-21
Generic.Formatting.DisallowMultipleStatements false positive for FOR loop with no body


gmponos:
Hi,

When I run phpcs against this file https://github.com/doctrine/coding-standard/pull/208/files#diff-9023c633b2425ded52916bf44f88ede1R17

The error I get are:

```
  7 | ERROR | [x] Useless semicolon. (SlevomatCodingStandard.PHP.UselessSemicolon.UselessSemicolon)
 10 | ERROR | [x] Useless semicolon. (SlevomatCodingStandard.PHP.UselessSemicolon.UselessSemicolon)
 10 | ERROR | [x] Each PHP statement must be on a line by itself (Generic.Formatting.DisallowMultipleStatements.SameLine)
 15 | ERROR | [x] Useless semicolon. (SlevomatCodingStandard.PHP.UselessSemicolon.UselessSemicolon)
 17 | ERROR | [x] Whitespace found before first semicolon of FOR loop
    |       |     (Squiz.ControlStructures.ForLoopDeclaration.SpacingBeforeFirst)
 17 | ERROR | [x] Space found before semicolon; expected "0;" but found "0 ;" (Squiz.WhiteSpace.SemicolonSpacing.Incorrect)
 17 | ERROR | [x] Each PHP statement must be on a line by itself (Generic.Formatting.DisallowMultipleStatements.SameLine)
 26 | ERROR | [x] Each PHP statement must be on a line by itself (Generic.Formatting.DisallowMultipleStatements.SameLine)
 26 | ERROR | [x] Useless semicolon. (SlevomatCodingStandard.PHP.UselessSemicolon.UselessSemicolon)
```

I care mainly for ` 17 | ERROR | [x] Each PHP statement must be on a line by itself (Generic.Formatting.DisallowMultipleStatements.SameLine)`

The autofix is this: 

https://github.com/doctrine/coding-standard/pull/208/files#diff-df7e22f20fb934baed4dd64e4ff1cd75R17

which seems wrong.

gsherwood:
The sample code to test this is:
```php
for ($i = 0 ; $i < 10; $i++);
{
}
```
The sniff auto-fixes this to be:
```php
for ($i = 0 ; $i < 10;
$i++);
{
}
```

gsherwood:
[...]

PR #2999, opened 2020-06-24
Detect loops with no body

gmponos:
Follow up of the comment here #2994 

I also discovered that the foreach-loop can also have a semicolon at the end.

### DIFF
--- a/src/Standards/Generic/Sniffs/CodeAnalysis/LoopWithNoBodySniff.php
+++ b/src/Standards/Generic/Sniffs/CodeAnalysis/LoopWithNoBodySniff.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * Detects loops (for and foreach) that do not have a body
+ * due to a semicolon
+ *
+ * The following scenarios most probably are typos.
+ *
+ * <code>
+ * foreach ($array as $for => $value);
+ * {
+ *   doSomething(); // doSomething will not be looped
+ * }
+ *
+ * for ($i =0; $i<10; $i++);
+ * {
+ *   doSomething(); // doSomething will not be looped
+ * }
+ * </code>
+ *
+ * @author    George Mponos <gmponos@gmail.com>
+ * @copyright 2020 George Mponos. All rights reserved.
+ * @license   https://github.com/squizlabs/PHP_CodeSniffer/blob/master/licence.txt BSD Licence
+ */
+
+namespace PHP_CodeSniffer\Standards\Generic\Sniffs\CodeAnalysis;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+
+class LoopWithNoBodySniff implements Sniff
+{
+
+
+    /**
+     * Registers the tokens that this sniff wants to listen for.
+     *
+     * @return int[]
+     */
+    public function register()
+    {
+        return [
+            T_FOR,
+            T_FOREACH,
+        ];
+
+    }//end register()
+
+
+    /**
+     * Processes this test, when one of its tokens is encountered.
+     *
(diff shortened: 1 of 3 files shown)


gmponos:
Not sure what's wrong with the failing build :/

jrfnl:
> Not sure what's wrong with the failing build :/

@gmponos As per: https://travis-ci.org/github/squizlabs/PHP_CodeSniffer/jobs/701538216#L544-L549 you didn't add the new files to the `package.xml` file which means they would not be included in the PEAR package (which is why that check is in place).

gmponos:
>May I suggest adding an additional check for there being a semi-colon being after the parenthesis_closer and a open curly after that ?

Don't quite get that... Is this the case you want to cover?

```
for($i =0; $i<10; $i++); {
}
```

If not can you provide an example similar to the above... 



> the only control structures which can be declared without a body are: while, for and declare. 

And `foreach`... https://3v4l.org/LHkDc

In anyway I must add `while` here as well.. https://3v4l.org/gZ64v

I don't think we should include `declare` in the current sniff.. It is specific that the sniff is only about loops.

jrfnl:
> And foreach... https://3v4l.org/LHkDc

Fair enough, but using a foreach without body doesn't really _do_ anything, so there wouldn't be any point doing that. The one exception being when using references, but that's a whole other snakepit.

Regarding my suggestion:

IMO, the sniff should never trigger on clearly deliberate use of a control structure without body:
```php
for($i =0; $i<10; $i++);
echo $i;
```

Only when it could conceivably be a mistake:
```php
for($i =0; $i<10; $i++); {
}
```

gmponos:
> IMO, the sniff should never trigger on clearly deliberate use of a control structure without body

Nowdays due to PSR2/12 it is common to add braces to loops... 
but not all standards use braces for control structures.. 

So IMHO both of the cases above should be reported.

jrfnl:
@gmponos This is not about braces or no braces. My example is 100% valid code without any mistakes.

And just like you say, as it is now common to use braces, the sniff should not trigger on intentional use of control structures without body.

This:
```php
for($i =0; $i<10; $i++);
echo $i; // 10
```

Is something completely different from:
```php
for($i =0; $i<10; $i++)
    echo $i; // 12345678910
```

or from this - which is what you stated you were targetting:
```php
for($i =0; $i<10; $i++);
{
    echo $i; // 10, but was intended to be 12345678910
}
```

jrfnl:
At the very least, please consider using a different error code for with/without curly brace as that way, people who use the short notation without body could still use the sniff, but exclude the "without braces" error code.

src/Standards/Generic/Sniffs/CodeAnalysis/LoopWithNoBodySniff.php
```diff
@@ -0,0 +1,69 @@
...
+     *                                               in the stack passed in $tokens.
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+        $token  = $tokens[$stackPtr];
+        if (isset($token['scope_opener']) === false) {
+            $phpcsFile->addError('The `%s` loop statement does not have a body', $stackPtr, 'NoBody', [$token['content']]);
```
gsherwood, line 63: Sniffs typically use the code `Found` so the message code reads better -  `Generic.CodeAnalysis.LoopWithNoBody.Found`

I'd also strtolower the token content because you never know what case it will be defined in.

gsherwood, line 63: You mentioned on the original issue that you thought this should be a warning, but you are generating an error here. Was that an intentional change or do you think warning is still the correct severity for this?

PR #3006, opened 2020-07-05
Tests: stabilize an expected global

jrfnl:
The global `$PHP_CODESNIFFER_PEAR` variable is used by the `Core` test and those test don't necessarily need to use the generic `AllTests` file as an entry point.
They can also be run using `phpunit ./tests/Core/AllTests.php`.

In that case, however, the `$PHP_CODESNIFFER_PEAR` variable is not defined leading to `Undefined index: PHP_CODESNIFFER_PEAR` errors.

By setting the global in the bootstrap, which is always loaded, this is avoided.

### DIFF
--- a/tests/AllTests.php
+++ b/tests/AllTests.php
@@ -9,16 +9,13 @@
 
 namespace PHP_CodeSniffer\Tests;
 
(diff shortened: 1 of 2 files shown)


gsherwood:
Thanks for this fix. Much cleaner as well.

Issue #3007, opened 2020-07-06
Directory exclude pattern improperly excludes directories with names that start the same


SteveTalbot:
This issue affects version 3.5.5 and is very closely related to #1920

## The problem

Imagine I want to exclude files in `bin` directories:
```xml
<exclude-pattern>*/bin/*</exclude-pattern>
```

This rule will also match and exclude any directory starting with `bin`, such as `./src/BingSearch/`

The problem appears to be in the same area of code that caused #1920, specifically [Filter.php#L221](https://github.com/squizlabs/PHP_CodeSniffer/blob/a957a73e3533353451eb9fd62ee58bd0aba2773c/src/Filters/Filter.php#L221):

```php
                if (substr($pattern, -2) === '/*') {
                    // Need to check this pattern for dirs as well as individual file paths.
                    $this->ignoreFilePatterns[$pattern] = $type;

                    $pattern = substr($pattern, 0, -2);
                    $this->ignoreDirPatterns[$pattern] = $type;
                } else {
                    // This is a file-specific pattern, so only need to check this
                    // for individual file paths.
                    $this->ignoreFilePatterns[$pattern] = $type;
                }
```

Removing the trailing slash with `substr` allows the pattern to match the directory, but it also allows it to match any directory with a name that starts the same.

## Proposed solution

A simple fix is to anchor the directory pattern to the end of the string, like this:
```php
                    $pattern = substr($pattern, 0, -2) . '$';
                    $this->ignoreDirPatterns[$pattern] = $type;
```

Or use a lookahead:
```php
                    $pattern = substr($pattern, 0, -2) . '(?=/|$)';
                    $this->ignoreDirPatterns[$pattern] = $type;
```

## Workaround

To exclude files in `bin` directories, use a lookahead in the exclude pattern in `phpcs.xml`:
```xml
<exclude-pattern>*/bin(?=/|$)/*</exclude-pattern>
```

PR #3008, opened 2020-07-06
Fixed bug #3007: Directory exclude pattern improperly excludes directories with names that start the same

SteveTalbot:
This is a suggested fix for #3007 

To unit test this properly, you'd need to mock the `is_dir` function on [Filter.php#L237](https://github.com/squizlabs/PHP_CodeSniffer/blob/a957a73e3533353451eb9fd62ee58bd0aba2773c/src/Filters/Filter.php#L237).  Sorry, I couldn't find a neat way to do that in the time I had available.


gsherwood:
Thanks a lot for this fix, and the detailed bug report you submitted as well.

PR #3009, opened 2020-07-07
PHP 8.0 | File::getMethodParameters(): support mixed type and union types

### LINKED ISSUE #2968: Support PHP 8's `mixed` type

jrfnl:
PHP 8.0 will introduce a new pseudo-type `mixed` as well as union types.

This PR adds tests for these to the `File::getMethodParameters()` method and makes minor changes to the method to allow for supporting these new type declarations.

Refs:
* https://wiki.php.net/rfc/mixed_type_v2
* https://wiki.php.net/rfc/union_types_v2

:point_right: **Important**: all sniffs using this method will need to be checked to see if they use the `type_hint` index and if so, if the check being done needs to take union types into account.

Partially addresses #2968

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -1442,6 +1442,9 @@ class File
                 }
                 break;
             case T_NS_SEPARATOR:
+            case T_BITWISE_OR:
+            case T_FALSE:
+            case T_NULL:
                 // Part of a type hint or default value.
                 if ($defaultStart === null) {
                     if ($typeHintToken === false) {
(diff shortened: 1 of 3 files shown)


jrfnl:
Test build which includes PR #3014 to which confirms that the build for this PR would pass on PHP 8/nightly: https://travis-ci.com/github/jrfnl/PHP_CodeSniffer/jobs/360290392

jrfnl:
Closing - this PR is partially replaced by PR #3018 (mixed type) and will for the remainder be replaced by a subsequent PR addressing union types. These have to be pulled in turn as they both add unit tests to the same files. See issue #2968 for the background of why the PR(s) are replaced.

PR #3013, opened 2020-07-10
PHP 8.0 | Tokenizer/PHP: support PHP8 magic constant dereferencing

jrfnl:
As of PHP 8, magic constants can be dereferenced, however, the square brackets are incorrectly tokenized as _short array_ brackets instead of as "normal" square brackets.

Fixed by:
* Adding a new `$magicConstants` array to the `Util\Tokens` calls.
* Adding the `$magicConstants` to the allowed tokens for leaving the brackets alone in the `Tokenizers\PHP` class.

Includes adding unit tests for this section of the `PHP::processAdditional()` method.

I've added as many relevant syntaxes as I could come up with, but won't claim that the unit test case file covers every possible scenario. Should be a good start though.

Ref: https://wiki.php.net/rfc/variable_syntax_tweaks#constants_and_magic_constants

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -113,6 +113,8 @@ http://pear.php.net/dtd/package-2.0.xsd">
       <file baseinstalldir="" name="BackfillFnTokenTest.php" role="test" />
       <file baseinstalldir="" name="BackfillNumericSeparatorTest.inc" role="test" />
       <file baseinstalldir="" name="BackfillNumericSeparatorTest.php" role="test" />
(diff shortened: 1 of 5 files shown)


jrfnl:
Test build which includes PR #3014 which confirms that the build for this PR would pass on PHP 8/nightly: https://travis-ci.com/github/jrfnl/PHP_CodeSniffer/jobs/360290642

gsherwood:
Thanks

jrfnl:
@gsherwood Should the new token array get a mention in the changelog ?

PR #2958, opened 2020-05-14
PHPCS 4.x: test against nightly

jrfnl:
In contrast to 60a91b37cb1705468439798856e5ce3e026d53f2, testing against PHP `nightly` can actually be done already, we just need to tell Composer to ignore the PHPUnit PHP requirements ;-)

Currently testing against nightly results in two failed tests:
```
1) PHP_CodeSniffer\Standards\Generic\Tests\PHP\DeprecatedFunctionsUnitTest::testSniff

[LINE 3] Expected 1 error(s) in DeprecatedFunctionsUnitTest.inc but found 0 error(s).

[LINE 4] Expected 1 error(s) in DeprecatedFunctionsUnitTest.inc but found 0 error(s).
```

Both test failures are caused by PHP 8 having removed the functions being tested, i.e. Reflection doesn't see them as deprecated anymore as the function doesn't exist anymore.

PR #3014, opened 2020-07-12
PHPCS 3.x: test against nightly

jrfnl:
PR #2958 turned running the unit tests on for the PHPCS 4.x branch, but this wasn't done for the PHPCS 3.x branch yet.

This commit basically does the same for PHPCS 3.x. PHPUnit 7 will run on PHP 8, even though unsupported, so ignoring platform requirements when installing via Composer allows for the tests for PHPCS 3.x to run on PHP 8.

Include applying the same fix for the failing tests as was applied in the PHPCS 4.x branch.

Loosely related to PR #3009, #3010, #3011 and #3013


gsherwood:
Nice one! Thanks a lot.

PR #3017, opened 2020-07-14
Generic.CodeAnalysis.UnusedFunctionParameter: Add possibility to specify excludes

o5:
Sometimes you have to implement interface but not all required parameters are needed to use. For example in symfony/console.

```php
public function execute(InputInterface $input, OutputInterface $output): int
{
    $output->writeln('Do something without using of $input');
    return 0;
}
```


gsherwood:
Thanks a lot for this contribution. The new sniff property will be available in 3.6.0, and is documented here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#genericcodeanalysisunusedfunctionparameter

PR #3018, opened 2020-07-15
PHP 8.0 | File::get[Method|Member][Parameters|Properties](): add unit test with "mixed" type

jrfnl:
PHP 8.0 will introduce a new pseudo-type `mixed`.

This PR adds tests to the `Core` testsuite for the `File::getMethodParameters()`, `File::getMethodProperties()` and `File::getMemberProperties()` utility methods confirming that the `mixed` type is handled correctly by these methods without further changes.

Refs:
* https://wiki.php.net/rfc/mixed_type_v2

Partially addresses #2968

Partially replaces  PR #3009, #3010, #3011


gsherwood:
Thanks

PR #3019, opened 2020-07-15
PHP 8.0 | Tokenizer/PHP: add support for `static` return type to arrow functions

jrfnl:
Follow up on #2952 which was merged for PHPCS `3.5.6` - @gsherwood with that in mind, can this PR also be earmarked for PHPCS 3.5.6 ?

Includes unit test.


gsherwood:
Thanks

PR #3025, opened 2020-07-20
Travis: change from "trusty" to "xenial"

jrfnl:
As the "trusty" environment is no longer officially supported by Travis, they decided in their wisdom to silently stop updating the PHP "nightly" image, which makes it next to useless as the last image apparently is from January....

This updates the Travis config to:
* Use the `xenial` distro, which at this time is the default.
* Sets the distro for low PHP versions explicitly to `trusty`.
* Makes the expected OS explicit (linux).
* Updates the `matrix` key to `jobs`, which is the canonical for which `matrix` is an alias.
* Removed `sudo: false` - this hasn't been supported for over a year now.

As the `xenial` images (again) appear to randomly be shipped with incompatible PHPUnit versions, I'm changing the setup to always run `composer install` and use the PHPUnit in the `vendor` directory.
This will make the build less prone to error out on incompatible Travis image changes in the future and should hardly make a difference in build time.

Note: the build against PHP nightly will now fail as it is run against a much more recent PHP build. This build failure will be fixed once the comment tokenizer change has been pulled & merged.
To pull that PR, merging this one is a prerequisite.

@gsherwood If you like, I'll rebase the other PHP 8 related PRs once this one has been merged to proof they are sound for PHP 8, though I have already tested that locally, so we should be good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,16 @@
-sudo: false
+os: linux
+dist: xenial
 language: php
-dist: trusty
 
-matrix:
+jobs:
   fast_finish: true
   include:
     - php: 5.4
+      dist: trusty
     - php: 5.5
+      dist: trusty
     - php: 5.5
+      dist: trusty
       env: CUSTOM_INI=1 XMLLINT=1
       addons:
         apt:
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks a lot for figuring this out

PR #3026, opened 2020-07-20
BackfillFnTokenTest: fix bug in the tests

jrfnl:
The scope closers weren't being tested properly as the local `$closer` variable was looking for the wrong array index.

Once that was enabled, it exposed another issue in the tests: nested arrow functions with a scope closer being shared with the "outer" arrow function did not have the correct expectations set.

Fixed now by:
* Selectively skipping the 'scope_condition' check for the scope closer in the `backfillHelper()` method.
* Checking that the `scope_opener` for such shared `scope_closer`s actually points to the "outer" arrow function.


gsherwood:
Thanks for picking up that error in the tests

jrfnl:
You're welcome.

PR #3027, opened 2020-07-20
PHP 8.0 | Tokenizer/PHP: stabilize comment tokenization

### LINKED ISSUE #3002: PHP 8: tokenization of trailing new line after comment token has changed
PHP 8 includes a change in how comment tokens with a trailing new line are tokenized.

>   . T_COMMENT tokens will no longer include a trailing newline. The newline will
>    instead be part of a following T_WHITESPACE token. It should be noted that
>    T_COMMENT is not always followed by whitespace, it may also be followed by
>    T_CLOSE_TAG or end-of-file.

Source: https://github.com/php/php-src/blob/47cf18ba4ef9fb9df6c3affff8f2a3f5a3f930a8/UPGRADING#L473-L477

A decision is needed to:
* Either backfill the old behaviour when PHPCS is run on PHP >= 8 - this can be done in the next 3.x version of PHPCS.
* Or emulate the change in PHP 8 when PHPCS is run on PHP < 8. This would be a big BC break for any sniff examining `T_COMMENT` tokens and should probably only be done in PHPCS 4.x.
* Or both, with the backfill maintaining existing behaviour for PHPCS 3.x and PHPCS 4.x introducing the new behaviour.

Depending on the decision taken, PHPCS native sniffs examining `T_COMMENT` tokens - including the PHPCS annotation tokens - will need to be reviewed and where needed, adjusted.

External standards which support more than just the latest & greatest version of PHPCS and examine `T_COMMENT` tokens - including the PHPCS annotation tokens - will be impacted by whatever decision is taken and will need to adjust their sniffs to allow for both situations.

Refs;
* https://github.com/php/php-src/pull/5182
* https://github.com/php/php-src/commit/47cf18ba4ef9fb9df6c3affff8f2a3f5a3f930a8#diff-7748eb3bfdd3bf962553f6f9f2723c45

jrfnl:
As described in issue #3002, in PHP 8 a trailing new line is no longer included in a `T_COMMENT` token.

This commit "forward-fills" the PHP 5/7 tokenization of `T_COMMENT` tokens for PHP 8, i.e. the PHP 8 comment tokenization is undone and changed back to how comments followed by a new line were tokenized in PHP 5/7.

Includes extensive unit tests. I'm hoping to have caught everything affected :fingers_crossed:

The initial set of unit tests `StableCommentWhitespaceTest` use Linux line endings `\n`.
The secondary set of unit tests `StableCommentWhitespaceWinTest` use Windows line endings `\r\n` to test that the fix is stable for files using different line ending.

For the tests with Windows line endings, both the test case file as well as the actual test file have been set up to use Windows line endings for all lines, not just the test data lines, to make it simpler to manage the line endings for the files.

The test file has been excluded from the line endings CS check for that reason and a directive has been added to the `.gitattributes` file to safeguard that the line endings of those files will remain Windows line endings.

Fixes #3002

This PR incidentally (well, not really) fixes the current build failure on PHP 8.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,7 @@ package.xml export-ignore
 phpunit.xml.dist export-ignore
 php5-testingConfig.ini export-ignore
 php7-testingConfig.ini export-ignore
+
+# Declare files that should always have CRLF line endings on checkout.
+*WinTest.inc text eol=crlf
+*WinTest.php text eol=crlf
--- a/package.xml
+++ b/package.xml
@@ -119,6 +119,10 @@ http://pear.php.net/dtd/package-2.0.xsd">
       <file baseinstalldir="" name="BackfillNumericSeparatorTest.php" role="test" />
       <file baseinstalldir="" name="ShortArrayTest.inc" role="test" />
       <file baseinstalldir="" name="ShortArrayTest.php" role="test" />
+      <file baseinstalldir="" name="StableCommentWhitespaceTest.inc" role="test" />
+      <file baseinstalldir="" name="StableCommentWhitespaceTest.php" role="test" />
+      <file baseinstalldir="" name="StableCommentWhitespaceWinTest.inc" role="test" />
+      <file baseinstalldir="" name="StableCommentWhitespaceWinTest.php" role="test" />
      </dir>
      <file baseinstalldir="" name="AbstractMethodUnitTest.php" role="test" />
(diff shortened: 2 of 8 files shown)


gsherwood:
Thanks so much for this

jrfnl:
Thanks for merging this. Should make live easier external standards which want to test on PHP 8 already (as well as the standards here of course).

PR #3030, opened 2020-07-22
Add option to allow variable usage in double quoted strings

cweiske:
The default for Squiz.Strings.DoubleQuoteUsage "allowVariables" is "false" so
everything stays the way it was.

I added an additional comment about the syntax error, because it cost me
20 minutes of debugging.

This PR needs a new entry in https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties


jrfnl:
@cweiske I'm not sure I understand the purpose of this change.

Doesn't this do the same as excluding the `ContainsVar` error code ? What am I missing ?

```xml
    <exclude name="Squiz.Strings.DoubleQuoteUsage.ContainsVar"/>
</rule>
```

cweiske:
You're right. I wasn't aware that there are different error types inside a single sniff, and that they can be excluded.

jrfnl:
@cweiske No worries, glad you got it working though.

PR #3033, opened 2020-07-23
Error generated during tokenizing of goto statements on PHP 8

jrfnl:
PHP 8.0 elevates a number of notices/warnings to errors, which now exposes a bug in the `goto` tokenizer logic.

Error: `Trying to access array offset on value of type null`

Fixed by making sure that the token being accessed is an array before trying to access it.

Discovered via a failing unit test in WordPressCS.


gsherwood:
Thanks for this fix

jrfnl:
You're welcome.

PR #3036, opened 2020-07-26
Tokenizer/PHP: add some missing tokens to the $knownLengths property

jrfnl:
Not sure if it really makes a difference, but I noticed some tokens were missing from this array which could safely be added.


gsherwood:
In theory, it makes tokenizing faster. It certainly did when I added it, but obviously depends on how often the tokens are used. Even if nobody notices, thank for making the change :)

Issue #3040, opened 2020-08-01
PHP 8.0 | Backfill the new T_NULLSAFE_OBJECT_OPERATOR token


jrfnl:
PHP 8 introduces a new object chaining operator `?->` which short-circuits moving to the next expression if the left-hand side evaluates to `null`.

Refs:
* https://wiki.php.net/rfc/nullsafe_operator
* https://github.com/php/php-src/commit/9bf119832dbf625174794834c71b1e793450d87f

### Impact for PHP_CodeSniffer

- [x] A backfill in the PHP tokenizer needs to be put in place to allow for this token. - PR #3046
    I have already prepared the necessary changes and will pull them once I've run some additional tests.

In addition to that, various sniffs will need to be adjusted to take the new token into account.

Initial inventory of affected sniffs:
- [x] Generic.NamingConventions.UpperCaseConstantName - PR #3079
- [x] Generic.PHP.ForbiddenFunctions - PR #3080
- [x] Generic.PHP.SAPIUsage - PR #3081
- [x] Generic.WhiteSpace.ScopeIndent - PR #3082
- [x] PEAR.WhiteSpace.ObjectOperatorIndent - PR #3083
- [x] PSR1.Files.SideEffects - PR #3084
- [x] PSR12.Classes.ClassInstantiation - PR #3085
- [x] Squiz.Formatting.OperatorBracket - PR #3086
- [x] Squiz.NamingConventions.ValidVariableName - PR #3087
- [x] Squiz.Operators.IncrementDecrementUsage - PR #3088
- [x] Squiz.PHP.DisallowSizeFunctionsInLoops - PR #3090
- [x] Squiz.PHP.LowercasePHPFunctions - PR #3091
- [x] Squiz.WhiteSpace.ObjectOperator - PR #3092
- [x] Zend.NamingConventions.ValidVariableName - PR #3093

I have commits ready for all sniffs with a checkmark and will pull those once the initial tokenizer changes have been merged.

Not claiming completeness, there may be more sniffs which need changes, but this should take care of the bulk of it.

jrfnl:
All the commits I had ready & waiting for this have now been pulled.

gsherwood:
That was an amazing set of PRs adding so much support. Thank you very much. I've still got a couple more to review but will get to those soon.

gsherwood:
Going to close this now as all linked PRs have been merged. If anything else needs doing, it can be handled as new a PR.

[...]

Issue #3041, opened 2020-08-04
PHP 8.0 | Proposal for handling "namespaced names as single token"


jrfnl:
### What changes in PHP 8

From the RFC:

>  PHP distinguishes four kinds of namespaced names:
> * Unqualified names like Foo, which coincide with identifiers.
> * Qualified names like Foo\Bar.
> * Fully qualified names like \Foo.
> * Namespace-relative names like namespace\Foo.
> 
> Each of these kinds will be represented by a distinct token: 
> ```php
> Foo;
> // Before: T_STRING
> // After:  T_STRING
> // Rule:   {LABEL}
 > 
> Foo\Bar;
> // Before: T_STRING T_NS_SEPARATOR T_STRING
> // After:  T_NAME_QUALIFIED
> // Rule:   {LABEL}("\\"{LABEL})+
>  
> \Foo;
> // Before: T_NS_SEPARATOR T_STRING
> // After:  T_NAME_FULLY_QUALIFIED
> // Rule:   ("\\"{LABEL})+
>  
> namespace\Foo;
> // Before: T_NAMESPACE T_NS_SEPARATOR T_STRING
> // After:  T_NAME_RELATIVE
> // Rule:   "namespace"("\\"{LABEL})+
> ```
> 
> Individual namespace segments may contain reserved keywords:
> ```php
> // This is interpreted as T_LIST (i.e., as a reserved keyword):
> List
> // All of the following are interpreted as legal namespaced names:
> \List
> FooBar\List
> namespace\List
> ```
> 
> Whitespace (or comments) is not permitted between namespace separators.

Ref: https://wiki.php.net/rfc/namespaced_names_as_token

### TL;DR

So, in short, PHP 8.0 introduces 3 new tokens for identifiers, on top of the existing `T_STRING`:
* `T_NAME_FULLY_QUALIFIED`
* `T_NAME_RELATIVE`
* `T_NAME_QUALIFIED`

### Perspective on this as a sniff author

In the long run, I think these new tokens will make writing sniffs analyzing function calls, class instantiation and the likes much easier, especially when namespace/use statement context needs to be taken into account when matching identifiers.

However, this is also a big breaking change which would require a significant change in how any such sniff is analyzing code.

A significant part of the current test failures against the `nightly` build for PHPCS itself can be traced back to this particular change. Same for several external standards which I examined.

### Proposal for how to handle this change in PHPCS

[...]

PR #3046, opened 2020-08-06
PHP 8.0 | Tokenizer/PHP: add support for nullsafe object operator

jrfnl:
This PR addresses the tokenizer changes needed for #3040. Once this PR has been merged, the sniff specific changes will be pulled with a PR per sniff.

PHP 8 introduces a new object chaining operator `?->` which short-circuits moving to the next expression if the left-hand side evaluates to `null`.

This operator can not be used in write-context, but that is not the concern of the PHPCS `Tokenizers\PHP` class.

This commit:
* Defines the token constant for PHP < 8.0.
* Adds a backfill for the nullsafe object operator for PHP < 8.0 to the PHP tokenizer.
* Adds the token to applicable token lists in the PHP and base tokenizer class, like the one used in the short array re-tokenization.
* Adds perfunctory unit tests for the nullsafe object operator backfill.
* Adds a unit test using the operator to the tokenizer tests for the short array re-tokenization.

Refs:
* https://wiki.php.net/rfc/nullsafe_operator
* https://github.com/php/php-src/commit/9bf119832dbf625174794834c71b1e793450d87f

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -121,6 +121,8 @@ http://pear.php.net/dtd/package-2.0.xsd">
       <file baseinstalldir="" name="BackfillFnTokenTest.php" role="test" />
       <file baseinstalldir="" name="BackfillNumericSeparatorTest.inc" role="test" />
       <file baseinstalldir="" name="BackfillNumericSeparatorTest.php" role="test" />
+      <file baseinstalldir="" name="NullsafeObjectOperatorTest.inc" role="test" />
+      <file baseinstalldir="" name="NullsafeObjectOperatorTest.php" role="test" />
       <file baseinstalldir="" name="ShortArrayTest.inc" role="test" />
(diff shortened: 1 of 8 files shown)


jrfnl:
@gsherwood Two FYIs:
1. The `nightly` build tests failure is not due to this PR, but to PHP 8 changes related to named identifiers as one token - see #3041 
2. While this does introduce a new token, that token is a backfill for a PHP native token, not a PHPCS specific one, so with that in mind, I believe this PR can be milestoned for `3.5.6`.

gsherwood:
Thanks. I think I've now unblocked those other PRs you have.

jrfnl:
@gsherwood One set of them for sure, though I also have sets ready for the union types and constructor property promotion.
I'll send in the ones which were blocked by this PR in a moment.

Issue #2770, opened 2019-12-17
Squiz.Commenting.FunctionComment.MissingReturn @inheritDoc not resolved?


kwhat:
Hi,

I have `@inheritDoc` as the docblock to one of my inherited methods.  The parent method is documented correctly, however, Squiz.Commenting does not appear to inherit the doc-block.

```
    /**
     * @inheritDoc
     */
    public function __toString(): string { /* Child Class */ }
```

```
    /**
     * @throws DatabaseException
     *
     * @return string
     */
    public function __toString(): string { /* Parent Class */ }
```

```
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 1 LINE
----------------------------------------------------------------------
 31 | ERROR | Missing @return tag in function comment
    |       | (Squiz.Commenting.FunctionComment.MissingReturn)
 31 | ERROR | Missing @throws tag in function comment
    |       | (Squiz.Commenting.FunctionCommentThrowTag.Missing)
----------------------------------------------------------------------
```

kwhat:
Tested with `{@inheritdoc}` and `{@inheritDoc}`, Possible regression of #214

kwhat:
Never mind, you learn something new every day.

bkdotcom:
Not sure why this was closed.   There's currently no open issue regarding inheritdoc

Seems like like if the sniff had a `$skipIfInheritdoc` (default `false`) option, everyone would be happy

`$skipIfInheritdoc` would address the common use case

```php
    /**
     * {@inheritdoc}
     */
```

kwhat:
I closed this issue after reading the PHP doc documentation, CS works according to the specification as it currently is. @inheritDoc is only for the title and description.

bkdotcom:
https://docs.phpdoc.org/latest/guides/inheritance.html

> However, it does make clear that an element has been explicitly documented (and thus not forgotten). As such we are working to include a new (normal) tag in the PHPDoc Standard @inheritDoc that will serve that purpose.

8 years have passed...

That may be the "official" specification, but the community has adopted it to include the the entire parent comment if the comment consists only of @inheritdoc

eclipse, PhpStorm, Symfony for starters support/ use it for this purpose

filips123:
Can this be reopened and supported? Basically any documentation builder (including phpDocumentor), framework and modern IDE supports `@inheritDoc` and `{@inheritDoc}`.

[...]

PR #3051, opened 2020-08-12
Add a property to Squiz.Commenting.FunctionComment to skip processing for {@inheritdoc} (issue #2770)

xjm:
https://github.com/squizlabs/PHP_CodeSniffer/issues/2770 and numerous previous closed issues have requested supporting `{@inheritdoc}` for function comments. This PR adds an optional property to the sniff that skips erroring on missing `@param`, `@return`, or `@throws` if the whole contents of the docblock are exactly `{@inheritdoc}` (case-insensitive).

Example usage:
```
    <rule ref="Squiz.Commenting.FunctionComment">                               
        <properties>
            <property name="skipIfInheritdoc" value="true"/>
        </properties>
    </rule>
```

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -18,6 +18,13 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
 {
 
     /**
+     * Whether to skip inheritdoc comments.
+     *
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks a lot for this new sniff property. This will be released in 3.6.0 and is documented here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#squizcommentingfunctioncomment

PR #3062, opened 2020-08-30
added support in JS Tokenizer for `??` and `??=`, fixed a bug in OperatorBracketSniff

ygottschalk:
I added support for the JS Operators [??](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) and [??=](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment).

In the process I found dead code in Sniff `Squiz.Formatting.OperatorBracket` which I considered a bug, fixed it and added test for PHP & JS.


ygottschalk:
I don't know why the build on PHP5.5 failed, but I think it was an issue at Travis and has nothing to do with the commit.

gsherwood:
I've removed support for JS and CSS checking from PHPCS v4 and have stopped accepting any changes to these features in v3. I'm sorry that I cannot accept the work you've submitted.

PHP_CodeSniffer is a terrible coding standards checker for JS and CSS as it doesn't support any modern syntax. If you are using PHPCS for this purpose, please migrate to a specialised tool.

PR #3063, opened 2020-08-30
PHP 8.0 | "undo" namespaced names as single token

### LINKED ISSUE #3041: PHP 8.0 | Proposal for handling "namespaced names as single token"

jrfnl:
As per the proposal in #3041.

This effectively "undoes" the new PHP 8.0 tokenization of identifier names for PHPCS 3.x.

Includes extensive unit tests to ensure the correct re-tokenization as well as that the rest of the tokenization is not adversely affected by this change.

Includes preventing `function ...` within a group use statement from breaking the retokenization.

Includes fixing the nullable tokenization when combined with any of the new PHP 8 identifier name tokens.

Fixes #3041 for PHPCS 3.x

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -121,6 +121,8 @@ http://pear.php.net/dtd/package-2.0.xsd">
       <file baseinstalldir="" name="StableCommentWhitespaceTest.php" role="test" />
       <file baseinstalldir="" name="StableCommentWhitespaceWinTest.inc" role="test" />
       <file baseinstalldir="" name="StableCommentWhitespaceWinTest.php" role="test" />
+      <file baseinstalldir="" name="UndoNamespacedNameSingleTokenTest.inc" role="test" />
+      <file baseinstalldir="" name="UndoNamespacedNameSingleTokenTest.php" role="test" />
      </dir>
      <file baseinstalldir="" name="AbstractMethodUnitTest.php" role="test" />
      <file baseinstalldir="" name="AllTests.php" role="test" />
@@ -1997,6 +1999,8 @@ http://pear.php.net/dtd/package-2.0.xsd">
    <install as="CodeSniffer/Core/Tokenizer/StableCommentWhitespaceTest.inc" name="tests/Core/Tokenizer/StableCommentWhitespaceTest.inc" />
    <install as="CodeSniffer/Core/Tokenizer/StableCommentWhitespaceWinTest.php" name="tests/Core/Tokenizer/StableCommentWhitespaceWinTest.php" />
    <install as="CodeSniffer/Core/Tokenizer/StableCommentWhitespaceWinTest.inc" name="tests/Core/Tokenizer/StableCommentWhitespaceWinTest.inc" />
+   <install as="CodeSniffer/Core/Tokenizer/UndoNamespacedNameSingleTokenTest.php" name="tests/Core/Tokenizer/UndoNamespacedNameSingleTokenTest.php" />
+   <install as="CodeSniffer/Core/Tokenizer/UndoNamespacedNameSingleTokenTest.inc" name="tests/Core/Tokenizer/UndoNamespacedNameSingleTokenTest.inc" />
    <install as="CodeSniffer/Standards/AllSniffs.php" name="tests/Standards/AllSniffs.php" />
    <install as="CodeSniffer/Standards/AbstractSniffUnitTest.php" name="tests/Standards/AbstractSniffUnitTest.php" />
   </filelist>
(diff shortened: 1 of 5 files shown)


jrfnl:
Found a few more small gotcha's, not directly related to the PHP 8 change, but for things the PHPCS tokenizer just didn't take into account while it should, which came to light while working on this.

Working on adding support for those now too.

jrfnl:
Opened PR #3066 for the other gotcha's.

jrfnl:
Note: the changes in this PR should **NOT** be cherry-picked to the PHPCS 4.x branch. I will submit a separate PR for PHPCS 4.x as per the proposal in #3041.

kukulich:
Just tested with Slevomat CS and it works perfect!

jrfnl:
@kukulich Thanks for testing this!

jrfnl:
Rebased to fix merge conflict.

gsherwood:
Thanks so much for this change

PR #3064, opened 2020-08-30
Ignore open tag if it's on the same line as the last close tag to allow inline non-PHP blocks

### LINKED ISSUE #2913: Generic.WhiteSpace.ScopeIndent false positive when opening and closing tag on same line inside conditional
The issue is reproducible on PHP_CodeSniffer version 3.5.4. Consider the following code example and standard configuration:

```php
<?php

if (true) {
    ?><?php echo addslashes(
        ''
    )?><?php
}
```
```xml
<ruleset>
    <rule ref="PSR2"/>
</ruleset>
```
The following violations are reported:
```
$ phpcs -s test.php

--------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------
 6 | ERROR | [x] Line indented incorrectly; expected at least 6 spaces, found 4
   |       |     (Generic.WhiteSpace.ScopeIndent.Incorrect)
--------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------
```
An attempt to auto-fix the file fails:
```
PHPCBF RESULT SUMMARY
/home/morozov/Projects/test/test.php                  FAILED TO FIX
A TOTAL OF 0 ERRORS WERE FIXED IN 1 FILE
PHPCBF FAILED TO FIX 1 FILE
```
The issue itself looks like a false positive since the expected number of spaces should be multiple of 4 (the tab width). The failure to fix most likely means that the sniff conflicts with some other (valid) sniff.

morozov:
Fixes #2913.

### DIFF
--- a/src/Standards/Generic/Sniffs/WhiteSpace/ScopeIndentSniff.php
+++ b/src/Standards/Generic/Sniffs/WhiteSpace/ScopeIndentSniff.php
@@ -1060,7 +1060,10 @@ class ScopeIndentSniff implements Sniff
                     $currentIndent = ($tokens[$i]['column'] - 1);
                 }
 
-                $lastOpenTag = $i;
+                // Ignore open tag if it's on the same line as the last close tag to allow inline non-PHP blocks.
+                if ($lastCloseTag === null || $tokens[$i]['line'] !== $tokens[$lastCloseTag]['line']) {
+                    $lastOpenTag = $i;
+                }
 
                 if (isset($adjustments[$i]) === true) {
                     $currentIndent += $adjustments[$i];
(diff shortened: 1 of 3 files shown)


gsherwood:
I need to look into this fix more. I don't like the idea of pretending that the open tag doesn't exist, but it obviously has no impact on the sniff unit tests so it is possibly fine. But I'd at least need to add debugging code around it to make sure I can track down any issues in the future.

gsherwood:
I investigated this further and found that there is already code that deals with open tags inside conditionals, but it makes the assumption that the open tag is the first thing on the line. To fix the original issue, I looked for the actual first content on the line (in this case it was the closing tag) and used that for the indent.

So I've decided not to merge this in as I don't think the opening tag needs to be ignored. Your fix also works, but I've just tried to limit the blast radius of the change a bit with this one. If more cases pop up where the opening tag is causing an issue, I may end up using this fix, but I'll close it for now.

Thanks for submitting the fix though.

morozov:
Thank you for taking the time to provide an alternative solution. I'm happy with whatever approach as long as the issue is fixed.

PR #3067, opened 2020-08-30
Travis: simpify PHPStan build

jrfnl:
As there is already a PHP 7.4 build in place running exactly the same build as the PHP 7.4 PHPStan build, with the exception of the actual PHPStan run, we can skip the "normal" build checks for the PHPStan build and just and only run PHPStan.

If so desired, I could also move the XML lint & PEAR validation check to a separate build running only those checks, simplifying the Travis script some.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,10 @@ jobs:
         apt:
           packages:
             - libonig-dev
+      script:
+      - composer require --dev phpstan/phpstan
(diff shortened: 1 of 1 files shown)


gsherwood:
Well that makes total sense :) Thanks a lot. I'd certainly accept more cleanup if you want to submit more.

jrfnl:
@gsherwood I've send in a follow-up PR with a further iteration on this. #3073

PR #3070, opened 2020-08-31
PHP 8.0 | PSR12/NullableTypeDeclaration: allow for static return type

jrfnl:
PHP 8.0 introduces `static` as a valid return type. This PR adjusts the `PSR12.Functions.NullableTypeDeclaration` sniff to handle that return type as well.


gsherwood:

PR #3071, opened 2020-08-31
Tokenizer: support hash comment for ignore annotations

jrfnl:
Tokenizer: support hash comment for ignore annotations

Until now, for the PHPCS native ignore annotations, only `//` slash or `/* */` star-style comments were supported.

This adds support for PHPCS native ignore annotations using `#` hash-style comments.

Includes unit tests and syncing of the `ltrim()` used in `File::process()` with the `ltrim()` used in `Tokenizer::createPositionMap()`.

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -356,7 +356,7 @@ class File
                 || $token['code'] === T_DOC_COMMENT_TAG
                 || ($inTests === true && $token['code'] === T_INLINE_HTML))
             ) {
-                $commentText      = ltrim($this->tokens[$stackPtr]['content'], ' /*');
+                $commentText      = ltrim($this->tokens[$stackPtr]['content'], " \t/*#");
                 $commentTextLower = strtolower($commentText);
                 if (strpos($commentText, '@codingStandards') !== false) {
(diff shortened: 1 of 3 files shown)


jrfnl:
Let's put this PR on hold for now until support for PHP 8.0 attributes has been investigated as it may complicate that.

gsherwood:
Sorry, I never came back to this after attribute support was added.

jrfnl:
Shall I rebase and add some tests to make sure they pass ?

jrfnl:
@gsherwood I'm also tempted to refactor that test file to use data providers to see more easily which cases are being tested. Would you be okay with that ? If so, I'll do it after we got this PR out of the way.

jrfnl:
PR updated, now including allowance for "late suppression" with `phpcs:ignoreFile` and extensive unit tests.

All seems to work fine.

gsherwood:
> @gsherwood I'm also tempted to refactor that test file to use data providers to see more easily which cases are being tested. Would you be okay with that ? If so, I'll do it after we got this PR out of the way.

Sounds like a good idea.

gsherwood:
Thanks a lot for finishing this off. Merged now.

PR #3073, opened 2020-09-01
Travis: further tweaks for cleaner output

jrfnl:
Further iteration after #3067.

This commit does a couple of things:
1. It adds a new build against PHP 7.4 which runs just and only the PEAR package file validation and the XML related checks.
    That build doesn't need to do a `composer install` and, as it has it's own script, will show just the output of those checks.
2. It adds a separate `script` section to the two builds being run with custom PHP `ini` files (and moves those down in the matrix).

The end result of this is:
* We can get rid of all the conditions in the `script` section.
* Builds will show much cleaner output, so it is easier to find out what went wrong.

For the "custom builds", the Travis script now has a separate "name" for these, so they can still be easily identified for what they are, without the environment variables (which is what used to show).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,35 +5,19 @@ language: php
 jobs:
   fast_finish: true
   include:
+    #############################################
+    # Builds using the default test script.
+    #############################################
     - php: 5.4
       dist: trusty
     - php: 5.5
       dist: trusty
-    - php: 5.5
-      dist: trusty
-      env: CUSTOM_INI=1 XMLLINT=1
-      addons:
-        apt:
-          packages:
-            - libxml2-utils
     - php: 5.6
     - php: 7.0
-    - php: 7.0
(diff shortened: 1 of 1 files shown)


jrfnl:
@gsherwood Would you like me to send in a similar PR for the PHPCS 4.x branch ?

jrfnl:
*Grumble* Starting to get increasingly annoyed with Travis/Packagist for install failures during builds. I've force pushed five times now trying to get the build to pass... It really isn't an issue with this PR. Wish that Travis would allow the person who sends in a PR to restart individual builds in the PR build...

gsherwood:
This is **soooo** much cleaner. I love it! Thank you.

> Would you like me to send in a similar PR for the PHPCS 4.x branch ?

If you want to take the time to do it, I'll very happily accept it :)

jrfnl:
Done, see #3119

PR #3077, opened 2020-09-01
PHPCS 4.x | Add missing commits

jrfnl:
These commits seem to be missing from the 4.x branch. Most are recent, but commit 9d690f0 is from quite a while ago.


gsherwood:
I haven't pushed my local branch in a few days. Will do now.

gsherwood:
Looks like that verbose message was all that was missed. Thanks for pointing that out - will get that in now.

gsherwood:
Have pushed up that verbose message now as well. Thanks for reviewing the commits. I did this in April (I think) when I realised I had been forgetting, but clearly missed one.

jrfnl:
@gsherwood Presuming that means 4.x is passing again, I'll rebase the other 4.x PRs to show the passing builds.

PR #3078, opened 2020-09-01
PHPCS 4.x | Fix inconsistent output

jrfnl:
Commit d4f33dbf45aa5bb78e077b0e73306746c8bd98cf changed the output from STDOUT to STDERR using `Common::printStatusMessage()` instead of plain `echo`'s.

In the mean time, some new `echo`'s have slipped in. This fixes those up.


jrfnl:
Rebased and updated for newly introduced `echo`s once again.

gsherwood:
Thanks for fixing these

PR #3082, opened 2020-09-01
PHP 8.0 | Generic/ScopeIndent: allow for nullsafe object operator

jrfnl:
Includes unit test.


gsherwood:
I always leave everything after what is currently line 1457 (I should comment that) and also keep the space and tab indented versions (the 1 and 2 files) in sync at all times. I can come back and do this later, but that's why it hasn't been merged yet.

jrfnl:
@gsherwood  Do you mean line 1457 or 1467 ? If 1457, it would mean that there are still two tests which would have to be renumbered each time.

jrfnl:
Either way, fixed based on line 1457 now.

gsherwood:
Thanks for this, and for moving those tests (and the comment).

PR #3083, opened 2020-09-01
PHP 8.0 | PEAR/ObjectOperatorIndent: sniff for nullsafe object operator

jrfnl:
This adds sniffing for the indentation of nullsafe object operators to the sniff.

Includes unit test.


gsherwood:
I noticed that none of the new tests have the nullsafe object operator indented incorrectly. I'm not sure if that's intentional, but it's something I was going to play around with before merging.

jrfnl:
> I noticed that none of the new tests have the nullsafe object operator indented incorrectly. I'm not sure if that's intentional, but it's something I was going to play around with before merging.

Added more explicit "OK" tests. The "need to be fixed" tests were already in place.

gsherwood:
Thanks for those extra tests

PR #3089, opened 2020-09-01
PHP 8.0 | Squiz/DisallowComparisonAssignment: remove redundant code

jrfnl:
While looking to add the `T_NULLSAFE_OBJECT_OPERATOR`, I realized the `T_OBJECT_OPERATOR` here did not have a function.

Removing it still allows all the tests to pass.

### DIFF
--- a/src/Standards/Squiz/Sniffs/PHP/DisallowComparisonAssignmentSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/DisallowComparisonAssignmentSniff.php
@@ -70,7 +70,6 @@ class DisallowComparisonAssignmentSniff implements Sniff
         $ignore = [
             T_STRING,
             T_WHITESPACE,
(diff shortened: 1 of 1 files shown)


gsherwood:
I think this is actually required, but was never implemented correctly.

The sniff throws an error for this code: `$var = $foo->something(!$var);`
> The value of a boolean operation must not be assigned to a variable

I think I forgot (11 years ago) to add `T_VARIABLE` in here. I will look into it more later, but I think there is actually a bug in here that needs fixing.

jrfnl:
In that case, when that bug gets fixed, the `T_NULLSAFE_OBJECT_OPERATOR` token will probably also need to be added.

gsherwood:
Have fixed up that bug, and added `T_NULLSAFE_OBJECT_OPERATOR` support. Thanks for that prompt.

PR #3096, opened 2020-09-02
Tokenizer/PHP: add missing scope closers

jrfnl:
Noticed (while working on something else) there were some typical scope closers missing from the array.


gsherwood:
Thanks for this. I feel like something should have broken without these tokens there, but never seen anything related.

PR #3102, opened 2020-09-05
PSR12.Squiz.OperatorSpacing false positive for default values of arrow functions

jrfnl:
The plus/minus in a default value of a function parameter declaration is ignored for normal functions and closures, but wasn't ignored yet for arrow functions.

Fixed now.

Includes unit test + test for the same for closures, which so far wasn't tested yet.

This also, by extension, fixes the same issue in the `PSR12.Operators.OperatorSpacing` sniff.


jrfnl:
Rebased for merge conflicts

gsherwood:
Thanks a lot for this fix. It has been merged in, but I was doing a rebase and the commit hashes got changed over.

Issue #2593, opened 2019-08-26
Add T_USE to the list of parenthesis openers


gsherwood:
T_USE can open parenthesis, or can be a scope opener. It currently doesn't get parenthesis opener/closer indexes when used as a USE group in a closure. Adding support isn't that hard, but may break backwards compatibility for some sniffs.

A diff of changes that could be required in the core:
```diff
diff --git a/src/Sniffs/AbstractVariableSniff.php b/src/Sniffs/AbstractVariableSniff.php
index 5dc8ba556..f2139b15a 100644
--- a/src/Sniffs/AbstractVariableSniff.php
+++ b/src/Sniffs/AbstractVariableSniff.php
@@ -118,19 +118,13 @@ abstract class AbstractVariableSniff extends AbstractScopeSniff
         if ($inFunction === false && isset($tokens[$stackPtr]['nested_parenthesis']) === true) {
             foreach ($tokens[$stackPtr]['nested_parenthesis'] as $opener => $closer) {
                 if (isset($tokens[$opener]['parenthesis_owner']) === false) {
-                    // Check if this is a USE statement for a closure.
-                    $prev = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($opener - 1), null, true);
-                    if ($tokens[$prev]['code'] === T_USE) {
-                        $inFunction = true;
-                        break;
-                    }
-
                     continue;
                 }

                 $owner = $tokens[$opener]['parenthesis_owner'];
                 if ($tokens[$owner]['code'] === T_FUNCTION
                     || $tokens[$owner]['code'] === T_CLOSURE
+                    || $tokens[$owner]['code'] === T_USE
                 ) {
                     $inFunction = true;
                     break;
diff --git a/src/Tokenizers/Tokenizer.php b/src/Tokenizers/Tokenizer.php
index bc904586c..677448e24 100644
--- a/src/Tokenizers/Tokenizer.php
+++ b/src/Tokenizers/Tokenizer.php
@@ -691,13 +691,26 @@ abstract class Tokenizer
                 $this->tokens[$i]['parenthesis_closer'] = null;
                 $this->tokens[$i]['parenthesis_owner']  = $i;
                 $openOwner = $i;
+
+                if (PHP_CODESNIFFER_VERBOSITY > 1) {
+                    echo str_repeat("\t", (count($openers) + 1));
+                    echo "=> Found parenthesis owner at $i".PHP_EOL;
+                }
[...]

PR #3104, opened 2020-09-05
PHPCS 4.x | Tokenizer: fix stray "parenthesis_..." indexes being set for non-closure use

jrfnl:
Commit 08824f327ce5e20d7528c5e838dd7fade3dcd11c (issue #2593) added support for `T_USE` tokens for closures being parentheses owners.

However, the net effect was that *every* `T_USE` token - including import/trait use statement `T_USE` tokens, would have the `parenthesis_owner`, `parenthesis_opener` and `parenthesis_closer` indexes set.

For closure `use` statements, those were set correctly.

However, for import/trait `use` statements, the `parenthesis_owner` would point to the `T_USE` keyword which doesn't have parentheses and the `parenthesis_opener` and `parenthesis_closer`  indexes would be `null` in most cases, but they would still be set.

Also, in some cases, the `parenthesis_opener` and `parenthesis_closer`  indexes for import/trait `use` statements would incorrectly be set if to arbitrary, unrelated parenthesis, if no open curly or open square bracket was encountered between the `use` statement and the next set of parenthesis.

That makes the `parenthesis_...` indexes harder and less intuitive to work with as any of the `parenthesis_...` indexes could be set, even when the token has no parenthesis.

I've fixed this now by:
* Not just resetting the `$openOwner` variable when we know there won't be any parenthesis, but by also resetting the `parenthesis_...` indexes of the incorrectly set owner.
* Doing the reset on more tokens, including doing it on a `T_OPEN_USE_GROUP` token, as the `tokenize()` method has already run, so group use tokens have already been retokenized to their dedicated token.
    Note: There may be some more tokens which can be added to this list, but this should be a reasonable start.

I've also added a few `continue` statements to skip the rest of the code within the loop when we know in advance none of the other conditions would match anyway.

Includes adding dedicated unit tests verifying the correct setting of the `parenthesis_...` indexes.

### DIFF
--- a/src/Tokenizers/Tokenizer.php
+++ b/src/Tokenizers/Tokenizer.php
@@ -651,6 +651,8 @@ abstract class Tokenizer
                 if (PHP_CODESNIFFER_VERBOSITY > 1) {
                     Common::printStatusMessage("=> Found parenthesis owner at $i", (count($openers) + 1));
                 }
+
+                continue;
             } else if ($this->tokens[$i]['code'] === T_OPEN_PARENTHESIS) {
                 $openers[] = $i;
                 $this->tokens[$i]['parenthesis_opener'] = $i;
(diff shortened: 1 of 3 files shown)


jrfnl:
Rebased yet again for merge conflicts. @gsherwood Would be great if this PR could get some attention as this bug is currently the main reason why external standards will fail against PHPCS 4.x when running tests.

gsherwood:
Sorry, merged now. Thanks a lot for fixing this up.

jrfnl:
@gsherwood Thanks for that!

PR #3105, opened 2020-09-05
File::isReference(): simplify logic

jrfnl:
The `File::getMethodParameters()` returns a `reference_token` index, so no need to do any token walking.


gsherwood:
Thanks for finding this improvement

PR #3106, opened 2020-09-05
Runner::printProgress(): fix alignment of file counts

### LINKED ISSUE #3058: Progress gets unaligned when 100% happens at the end of the available dots
### Expected vs Actual output diff

```diff
 $ vendor/bin/phpcs
 ............................................................  60 / 120 (50%)
-............................................................  120 / 120 (100%)
+............................................................                                                             120 / 120 (100%)
```

jrfnl:
... when the total number of files is exactly dividable by 60.

Fixes #3058

For a "before" screenshot see the original issue.

"After" screenshot tested with various file counts:

image

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -874,7 +874,7 @@ class Runner
 
         $percent = round(($numProcessed / $numFiles) * 100);
         $padding = (strlen($numFiles) - strlen($numProcessed));
-        if ($numProcessed === $numFiles && $numFiles > $numPerLine) {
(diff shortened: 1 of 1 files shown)


gsherwood:
Sorry :(

I got a bit excited when I saw the original report because I remember finding and fixing it in 4.0 (but forgetting to backport). I picked it in before noticing this.

Looks like we had the same fix in the end.

Thanks for taking the time to look into this though.

PR #3113, opened 2020-09-15
Allow variables to be used in double quotes

duncan3dc:
We would like to replace/error on unnecessary use of double quotes, however if there is a variable then we don't want to enforce concatenation. This setting allows double quoted strings to be accepted if they contain variables. Which also means this sniff can run in a mode that only raises fixable errors


duncan3dc:
It's been brought to my attention that this is possible by just excluding the `Squiz.Strings.DoubleQuoteUsage.ContainsVar` rule, sorry for the noise

PR #3119, opened 2020-09-21
PHPCS 4.x | Travis: tweaks for cleaner output

jrfnl:
Sister-PR to PRs #3067 and #3073 for the PHPCS 4.x branch.

Basically does the same thing:
1. Skip the "normal" build checks for the PHPStan build and just and only run PHPStan.
2. Add a new build against PHP 7.4 which runs just and only the XML related checks.
    That build doesn't need to do a `composer install` and, as it has it's own script, will show just the output of those checks.
3. Add a separate `script` section to the build being run with custom a PHP `ini` file (and moves those down in the matrix).

For the "custom builds", the Travis script now has a separate "name" for these, so they can still be easily identified for what they are, without the environment variables (which is what used to show).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,25 +5,12 @@ language: php
 jobs:
   fast_finish: true
   include:
-    - php: 7.2
-      env: CUSTOM_INI=1 XMLLINT=1
-      addons:
-        apt:
-          packages:
-            - libxml2-utils
+    #############################################
+    # Builds using the default test script.
+    #############################################
     - php: 7.2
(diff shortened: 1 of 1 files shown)


gsherwood:
I really love the Travis output with the new build names. Thanks a lot for this.

jrfnl:
We could give all builds names, but for now, I just added them for the "different" (from default script) builds. And yes, I find the cleaner output, especially when a build fails much easier to scan through as well, so happy to get this set up.

PR #3121, opened 2020-09-21
PHPCS 4.x: let `T_STATIC` be `T_STATIC`

### LINKED ISSUE #3115: PHPCS 4.x: let `T_STATIC` be `T_STATIC`
The `static` keyword in PHP is used for a number of different things:
* Static properties
* Static methods
* Static closures and arrow functions
* Static variables
* Late static binding

Now, for all these uses, the token is tokenized as `T_STATIC`, with one exception: late static binding in combination with the `instanceof` operator.

This defies the principle of _least astonishment_ as both `self` and `parent` in combination with `instanceof` _are_ tokenized as `T_SELF` and `T_PARENT` respectively.

I've been caught out by this a couple of times already and I imagine that others have too, or if they haven't, that sniffs may be throwing false positives/negatives because of it.

It's not completely clear to me why this very specific exception was put in place, but I'd like to propose to remove this exception in PHPCS 4.x and let the `static` keyword be tokenized as `T_STATIC` when preceded by `instanceof`.

Based on a history search, it looks to have been put in place in response to [PEAR#19726](https://pear.php.net/bugs/bug.php?id=19726).
In my opinion, that issue should have been fixed in the sniff code for the `Squiz.WhiteSpace.ScopeKeywordSpacing` sniff, not in the Tokenizer, same as is done in the sniff with other typical late static binding occurrences.

https://github.com/squizlabs/PHP_CodeSniffer/blob/9cce6859a595b3fa77fc823d9b0d451951dcf9d2/src/Tokenizers/PHP.php#L2086-L2101

**Code sample**
```php
if ($geometry instanceof static) {
    echo 'foo';
}
```
`static` in the above code sample is tokenized as `T_STRING`, not `T_STATIC`.

jrfnl:
Implementation of the proposal outlined in issue #3115.

1. Removes the Tokenizer/PHP code which made an exception for `instanceof static`.
2. Adds an exception for `instanceof static` to the `Squiz.WhiteSpace.ScopeKeywordSpacing` sniff for which the tokenizer code was originally put in place.

The tests for all other sniffs still pass.

I've had a quick look through the other sniffs which refer to the `T_STATIC` token, but didn't see any which should be impacted by this change (and didn't have unit tests with `instanceof static`).

Fixes #3115

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/ScopeKeywordSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/ScopeKeywordSpacingSniff.php
@@ -55,6 +55,7 @@ class ScopeKeywordSpacingSniff implements Sniff
         if ($tokens[$stackPtr]['code'] === T_STATIC) {
             if (($nextToken === false || $tokens[$nextToken]['code'] === T_DOUBLE_COLON)
                 || $tokens[$prevToken]['code'] === T_NEW
+                || $tokens[$prevToken]['code'] === T_INSTANCEOF
             ) {
                 // Late static binding, e.g., static:: OR new static() usage or live coding.
                 return;
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -2800,24 +2800,6 @@ class PHP extends Tokenizer
                 }
 
                 continue;
-            } else if ($this->tokens[$i]['code'] === T_STATIC) {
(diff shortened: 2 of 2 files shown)


jrfnl:
Rebased for imaginary merge conflict.

jrfnl:
Rebased for merge conflict.

jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/37

PR #3129, opened 2020-09-29
Squiz/PSR12/OperatorSpacing: bug fix - unary plus/minus in arrow function return

### LINKED ISSUE #3043: Squiz.WhiteSpace.OperatorSpacing false positive for negation in arrow function
**Describe the bug**
Minus `-` used as negation of brackets result is okay in `function` but not in `fn`

**Code sample**

This **is** reported
```php
static fn(DateTime $a, DateTime $b): int => -($a->getTimestamp() <=> $b->getTimestamp())
```

This **is not** reported
```php
static function (DateTime $a, DateTime $b): int {
	return -($a->getTimestamp() <=> $b->getTimestamp());
}
```

**Custom ruleset**
```xml
<ruleset>
  <rule ref="Squiz.WhiteSpace.OperatorSpacing">
		<properties>
			<property name="ignoreNewlines" value="true"/>
		</properties>
	</rule>
</ruleset>
```

**To reproduce**
2. Create a file called `ruleset.xml` with custom ruleset above
2. Run `phpcs test.php --standard=ruleset.xml`
```
Expected 1 space after "-"; 0 found (Squiz.WhiteSpace.OperatorSpacing.NoSpaceAfter)
```

**Expected behavior**

Code is not reported as invalid or at least have same behavior for both contexts (`function` and `fn`)

jrfnl:
A plus/minus directly after the arrow of an arrow function, will always be a unary plus/minus and should therefore be ignored by these sniffs.

Includes unit tests.

Includes two additional simplifications:
* `T_DOUBLE_ARROW` is already part of the `Tokens::$assignmentTokens` array, so doesn't need to be added separately.
* The cast tokens array was incomplete (missing `T_BINARY_CAST`) and can be replaced by the more complete `Tokens::$castTokens` array anyhow.

Fixes #3043

### DIFF
--- a/src/Standards/PSR12/Tests/Operators/OperatorSpacingUnitTest.inc
+++ b/src/Standards/PSR12/Tests/Operators/OperatorSpacingUnitTest.inc
@@ -59,3 +59,5 @@ $a = [ 'a' => &$something ];
 
 $fn = fn(array &$one) => 1;
 $fn = fn(array & $one) => 1;
+
+$fn = static fn(DateTime $a, DateTime $b): int => -($a->getTimestamp() <=> $b->getTimestamp());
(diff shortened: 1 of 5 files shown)


jrfnl:
@gsherwood As there are already three other fixes for the operator spacing sniffs in `3.5.7`, including two others related to arrow functions, I imagine this one can go into `3.5.7` as well.

gsherwood:
Thanks for the fix and general sniff cleanup

PR #3132, opened 2020-09-30
Fix docblocks for File::findPrevious and File::findNext

mmoll:


mmoll:
Otherwise this breaks https://github.com/mayflower/mo4-coding-standard/pull/159 for me:
```
Argument 3 ($end) is null of type null but \PHP_CodeSniffer\Files\File::findNext() takes int (no real type) defined at vendor/squizlabs/php_codesniffer/src/Files/File.php:2202 (the inferred real argument type has nothing in common with the parameter's phpdoc type)
Argument 5 ($value) is null of type null but \PHP_CodeSniffer\Files\File::findPrevious() takes string (no real type) defined at vendor/squizlabs/php_codesniffer/src/Files/File.php:2121 (the inferred real argument type has nothing in common with the parameter's phpdoc type)
```

gsherwood:
Thanks for this doc fix

PR #3135, opened 2020-10-02
Ignore annotations are broken on PHP 8.0

jrfnl:
A very recent change in PHP 8.0 changes the possible return values of the `substr()` function from:

> Pre-PHP 8:
> Returns the extracted part of string; or FALSE on failure, or an empty string.

> PHP 8-RC1:
> Returns the extracted part of string; or an empty string.

This is an insidious change as basically all code (strict) checking the return value of `substr()` against `false` will now be broken.

Checking the return value with `empty()` will fix this in a cross-version compatible manner as it allows for both `false` as well as an empty string being returned.

This change broke the ignore annotations as implemented in PHPCS.

The existing unit tests for the ignore annotations cover this change.

Includes removing some unnecessary, duplicate function calls to `substr()`.

Ref: https://github.com/php/php-src/pull/6182

### DIFF
--- a/src/Tokenizers/Tokenizer.php
+++ b/src/Tokenizers/Tokenizer.php
@@ -424,10 +424,10 @@ abstract class Tokenizer
                         $disabledSniffs = [];
 
                         $additionalText = substr($commentText, 14);
-                        if ($additionalText === false) {
+                        if (empty($additionalText) === true) {
                             $ignoring = ['.all' => true];
                         } else {
-                            $parts = explode(',', substr($commentText, 13));
+                            $parts = explode(',', $additionalText);
                             foreach ($parts as $sniffCode) {
                                 $sniffCode = trim($sniffCode);
                                 $disabledSniffs[$sniffCode] = true;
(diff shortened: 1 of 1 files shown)


Ayesh:
I was bit with this too. [`substr`](https://php.watch/versions/8.0/substr-out-of-bounds#substr) and [`iconv_substr`](https://php.watch/versions/8.0/substr-out-of-bounds#iconv_substr) functionality is changed recently in PHP `master` branch. [`mb_substr`](https://php.watch/versions/8.0/substr-out-of-bounds#mb_substr) and [`grapheme_substr`](https://php.watch/versions/8.0/substr-out-of-bounds#grapheme_substr) has not.

jrfnl:
@Ayesh Yup, it's a nasty one. At least it's now mentioned in the changelog (after I made remark about it).

gsherwood:
I'm very thankful you found and fixed this. Bit of a rough last minute change.

jrfnl:
Well, at least there were tests in place to find it quickly ;-)

From a PHP Core perspective, I do think it is a good change as it makes the function return type of these functions more consistent going forward, but yes, BC-break-wise it's nasty.

Issue #2974, opened 2020-05-28
Wrong error on Squiz.ControlStructures.SwitchDeclaration break indentation


djibarian:
The documentation states that "*the terminating statement is indented 4 spaces from the case or default keyword*", but doesn’t seem to be the case:

```
switch(a){
  case “b”:
    foo();
    break;
}
```

generates an error saying that the `break` statement must be indented one stop from the SWITCH statement, which would be then:

```
switch(a){
  case “b”:
  foo();
  break;
}
```

So whether the rule or the docs should be fixed.

jrfnl:
> The documentation states that "the terminating statement is indented 4 spaces from the case or default keyword", but doesn’t seem to be the case:

@djibarian Which documentation are you referring to ? There is no such phrase/documentation in the `Squiz.ControlStructures.SwitchDeclaration` sniff.
The error thrown by the sniff - `CASE keyword must be indented 4 spaces from SWITCH keyword` - and the fixes made are in line with each other, so I don't understand what you are referring to.

If you want a sniff which indents the `break` to the same level as the `case`/`default` body, use the `PSR2.ControlStructures.SwitchDeclaration` sniff instead.

danhunsaker:
https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#squizcontrolstructuresswitchdeclaration

gsherwood:
@danhunsaker Thanks for pointing that out. I have corrected the description over there to reflect what the sniff actually enforces.

PR #3137, opened 2020-10-07
Fix for #2974

danhunsaker:
This commit brings the code in line with the documentation.

### DIFF
--- a/src/Standards/Squiz/Sniffs/ControlStructures/SwitchDeclarationSniff.php
+++ b/src/Standards/Squiz/Sniffs/ControlStructures/SwitchDeclarationSniff.php
@@ -66,11 +66,12 @@ class SwitchDeclarationSniff implements Sniff
             return;
         }
 
-        $switch        = $tokens[$stackPtr];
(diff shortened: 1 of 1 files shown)


jrfnl:
@danhunsaker This is not the intention of that particular sniff and this change would be a huge BC-break.

If you want a sniff which indents the `break` to the same level as the `case`/`default` body, use the `PSR2.ControlStructures.SwitchDeclaration` sniff instead.

danhunsaker:
I strongly recommend fixing the documentation, then, because it states otherwise.

gsherwood:
> I strongly recommend fixing the documentation, then, because it states otherwise.

Which documentation do you mean?

danhunsaker:
https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#squizcontrolstructuresswitchdeclaration

gsherwood:
@danhunsaker Thanks for pointing that out. I have corrected the description over there to reflect what the sniff actually enforces.

PR #3139, opened 2020-10-10
Add support to use short forms of type keywords

### LINKED ISSUE #1434: IncorrectParamVarName: Problem with short and long types (int / integer)
Hi!
Version: 3.0.0RC4
Ubuntu - Jenkins

Problem:

I get a Warning: Expected "integer" but found "int" for parameter typ

`/**
* Load video
*
* @param int $gid Gid.
* @return \stdClass
*/
private function loadVideo(int $gid): \stdClass`

My phpcs.xml looks like this:
`<?xml version="1.0"?>
<ruleset name="PSR2 + PhpDoc">

    
    <rule ref="PSR2"/>

    <description>My custom coding standard.</description>
    <rule ref="Squiz.Commenting.FunctionComment">
        <exclude name="Squiz.Commenting.FunctionComment.MissingParamComment"/>
        <exclude name="Squiz.Commenting.FunctionComment.SpacingAfterParamType"/>
        <exclude name="Squiz.Commenting.FunctionComment.ThrowsNotCapital"/>
    </rule>
    <rule ref="Squiz.Commenting.FunctionCommentThrowTag" />
    <rule ref="Squiz.Commenting.ClassComment">
        <exclude name="Squiz.Commenting.ClassComment.TagNotAllowed"/>
    </rule>
    <rule ref="Squiz.Commenting.VariableComment" />
</ruleset>`

Is this a bug or do I need to use another rule?

Thanks!

### LINKED ISSUE #1864: Type hinting: bool or boolean
The sniffer makes a pretty strange requirement. He is always want to see the `boolean` as a type hint. But, at the same time PHP 7.1 supports only `bool` as a type hinting. So, to make everyone happy I need to write the hinting in two versions:

```php 
   /**
     * Some comment here
     *
     * @return boolean
     */
    public function isUserActive(): bool
    {
        return false;
    }
```

This is weird from the coding viewpoint.

Here is a discussion about the right hinting <https://stackoverflow.com/a/44009101/5497749>

filips123:
This adds support to use short forms of type keywords in comments for function parameters and returns and variable types. It fixes #1864 and fixes #1434.

It adds `useShortTypes` property, which is by default `false`, to `Squiz.Commenting.FunctionComment` and `Squiz.Commenting.VariableComment` to allow users to use short type checks. If it is enabled, it will force short types in comments and suggest them when long ones are used.

P.S. I would appreciate if you add `hacktoberfest-accepted` label to this PR, so it will count for this year's Hacktoberfest.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -24,6 +24,13 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
      */
     private $phpVersion = null;
 
+    /**
+     * Whether to use short forms of type keywords.
+     *
+     * @var boolean
+     */
+    public $useShortTypes = false;
+
 
     /**
      * Process the return comment of this function comment.
@@ -77,7 +84,7 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
                 $typeNames      = explode('|', $returnType);
                 $suggestedNames = [];
                 foreach ($typeNames as $i => $typeName) {
-                    $suggestedName = Common::suggestType($typeName);
+                    $suggestedName = Common::suggestType($typeName, $this->useShortTypes);
                     if (in_array($suggestedName, $suggestedNames, true) === false) {
                         $suggestedNames[] = $suggestedName;
                     }
@@ -382,7 +389,7 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
                     $typeName = substr($typeName, 1);
                 }
 
-                $suggestedName        = Common::suggestType($typeName);
+                $suggestedName        = Common::suggestType($typeName, $this->useShortTypes);
                 $suggestedTypeNames[] = $suggestedName;
 
                 if (count($typeNames) > 1) {
--- a/src/Standards/Squiz/Sniffs/Commenting/VariableCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/VariableCommentSniff.php
@@ -16,6 +16,13 @@ use PHP_CodeSniffer\Util\Common;
 class VariableCommentSniff extends AbstractVariableSniff
 {
 
+    /**
+     * Whether to use short forms of type keywords.
+     *
+     * @var boolean
+     */
+    public $useShortTypes = false;
+
 
     /**
      * Called to process class member vars.
@@ -113,7 +120,7 @@ class VariableCommentSniff extends AbstractVariableSniff
         $typeNames      = explode('|', $varType);
         $suggestedNames = [];
         foreach ($typeNames as $i => $typeName) {
-            $suggestedName = Common::suggestType($typeName);
+            $suggestedName = Common::suggestType($typeName, $this->useShortTypes);
             if (in_array($suggestedName, $suggestedNames, true) === false) {
                 $suggestedNames[] = $suggestedName;
             }
--- a/src/Util/Common.php
+++ b/src/Util/Common.php
@@ -29,6 +29,23 @@ class Common
         'callable',
     ];
 
+    /**
+     * An array of short variable types for param/var we will check.
+     *
+     * @var string[]
+     */
+    public static $allowedShortTypes = [
+        'array',
+        'bool',
+        'float',
+        'int',
+        'mixed',
+        'object',
+        'string',
+        'resource',
+        'callable',
+    ];
+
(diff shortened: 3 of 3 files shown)


src/Standards/Squiz/Sniffs/Commenting/VariableCommentSniff.php
```diff
@@ -16,6 +16,13 @@
 class VariableCommentSniff extends AbstractVariableSniff
 {
 
+    /**
+     * Whether to use short forms of type keywords.
+     *
+     * @var boolean
+     */
+    public $useShortTypes = false;
+
```
halfpastfouram, line 25: Remove this empty line.

```suggestion
```

src/Util/Common.php
```diff
@@ -29,6 +29,23 @@ class Common
...
+        'bool',
+        'float',
+        'int',
+        'mixed',
+        'object',
+        'string',
+        'resource',
+        'callable',
+    ];
+
```
halfpastfouram, line 48: Remove this empty line.

```suggestion
```

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -24,6 +24,13 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
      */
     private $phpVersion = null;
 
+    /**
+     * Whether to use short forms of type keywords.
+     *
+     * @var boolean
+     */
+    public $useShortTypes = false;
```
halfpastfouram, line 32: Shouldn't this default to true due to PSR-12?

halfpastfouram (changes requested):

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
jrfnl: Please don't request changes which are in violation with the coding standard used by the library.

src/Standards/Squiz/Sniffs/Commenting/VariableCommentSniff.php
jrfnl: Please don't request changes which are in violation with the coding standard used by the library.

src/Util/Common.php
jrfnl: Please don't request changes which are in violation with the coding standard used by the library.

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
filips123: > Shouldn't this default to true due to PSR-12?

Problem is that this might be considered as a breaking change. I'm not sure how this was done in the past, but maybe it should remain `false` in PHPCS 3.x and changed to `true` in 4.x.

halfpastfouram: I understand. In the mean time it's easy to just change that property in the config file.

bramstroker:
Any chance to get this merged? Short return types are the way to go, especially because it does not comply to the rules of PSR-12 currently. "Short form of type keywords MUST be used i.e. bool instead of boolean, int instead of integer etc.".

ZebraNorth:
Until such time as this can be merged, I've added a workaround here: https://packagist.org/packages/zebra-north/phpcs-short-types

It's not a proper fix, it would be great if this PR could be merged.

drdrak3:
I would love for this to be merged, we are using PHPStorm and it automatically (and correctly) uses short types which we have to manually update to pass our code sniffers

richard-hp:
Merging this in would be useful

Justintime50:
Would definitely love to see this merged and released! Super odd to have code type hints be one thing and docblock types not enforced to match.

fredden:
@Justintime50 it is very unlikely that this will be merged here. See #3932.

Justintime50:
Seeing as it's unlikely this makes it into PHPCS, I've created a sniff that enforces short scalar type names with auto-fix as well: https://github.com/Justintime50/phpcs-short-scalar-types. This can be used in lieu of the missing 1st party support.

This sniff will error if `boolean` or `integer` are used in comments/docblocks and instead instruct to use `bool` and `int`. Run the auto-fixer to cleanup a project real quick.

Hope it's helpful for future travelers!

filips123:
@fredden Should this PR then be moved to the new repository?

PR #3142, opened 2020-10-14
Keep correct indent when moving brace to a new line

chindit:
When using this rule, brace is moved at the beginning of a new line.
Example:
```
    public function foo()
{
        // stuff
    }
```

With this fix, correct indent is used:
```
    public function foo()
    {
        // stuff
    }
```

### DIFF
--- a/src/Standards/Generic/Sniffs/Functions/OpeningFunctionBraceBsdAllmanSniff.php
+++ b/src/Standards/Generic/Sniffs/Functions/OpeningFunctionBraceBsdAllmanSniff.php
@@ -121,6 +121,9 @@ class OpeningFunctionBraceBsdAllmanSniff implements Sniff
                         $phpcsFile->fixer->replaceToken(($openingBrace - 1), '');
                     }
 
+	                $indentSize = $tokens[$phpcsFile->findPrevious([T_IF, T_ELSEIF, T_ELSE], $stackPtr)]['column'];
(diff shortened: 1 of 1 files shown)


gsherwood:
This sniff (and many others) intentionally don't spend time setting indents because they don't make assumptions about how the code should be indented, and they don't have code to take a whole range of things into account when position a brace (like comments, special code structures, phpcs annotations etc). They do the simplest thing they can and then let another sniff come along and fix the brace position later. This is why PHPCBF loops multiple times while fixing.

If this sniff was going to position the brace under to align with the start of the function, it would need unit tests added to ensure that the fix was working correctly, so I couldn't accept a PR without them.

I'm also curious why your code goes searching back through the token stack looking for if/elseif/else tokens to position the brace. I don't see how that code will move the brace under `public` if your example code.

chindit:
This sniff already fix the indent but you were right, checking for if/elseif token was not the right solution.

I've changed the code, which is much more cleaner.

About tests, I admin I don't know how I can test tab indent in this case :confused:

gsherwood:
I obviously misread the sniff and didn't notice it is checking indent. Sorry about that.

So it looks to me like you are just trying to add tab indent support to this sniff. If that's the case, it's not something this sniff needs to worry about. As long as you set `--tab-width` via the command line, and make sure you are including the sniff Generic.WhiteSpace.DisallowSpaceIndent then PHPCBF will convert that space indent back into a tab indent for you.

PR #3146, opened 2020-10-25
PSR-2 is officially deprecated, use PSR-12

danger89:
PSR-2 is deprecated, they advice to use PSR-12 from now onwards.


gsherwood:
Good idea to change the docs. Thanks.

PR #3150, opened 2020-10-29
PHPCS 4.x | Travis: remove composer nightly toggle

jrfnl:
PHPUnit 9.3 which came out beginning of August is compatible with PHP 8.0, so no need for the toggle anymore.


gsherwood:
Sorry, cherry-picked this commit and pushed - but this change is in now. Thanks for tracking that release and fixing this.

PR #3152, opened 2020-11-02
PHP 8.0 | File::getMethodParameters(): add support for PHP 8 constructor property promotion

### LINKED ISSUE #2968: Support PHP 8's `mixed` type

jrfnl:
PHP 8 introduces constructor property promotion.

This commit adds support for constructor property promotion to the `File::getMethodParameters()` method.

This change introduces two new keys - `property_visibility` and `visibility_token` - to the return array which will only be included if constructor property promotion is detected.

The method does not check whether property promotion is _valid_ in the function/method in which it is used. That is not the concern of this method.

Includes unit tests.

Ref: https://wiki.php.net/rfc/constructor_promotion

Implementation as discussed in #2968. Partially fixes #2968.

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -1301,10 +1301,14 @@ class File
      *        )
      * </code>
      *
-     * Parameters with default values have an additional array indexes of:
+     * Parameters with default values have additional array indexes of:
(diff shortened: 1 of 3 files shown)


gsherwood:

PR #3155, opened 2020-11-03
PHPCS 4.x | PHP 8.0: backfill the PHP8 namespaced name tokenization

### LINKED ISSUE #3041: PHP 8.0 | Proposal for handling "namespaced names as single token"

jrfnl:
Now PR #3032 and PR #3152 have been merged, I'm pulling the PR to backfill the PHP 8 tokenization of "namespaced names as single token" in PHPCS 4.x.

Unfortunately I have to pull all the changes in one go as otherwise we wouldn't be able to get a passing build.

@gsherwood It would be greatly appreciated if this PR could get some priority. As it touches so many files, the chances of it getting into a conflict state if left open for a while, are pretty large.

Fixes #3041

### PHPCS 4.x / PHP 8.0: namespaced names as single token

As per the proposal in #3041.

This effectively backfills the new PHP 8.0 tokenization of identifier names for all supported PHP versions in PHPCS 4.x.

Note: this backfills takes into account that reserved keywords are now allowed to be used in namespaced names.
For that reason no check is done of the token _type_ of the token between namespace separators, only on the contents.

Includes extensive unit tests to ensure the correct re-tokenization as well as that the rest of the tokenization is not adversely affected by this change.

Includes making sure that the following (re-)tokenizations are still stable:
* Nullable vs ternary.
* Function names are `T_STRING` - but not necessarily so when in a group use statement, i.e. `use Package\{ function Level\functionName };.
* Union types.

### Tokenizer/PHP: update the arrow function backfill to allow for PHP 8 identifier tokens

Includes updating the unit tests for the new reality and adding some additional tests.

### Tokenizer/PHP: update the short array tokenization to allow for PHP 8 identifier tokens

Includes adding some additional tests.

### File::getMethodParameters(): fix method to work with the PHP 8 identifier tokens

Includes adding some additional tests for PHP 8 identifier tokens previously not covered by tests.

### File::getMethodProperties(): fix method to work with the PHP 8 identifier tokens

Includes adding some additional tests for PHP 8 identifier tokens previously not covered by tests.

### File::getMemberProperties(): fix method to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

### File::isReference(): fix method to work with the PHP 8 identifier tokens

Includes adding some additional tests for PHP 8 identifier tokens previously not covered by tests.

### File::findEndOfStatement(): update expectations in test

... to allow for the different tokenization now expected.

### File::findExtendedClassName(): fix method to work with the PHP 8 identifier tokens

Includes adding some additional tests for PHP 8 identifier tokens previously not covered by tests.

### File::findImplementedInterfaceNames(): fix method to work with the PHP 8 identifier tokens

Includes adding some additional tests for PHP 8 identifier tokens previously not covered by tests.

### Generic/DuplicateClassName: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

Includes minor additional fix to make the sniff more tolerant of comments.

### Generic/ForbiddenFunctions: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain a test covering this change.

Includes minor additional fix to make the sniff more tolerant of comments.

### Generic/LanguageConstructSpacing: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain a test covering this change.

### PSR2/ClassDeclaration: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

By extension, this also fixes the `PSR12.Classes.AnonClassDeclaration` and the `Squiz.Classes.ClassDeclaration` sniffs.

### PSR12/ClassInstantiation: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

### PSR12/ImportStatement: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

### PSR12/NullableTypeDeclaration: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

### PSR12/CompoundNamespaceDepth: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

The tests also include a test which would be considered a parse error in PHP 8.0. The current fix will handle this the same in both PHP < 8 as well as PHP 8.0+ and will throw the expected errors even so.

The parse error in PHP 8 has been annotated in the test case file.

### Squiz/LowercasePHPFunctions: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

### Squiz/SelfMemberReference: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

The sniff has slightly different behaviour PHPCS 4.x vs the previous behaviour in PHPCS 3.x.

Notable differences:
* The tolerance for inline names interlaced with whitespace and comments (parse error in PHP 8) has been removed.
    Previously, the sniff would still fix such a name if there was a match. As of PHPCS 4.x it no longer will and won't throw an error for this anymore either.
* The `getDeclarationNameWithNamespace()` method has been removed and the `getNamespaceOfScope()` method has been renamed to `getNamespaceName()`.
    Sniffs which extend this sniff and overload(ed) either method, will need to be aware of this.

Includes:
* Fixing an incorrect unit test.

### Squiz/FunctionCommentThrowTag: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

Includes minor additional fix to make the sniff more tolerant of comments.

### Squiz/VariableComment: fix sniff to work with the PHP 8 identifier tokens

Includes adding unit tests covering the change.

### Squiz/OperatorBracket: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain a test covering this change.

### Squiz/DisallowMultipleAssignments: fix sniff to work with the PHP 8 identifier tokens

The existing unit tests already contain tests covering this change.

### Generic/ArbitraryParenthesesSpacing: add tests with the PHP 8 identifier tokens

... to make sure that the spacing for the parentheses of function calls are ignored by this sniff.

### PEAR/FunctionCallSignature: add tests with the PHP 8 identifier tokens

... to make sure that function calls using the new tokens are handled correctly by the sniff.

### Squiz/ComparisonOperatorUsageUnitTest: adjust tests to use the PHP 8 identifier tokens

... to make sure that the sniff correctly skips to the end of the function call for function calls using the new tokens.


gsherwood:
tenor

Not sure "thanks for the PR" is going to cut it for this one... but thanks!

Note that I haven't updated the changelog yet as I don't have time, but have merged this to ensure it doesn't conflict.

jrfnl:
@gsherwood Thanks for merging this this quickly. One more down, only three more to go.... (`match`, named params and attributes).

PR #3156, opened 2020-11-03
PHPCS 4.x | .gitattributes: minor tweak

jrfnl:
Commit 5c40919e9cbc32db69e7e31843c2d19a038acd65 removed a little too much when ported to the 4.x branch, as there is still a committed `testingConfig.ini` file which is used by the test runs and is not needed for end-users.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -14,6 +14,7 @@
 phpcs.xml.dist export-ignore
 phpstan.neon export-ignore
 phpunit.xml.dist export-ignore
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for finding this. That file looks to have never been excluded in the 4.0 branch - I must have forgot when the test changes went in.

jrfnl:
Not an issue though, it's only a `dev` branch for now after all.

PR #3170, opened 2020-11-21
Squiz.WhiteSpace.OperatorSpacing false positive when using negation with string concat

simPod:
~I don't know how to fix but OperatorSpacing sniff wrongly requires space when negating var/constant in string concatenation~

Ok, I don't understand how this test was meant. My point I was trying to make is that

```php
private const A = -PHP_INT_MAX; // is valid
private const B = 'a ' . -PHP_INT_MAX; // is not valid
private const C = 'a ' . - PHP_INT_MAX; // is valid
```

I cannot negate a constant when used within string concatenation without having a space after negation.

### DIFF
--- a/src/Standards/PSR12/Tests/Operators/OperatorSpacingUnitTest.inc
+++ b/src/Standards/PSR12/Tests/Operators/OperatorSpacingUnitTest.inc
@@ -49,8 +49,10 @@ try {
 
 if (strpos($tokenContent, 'b"') === 0 && substr($tokenContent, -1) === '"') {}
 
-$oldConstructorPos = +1;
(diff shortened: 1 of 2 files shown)


gsherwood:
The issues exists in the Squiz sniffs, but it also impacts the PSR12 sniff as it extends that Squiz one.

gsherwood:
Thanks a lot of this bug report. I didn't end up merging you PR as I put tests into the main Squiz sniff where the fix needed to go, but the code samples were clear and helpful. This fix will be in the 3.6.0 release.

simPod:
Thank you, appreciated

PR #3172, opened 2020-11-23
PHP 8.0 | Tokenizer/PHP: support dereferencing of text strings with interpolated variables

jrfnl:
As of PHP 8, interpolated text strings can be dereferenced, however, the square brackets are incorrectly tokenized as _short array_ brackets instead of as "normal" square brackets.

Fixed by adding the `T_DOUBLE_QUOTED_STRING` token to the allowed tokens for leaving the brackets alone in the PHP Tokenizer class.

Includes unit test.

Ref: https://wiki.php.net/rfc/variable_syntax_tweaks#interpolated_and_non-interpolated_strings


gsherwood:
Thanks! I wasn't aware of this change.

jrfnl:
One of those which easily slips through the cracks, only just noticed it the other day myself.

Issue #3173, opened 2020-11-25
empty cache file === PHP Notice


voku:
**Describe the bug**

If the cache file (--cache=) is empty, I see this on the first run. 

```
PHP Notice:  Trying to access array offset on value of type null in phar:///home/lmoelleken/meerx/Framework/thirdparty/phpcs.phar/src/Util/Cache.php on line 2
PHP Stack trace:
PHP   1. {main}() /home/lmoelleken/meerx/Framework/thirdparty/phpcs.phar:0
PHP   2. PHP_CodeSniffer\Runner->runPHPCS() /home/lmoelleken/meerx/Framework/thirdparty/phpcs.phar:6
PHP   3. PHP_CodeSniffer\Runner->run() phar:///home/lmoelleken/meerx/Framework/thirdparty/phpcs.phar/src/Runner.php:2
PHP   4. PHP_CodeSniffer\Util\Cache::load() phar:///home/lmoelleken/meerx/Framework/thirdparty/phpcs.phar/src/Runner.php:2
```

```
        => Using cache file: /home/lmoelleken/meerx/App/tmp/cache/phpcs_6fc27bc317b46a21ef95763734f89a80
        * cache was invalid and has been cleared *
DONE (0 files in cache)
```

---

https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Util/Cache.php
```php
        if (file_exists(self::$path) === true) {
            self::$cache = json_decode(file_get_contents(self::$path), true); // NULL for empty files // https://3v4l.org/ADgWv

            // Verify the contents of the cache file.
            if (self::$cache['config'] !== $configData) { // <- we can get a PHP Notice here because of $cache['config'] === NULL
                self::$cache = [];
                if (PHP_CODESNIFFER_VERBOSITY > 1) {
                    echo "\t* cache was invalid and has been cleared *".PHP_EOL;
                }
            }
        } else if (PHP_CODESNIFFER_VERBOSITY > 1) {
            echo "\t* cache file does not exist *".PHP_EOL;
        }

        self::$cache['config'] = $configData;
```

**Versions (please complete the following information):**
 - OS: Ubuntu 20.04
 - PHP: 7.4
 - PHPCS: 3.5.8

jrfnl:
@voku Could you provide steps to reproduce this issue ?

I've not been able to reproduce this unless I manually cleared out the content of a cache file (or replaced it with nonsense), but left the file in place.

Things I've tried:
[...]

PR #3174, opened 2020-11-25
Cache: Fix PHP Notice

voku:
try to fix this issue: https://github.com/squizlabs/PHP_CodeSniffer/issues/3173


jrfnl:
Without a reproduction scenario which leads to the error this is supposed to fix, I'm not sure this should be merged.

PR #2991, opened 2020-06-19
remove (from publishing) files that are not needed for end user

vladyslavstartsev:

Issue #1908, opened 2018-02-22
Remove tests from the composer package


neomerx:
Firstly, thank you for this tool.

Could you please exclude test classes/methods/ etc from the release and use dedicated namespaces instead of the global one?

IMHO the nastiest are [these ones](https://github.com/squizlabs/PHP_CodeSniffer/blob/5f2cdef533ce14dc2d74252b3bbadac05076c1c9/src/Standards/Squiz/Tests/Functions/FunctionDeclarationUnitTest.inc#L70-L74) as they constantly show up in IDE intellisense.

jrfnl:
See #548

neomerx:
@Craige 

> the tests/ directory is very valuable

without explaining why however later he mentioned he needed them because he was developing some coding standard

> I'm really only concerning myself with development of this standard as this use-case.

As far as I understand, having those tests included serves uncommon use case and punishes all regular users who are OK with included standards.

Also, the necessity of having those test do not justify polluting the global namespace.

gsherwood:
Every once and a while, someone asks for the tests directory to be excluded. If I do that, someone asks for the tests directory to be included. I can't win.

So I'm leaving the tests directory in because it reflects the full code base of PHPCS. If there comes a time (or someone knows how right now) when composer can be told to include the tests directory, then I'll exclude it by default and sniff developers can selectively include it.

I'll leave this open as a request to add namespace lines to all test files, and adjust all error message lines accordingly. But I have to admit that this is a low priority for me given the mountain of other requests I have, so I'd be relying on someone else to submit a PR for it.

jrfnl:
> If there comes a time (or someone knows how right now) when composer can be told to include the tests directory, then I'll exclude it by default and sniff developers can selectively include it.

Luckily, that is already available ;-)

The default for Composer is `--prefer-dist` and if a `.gitattributes` file would be added with `export-ignore`s for the test directory and other dev related files, those wouldn't be included anymore by default.

For people who develop for PHPCS, all they'd need to do is run Composer with `--prefer-source` and they'd get the complete package, including tests etc.

gsherwood:
> For people who develop for PHPCS, all they'd need to do is run Composer with --prefer-source and they'd get the complete package, including tests etc.

I know about that option, but I was told previously that this is not how people wanted to install PHPCS, even with the tests dir. I think it was something about caching. I'd be happy to create two version of PHPCS (with and without tests) if there was some way of installing the one you want.

[...]

Issue #548, opened 2015-04-14
Can we get the tests/ directory in the zip download


Craige:
When downloading the zip from github (as Composer will do by default), the tests/ directory is absent. I assume this has to do with the build prepared for download. Unfortunately, when using php_codesniffer as a dependency in composer.json (such as when writing an internal standard with new sniffs), the tests/ directory is very valuable, if only for the `AllTests.php` and `TestSuites.php`

gsherwood:
I added this to the export-ignore list because people specifically requested that I exclude the tests. I can't satisfy both requests. Is there any way you can use a clone, or tell composer to use a clone during development?

Craige:
I can specify the `--prefer-source flag` when running composer install, but that will install all dependencies as a full clone (no zips). Furthermore, a full clone of the phpcs repository seems to take a long time for me. Long enough that it trips the 300 timeout:

```
     [exec]   - Installing squizlabs/php_codesniffer (2.3.0)
     [exec]     Cloning 5046b0e01c416fc2b06df961d0673c85bcdc896c
     [exec]     Failed to download squizlabs/php_codesniffer from source: The process "git clone --no-checkout 'git://github.com/squizlabs/PHP_CodeSniffer.git' '/Users/craigeleeder/Source/Symfony2-coding-standard/vendor/squizlabs/php_codesniffer' && cd '/Users/craigeleeder/Source/Symfony2-coding-standard/vendor/squizlabs/php_codesniffer' && git remote add composer 'git://github.com/squizlabs/PHP_CodeSniffer.git' && git fetch composer" exceeded the timeout of 300 seconds.
```

Konafets:
You could maintain two installations of PHPCS. 

1. The first one is a clone of PHPCS with your custom standard inside. 
2. The 2nd one is a composer installation. 

As long vendor/bin/ is in your $PATH you will get the right installation of PHPCS for production use. For testing purpose just use the binary from the clone.

You can go one step further and symlink your custom standard into the clone of PHPCS. 
With this setup you are able to execute your custom sniffs within both installations and you are able to make changes on the custom standard which are covered by tests and they are immediately available in the production installation.

Craige:
@Konafets You may be misunderstanding my requirements.

Right now, I have php_codesniffer and phpunit as dependencies in the repository for this standard. Upon installation, I run:

    $ ./bin/phpcs --config-set installed_paths `pwd`/src # My standard is located under ./src

[...]

PR #3176, opened 2020-11-27
Add tests and .github to export-ignore

fezfez:


jrfnl:
Related to #548, #1908, #2991 - the change for `tests` will not be accepted in the 3.x branch. The tests will be `export-ignore`d in the 4.x branch.

fezfez:
I understand the version policy but my intention is not to fix code but save a lot of CO2.
the 3.x branch will be downloaded million time even if 4.x is release because a lot of project will be stuck with it and a lot of CO2 will be burn for nothing more than a version policy.....

gsherwood:
Thanks, but I absolutely cannot make this change in version 3. I've already committed to this.

PR #3177, opened 2020-11-27
Incorrect tokenization of GOTO statements in mixed PHP/HTML files

jrfnl:
The logic could get confused when `goto` would be used in a mixed PHP/HTML file.

Includes adding a limited set of unit tests for the `T_GOTO_LABEL` tokenization.


gsherwood:
Thanks for this fixed. I got bitten by the new `T_NAME_FULLY_QUALIFIED` when bringing it into the 4.0 branch :)

PR #3178, opened 2020-11-28
PHP 8.0 | Add support for named function call arguments

### LINKED ISSUE #3159: PHP 8.0 | Add support for named function call parameters
## Introduction

PHP 8.0 introduces named function call parameters:
```php
array_fill(start_index: 0, num: 100, value: 50);

// Using reserved keywords as names is allowed.
array_foobar(array: $array, switch: $switch, class: $class);
```

Ref: https://wiki.php.net/rfc/named_params

## Current tokenization

PHP 8.0 natively tokenizes the parameter label as `T_STRING`. This is the same in PHP 7.4.
When a reserved keyword is used as a parameter label, the tokenizer will tokenize it as the token for the reserved keyword.

When run through the PHPCS tokenizer, the parameter label, in most cases, will currently be incorrectly retokenized as `T_GOTO_LABEL`.
The reserved keyword tokenization as the reserved keyword token remains the same in most cases, with a few exceptions where the token is retokenized as `T_STRING`.

The `:` will currently tokenize to `T_COLON` or `T_INLINE_ELSE` in PHPCS, depending on the wider context of the code sample.

## Proposal for handling this in PHPCS

I'm currently looking into adding support for named function call parameters to PHPCS.

Based on my initial findings, I'd like to propose the following:
* Introduce a new PHPCS native token constant `T_PARAM_NAME` (or `T_PARAM_NAME_LABEL` or `T_PARAM_LABEL`) and consistently tokenize the parameter label as such.
* Ensure that the `:` following a parameter label is always tokenized as `T_COLON`.

Reasoning:
* The new `T_PARAM_NAME` token will make it a lot easier to prevent confusion in the ternary tokenizer and can prevent the colon from being misidenitified as `T_INLINE_ELSE`.
* `T_STRING` is already used for a lot of things and often enough it is hard to determine what a `T_STRING` signifies.
    Without the new token, any sniff looking for `T_STRING` would need to be checked and possibly adjusted to take parameter name labels into account when determining if this is a `T_STRING` the sniff should handle.
    With the new token, existing sniffs looking for `T_STRING` would not need any adjustments and would continue to work as they are.
* For sniffs looking for the `T_COLON` token, with the new `T_PARAM_NAME` token, it will be easy to determine the context of the `T_COLON`.

@gsherwood What's your opinion on this proposal ?

## Affected sniffs

I haven't made a full inventory yet of sniffs which will be affected.

At the very least, I imagine any and all sniffs which examine and fix "function call argument spacing" will need to be reviewed and adjusted to allow for parameter name labels.

jrfnl:
PHP 8.0 introduces named function call parameters:
```php
array_fill(start_index: 0, num: 100, value: 50);

// Using reserved keywords as names is allowed.
array_foobar(array: $array, switch: $switch, class: $class);
```

Ref: https://wiki.php.net/rfc/named_params

This PR adds support to PHPCS for named function call arguments by adding a special custom token `T_PARAM_NAME` and tokenizing the _labels_ in function calls using named arguments to that new token, as per the proposal in #3159.

I also ensured that the colon _after_ a parameter label is always tokenized as `T_COLON`.

Includes some minor efficiency fixes to the code which deals with the colon vs inline else determination as there is no need to run the "is this a return type" or the "is this a `case` statement" checks if it has already been established that the colon is a colon and not an inline else.

Includes a ridiculous amount of unit tests to safeguard the correct tokenization of both the parameter label as well as the colon after it (and potential inline else colons in the same statement).
Please also see my comment about this here: https://github.com/squizlabs/PHP_CodeSniffer/issues/3159#issuecomment-735247066

**Note**: The only code samples I could come up with which would result in "incorrect" tokenization to `T_PARAM_NAME` are all either parse errors or compile errors. I've elected to let those tokenize as `T_PARAM_NAME` anyway as:
1. When there is a parse error/compile error, there will be more tokenizer issues anyway, so working around those cases seems redundant.
2. The code will at least tokenize consistently (the same) across PHP versions. (which wasn't the case for parse errors/compile errors with numeric literals or arrow functions, which is why they needed additional safeguards previously).

Fixes #3159

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -894,6 +894,62 @@ class PHP extends Tokenizer
             }//end if
 
             /*
+                Tokenize the parameter labels for PHP 8.0 named parameters as a special T_PARAM_NAME
+                token and ensure that the colon after it is always T_COLON.
+            */
+
+            if ($tokenIsArray === true
+                && preg_match('`^[a-zA-Z_\x80-\xff]`', $token[1]) === 1
+            ) {
+                // Get the next non-empty token.
+                for ($i = ($stackPtr + 1); $i < $numTokens; $i++) {
+                    if (is_array($tokens[$i]) === false
+                        || isset(Util\Tokens::$emptyTokens[$tokens[$i][0]]) === false
+                    ) {
+                        break;
+                    }
+                }
+
+                if (isset($tokens[$i]) === true
+                    && is_array($tokens[$i]) === false
+                    && $tokens[$i] === ':'
+                ) {
+                    // Get the previous non-empty token.
+                    for ($j = ($stackPtr - 1); $j > 0; $j--) {
+                        if (is_array($tokens[$j]) === false
+                            || isset(Util\Tokens::$emptyTokens[$tokens[$j][0]]) === false
+                        ) {
+                            break;
+                        }
+                    }
+
+                    if (is_array($tokens[$j]) === false
+                        && ($tokens[$j] === '('
+                        || $tokens[$j] === ',')
+                    ) {
+                        $newToken            = [];
+                        $newToken['code']    = T_PARAM_NAME;
+                        $newToken['type']    = 'T_PARAM_NAME';
+                        $newToken['content'] = $token[1];
+                        $finalTokens[$newStackPtr] = $newToken;
+
+                        $newStackPtr++;
+
+                        // Modify the original token stack so that future checks, like
+                        // determining T_COLON vs T_INLINE_ELSE can handle this correctly.
+                        $tokens[$stackPtr][0] = T_PARAM_NAME;
+
+                        if (PHP_CODESNIFFER_VERBOSITY > 1) {
+                            $type = Util\Tokens::tokenName($token[0]);
+                            echo "\t\t* token $stackPtr changed from $type to T_PARAM_NAME".PHP_EOL;
+                        }
+
+                        continue;
+                    }
+                }//end if
+            }//end if
+
+            /*
                 Before PHP 7.0, the "yield from" was tokenized as
                 T_YIELD, T_WHITESPACE and T_STRING. So look for
                 and change this token in earlier versions.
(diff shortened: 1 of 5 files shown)


src/Tokenizers/PHP.php
```diff
@@ -893,6 +893,62 @@ protected function tokenize($string)
                 continue;
             }//end if
 
+            /*
+                Tokenize the parameter labels for PHP 8.0 named parameters as a special T_PARAM_NAME
+                token and ensure that the colon after it is always T_COLON.
+            */
+
+            if ($tokenIsArray === true
+                && preg_match('`^[a-zA-Z_\x80-\xff]`', $token[1]) === 1
```
gsherwood, line 902: I'm a bit worried about running a preg_match on so many tokens, and then the following loop to find the next non-empty looking for a colon.

I probably would have tried tackling this problem starting with the colon and looking backwards, although I know that's going to be complex due to the different ways it can be used. I possibly would have also tried in processAdditional, but that might have been even more complex to unravel.

I'm curious to know if you tried any alternative implementations before I go and play around with the code to see how it's working.

jrfnl: > I'm a bit worried about running a preg_match on so many token

Well, as regexes go, this is as fast you can get: a regex checking for one character on a string of one character, so no chance of backtracing at all, it either matches or it doesn't.
The alternative would be to have a list of the characters we want to exclude and do either a `strpos()` or `isset()` or something, but then that would break straight away if a new special character would get meaning in PHP. It would definitely be less stable.

> I probably would have tried tackling this problem starting with the colon and looking backward

That was my first choice for token to search for, but then the "final token" for the _previous_ effective token (which is the one we need to chance) may already have been set in a previous loop, and what with whitespace and comments allowed everywhere, that seemed like it was going to be pretty complicated.

> I possibly would have also tried in processAdditional, but that might have been even more complex to unravel.

I consider that as well, but that would have been too late to prevent misidentified ternary "else" tokens and trying to recreate the whole if/else determination in `processAdditional()` after the fact, seemed like it would cause a huge overhead, aside from it probably not being stable.

jrfnl: Just thinking - I could possibly make the regex condition a "T_STRING _or_ matching the regex" condition, that way the most common token for identifiers would not be run through the regex, which would possibly be slightly faster.

jrfnl: And yes, I thought long and hard before deciding on this way to do it, considered sniffing based on the `(` or `,` before it too. Every time I tried, there was something which would make it less stable or more complex to set the final token, which is why I ended up with this.

gsherwood: I assumed you'd thought through those options, but just wanted to check before I dug into it. Bypassing the regex for T_STRING tokens is a good idea - I might give it a go. The *extensive* unit tests make this very easy to play around with, so thank you so much for coming up with all those test cases.

gsherwood:
Thanks so much for this. I did end up putting in your suggestion for checking for T_STRING before doing the regex as I could see a minor performance improvement overall in my test runs.

PR #3179, opened 2020-12-04
Travis: add build against PHP 8.0

jrfnl:
PHP 8.0 has been branched off two months ago, so `nightly` is now PHP 8.1 and in the mean time PHP 8.0 was released last week.

As of today, there is a PHP 8.0 image available on Travis.

This PR adds a new build against PHP 8.0 to the matrix and, as PHP 8.0 has been released, that build is not allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ jobs:
     - php: 7.2
     - php: 7.3
     - php: 7.4
-    # Nightly is PHP 8.0 since Feb 2019.
+    - php: 8.0
+    # Nightly is PHP 8.1 since Oct 2020.
     - php: nightly
       addons:
         apt:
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks a lot for this. Travis (again) has a massive backlog of jobs, so going to be a while before I can see if this has processed correctly.

jrfnl:
I'll be looking into moving to GH Actions for some other repos soon due to all the problems with Travis (.org being excessively slow and being shut down soon anyway and .com not giving out anywhere near the open source tokens needed - they are basically stopping open source support, just not being honest about it at the moment).
If you like, I can look at moving this script over as well when I get to it. Compared to the other scripts I need to migrate, this one seems simple by comparison.

PR #3184, opened 2020-12-09
PSR2.Namespace.NamespaceDeclaration false positive on namespace operator

jrfnl:
When the `namespace` keyword was used as an operator, the sniff would incorrectly throw an error too.

Fixed now.

Includes unit test.


gsherwood:
Thanks for fixing this

Issue #2800, opened 2020-01-06
False positive in SwitchDeclaration TerminatingComment with all branches covered


simPod:
```php
switch ($x) {
    case T_CONST:
        switch ($this->getVisibilityForToken($file, $pointer)) {
            case T_PUBLIC:
                return self::STAGE_PUBLIC_CONSTANTS;
            case T_PROTECTED:
                return self::STAGE_PROTECTED_CONSTANTS;
            default:
                return self::STAGE_PRIVATE_CONSTANTS;
        }
}
```

As you can see, there's no way the code would continue executing after running inner `switch`. However, this error is reported 

> There must be a comment when fall-through is intentional in a non-empty case body

on line `    case T_CONST:`.

Changing the code this way satisfies the sniff:

```diff
            case T_CONST:
                switch ($this->getVisibilityForToken($file, $pointer)) {
                    case T_PUBLIC:
                        return self::STAGE_PUBLIC_CONSTANTS;
                    case T_PROTECTED:
                        return self::STAGE_PROTECTED_CONSTANTS;
                    default:
                        return self::STAGE_PRIVATE_CONSTANTS;
                }
+
+                // no way to get here
```

gsherwood:
The sniff doesn't attempt to figure out if every possible execution path would lead to an exit condition, which is why you get this error. The intention of the sniff is to always provide the comment so that it is clear what is happening in the code without a developer also having to follow all the code paths.

I'm going to close this as it's not a use case that this sniff has been designed for and the effort to attempt to support it is too high for me to take on.

gsherwood:
PR #3186 has now implemented this request. It will be released in version 3.6.0.

PR #3186, opened 2020-12-15
[SwitchDeclarationSniff] Handle nested switch

### LINKED ISSUE #3192: findStartOfStatement doesn't work correctly inside switch
**Code sample**
```php
switch ($foo) {
    case 1:
        return 2;
    case 2:
        return 4;
}
```

**Describe the bug**
When working on https://github.com/squizlabs/PHP_CodeSniffer/pull/3186, I got a weird behavior with the scope opener/closer of switch/case. In the code sample, when I dump the tokens, `:` is the scope opener of the `scope_condition` `case`, and `return` is the scope closer.

Seems like it's because of this: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Tokenizers/PHP.php#L220

But then, when using the `findStartOfStatement` with the index of the token `4`, it will return the token `4` instead of the `return` one, because of this condition: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Files/File.php#L2311-L2319
Return is a scope closer here so it return the value before `return` (or after, since we're looking backward), which is `4`.

So I think we should ignore the scope closer of the `switch/case`, the right fix would be this: https://github.com/squizlabs/PHP_CodeSniffer/pull/3186/commits/e790378dfd15b0dcbb21eefc8619ac33218b28d3

VincentLanglet:
This fix the issue reported here: https://github.com/squizlabs/PHP_CodeSniffer/issues/2800
I added tests.

Edit: Also close #3192

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -2313,6 +2313,11 @@ class File
                 && $this->tokens[$i]['code'] !== T_CLOSE_PARENTHESIS
                 && $this->tokens[$i]['code'] !== T_END_NOWDOC
                 && $this->tokens[$i]['code'] !== T_END_HEREDOC
+                && $this->tokens[$i]['code'] !== T_BREAK
+                && $this->tokens[$i]['code'] !== T_RETURN
+                && $this->tokens[$i]['code'] !== T_CONTINUE
+                && $this->tokens[$i]['code'] !== T_THROW
+                && $this->tokens[$i]['code'] !== T_EXIT
             ) {
                 // Found the end of the previous scope block.
                 return $lastNotEmpty;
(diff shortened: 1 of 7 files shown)


VincentLanglet:
Friendly ping @jrfnl @gsherwood.

This PR is working for
```
switch ($foo) {
    case 1:
        switch ($bar) {
            case 1:
                return;
            default:
                return;
        }
    case 2:
        return 2;
}
```
But not for
```
switch ($foo) {
    case 1:
        switch ($bar) {
            case 1:
                return;
            default:
                throw new \Exception();
        }
    case 2:
        return 2;
}
```

After more investigation, it seems that the `T_COLON` of `default` has a `scope_opener` value and the `scope_closer` is the index of the `throw` token. So `findStartOfStatement` is returning `new` because it's the first token after a `scope_closer`.

Seems like the same happens with 
```
switch ($foo) {
    case 1:
        switch ($bar) {
            case 1:
                return;
            default:
                return 3;
        }
    case 2:
        return 2;
}
```
Here `findStartOfStatement` is returning `3` because the `T_RETURN` token is the scope closer of the `T_COLON` of default.

So I made a minimal reproducer:
```
switch ($foo) {
    case 1:
        return 2;
}
```
When using `findStartOfStatement` on the token `2`, it return the index of the token `2` instead of the index of the token `return`.

I made an issue to fix to relate this: https://github.com/squizlabs/PHP_CodeSniffer/issues/3192

gsherwood:
Thanks a lot for writing that code to check all the branches. Looks to work nicely.

Thanks for fixing that bug in findStartOfStatement() as well. I'm a bit surprised it's never been a problem before.

PR #3187, opened 2020-12-19
CI: switch from Travis to GH Actions

jrfnl:
This switches the complete Travis script out for the same but run via GH Actions.

The status won't be shown in this repo yet until this PR has been merged.
You can see the output of each of the workflows in my fork:
* Main Test workflow: https://github.com/jrfnl/PHP_CodeSniffer/actions/runs/432805649
* XML and PEAR validation: https://github.com/jrfnl/PHP_CodeSniffer/actions/runs/432805647
* PHPStan: https://github.com/jrfnl/PHP_CodeSniffer/actions/runs/432805646

To do once this PR has been merged:
- [ ] If applicable: update required statuses for PRs.
- [ ] Remove the webhook/integration/app for Travis and turn the service off.
- [ ] For those PRs which are currently unmergable due to status checks not reporting, ask the OP to rebase them.

--- 

### CI: switch to GitHub Actions - step 1: phpstan

This commit:
* Adds a GH  Actions workflow for the PHPStan check which was previously run on Travis.
*  Removes that part of the `.travis.yml` configuration.

Notes:
Builds will run on all pushes and on pull requests, with the exception  of those just modifying files which are irrelevant to this workflow.

### CI: switch to GitHub Actions - step 2: XML and PEAR validation

This commit:
* Adds a GH  Actions workflow for the build which validated the XML file against schema and checked their code style consistency, as well as validate the PEAR package file, as was previously run on Travis.
*  Removes that part of the `.travis.yml` configuration.

Notes:
Builds will run on all pushes and on pull requests, with the exception  of those just modifying files which are irrelevant to this workflow.

### CI: switch to GitHub Actions - step 3: move the PEAR package validation

The PEAR native package validation script used to be run on every PHP build, but the results should be no different on any particular PHP version.

To that end, move the running of the script from the "test" cycles, to the one-time only run PEAR specific job.

### CI: switch to GitHub Actions - step 4: tests

This commit:
*  Adds a GH Actions workflow for the CI checks which were previously run on Travis as the default test script.
    Includes only running the base tests for the builds using custom ini setting.
* Removes the, now redundant,  `.travis.yml` configuration.
* Remove the custom PHP ini configurations.
    Setting these is now handled in the script itself.
* Updates the `.gitattributes` file.
* Updates the  "Build Status" badge in the Readme to use the results from the GH `Test`  Actions runs.


jrfnl:
**Open question:** as the phar is created during the test run, I could adjust the script to automatically upload it as an artifact result of the CI run. That way a phar for each PR/push would automatically be available for download.

I'd presume using the phar created during the PHP 7.4 run would suffice.
Default retention period is 90 days, but that can be changed.

gsherwood:
This is terrific. Thank you so much!

There are a couple of 8.1 issues I need to look into for master and 4.0, but everything else is working really well.

I'm not too fussed about keeping the phars, but happy to do so if someone else would find them useful.

PR #3204, opened 2021-01-30
Ignore files when exporting package

simonschaufi:
This commit is part of a campaign to reduce the amount of data transferred to save global bandwidth and reduce the amount of CO2. See https://github.com/Codeception/Codeception/pull/5527 for more info.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,11 +1,15 @@
+.github/ export-ignore
+scripts/ export-ignore
+src/Standards/**/Tests/ export-ignore
+tests/ export-ignore
 .cspell.json export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
+CONTRIBUTING.md export-ignore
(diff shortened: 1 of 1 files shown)


simonschaufi:
What about `CodeSniffer.conf.dist`? Can this file also get excluded?

jrfnl:
Related to #548, #1908, #2991, #3176 - the change for tests will not be accepted in the 3.x branch. The tests will be `export-ignore`d in the 4.x branch.

The update for `.github/` is justified. This is a recently new directory.

simonschaufi:
Ok I see but I really wonder why people depend on third party tests... When can I expect a 4.0 release?

gsherwood:
I've added the `.github` dir to the export ignore list. Thanks for pointing that out.

The test dirs are ignored in the 4.0 branch but I do not have a release date for that version.

PR #3213, opened 2021-02-13
JSLint: bug fix

jrfnl:
The `$name` passed to `Config::getExecutablePath()` appears to be incorrect which would result in a duplicate `jslint jslint` command.


gsherwood:
Thanks for the fixing, although I'm not too worried about this sniff not working given I've been trying to push people away from using PHPCS for JS and CSS files, but makes sense to fix it given you've done the work. I didn't include it in the changelog though as I don't want to draw attention to any of the JS or CSS sniffs.

PR #3214, opened 2021-02-13
Windows: fix escaping of external commands

jrfnl:
The `escapeshellcmd()` function apparently does not escape spaces within a path on Windows which can result in broken functionality.

While the sniffs and report affected by this are apparently not used that much based on the lack of bug reports, fixing it still seemed like the _right thing to do_.

Noticed while running the unit tests on a fresh install on Windows 10.

At some point over the past years, Node has apparently changed their default install directory on Windows and the order in which they register their paths to the Windows system `PATH`.
This means that `where csslint` may result in a `$cmd` path like `C:\Program Files\nodejs\csslint.cmd`, which would be escaped to `C:^\Program Files^\nodejs^\csslint.cmd` on Windows, which in turn results in a `'C:\Program' is not recognized as an internal or external command, operable program or batch file.` error.

I could have changed the install path for NVM on my machine, but that would just have hidden the underlying issue.

It does appear to be a known issue with the function based on the last two comments in this upstream bug report: https://bugs.php.net/bug.php?id=43261, however as that issue is closed, I don't expect this to be fixed in PHP itself, though it might be worth it to open a new issue upstream about it (as those two comments were left on a closed issue years after the close).

Fixed now by checking an escaped path for unescaped spaces when on Windows and if necessary, escaping them.
The escaping is done in such a way that, even if PHP itself would start escaping these spaces, the `Common::escapeshellcmd()` function will still handle this correctly.

### DIFF
--- a/src/Reports/Notifysend.php
+++ b/src/Reports/Notifysend.php
@@ -18,6 +18,7 @@ namespace PHP_CodeSniffer\Reports;
 
 use PHP_CodeSniffer\Config;
 use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Util\Common;
 
 class Notifysend implements Report
 {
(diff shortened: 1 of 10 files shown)


gsherwood:
There are still 3 of these paths in version 4, so very much worth doing. Thanks.

jrfnl:
Yvw

PR #3215, opened 2021-02-13
Arrow function tests: simplify and improve the tests

jrfnl:
### Arrow function tests: remove lots of code duplication

* Simplify the arrow function testing by using a helper function to check the token positions.
* Fix incorrect parameter order of the `$expected` vs `$result` parameters in `assertSame()`.
    While when the test passes, this makes no difference, it does make a difference in the error output when a test fails.

### Arrow function tests: some simplifications in the rest of the tests

This addresses the same issues as in the previous commit for those functions which need custom "error" messages or have a shared scope closer, resulting in the `opener` for the `scope_closer` pointing to another construct

* Remove duplicate calculations of the scope opener/closer positions.
* Fix incorrect parameter order of the `$expected` vs `$result` parameters in `assertSame()`.
    While when the test passes, this makes no difference, it does make a difference in the error output when a test fails.

### Arrow function tests: minor documentation fix

### DIFF
--- a/tests/Core/Tokenizer/BackfillFnTokenTest.php
+++ b/tests/Core/Tokenizer/BackfillFnTokenTest.php
@@ -29,17 +29,7 @@ class BackfillFnTokenTest extends AbstractMethodUnitTest
         foreach (['/* testStandard */', '/* testMixedCase */'] as $comment) {
             $token = $this->getTargetToken($comment, T_FN);
             $this->backfillHelper($token);
-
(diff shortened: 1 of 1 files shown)


gsherwood:
That looks a lot cleaner :) Thanks

jrfnl:
@gsherwood We could now even iterate on it further and use a dataprovider for the majority of tests, if you like.

PR #3216, opened 2021-02-13
Generic/FunctionCallArgumentSpacing: add test with named function call args

jrfnl:
This verifies that named function call arguments do not affect the functioning of the sniff as-is.


gsherwood:
Thanks

PR #3217, opened 2021-02-13
Tokenizer/PHP: minor tweak

jrfnl:
Prevents some unnecessary token changes of the type `token # changed from T_STRING to T_STRING`.


gsherwood:
Nice catch, thanks

PR #3218, opened 2021-02-13
Tokens: fix up some PHPCS native token values

jrfnl:
The values of the PHPCS native token types are normally prefixed with `PHPCS_`. This wasn't the case for the three most recent additions, `T_FN_ARROW` as introduced in PHPCS 3.5.3, `T_TYPE_UNION` and `T_PARAM_NAME` as will be introduced in PHPCS 3.6.0 (not yet released).

While the change to the value for `T_FN_ARROW` could be considered a breaking change, it is exceedingly rare for a sniff to use the _value_ of a token constant, so IMO opinion, this is a safe change to make.

As for the other two tokens, as they have not been in a release yet, they can be safely updated no matter what.

### DIFF
--- a/src/Util/Tokens.php
+++ b/src/Util/Tokens.php
@@ -74,9 +74,9 @@ define('T_OPEN_USE_GROUP', 'PHPCS_T_OPEN_USE_GROUP');
 define('T_CLOSE_USE_GROUP', 'PHPCS_T_CLOSE_USE_GROUP');
 define('T_ZSR', 'PHPCS_T_ZSR');
 define('T_ZSR_EQUAL', 'PHPCS_T_ZSR_EQUAL');
-define('T_FN_ARROW', 'T_FN_ARROW');
-define('T_TYPE_UNION', 'T_TYPE_UNION');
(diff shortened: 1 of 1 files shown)


jrfnl:
@gsherwood I think this one should definitely go into 3.6.0. Would you mind milestoning it ?

gsherwood:
Well spotted, thanks for fixing those.

Agree that nobody should actually be looking at the value. They are only prefixed like that because PHPUnit (I think) introduced the same tokens with different values at some point in the past and they conflicted with all the existing PHPCS ones, so it was for library compatibility more than anything else.

jrfnl:
Makes sense and all the more relevant now PHPUnit has started using PHP_Parser since PHPUnit 9.3 for code coverage. I already ran into breakage in an external library (PHPCompatibility) caused by PHP_Parser backfilling tokens while the sniff was relying on detecting the combination of PHP/PHPCS by checking `defined()` for a token constant.

Reason I noticed it now was, that I was getting some weird token names (`H_DEFAULT` for something in the match WIP PR) as the `Tokens::tokenName()` method expects the `PHPCS_` prefix and removes it to get the name 😀

PR #3223, opened 2021-02-17
new argument for the command line

RezaT1994:
I've added a new argument for the command line to write the documentation into a file.

```php
bin\phpcs --standard=Generic --generator=html --output-path=.\generic.html
```


src/Standards/Generic/Docs/Arrays/ArrayIndentStandard.xml
```diff
@@ -0,0 +1,26 @@
+<documentation title="Array Indent">
```
jrfnl, line 1: This file seems out of place in this PR ?

RezaT1994: How can I add a new documentation?

jrfnl: Separate PR ?

RezaT1994: Yes

RezaT1994: I just remove the this file in this PR.
So, should I add new documentation in a new PR?

gsherwood:
I'm not sure this feature is required. It's not hard to get content into a file, and it's not something that is done often enough (like in CI) to make a dedicated CLI command worth it. Would be good to know what use case you are solving for though.

RezaT1994:
I know it's not required. because we can just copy the content from terminal. But it's hard sometimes 🙁. The easy way is that we just provide a path to save the content to a file instead of copy and paste the content from terminal.

jrfnl:
@RezaT1994 Have you tried piping the output to a file ? In my experience that saves you from having to manually copy/paste.

```
bin\phpcs --standard=Generic --generator=html > ./generic.html
```

The `>` basically says "take the output from this command and save it to a file with this name".

RezaT1994:
Well I didn't know that. Thank you.
Looks like we didn't need that argument after all 😅

Should I close the PR?

jrfnl:
> Looks like we didn't need that argument after all 😅

@RezaT1994 No worries, we all learn every day ;-)

> Should I close the PR?

You may as well. Would be nice if you would open a new PR for the docs for the ArrayIndent sniff which were originally included in this PR. 👍

RezaT1994:
Ok I will

PR #3225, opened 2021-02-20
Arrow function tests: remove some stray function calls

jrfnl:
Follow-up on #3215 which made these redundant, but didn't remove them.


gsherwood:
Thanks

PR #3250, opened 2021-02-28
PHP 8.1: fix deprecation notice

jrfnl:
The `Util\Standards::getInstalledStandardPath()` method returns either a string or `null`, however the `strpos()` method which is used in both the `Util\Common::isPharFile()` method, as well as in the follow-on condition, only accepts a `string` as `$haystack`.

As of PHP 8.1, this will generate a deprecation notice `strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated`.

Discovered while testing an external standard against PHPCS `master` on PHP 8.1.

Fixed now.

### DIFF
--- a/src/Ruleset.php
+++ b/src/Ruleset.php
@@ -726,7 +726,7 @@ class Ruleset
         } else {
             // See if this is a whole standard being referenced.
             $path = Util\Standards::getInstalledStandardPath($ref);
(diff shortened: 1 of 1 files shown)


gsherwood:
Nice find, thanks

jrfnl:
Thanks for merging this! Retriggering the build via which I found it now to see if it goes 🟢

jrfnl:
Nope, still failing, now on a similar, but different error. Follow-up PR upcoming.

PR #3251, opened 2021-02-28
GH Actions: various tweaks to the scripts

jrfnl:
### GH Actions: simplify the PHPStan workflow

No need to install via Composer, the `setup-php` action can make PHPStan globally available.

### GH Actions: improve fix to run the tests on PHP 8.1

There is no need to remove the config file and have all settings on the command-line as PHPUnit offers a `--no-configuration` option to ignore the config file.

Enabling that simplifies the action steps a little and isolates the customization for PHP 8.1 to PHP 8.1.

### GH Actions: report CS violations in the PR

The cs2pr tool will allow to display the results from an action run in checkstyle format in-line in the PR code view, which should improve usability of the workflow results.

As the CS run is run in every test run in the matrix, passing the results on to CS2PR would result in violations being displayed multiple times, so using conditions to only pass the CS violations on to the PR for one of the runs.

Ref: https://github.com/staabm/annotate-pull-request-from-checkstyle

Example of what this looks like in a PR:
image

### GH Actions: allow for manually triggering a workflow

Triggering a workflow for a branch manually is not supported by default in GH Actions, but has to be explicitly allowed.

This is useful if, for instance, an external action script or composer dependency has broken.
Once a fix is available, failing builds for open PRs can be retriggered manually instead of having to be re-pushed to retrigger the workflow.

Ref: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -9,6 +9,8 @@ on:
   pull_request:
     paths-ignore:
       - '**.md'
+  # Allow manually triggering the workflow.
+  workflow_dispatch:
 
 jobs:
   phpstan:
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -25,14 +25,13 @@ jobs:
         with:
           php-version: '7.4'
           coverage: none
-
-      - name: 'Composer: require PHPStan'
-        run: composer require --no-update --dev phpstan/phpstan
+          tools: phpstan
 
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks for cleaning that up, and for all the extras. I'm keen to see the CS errors being displayed.

jrfnl:
You're welcome. I'm learning more about GH actions every day, so who knows, might come up with some more things at some point in the future.

PR #3254, opened 2021-03-04
 PHP 8.0 | Pear+PSR2/FunctionCallSignature: support named parameters

jrfnl:
The `File::findEndOfStatement()` method regards a `T_COLON` as the end of a statement, while for function call arguments using named parameters, the colon is part of the parameter name declaration and should be disregarded when determining the end of the statement.

Fixed now.

Includes unit tests for both affected sniffs.


gsherwood:
Thanks a lot for fixing this so quickly. I didn't even get a chance to look at the report before it was fixed and verified!

jrfnl:
You're welcome. Note: there is a follow-up bug report related to this: #3255 which should probably also be fixed before releasing 3.6.0.

PR #3256, opened 2021-03-04
Normalize and fix phpdoc typehints across the code

Riimu:
While writing my own github actions reporter format (to get annotations properly in github), I noticed that my PHPStan was complaining about types in some of the PHP_CodeSniffer code. As a result I noticed that there were several issues when it came to phpdoc types in the code base.

This PR introduces bunch of normalization and fixes to the type hints in the phpdocs of various methods. I tried to normalize the types to what was already the most common format. The following changes have been applied:

* All occurrences of `boolean` have been replaced with `bool`
* All occurrences of `integer` have been replaced with `int`
* Several instances of `\PHP_CodeSniffer\File` have been fixed to the correct type `\PHP_CodeSniffer\Files\File`
* Couple instanced of `File` have been replaced with `\PHP_CodeSniffer\Files\File` since rest of the code base used FQCNs exclusively
* Couple array types were normalized to what rest of the code base uses

I have not touched any actual code. Only phpdoc comments. Additionally, this PR does not modify any files in test directories, only those that are part of the actual code.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -1540,9 +1540,9 @@ class Config
      * @param string|null $value The value to set. If null, the config
      *                           entry is deleted, reverting it to the
      *                           default value.
-     * @param boolean     $temp  Set this config data temporarily for this
-     *                           script run. This will not write the config
-     *                           data to the config file.
+     * @param bool        $temp  Set this config data temporarily for this
+     *                           script run. This will not write the
+     *                           config data to the config file.
      *
      * @return bool
      * @see    getConfigData()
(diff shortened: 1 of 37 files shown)


jrfnl (commented):
@Riimu I vaguely seem to remember this codebase preferred long form `boolean` and `integer` over short form. I believe it used to actual throw errors about it, but clearly doesn't anymore. So that may need clarification. Other than that, looking good!

Riimu:
Ah. I assumed `int` and `bool` would be preferred since those are what PHP now uses and there seemed to be more usages of `int` and `bool` than `integer` and `boolean`.

I can swap those around, if requested, or remove those changes entirely too. The real important part is the fixing of those `\PHP_CodeSniffer\File` references, since they refer to non-existing classes.

PR #3258, opened 2021-03-05
Squiz.Formatting.OperatorBracket duplicate error messages for unary minus

jrfnl:
### PHP 8.0 | Squiz/OperatorBracket: add tests with named function call parameters

... to confirm that the sniff does not break on these.

### Squiz/OperatorBracket: bug fix - improve recognition of unary minus

The new test was giving _three_ instead of _two_ errors on line 194 for this snippet `$padding * -1 + 3`, with the errors being thrown on the `*`, `-` and the `+` operators.

The `-` operator, however, is a unary operator and should not trigger this error.

Fixed now.


gsherwood:
Thanks for fixing this bug, and for those additional test cases.

jrfnl:
You're welcome.

PR #3259, opened 2021-03-05
PHP 8.0 | Squiz/ScopeKeywordSpacing: add support for constructor property promotion

jrfnl:
### PHP 8.0 | Squiz/ScopeKeywordSpacing: add support for constructor property promotion

Prevent the sniff from ignoring the spacing after the scope keyword in case of PHP 8.0 constructor property promotion.

As it was, the sniff would presume "normal" property syntax, which meant that in the case of constructor property promotion in a multi-line constructor declaration, the sniff would look for a semicolon to end the statement and bow out when the semicolon wasn't found.

By explicitly checking for constructor property promotion and skipping the above mentioned check in that case, we ensure that scope keywords in constructors are still handled correctly.

Includes tests.

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/ScopeKeywordSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/ScopeKeywordSpacingSniff.php
@@ -84,7 +84,23 @@ class ScopeKeywordSpacingSniff implements Sniff
             return;
         }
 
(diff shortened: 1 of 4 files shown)


gsherwood:
Thank you! I think that would have gone completely unnoticed if not for you.

jrfnl:
@gsherwood For most new PHP syntaxes, I've done some keyword/function call searches for tokens/function calls which could indicate a sniff would need changing. This was just one of the results of those.

PR #3262, opened 2021-03-05
PHP 8.0 | Generic/ArrayIndent: add tests with named function call parameters

jrfnl:
... to confirm that the sniff does not break on these.


gsherwood:
Thanks

PR #3268, opened 2021-03-15
PHP 8.0 | Tokenizer/PHP: bugfix for union type operator tokenization

### LINKED ISSUE #3267: PHP 8 / PSR12: Variable-length union arguments generate invalid OperatorSpacing errors
**Describe the bug**
Variable-length union arguments generate invalid "expected at least one space" error. 

**Code sample**
```php
<?php

function test(string|int ...$values)
{
}
```

**To reproduce**
2. Run `phpcs -s test.php --standard=PSR12`
```
FILE: test.php
---------------------------------------------------------------------------------------------------------------
---------------------------------------------------------------------------------------------------------------
 3 | ERROR | [x] Expected at least 1 space before "|"; 0 found
   |       |     (PSR12.Operators.OperatorSpacing.NoSpaceBefore)
 3 | ERROR | [x] Expected at least 1 space after "|"; 0 found
---------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
---------------------------------------------------------------------------------------------------------------

Time: 41ms; Memory: 6MB
```

**Expected behavior**
The code is correct as-is and should not be flagged as an error to be fixed.

 - OS: Linux Ubuntu 18.04.1
 - PHP: 8.0.3
 - PHPCS: dev-master 46715ed9eb3842096581a9d88c6b766696951b19
 - Standard: PSR12

**Additional context**
Removing the `...` makes the error go away. Running `phpcbf` changes the code to `(string | int ...$values)` and `phpcs` stops warning of an error.

jrfnl:
In `PHP::processAdditional()`, in the part which retokenizes `T_BITWISE_OR` tokens to `T_TYPE_UNION`, the code already took reference operators and spread operators between the type declaration and the variable into account, but would also increment the current position in the loop, while this is already done in the `for()`. My bad.

Fixed now.

Includes additional unit tests for all relevant functions within the chain, as well as for the sniff for which the issue was reported.

Fixes #3267

### DIFF
--- a/tests/Core/File/GetMethodParametersTest.inc
+++ b/tests/Core/File/GetMethodParametersTest.inc
@@ -45,6 +45,9 @@ function namespaceOperatorTypeHint(?namespace\Name $var1) {}
 /* testPHP8UnionTypesSimple */
 function unionTypeSimple(int|float $number, self|parent &...$obj) {}
 
+/* testPHP8UnionTypesWithSpreadOperatorAndReference */
+function globalFunctionWithSpreadAndReference(float|null &$paramA, string|int ...$paramB) {}
+
 /* testPHP8UnionTypesSimpleWithBitwiseOrInDefault */
 $fn = fn(int|float $var = CONSTANT_A | CONSTANT_B) => $var;
 
--- a/src/Standards/PSR12/Tests/Operators/OperatorSpacingUnitTest.inc
+++ b/src/Standards/PSR12/Tests/Operators/OperatorSpacingUnitTest.inc
@@ -61,3 +61,5 @@ $fn = fn(array &$one) => 1;
 $fn = fn(array & $one) => 1;
 
 $fn = static fn(DateTime $a, DateTime $b): int => -($a->getTimestamp() <=> $b->getTimestamp());
+
+function issue3267(string|int ...$values) {}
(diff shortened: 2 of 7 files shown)


tests/Core/File/GetMethodParametersTest.inc
```diff
@@ -45,6 +45,9 @@ function namespaceOperatorTypeHint(?namespace\Name $var1) {}
 /* testPHP8UnionTypesSimple */
 function unionTypeSimple(int|float $number, self|parent &...$obj) {}
 
+/* testPHP8UnionTypesWithSpreadOperatorAndReference */
+function globalFunctionWithSpreadAndReference(string|int ...$paramA, float|null &$paramB) {}
```
zimzat, line 49: This is normally invalid syntax. The reference would have to be listed first.

```php
php > function globalFunctionWithSpreadAndReference(string|int ...$paramA, float|null &$paramB) {}
PHP Fatal error:  Only the last parameter can be variadic in php shell code on line 1
```

jrfnl: @zimzat Agreed. For context: the code in test case files for PHPCS doesn't actually have to be valid code. It just has to be code which allow us to test whether PHPCS handles certain syntax combinations correctly (in this case), and as this test confirms, it now does.

jrfnl: FYI: fixed it up now either way ;-)

gsherwood:
Thanks for fixing this

jrfnl:
You're welcome.

PR #3270, opened 2021-03-16
PHP 8.0 | Generic/UnusedFunctionParameter: add support for constructor property promotion

### LINKED ISSUE #3269: PHP 8: Constructor property promotion support
I see you already have closed many PHP 8 related tickets, when can we expect `PHP 8: Constructor property promotion`?
I'm getting error  `ERROR | [ ] Unused parameter $id. 9` , for code like this:
```
abstract class AggregateId implements Equality, IntegerId
{
    public function __construct(protected int $id) {}
}
```
 - OS: MacOS 10.15
 - PHP: 8
 - PHPCS: 3.5.8

jrfnl:
When constructor property promotion is used, the properties declared in the class constructor should never be marked as unused by this sniff.

Includes tests.

Fixes #3269


gsherwood:
Thanks for this. Makes sense that the property value doesn't need to be used in the code if it's declaring a class prop.

PR #3271, opened 2021-03-16
PHP 8.0 | Generic/LowerCaseType: add tests with Constructor Property Promotion

jrfnl:
The sniff already handles this correctly. This just adds some tests to confirm it and safeguard it for the future.


gsherwood:
Thanks

PR #3272, opened 2021-03-16
PHP 8.0 | PEAR/ValidDefaultValue: add tests with constructor property promotion

jrfnl:
~~Properties declared using constructor property promotion, should be ignored for the purpose of determining whether an optional parameter is declared before a required one.~~

~~Includes tests.~~

The sniff already handles this correctly. This just adds some tests to confirm it and safeguard it for the future.

### DIFF
--- a/src/Standards/PEAR/Tests/Functions/ValidDefaultValueUnitTest.inc
+++ b/src/Standards/PEAR/Tests/Functions/ValidDefaultValueUnitTest.inc
@@ -100,5 +100,20 @@ $closure = function(array $arg2=array(), array $arg1) {}
 
 $fn = fn($a = [], $b) => $a[] = $b;
 
+class OnlyConstructorPropertyPromotion {
+    public function __construct(
+        public string $name = '',
+        protected $bar
+    ) {}
+}
+
+class ConstructorPropertyPromotionMixedWithNormalParams {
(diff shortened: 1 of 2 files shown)


gsherwood:
The following code issues a deprecation notice in PHP 8+ for me:
```php
class ConstructorPropertyPromotionValid {
    public function __construct(
        public string $name = '',
        protected $bar = [],
        mixed $requiredParam,
        ?int $optionalParam = null
    ) {}
}
```
The constructor property promotion RFC also describes how the desugaring process works, which basically removes the scope and writes out the class props as normal (minus default values). So the result appears to be the same - having a required property after optional properties makes them all required as you still need to pass all the values.

So I don't think promoted properties should be ignored by this sniff, and the above code sample should produce an error.

But I haven't used this in any real code - just test code - so please let me know if I've missed something here.

jrfnl:
@gsherwood Thanks for catching that. It's weird, I'm sure I had run tests with this before for PHPCompatibility, but clearly I'm remembering it wrong.

Let me change this PR to just adding tests instead, to confirm and safeguard that the sniff was already handling this correctly.

jrfnl:
PR has been updated to just add tests.

gsherwood:
Thanks!

PR #3279, opened 2021-03-23
PHP 8.0 | PEAR&Squiz/ScopeClosingBrace: braces for constructor property promotion

ElisDN:
Fixes error for empty `{}` body in methods with  multiline parameters:

```php
class ConstructorPropertyPromotion
{
    public function __construct(
        protected int $id
    ) {} // ERROR: Closing brace must be on a line by itself
}
```

### DIFF
--- a/src/Standards/PEAR/Sniffs/WhiteSpace/ScopeClosingBraceSniff.php
+++ b/src/Standards/PEAR/Sniffs/WhiteSpace/ScopeClosingBraceSniff.php
@@ -92,6 +92,18 @@ class ScopeClosingBraceSniff implements Sniff
             }
         }
 
+        // Allow empty braces with multiline parenthesis.
+        if ($tokens[$stackPtr]['type'] === 'T_FUNCTION') {
+            $parenthesisStart = $tokens[$stackPtr]['parenthesis_opener'];
+            $parenthesisEnd   = $tokens[$stackPtr]['parenthesis_closer'];
+            if ($tokens[$parenthesisStart]['line'] !== $tokens[$parenthesisEnd]['line']
+                && $tokens[$scopeStart]['line'] === $tokens[$scopeEnd]['line']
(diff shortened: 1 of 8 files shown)


src/Standards/PEAR/Sniffs/WhiteSpace/ScopeClosingBraceSniff.php
```diff
@@ -92,6 +92,18 @@ public function process(File $phpcsFile, $stackPtr)
             }
         }
 
+        // Allow empty braces with multiline parenthesis.
+        if ($tokens[$stackPtr]['type'] === 'T_FUNCTION') {
+            $parenthesisStart = $tokens[$stackPtr]['parenthesis_opener'];
+            $parenthesisEnd   = $tokens[$stackPtr]['parenthesis_closer'];
+            if ($tokens[$parenthesisStart]['line'] !== $tokens[$parenthesisEnd]['line']
```
ElisDN, line 99: Perhaps we can allow empty body for all methods and functions without checking of parenthesis lines diff:

```php
class ConstructorPropertyPromotion
{
    public function __construct(protected int $id) {}
}
```

gsherwood:
This PR changes the behaviour of these sniffs so they now violate their respective coding standards, so I can't accept it. The standards don't allow for any case where the positioning of the braces is on the same line and directly next to each other.

If this is something you're looking to enforce, you will likely want to write a custom sniff for it and include it with your project's own coding standard.

This could also be implemented in the existing sniffs using a sniff property, but I wouldn't want to do this without knowing if this is something that will make it into an official coding standard. Otherwise, it's just an option that needs to be supported but may not be in widespread use.

If I've misunderstood the purpose of the PR, please correct me.

ElisDN:
> If this is something you're looking to enforce, you will likely want to write a custom sniff for it and include it with your project's own coding standard.

OK. I will do it in my project. Thanks, @gsherwood!

PR #3281, opened 2021-03-25
PHP 8.0 | Tokenizer/PHP: efficiency tweaks

jrfnl:
This PR contains two separate, small efficiency tweaks for the `match` expression retokenization.

1. The tokenizer was walking backwards in the `match` backfill, while it could have just used `$finalTokens[$lastNotEmptyToken]` to get the last non-empty token.
2. There was some redundancy in the tokens which were being checked for the token "before".
    For tokens like, `T_CLASS` and `T_CONST`, the next token after the `match` keyword could never be an open parenthesis, so they would already be marked as "not match".

Includes additional tests.
Additionally, when I initially started the work on backfilling `match` expressions, the nullsafe object operator had not yet been backfilled, so there was not test covering that case. Added now just to be on the safe side.

Note: all extra tests would already pass prior to this change. These are just extra safeguards.

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -1288,33 +1288,22 @@ class PHP extends Tokenizer
                         break;
                     }
 
-                    // Next was an open parenthesis, now check what is before the match keyword.
-                    for ($y = ($stackPtr - 1); $y >= 0; $y--) {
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks

PR #3282, opened 2021-03-25
PHP 8.0 | Tokenizer/PHP: improve retokenization of T_DEFAULT

### LINKED ISSUE #3280: False positive for Generic.PHP.LowerCaseKeyword.Found when having constant with key
**Describe the bug**

**Code sample**
```php
class SomeClass {
    public const DEFAULT = 'whatever';
}
...
		$arr = [
			SomeClass::DEFAULT => 1,
		]
SomeClass::DEFAULT //trigers error PHP keywords must be lowercase; expected "default" but found "DEFAULT"
```

**To reproduce**
2. Run `phpcs test.php ...`
```
ERROR | [x] PHP keywords must be lowercase; expected "default" but found "DEFAULT"
    |       |     (Generic.PHP.LowerCaseKeyword.Found)
ERROR | [ ] No key specified for array entry; first entry specifies key
    |       |     (Squiz.Arrays.ArrayDeclaration.NoKeySpecified)

```

**Expected behavior**
No error triggered
 - OS: Ubuntu 20.4
 - PHP: [8.0]
 - PHPCS: [* dev-master, * 3.x-dev]
 - Standard: [e.g., PSR2, PSR12, Squiz]

jrfnl:
### Tokenizer/PHP: add more tests for default keyword and double arrow

In particular, this adds tests where the `default` keyword is not used as the keyword and should be tokenized as `T_STRING`.

### PHP 8.0 | Tokenizer/PHP: improve retokenization of T_DEFAULT

This commit improves the tokenization of "default" keywords, by making sure that the keyword is not used as a constant or property.
If it is, it will never be either `T_DEFAULT` or `T_MATCH_DEFAULT`, so it will be retokenized to `T_STRING` early.

Fixes #3280


jrfnl:
Note: the build failure is unrelated to this PR. I suspect it may be fixed by just rerunning the workflow. If not the XML schema location may need adjusting, but that is outside of the scope of this PR.

gsherwood:
Thanks for fixing this

jrfnl:
Yvw, just very glad this was found before the release.

PR #3293, opened 2021-04-08
Squiz.PHP.DisallowMultipleAssignmentsSniff: Fixed false positives

kukulich:


gsherwood:
Thanks for finding and fixing these issues, and for splitting them up into different commits. It made it very easy to review.

PR #3295, opened 2021-04-11
Add support to show the alternative to deprecated functions

stronk7:
The Generic DeprecatedFunctionsSniff is basically a wrapper
over the Generic ForbiddenFunctionsSniff, with only some
details making them different.

One of those details (that may be could be unified) is the
addError() method.

While the Forbidden Sniff supports showing the defined alternative,
curiously the Deprecated Sniff is missing that nice feature.

This PR simply implements it, when an alternative has been
defined, let's show it, like the Forbidden Sniff does.

Again, a nice thing would be to unify both Sniffs addError(), but
that falls out from this specific issue and, particularly, from
my time availability.

### DIFF
--- a/src/Standards/Generic/Sniffs/PHP/DeprecatedFunctionsSniff.php
+++ b/src/Standards/Generic/Sniffs/PHP/DeprecatedFunctionsSniff.php
@@ -61,6 +61,18 @@ class DeprecatedFunctionsSniff extends ForbiddenFunctionsSniff
         $error = 'Function %s() has been deprecated';
         $type  = 'Deprecated';
 
+        if ($pattern === null) {
+            $pattern = strtolower($function);
+        }
+
+        if ($this->forbiddenFunctions[$pattern] !== null
+            && $this->forbiddenFunctions[$pattern] !== 'null'
+        ) {
+            $type  .= 'WithAlternative';
+            $data[] = $this->forbiddenFunctions[$pattern];
+            $error .= '; use %s() instead';
+        }
+
         if ($this->error === true) {
             $phpcsFile->addError($error, $stackPtr, $type, $data);
         } else {


stronk7:
TODO: Add some test... will try after getting the checks run...

stronk7:
Ok, so it seems that adding unit tests for this is not as simple as expected because of how the tester and the sniff are working.

I've prepared another branch where you can see the test that I had prepared, that fails... and another commit with a dirty hack to make it pass.

https://github.com/stronk7/PHP_CodeSniffer/compare/deprecated_with_alternative_and_test~2...deprecated_with_alternative_and_test

The explanation is, basically, from commit message:

> Note that right now this doesn't work because of the following:
> 
> 1. We are setting forbiddenFunctions via phpcs:set, but the
>   DeprecatedFunctions Sniff does all its work using protected
>   forbiddenFunctionNames instead.
> 2. forbiddenFunctionNames is calculated on register(), not on
>   process, and AbstractSniffUnitTest only instantiates/runs it
>   once, so no matter we set forbiddenFunctions... the Sniff
>   continues working with the original functions.
> 
> I only can imagine 2 ways to make this work.
> 
> 1. Or AbstractSniffUnitTest is modified to instantiate the
>    Sniff as many times as needed (number of .inc files).
> 2. Or the ForbiddenFunctionsSniff is modified to re-calculate
>    forbiddenFunctionNames if differences are found.

So I'd suggest to accept this without the tests, because it's passing both in the "dirty hack" branch commented above, and also [here, because we use a completely different testing runner](https://github.com/stronk7/moodle-local_codechecker/commit/110913d7bead91a8c265d2f1ce2a082dbe43aa07) from `AbstractSniffUnitTest`, able to test together both the internal deprecations and the custom ones (number of errors/warnings and/or message contents).

For your consideration!

jrfnl:
> The Generic DeprecatedFunctionsSniff is basically a wrapper over the Generic ForbiddenFunctionsSniff, with only some details making them different.

> While the Forbidden Sniff supports showing the defined alternative, curiously the Deprecated Sniff is missing that nice feature.

@stronk7 I think you may be misunderstanding the `Generic.PHP.DeprecatedFunctions` sniff. As per the file docblock and as can also be seen in the `__construct` method, this sniff will automatically fill the `$forbiddenFunctions` array with the names of deprecated PHP native functions based on the PHP version on which you are running PHPCS. This property is **not** intended to be set from the ruleset for this particular sniff. This is also the reason why the property is not listed in the [wiki on customisable properties](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties).

As far as I know, there is no easy way to get suggested alternative functions from the PHP ReflectionAPI, so this array will never contain alternative, which is exactly why the "alternative" is not handled in the error messaging.

If you want a `DeprecatedFunctions` sniff for your own project/framework, you basically need to create your own by extending the `Generic.PHP.ForbiddenFunctions` sniff and providing it the list of deprecated functions with their alternatives.
The `Generic.PHP.DeprecatedFunctions` sniff should not be used for this.

stronk7:
Hi @jrfnl ,

the behavior that you've defined above is exactly what I wanted here. ONE unique Sniff able to check for both PHP internal deprecations and OWN (product) ones. And (children of) `Generic.PHP.DeprecatedFunctions` do support that perfectly, because the PHP internal deprecations just ADD new elements to the `$forbiddenFunctions` property, don't replace.

So, in our extending (from Deprecated, not from Forbidden) Sniff, we have done exactly that and it is detecting both deprecations (PHP and OWN) perfectly.

We *don't need/want to set that property in ruleset at all,* just when trying to cover it with tests using PHPCS `AbstractSniffUnitTest` I faced the problems commented in the last post. Using our own testing machinery (link below) it's passing perfectly (it shows that the extended Sniff supports both PHP and OWN without problem.

https://github.com/moodlehq/moodle-local_codechecker/pull/141/commits/09e69be5542814c48ee39ea48dc2d1565e7e3e1b#diff-bf5c9b5383660435e6aae89199370996be5636dcf2dfe45d13138e5b4a2e402d

In any case, this issue is only about to add support for show alternatives when there are alternatives. Something that the Sniff can support without trouble (its "parent" `Generic.PHP.ForbiddenFunctions` supports it).

More yet, it makes perfect sense that something deprecated can have an alternative, and phpdocs say so:

```
    /**
     * A list of forbidden functions with their alternatives.
     *
     * The value is NULL if no alternative exists. IE, the
     * function should just not be used.
     *
     * @var array<string, string|null>
     */
    public $forbiddenFunctions = [];
```

Why not?

Anyway, I'm happy if you want to keep this Sniff only supporting PHP deprecations, and without the ability of show alternatives. Just thought that it may be interesting to add the ability to report alternatives when available, hence the PR, because we have implemented it in our extended Sniff above (and assumed that it was legit use).

That's all, as said, all the complications came from the attempts to cover the PR with PHPCS `AbstractSniffUnitTest`, just that.

So, 100% up to you, thanks for the support!

Ciao :-)

Issue #3294, opened 2021-04-11
Bug in attribute tokenization when content contains PHP end token or attribute closer on new line


jrfnl:
**Describe the bug**
Tokenization of attributes is not always consistent across PHP versions.

While the given code sample may be an edge case, it is 100% valid PHP 8.0 code and is in actual fact one of the examples given in one of the related RFCs.
(last example in this section: https://wiki.php.net/rfc/shorter_attribute_syntax_change#discussion_of_forwards_compatibility_procons )

**Code sample**
```php
<?php
#[DeprecationReason('reason: <https://some-website/reason?>')]
function main() {}
```

Tokenizes on PHP 8.0 with PHPCS 3.6.0 as:
```
Ptr | Ln | Col  | Cond | ( #) | Token Type                 | [len]: Content
-------------------------------------------------------------------------
  0 | L1 | C  1 | CC 0 | ( 0) | T_OPEN_TAG                 | [5]: <?php

  1 | L2 | C  1 | CC 0 | ( 0) | T_ATTRIBUTE                | [2]: #[
  2 | L2 | C  3 | CC 0 | ( 0) | T_STRING                   | [17]: DeprecationReason
  3 | L2 | C 20 | CC 0 | ( 0) | T_OPEN_PARENTHESIS         | [1]: (
  4 | L2 | C 21 | CC 0 | ( 1) | T_CONSTANT_ENCAPSED_STRING | [40]: 'reason: <https://some-website/reason?>'
  5 | L2 | C 61 | CC 0 | ( 0) | T_CLOSE_PARENTHESIS        | [1]: )
  6 | L2 | C 62 | CC 0 | ( 0) | T_ATTRIBUTE_END            | [1]: ]
  7 | L2 | C 63 | CC 0 | ( 0) | T_WHITESPACE               | [0]:

  8 | L3 | C  1 | CC 0 | ( 0) | T_FUNCTION                 | [8]: function
  9 | L3 | C  9 | CC 0 | ( 0) | T_WHITESPACE               | [1]:
 10 | L3 | C 10 | CC 0 | ( 0) | T_STRING                   | [4]: main
 11 | L3 | C 14 | CC 0 | ( 0) | T_OPEN_PARENTHESIS         | [1]: (
 12 | L3 | C 15 | CC 0 | ( 0) | T_CLOSE_PARENTHESIS        | [1]: )
[...]

PR #3299, opened 2021-04-13
Fix bug in attributes tokenization on PHP < 8.0 (#3294)

alekitto:
As per title, this fixes the bug described in #3294.

Only if the PHP version is less than 8.0 AND the attribute closer token is not found (not being on the same line or placed after a "?>" token which is interpreted as a php close tag for example), the remainder of the file will be re-tokenized and the attribute closed is searched in the new token stream.
If found the new token stream is injected into the token array and parsing continues, otherwise the re-tokenized stream will be ignored.


gsherwood:
Thanks a lot for this fix

PR #3302, opened 2021-04-20
PHP 8.0 | Tokenizer/PHP: bugfix for union types using namespace operator

jrfnl:
Type declarations can use namespace relative Types, i.e. `namespace\Sub\Name`.

However, in that case, the `T_BITWISE_OR` token was incorrectly not converted to `T_TYPE_UNION`.

Includes unit tests.
Includes additional tests with all other "namespaced identifier name" types to safeguard this for the PHPCS 4.x change to the PHP 8.0 tokenization of identifier names.

**Note**: the tests should be forward-ported to the 4.x branch. The change to the `Tokenizer/PHP` file should **not**.


gsherwood:
Thanks for finding and fixing this. Merged into 4.0 as per instructions and all went well.

PR #3304, opened 2021-04-22
Generic/ExecutableFile: check files using short open echo tag

jrfnl:
I can't think of any good reason why "template files", which typically use the short open echo tag, should be excluded from the check done by this sniff.

Includes unit tests.


gsherwood:
Thanks for this PR, and all the other related ones I just merged. Given they would check files if they had a mix of standard open and short open echo tags in them, it makes sense they would check files that only have short open echo tags in them as well.

jrfnl:
@gsherwood Thanks for the merges and yes, that's exactly why I pulled them.
There are more sniffs which look just and only for the "normal" open tag, but for the others making this change seemed more contentious.

PR #3315, opened 2021-04-22
Squiz/BlockComment: also check for short open echo tag

jrfnl:
The sniff has different behaviour when there is a PHP open tag before the block comment, however, this behaviour was not applied when the PHP short open echo tag was encountered.

Includes unit tests.


gsherwood:
Thanks

PR #3319, opened 2021-04-22
Update Runner.php

WinterSilence:
1. Fix `@return`
2. Add optional parameter to `run*` methods to pass custom configuration settings. (to integrate into web/http applications)

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -49,22 +49,23 @@ class Runner
     /**
      * Run the PHPCS script.
      *
-     * @return array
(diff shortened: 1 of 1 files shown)


gsherwood:
Possibly helpful as well: https://gist.github.com/gsherwood/aafd2c16631a8a872f0c4a23916962ac

WinterSilence:
@gsherwood yes, but it's longer than `(new \PHP_CodeSniffer\Runner())->runPHPCBF(['--standard=PSR12', $file]);`, now constructor parameters are unuseful

PR #3320, opened 2021-04-23
PHP 8.0 | File::getMethodParameters(): bug fix for attributes leaking into typehint

### LINKED ISSUE #3298: PHP 8.0 | File::getMethodParameters() needs to support attributes
**Describe the bug**
Parameters in a function declaration can have attributes attached to them. This is currently not handled correctly in the `File::getMethodParameters()` method.

**Code sample**
```php
class ParametersWithAttributes(
    public function __construct(
        #[\MyExample\MyAttribute] private string $constructorPropPromTypedParamSingleAttribute,
        #[MyAttr([1, 2])]
        Type|false
        $typedParamSingleAttribute,
        #[MyAttribute(1234), MyAttribute(5678)] ?int $nullableTypedParamMultiAttribute,
        #[WithoutArgument] #[SingleArgument(0)] $nonTypedParamTwoAttributes,
        #[MyAttribute(array("key" => "value"))]
        &...$otherParam,
    ) {}
}
```

For the above code, the `T_STRING` tokens from within the attribute would be added to the `type_hint` array key in the return value, so for example, for the first parameter, the `type_hint` would (incorrectly) come back as `'type_hint' => '\MyExample\MyAttributestring'`

**Proposed behavior**
I'm currently looking into fixing this.

My current thinking hinges on four possible "solutions".

1. Ignore attributes in function declarations altogether. Just skip over them.
2. Add an `attributes` index key to the array for each parameter with a boolean flag to indicate whether there are attribute(s) attached to the parameter. This index key would always be set and defaults to `false`.
    The `content` key would remain the same as it is now and not contain the attribute(s).
    This option will give sniff writers an indication whether attributes are attached to the parameter. If the sniff writer would need the attribute details, they can do a `findPrevious()` for an `T_ATTRIBUTE_END` token before `'token'` and walk the attribute tokens from there.
3. Add three new index keys to the array for each parameter which would only be set when there are attribute(s) attached to the parameter.
    The `content` key would now also contain the complete attribute(s).
    - `attributes` containing a string representation of the attribute(s).
    - `attributes_start` containing the stack pointer to the first attribute opener for this parameter
    - `attributes_end` containing the stack pointer to the last attribute closer for this parameter.
4. Add a new multi-level array index key `attributes`, which would only be set when there are attribute(s) attached to the parameter. The array format would be along the lines of the below.
    The `content` key would now also contain the complete attribute(s).
    ```
    'attributes' => [
        [0] => [
            - `attribute` containing a string representation of the first attribute.
            - `attribute_start` containing the stack pointer to the attribute opener for the first attribute.
            - `attribute_end` containing the stack pointer to the attribute closer for the first attribute.
        ],
        [1] => ...
    ]
    ```

I'm personally leaning towards either option 1 or option 2.

I don't think option 3 is a good solution as there can be multiple attributes attached to the parameter and that solution does not do that justice.

As for option 4, I'm not so sure that this is really needed as sniffs examining attributes can just listen to `T_ATTRIBUTE` instead and for sniffs specifically only examining attributes in function declarations, having the `attributes` indicator (option 2) or walking the tokens between the open/close parentheses looking for `T_ATTRIBUTE` tokens, is probably enough and would prevent the performance hit of processing potential attributes on each call to `File::getMethodParameters()`.

@gsherwood Your input/opinion on which solution is preferred would be much appreciated.

 - PHPCS: `master`

jrfnl:
### File::getMethodParameters(): add test documenting behaviour for comments

This test demonstrates and documents the existing behaviour of the method when comments are encountered within a parameter declaration.

### File:;getMethodParameters(): bug fix for attributes leaking into typehint

This commit adds handling of parameter attributes to the `File::getMethodParameters()` method as per option [2] discussed in issue #3298.

In practice this means that:
* [New] A new `attributes` index is introduced into the returned array which will hold a boolean value indicating whether attributes are attached to the parameter.
* [Unchanged] The `content` index in the returned array includes the textual representation of any attributes attached to a parameter.
* [Fixed] The `type_hint` and `type_hint_token` indexes will no longer be polluted (set incorrectly) with information belonging to the attribute(s) instead of to the type declaration.

Includes minor efficiency fix for handling of parenthesis and brackets in default values.

Includes dedicated unit test.

Fixes #3298

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -1283,6 +1283,7 @@ class File
      *         'name'                => '$var',  // The variable name.
      *         'token'               => integer, // The stack pointer to the variable name.
      *         'content'             => string,  // The full content of the variable definition.
+     *         'attributes'          => boolean, // Does the parameter have one or more attributes attached ?
      *         'pass_by_reference'   => boolean, // Is the variable passed by reference?
      *         'reference_token'     => integer, // The stack pointer to the reference operator
      *                                           // or FALSE if the param is not passed by reference.
@@ -1355,6 +1356,7 @@ class File
         $defaultStart    = null;
         $equalToken      = null;
         $paramCount      = 0;
+        $attributes      = false;
         $passByReference = false;
         $referenceToken  = false;
         $variableLength  = false;
@@ -1373,18 +1375,25 @@ class File
             if (isset($this->tokens[$i]['parenthesis_opener']) === true) {
                 // Don't do this if it's the close parenthesis for the method.
                 if ($i !== $this->tokens[$i]['parenthesis_closer']) {
-                    $i = ($this->tokens[$i]['parenthesis_closer'] + 1);
+                    $i = $this->tokens[$i]['parenthesis_closer'];
+                    continue;
                 }
             }
 
             if (isset($this->tokens[$i]['bracket_opener']) === true) {
-                // Don't do this if it's the close parenthesis for the method.
                 if ($i !== $this->tokens[$i]['bracket_closer']) {
-                    $i = ($this->tokens[$i]['bracket_closer'] + 1);
+                    $i = $this->tokens[$i]['bracket_closer'];
+                    continue;
                 }
             }
 
             switch ($this->tokens[$i]['code']) {
+            case T_ATTRIBUTE:
+                $attributes = true;
+
+                // Skip to the end of the attribute.
+                $i = $this->tokens[$i]['attribute_closer'];
+                break;
             case T_BITWISE_AND:
                 if ($defaultStart === null) {
                     $passByReference = true;
@@ -1501,6 +1510,7 @@ class File
                     $vars[$paramCount]['default_equal_token'] = $equalToken;
                 }
 
+                $vars[$paramCount]['attributes']          = $attributes;
                 $vars[$paramCount]['pass_by_reference']   = $passByReference;
                 $vars[$paramCount]['reference_token']     = $referenceToken;
                 $vars[$paramCount]['variable_length']     = $variableLength;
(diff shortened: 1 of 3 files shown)


gsherwood:
Do you think the new array index would be better named "has_attributes" to indicate that it's a boolean? I could make that change once I go to merge this if you agree.

jrfnl:
> Do you think the new array index would be better named "has_attributes" to indicate that it's a boolean?

I'm not fussed about the name either way, but yes, `has_attributes` did occur to me to.

I based the current name `attributes` on the existing naming of index keys used in this method. To be more specific: the other boolean values don't have a key with a verb attached.
I.e. the "is the variable passed by reference ?" index is `pass_by_reference` not `is_reference`. Similarly, the "is the type nullable" indicator is called `nullable_type` not `is_nullable`.

jrfnl:
@gsherwood Shall I otherwise just remove the extra array index key for now ? We can always circle back to that if and when there are explicit requests for it.

gsherwood:
> Shall I otherwise just remove the extra array index key for now ? We can always circle back to that if and when there are explicit requests for it.

I think it's fine to leave it - the work has been done - but I will likely rename it.

I think `pass_by_reference` and `nullable_type` aren't great, but I don't think they would be confused with actually containing a non-boolean value because there is already another array index with the value extracted into it. But I think `attributes` could be confused with actually containing the value of the attributes themselves.

The actual value of the MR is in the fix itself, but I still think the new index is a nice addition if we can agree on a name. Given you had the exact same naming thought, if you're ok with some inconsistency in the naming (hopefully I explained why I am) then I think the index should stay.

gsherwood:
Thanks a lot for this. I did end up changing the name of the array index to `has_attributes` in case `attributes` is ever needed to contain the attribute value itself.

PR #3334, opened 2021-05-06
PHP 8.0 | Squiz.WhiteSpace.MemberVarSpacing: add support for attributes

yariknechyporuk:


src/Standards/Squiz/Tests/WhiteSpace/MemberVarSpacingUnitTest.inc
```diff
@@ -332,3 +332,34 @@ class CommentedOutCodeAtStartOfClassNoBlankLine {
...
+     * Description.
+     */
+
+    #[Attribute1]
+
+    #[Attribute2]
+
+
+    private $var3 = 'value';
+}
```
jrfnl, line 365: Just some suggestions:
* What about a test case where the attribute is _before_ the docblock ?
* What about a test case with one attribute on the same line as the property ?
* What about a test case with two attributes on the same line as the property ?

Would be good to verify that the sniff will handle those cases correctly as well.

yariknechyporuk: Despite there is no PSR for Attributes for now, I suggest to always keep phpDoc before attributes.

jrfnl:
For visibility: there's a discussion open about this PR here: https://github.com/squizlabs/PHP_CodeSniffer/discussions/3419

PR #3335, opened 2021-05-08
ErrorSuppressionTest: refactor to use data providers + bug fixes for the tests

jrfnl:
Follow up to https://github.com/squizlabs/PHP_CodeSniffer/pull/3071#issuecomment-828109667

This refactors all test methods in the `ErrorSuppresionTest` class to use data providers.

For each of these refactors, the following comment applies:

* Maintains the exact same existing tests, now using a data provider.
    The data provider uses keys which allows for much more descriptive output when using the PHPUnit `--testdox` option, as well as for easier debugging if/when a test would fail.
* Orders the tests in logical groups in the data provider.
* Switches out `assertEquals()` (loose type comparison) for `assertSame()` (strict type comparison).
* Caches the `$config` and `$ruleset` for the test via static local variables to prevent the test run from becoming slower due to the change to the data provider.

### ErrorSuppressionTest: move one test case out of testSuppressLine()

... as it uses a different code pattern, so doesn't fit with the other tests for the (upcoming) data provider.

### ErrorSuppressionTest: fix bug in testSuppressScope()

As the "no suppression" test would yield no errors, the test wasn't actually testing anything as there were no errors to suppress.

Fixed now by using a different sniff to test against, which does yield an error on the line being suppressed by the rest of the tests.

### ErrorSuppressionTest::testSuppressFile(): fix bugs in test

This fixes two bugs in two test cases for the `testSuppressFile()` method:
1. The code snippet as was, contained a parse error (missing semi-colon after class instantiation).
2. The code snippet did not contain anything which would trigger the warning the test is looking for in the first place, so these two tests would always pass.

### ErrorSuppressionTest::testDisableSelected(): refactor to data provider

**Note**: the "docblock" tests have changed to use the same basic code sample as the other tests. In practice this means that instead of having 0 errors and 0/1 warnings, they will now yield 1 error and 0/1 warnings. Functionally these tests still test the same principle.

### ErrorSuppressionTest::testDisableSelected(): add some more tests

By the looks of it, combining disabling at different levels and disabling error codes wasn't covered in the tests.

### DIFF
--- a/tests/Core/ErrorSuppressionTest.php
+++ b/tests/Core/ErrorSuppressionTest.php
@@ -579,7 +579,7 @@ EOD;
     {
         $config            = new Config();
         $config->standards = ['PEAR'];
-        $config->sniffs    = ['PEAR.NamingConventions.ValidVariableName'];
+        $config->sniffs    = ['PEAR.Functions.FunctionDeclaration'];
 
         $ruleset = new Ruleset($config);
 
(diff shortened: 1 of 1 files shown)


gsherwood:
Was a fun merge into 4.0 :) I ended up squashing the commits to make it easier, but think I captured all the changes in the commit message. Thanks - so much cleaner.

jrfnl:
@gsherwood I imagine so 😂  And yes, the separate commits were only there to allow for easier review of this PR as it's so huge and the review was done for 3.x branch anyhow.

PR #3339, opened 2021-05-09
NamedFunctionCallArgumentsTest: minor fixes

jrfnl:
The error message would display the token code, which is not helfpul for a human, so changing it to the token `type`.


gsherwood:
Thanks

PR #3340, opened 2021-05-10
Ensure interface and trait names are always tokenized as T_STRING

jrfnl:
Interface and trait names should always be `T_STRING`.


gsherwood:
Did this come up in actual code I can use for a test, or is it something you just noticed?

gsherwood:
Thanks

PR #3341, opened 2021-05-10
Tokenizer/PHP: stabilize T_FINALLY backfill

jrfnl:
Make the backfill for T_FINALLY a little more stable by preventing re-tokenizing non-keyword `finally` strings to `T_FINALLY`.

Includes unit tests.


jrfnl:
Oh dear, appears the lined endings got switched and not corrected. Thanks for pointing that out. Fixed now.

gsherwood:
Thanks

PR #3346, opened 2021-05-11
Fixer/Cbf: add missing new line in level >=2 debug output

jrfnl:
... to make the output slightly more readable.

### Before:
image

### After:
image

### DIFF
--- a/src/Fixer.php
+++ b/src/Fixer.php
@@ -184,6 +184,9 @@ class Fixer
                 }
 
                 echo ']... ';
+                if (PHP_CODESNIFFER_VERBOSITY > 1) {
+                    echo PHP_EOL;
+                }
             }
 
             if ($this->numFixes === 0 && $this->inConflict === false) {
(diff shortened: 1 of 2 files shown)


gsherwood:
These newlines could only be added if the verbosity level is greater than 1 because the line is rewritten during the run. 

So running with `-v` would show:
```
Processing temp.php [PHP => 46 tokens in 7 lines]... DONE in 6ms (6 fixable violations)
	=> Fixing file: 6/6 violations remaining
        => Fixing file: 0/6 violations remaining [made 2 passes]... DONE in 3ms
	=> File was overwritten
```
Instead of:
```
Processing temp.php [PHP => 46 tokens in 7 lines]... DONE in 6ms (6 fixable violations)
        => Fixing file: 0/6 violations remaining [made 2 passes]... DONE in 3ms
	=> File was overwritten
```

gsherwood:
I do agree it's a good change though :) Strange that I've never noticed.

jrfnl:
Good catch on the level 1 output. I've fixed up the commit and added the new lines within additional debug level conditionals now.

gsherwood:
Thanks for this. I applied the same change in 4.0 but the messages are printed differently so I couldn't just cherry pick the commit.

PR #3348, opened 2021-05-11
Squiz/ObjectInstantiation: bug fix - account for null coalesce

### LINKED ISSUE #3333: Squiz.Objects.ObjectInstantiation: null coalesce operators are not recognized as assignment
**Describe the bug**
Sniff `Squiz.Objects.ObjectInstantiation` does not recognizes `??` and `??=` operators as assignment.

**Code sample**
```php
$time1 = new \DateTime();
$time2 ??= new \DateTime();
$time3 = $time1 ?? new \DateTime();
```

**Custom ruleset**
```xml
<?xml version="1.0"?>
  <rule ref="Squiz.Objects.ObjectInstantiation" />
</ruleset>
```

**To reproduce**
2. Run `phpcs test.php ...`
```
 4 | ERROR | New objects must be assigned to a variable
 5 | ERROR | New objects must be assigned to a variable
```

**Expected behavior**
Error `Squiz.Objects.ObjectInstantiation.NotAssigned` is not reported for any of the lines.

 - OS: Ubuntu Linux 20.10
 - PHP: 8.0.5
 - PHPCS: 3.6.0
 - Standard: custom

jrfnl:
The `Squiz.Objects.ObjectInstantiation` did not take null coalesce into account.

It also was potentially a little too lenient for ternaries as an inline then or inline else token before the `new` keyword would be considered an assignment, while it was never checked that the result of the ternary was actually assigned to something.

While still not 100%, this commit improves the sniffs to:
1. Allow for null coalesce assignments `??=`.
2. For ternary tokens + null coalesce `??`: verify if the result of either of these is actually "assigned" to something (or preceded by one of the other "allowed" tokens).

This should reduce the number of both false positives as well as false negatives, though there is still the potential for some false negatives - see the test on line 32 of the test case file.
This could be further reduced in a future iteration on this sniff, but for now, this fix can be considered a significant step forward.

Includes unit tests.

Fixes #3333

### DIFF
--- a/src/Standards/Squiz/Sniffs/Objects/ObjectInstantiationSniff.php
+++ b/src/Standards/Squiz/Sniffs/Objects/ObjectInstantiationSniff.php
@@ -48,20 +48,36 @@ class ObjectInstantiationSniff implements Sniff
         $prev = $phpcsFile->findPrevious($allowedTokens, ($stackPtr - 1), null, true);
 
         $allowedTokens = [
-            T_EQUAL        => true,
-            T_DOUBLE_ARROW => true,
-            T_FN_ARROW     => true,
-            T_MATCH_ARROW  => true,
-            T_THROW        => true,
-            T_RETURN       => true,
-            T_INLINE_THEN  => true,
-            T_INLINE_ELSE  => true,
+            T_EQUAL          => T_EQUAL,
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks for fixing this.

Line 32 probably is a false negative, although it is an interesting one and could go either way.

PR #3349, opened 2021-05-12
DoubleArrowTest: minor bug fix

jrfnl:
The test case file contained one duplicate test marker and the actual test file contained a few duplicate array keys, which meant that a few cases weren't actually being tested.

Fixed now.

Note: the diff looks bigger than it really is due to the array arrow alignment. Best viewed while ignoring whitespace changes.


gsherwood:
Nice find, thank you.

jrfnl:
@gsherwood Well, there's a sniff for that 😂 
https://github.com/PHPCSStandards/PHPCSExtra/blob/develop/Universal/Sniffs/Arrays/DuplicateArrayKeySniff.php

PR #3350, opened 2021-05-12
 Tokenizer/PHP: bug fix - fix performance issue

jrfnl:
### Tokenizer/PHP: bug fix - fix performance issue

Happened to come across this while investigating something else.

As it was, as long as no open parenthesis or variable was encountered, this snippet would loop to the end of the file for each `T_ARRAY` token encountered as it would only `break` in the `if/elseif` and there was no `else` clause.

Basically, we only want the `array` keyword to be tokenized as `T_ARRAY` if it is an actual array declaration. In all other cases, it should be tokenized as `T_STRING`.

This fixes the performance leak by only looping to the first non-empty token after the keyword, checking if it's an open parenthesis and retokenizing the `T_ARRAY` to `T_STRING` in all other cases.

It also removes the need for the separate _return type_ retokenization of the array keyword.

Includes adding unit tests specifically for the array keyword.

### Sniff test: adjust two test files to match

This removes two expected errors due to the tokenizer change made in the previous commit.

#### `Generic.Arrays.DisallowLongArraySyntax`

The code on line 13 is a parse error, so the sniff no longer throwing an error for it should not be our concern.
```php
$var = array;
```

#### `Squiz.PHP.CommentedOutCode`

The code which triggered the warning on line 35 was a docblock written as a block comment and not really commented out code anyway, so this could be considered a fix for a false positive.
```php
/*
 * The listeners array.
 *
 * @var array(PHP_CodeSniffer_Sniff)
 */
```

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -150,6 +150,8 @@ http://pear.php.net/dtd/package-2.0.xsd">
      <dir name="Tokenizer">
       <file baseinstalldir="" name="AnonClassParenthesisOwnerTest.inc" role="test" />
       <file baseinstalldir="" name="AnonClassParenthesisOwnerTest.php" role="test" />
+      <file baseinstalldir="" name="ArrayKeywordTest.inc" role="test" />
+      <file baseinstalldir="" name="ArrayKeywordTest.php" role="test" />
       <file baseinstalldir="" name="AttributesTest.inc" role="test" />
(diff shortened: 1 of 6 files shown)


gsherwood:
Sorry this took a while to go through, but those changes all made sense to me. Thanks a lot for the improvement.

jrfnl:
No worries, glad it's merged now though.

Issue #683, opened 2015-08-15
Comments are incorrectly reported by PSR2.ControlStructures.SwitchDeclaration sniff


OndraM:
Hi,
since 2.3.3 (and caused probably by #621) comments are incorrectly reported as an error in following structure:

```php
switch ($foo) {
    case 'bar': // some comment
        ...
        break;
    case 'baz': // other comment
        ...
        break;
    default: // other comment
        ...
        break;
}
```

On the lines with the comment following error occurs:
`ERROR | [x] The CASE body must start on the line following the statement`

... though the case body actually starts on a proper line.

gsherwood:
Thanks for reporting.

rmbl:
There still seems to be a problem with comments in the next line after the case statement, which triggers errors for our codebase.

```php
switch ($foo) {
    case 'bar': 
        // some comment
        ...
        break;
}
```

gsherwood:
Thanks again. I think I've got the problem this time.

PR #3354, opened 2021-05-13
PSR2/SwitchDeclaration: various bug fixes

### LINKED ISSUE #3297: PSR2.ControlStructures.SwitchDeclaration.TerminatingComment does not handle try/finally blocks
**Describe the bug**
Using the sniff `PSR2.ControlStructures.SwitchDeclaration` brings in an edge case for the error code `TerminatingComment` when having try/finally in the switch case body

**Code sample**
```php
<?php

$statusValue = mt_rand();
$resource = SomeClass::getWorker();

switch ( $statusValue ) {
	case 0:
		// fall-through
	case 2:
		try {
			return $resource->doWork();
		} finally {
			$resource->release();
		}
	case 3:
		$other = $code;
		break;
}
```

**Custom ruleset**
```xml
<?xml version="1.0"?>
  <rule ref="PSR2.ControlStructures.SwitchDeclaration" />
</ruleset>
```

**To reproduce**
2. Run `phpcs test.php ...`
```
 9 | ERROR | There must be a comment when fall-through is intentional
   |       | in a non-empty case body
   |       | (PSR2.ControlStructures.SwitchDeclaration.TerminatingComment)
```

**Expected behavior**
The try/finally always returns, that means there is no fall-through which does not need a comment.
That also the same when using finally + catch or catch alone

 - OS: Windows 10
 - PHP: 8.0
 - PHPCS: 3.6.0
 - Standard: -

**Additional context**
-

### LINKED ISSUE #3352: PSR2.ControlStructures.SwitchDeclaration can remove comments on the same line as the case statement while fixing
**Describe the bug**
Running `PSR2.ControlStructures.SwitchDeclaration.BodyOnNextLineCASE` can remove comments which are on the same line as the `case` statement

**Code sample**
```php
$test = 2;

switch ( $test ) {
	case 2: // comment followed by empty line

		echo "test 2";
}
```

**Custom ruleset**
```xml
<?xml version="1.0"?>
  <rule ref="PSR2.ControlStructures.SwitchDeclaration" />
</ruleset>
```

**To reproduce**
2. Run `phpcbf test.php ...`
3. The comment is gone.
```php
$test = 2;

switch ( $test ) {
	case 2:
		echo "test 2";
}

```

**Expected behavior**
The remove of the empty line is correct, but the comment should stay from my point of view.

 - OS: Windows 10
 - PHP: 8.0
 - PHPCS: 3.6.0
 - Standard: -

**Additional context**
The fixer starts directly after the colon/scope opener to remove everything there, not at the end of the line

Seems an issue since the fix for #683, which ignores the comments, but does not adjust the fixer part.

jrfnl:
### PSR2/SwitchDeclaration: bug fix - don't remove trailing comments

As reported in #3352, the `PSR2.ControlStructures.SwitchDeclaration.BodyOnNextLineCASE` fixer could inadvertently remove trailing comments.

This commit improves on the fix which was put in place in response to #683, by making the detection of what should be considered the start of the body more precise:
* Any amount of trailing comments on the same line as the `case` will be ignored.
* Any type of comment will be taken into consideration, not just PHPCS annotations and `T_COMMENT` tokens.
* Comments _not_ on the same line as the `case` and non-empty tokens on the same line as the `case` will both be considered as the "start of the body".

As for the fixer. The "body starts on same line" fixer was already okay. The "empty lines between case and body" fixer, however, will now ignore anything on the same line as the `case`, which prevents us having to account for comments which already include a new line.

Includes unit tests.

Fixes #3352

### PSR2/SwitchDeclaration: bug fix - improve handling of comments

Comments in certain places in the code flow would throw the `findNestedTerminator()` method used for determining the `TerminatingComment` error off.

Fixed now.

Includes unit tests.
Without this fix, the first test would not throw an error (false negative), while the second would (false positive).

### PSR2/SwitchDeclaration: bug fix - handle try/catch/finally correctly

Add code to handle `try-catch-finally` statements correctly when determining whether a `case` needs a terminating comment.

As per #3297 (comment):

> * If there is a `finally` statement and the body of that contains a "terminating statement" (`return`, `break` etc), we don't need to check the body of the `try` or any of the `catch` statements as, no matter what, all execution branches are then covered by that one terminating statement.
> * If there is **_no_** `finally` or if there is, but it does not contain a terminating statement, then all `try` and `catch` structures should each contain a terminating statement in the body.

Includes plenty of unit tests.

Fixes #3297

### DIFF
--- a/src/Standards/PSR2/Sniffs/ControlStructures/SwitchDeclarationSniff.php
+++ b/src/Standards/PSR2/Sniffs/ControlStructures/SwitchDeclarationSniff.php
@@ -107,13 +107,13 @@ class SwitchDeclarationSniff implements Sniff
                     }
                 }
 
-                $next = $phpcsFile->findNext(T_WHITESPACE, ($opener + 1), null, true);
-                if ($tokens[$next]['line'] === $tokens[$opener]['line']
-                    && ($tokens[$next]['code'] === T_COMMENT
-                    || isset(Tokens::$phpcsCommentTokens[$tokens[$next]['code']]) === true)
-                ) {
-                    // Skip comments on the same line.
-                    $next = $phpcsFile->findNext(T_WHITESPACE, ($next + 1), null, true);
+                for ($next = ($opener + 1); $next < $nextCloser; $next++) {
+                    if (isset(Tokens::$emptyTokens[$tokens[$next]['code']]) === false
+                        || (isset(Tokens::$commentTokens[$tokens[$next]['code']]) === true
+                        && $tokens[$next]['line'] !== $tokens[$opener]['line'])
+                    ) {
+                        break;
+                    }
                 }
 
                 if ($tokens[$next]['line'] !== ($tokens[$opener]['line'] + 1)) {
@@ -126,6 +126,11 @@ class SwitchDeclarationSniff implements Sniff
                         } else {
                             $phpcsFile->fixer->beginChangeset();
                             for ($i = ($opener + 1); $i < $next; $i++) {
+                                if ($tokens[$i]['line'] === $tokens[$opener]['line']) {
+                                    // Ignore trailing comments.
+                                    continue;
+                                }
+
                                 if ($tokens[$i]['line'] === $tokens[$next]['line']) {
                                     break;
                                 }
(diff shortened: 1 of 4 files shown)


jrfnl:
I've updated this PR and added three more commits.
* One fix for a bug with the token walking not handling comments correctly discovered while researching #3297.
* One with a fix for #3297.
* And one small performance fix.

Pulling these separately would mean that the test files would conflict, so adding the commits to this open PR seemed like the simplest way forward.

I've updated the PR title + description to match the new content.

gsherwood:
Thanks a lot for all these fixes.

PR #3356, opened 2021-05-14
Normalise License file

exussum12:
As the BSD-3 licence has been adjusted github doesn't notice this is the BSD-3 licence, after these changes github will pick up the correct licence


gsherwood:
Thanks. Github used to identify the licence even modified but I didn't notice that this changed.

PR #3358, opened 2021-05-15
Generic/OpeningFunctionBraceBsdAllman: bug fix - prevent removing return types

### LINKED ISSUE #3357: Generic.Functions.OpeningFunctionBraceBsdAllman removes return type when additional lines are present
**Describe the bug**
When a function has a return type and an additional new line between the brace and function signature, `phpcbf` will remove the return type.

**Code sample**
```php
<?php

class FormattingBug
{
    public function extraLine(string: $a): void

    {
        // code here.
    }
}
```

**To reproduce**
2. Run `phpcbf --standard=PSR12 test.php`

phpcbf will update the problematic formatting, however the resulting code looks like:

```php
<?php

class FormattingBug
{
    public function extraLine(string: $a)
    {
        // code here.
    }
}
```

**Expected behavior** 

Formatting to be updated without removing the return type.

jrfnl:
As reported in #3357, the `Generic.Functions.OpeningFunctionBraceBsdAllman` sniff would remove return types (and comments) when fixing code where blank lines existed between the end of the function declaration and the open brace.

This commit fixes that bug.

In the case of comments, the `BraceSpacing` error will no longer auto-fix as a dev should decide where the comment should go and/or whether it should be removed.

Includes unit tests.

Fixes #3357

Previously #1931, #1938

### DIFF
--- a/src/Standards/Generic/Sniffs/Functions/OpeningFunctionBraceBsdAllmanSniff.php
+++ b/src/Standards/Generic/Sniffs/Functions/OpeningFunctionBraceBsdAllmanSniff.php
@@ -139,17 +139,31 @@ class OpeningFunctionBraceBsdAllmanSniff implements Sniff
         } else if ($lineDifference > 1) {
             $error = 'Opening brace should be on the line after the declaration; found %s blank line(s)';
             $data  = [($lineDifference - 1)];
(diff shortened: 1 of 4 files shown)


gsherwood:
Thanks for this fix

PR #3370, opened 2021-06-03
Replace hot Tokenizer loop with a mathematical expression

thiemowmde:
This piece of code is very hot, executed thousands, possibly millions of times.

I found this when I benchmarked our custom sniffs. The replaceTabsInToken() method is consistently reported as one of the absolute hottest ones. I looked at the method and was wondering why this particular piece is a loop when it can be a mathematical expression. According to my micro-benchmark (only measuring the changed piece) it's about twice as fast now. This will probably not have much of an effect in a real-world scenario. Still I find it worth it.

### DIFF
--- a/src/Tokenizers/Tokenizer.php
+++ b/src/Tokenizers/Tokenizer.php
@@ -643,7 +643,7 @@ abstract class Tokenizer
 
                 // Move the pointer to the next tab stop.
                 $length      = $tabWidth - ($currColumn + $tabWidth - 1) % $tabWidth;
-	            $currColumn += $length;
+                $currColumn += $length;
                 $newContent .= $prefix.str_repeat($padding, ($length - 1));
             }//end foreach
         }//end if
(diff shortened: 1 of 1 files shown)


src/Tokenizers/Tokenizer.php
```diff
@@ -642,21 +642,10 @@ public function replaceTabsInToken(&$token, $prefix=' ', $padding=' ', $tabWidth
...
-                } else {
-                    $currColumn++;
-                    while (($currColumn % $tabWidth) !== 0) {
-                        $currColumn++;
-                    }
-
-                    $currColumn++;
-                }
-
-                $length     += ($currColumn - $lastCurrColumn);
```
longwave, line 658: `$lastCurrColumn` appears to be unused now and the declaration above can be removed

gsherwood:
Thanks for this change. Code bases with lots of strings or comments (including docblocks) with tabs may see some improvement, so I absolutely think it's a worthwhile change.

Issue #2551, opened 2019-07-14
Rule override for each directory (feature request)


top-master:
Taking just one rule-set XML file from the command line is really not enough to cover our entire project, its unit test dir specific rules, its sub-module related rules, and so on; all this in a single XML file can get quite complex.

For example, when `git add .` command is executed it handles current and all sub-directories and:
   - It searches for `.gitignore` files in each directory (starting from the nearest root directory which contains the `.git` directory until currently being handled directory).
   - When `.gitignore` exists its rules get added to the rule-stack, and once that directory is finished those rules are removed again from the rule-stack (it is using "Last In, First Out" queue or something).

I would like `PHPCS` to handle the `.phpcs.xml` files in a similar way, I mean:
   -  `PHPCS` should find its root directory by searching in current and parent directories for the nearest `composer.json` file, which's rule-set is not set to "`<use-parent>`" (The `<use-parent>` rule-set should be a reserved keyword to allow sub-modules to reuse the settings from a parent directory, which might be the main project).
   - Then `PHPCS` should get its initial rule-set from that `composer.json` file (and ignore `.phpcs.xml` files of any parent directories).
   - At last, like `git` does for `.gitignore` files, `PHPCS` should search for `.phpcs.xml` files in each directory (starting from the root directory until currently being handled directory).
   - When `.phpcs.xml` exists its rules should get added to the rule-stack, and once that directory is finished those rules should get removed from the rule-stack again.
   - Once all this is done, each directory can have its own rules without need to change parent directories rules, and users only need to execute the `phpcs .` or `phpcbf .` command (for backward compatibility, when no `composer.json` file in path contains the initial rule-set `PHPCS` should just do as it does currently).

westonruter:
I believe this is the same issue as is being described in the JetBrains issue tracker (for PhpStorm): https://youtrack.jetbrains.com/issue/WI-36028

My use case is that I have a project (a WordPress site) which contains multiple subprojects (WordPress themes and plugins), each of which may have their own configurations for PHPCS,  ESLint, etc. When editing a JS file in a subproject, the IDEA lints the file according to the nearest ESLint config in that file's directory tree. In order to get the same behavior with PHPCS, as [noted](https://youtrack.jetbrains.com/issue/WI-36028#focus=Comments-27-4625951.0-0) by @MaXal: 

[...]

PR #3378, opened 2021-06-14
Allow the filter to return LocalFile objects to FileList

anomiex:
This gets us most of the way to fixing #2551. Now it's just a matter of writing a filter that updates the Config and/or Ruleset as appropriate while processing the directory tree.

I've written such a filter in [anomiex/phpcs-filter-draft](https://github.com/anomiex/phpcs-filter-draft/tree/master/phpcs-filter), which you can look at if you want to see how we intend to make use of this if this PR is accepted. Other people could do something similar but make other decisions, such as the looking for `composer.json` as suggested in #2551.

Since this change to `Filter::current()` may break expectations of existing filters, a filter must opt in to this change by overriding the protected `$this->produceLocalFileObjects` to true.

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -29,9 +29,9 @@ class File
     /**
      * The content of the file.
      *
-     * @var string
+     * @var string|null
      */
-    protected $content = '';
+    protected $content = null;
 
     /**
      * The config data for the run.
@@ -552,12 +552,18 @@ class File
      * Tokenizes the file and prepares it for the test run.
      *
      * @return void
+     * @throws \PHP_CodeSniffer\Exceptions\RuntimeException If no content is set.
      */
     public function parse()
     {
         if (empty($this->tokens) === false) {
             // File has already been parsed.
             return;
+        }
+
+        if ($this->content === null) {
+            // No content has been set.
+            throw new RuntimeException('No content has been set');
         }
 
         try {
--- a/src/Files/FileList.php
+++ b/src/Files/FileList.php
@@ -84,7 +84,12 @@ class FileList implements \Iterator, \Countable
                 $iterator = new \RecursiveIteratorIterator($filter);
 
                 foreach ($iterator as $file) {
-                    $this->files[$file->getPathname()] = null;
+                    if ($file instanceof LocalFile) {
+                        $this->files[$file->getFilename()] = $file;
+                    } else {
+                        $this->files[$file->getPathname()] = null;
+                    }
+
                     $this->numFiles++;
                 }
             } else {
@@ -125,7 +130,12 @@ class FileList implements \Iterator, \Countable
         $iterator = new \RecursiveIteratorIterator($filter);
 
         foreach ($iterator as $path) {
-            $this->files[$path] = $file;
+            if ($path instanceof LocalFile) {
+                $this->files[$path->getFilename()] = ($file ?? $path);
+            } else {
+                $this->files[$path] = null;
+            }
+
             $this->numFiles++;
         }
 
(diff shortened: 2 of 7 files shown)


kraftbj:
Howdy @gsherwood — I wanted to check-in to see your thoughts on this PR and set my expectations if this seems feasible to be included or if we needed to figure out an alternative approach toward 2551. Our particular goal is to be able to have per-directory rules or overrides. We have a similar setup for eslint and hopeful to have similar flexibility for phpcs.

Thank you!

anomiex:
@gsherwood FYI, we're now making use of this patch in production in our monorepo at https://github.com/Automattic/jetpack. It'd be nice if this could get reviewed.

anomiex:
@gsherwood We've been using this in production in our monorepo for three months now. It seems to work well.

jrfnl:
Without having an express opinion on this PR, I would like to warn/point out that this contains a BC-break for external standards and other integrations, at the very least for those external standards using their own testing framework, but also for integrations using a custom runner.
For that reason, if this change would be accepted, IMO it should go into 4.0.0.

anomiex:
@jrfnl Would you clarify? I tried to avoid having a BC break in here by making the filter have to specifically request the new behavior. 

Or is it the thing about LocalFile needing to have reloadContent() called? I suppose we could change it to call that on demand in `parse()` if it hasn't already been called instead of throwing an exception.

jrfnl:
> Or is it the thing about LocalFile needing to have reloadContent() called?

Exactly that.

anomiex:
@jrfnl How's it look to you now?

kraftbj:
@gsherwood or @jrfnl — Just wanted to check-in to see if this PR looked suitable now or we needed to do more work until it could be considered?

jrfnl:
@kraftbj I think the question which needs to be answered first is whether #2551 is a feature which PHPCS wants to support. That question never got answered and the #2551 ticket has seen little traction, so there doesn't seem to be that much community interest in it either.

This PR is an implementation detail, which isn't interesting to look at, unless the underlying question is first answered with "yes, the feature is a candidate to be added to PHPCS". (and no, I'm not the right person to answer that question)

gsherwood:
I like that general idea of #2551, but it's never been requested enough to put time into it and is not something I'd be looking to support in the core at this time. If I was, I'd make that change in version 4 regardless of BC breaks.

If this PR makes it easier to support outside of PHPCS, then I'm happy to consider that for inclusion sooner, but I have not looked into the purpose of these changes yet.

anomiex:
> If this PR makes it easier to support outside of PHPCS, then I'm happy to consider that for inclusion sooner, but I have not looked into the purpose of these changes yet.

That's exactly what this does.

PHPCS already uses a Ruleset and Config referenced by the File object rather than using a global Ruleset and Config. I didn't change anything with respect to that.

The main thing this PR does is allow an implementation of `PHP_CodeSniffer\Filters\Filter` to return a `PHP_CodeSniffer\Files\LocalFile` object rather than a string filename (or SplFileInfo object) that PHPCS will later wrap in a LocalFile anyway. It also, as a resource management measure, updates LocalFile to load the file content when it's needed instead of on construction. That's all.

All the logic for reading the per-directory configuration and constructing derived Rulesets and Configs can then be in the filter.

We've been using this since December in production in our monorepo, and it has worked really well. The main use for us has been to configure the properties of an i18n-related sniff with different text domains for different projects. We've also found it easier to use than the normal syntax when we need to disable a rule in a subdirectory.

sdixon194:
@gsherwood Just wondering if there's any further thoughts on this moving forward, or if there's any other info we can provide in the meanwhile.

PR #3380, opened 2021-06-21
GH Actions: set error reporting to E_ALL

jrfnl:
Turns out the default setting for `error_reporting` used by the SetupPHP action is `error_reporting=E_ALL & ~E_DEPRECATED & ~E_STRICT` and `display_errors` is set to `Off`.

For the purposes of CI, I'd recommend running with `E_ALL` and `display_errors=On` to ensure **all** PHP notices are shown.

Includes minor fix to the documentation of that part of the script.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,15 +51,14 @@ jobs:
       - name: Setup ini config
         id: set_ini
         run: |
-          # On stable PHPCS versions, allow for PHP deprecation notices.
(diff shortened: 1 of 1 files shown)


gsherwood:
Nice find. Thanks a lot for sorting this out.

jrfnl:
Yeah, this was an interesting one to debug... had two repos which were locally showing failures on PHP 8.1, but not in CI... once I started investigating and asking questions, this was the result.

PR #3381, opened 2021-06-21
PHP 8.1: Generic/ConstructorName: fix deprecation notices

jrfnl:
On PHP 8.1, passing `null` to `strtolower()` generates a `strtolower(): Passing null to parameter #1 ($string) of type string is deprecated` notice.

The `File::getDeclarationName()` will return `null` for anonymous classes, so we need an extra safeguard here.

This fix is covered by the existing unit tests and was exposed when running the tests on PHP 8.1 with `error_reporting` set to `E_ALL`. (see the GH actions script output for #3380)


gsherwood:
Thanks a lot for fixing this

PR #3396, opened 2021-07-23
PHP 8.0 | PEAR/FunctionComment: prevent false positives with attributes

jrfnl:
PHP 8.0+ attributes can be placed between a docblock and the function declaration it applies to.

The `PEAR.Commenting.FunctionComment` sniff - and by extension the `Squiz.Commenting.FunctionComment` sniff - did not yet take this into account.

This would result in a false positive `Missing doc comment for function ... ` error.

Resolving that error though, would result in a new `There must be no blank lines after the function comment` error, so the blank line check also needed to be fixed.

With the current fix, blank lines between the docblock and the function declaration are still not allowed, but non-blank lines between the two (i.e. lines containing attributes) will be ignored.

Only one "no blank lines between" error will be thrown per function declaration.

### DIFF
--- a/src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php
@@ -68,11 +68,25 @@ class FunctionCommentSniff implements Sniff
             return;
         }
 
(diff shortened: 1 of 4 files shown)


gsherwood:
Thanks for fixing this

PR #3397, opened 2021-07-23
PHP 8.0 | Squiz/InlineComment: prevent false positives with attributes

jrfnl:
PHP 8.0+ attributes can be placed between a docblock and the function/class declaration it applies to.

The `Squiz.Commenting.InlineComment` sniff did not yet take this into account when determining whether a comment was a docblock or an single line comment incorrectly using the docblock syntax.

This would result in false positive `Inline doc block comments are not allowed; use "/* Comment */" or "// Comment" instead` errors.

Fixed now.


gsherwood:
Thanks for fixing this

PR #3398, opened 2021-07-23
PHP 8.0 | Squiz/BlockComment: prevent false positives with attributes

jrfnl:
PHP 8.0+ attributes can be placed between a docblock and the function/class declaration it applies to.

The `Squiz.Commenting.BlockComment` sniff did not yet take this into account when determining whether a comment was a docblock or an block comment incorrectly using the docblock syntax.

This would result in false positive `Block comments must be started with /*` errors.

Fixed now.


gsherwood:
Thanks for fixing this

PR #3399, opened 2021-07-23
PHP 8.0 | PEAR/VariableComment: prevent false positives with attributes

jrfnl:
PHP 8.0+ attributes can be placed between a docblock and the property declaration it applies to.

The `Squiz.Commenting.VariableComment` sniff did not yet take this into account.

This would result in a false positive `Missing member variable doc comment` error.

Fixed now.


gsherwood:
Thanks for this fix. I found a related one in the Squiz MemberVarSpacing sniff which I also fixed.

jrfnl:
@gsherwood Ah, yes there are probably still more. I looked through and fixed all sniffs in the various `Commenting` categories now, triggered by PR #3400.

Issue #3395, opened 2021-07-23
Refactor/modernize test suite setup


jrfnl:
As I've been seeing new issues popping up elsewhere with some of the recent changes in PHP 8.1, I had a look at the latest test run for this repo and as I feared, the tests break on PHP 8.1 due to PHPUnit 7.5 being used.

See: https://github.com/squizlabs/PHP_CodeSniffer/runs/3118862021?check_suite_focus=true

As PHPUnit 7.5 is no longer supported, this will not be fixed on the PHPUnit side of things.

I think a refactor of the test suite setup is in order to allow it to be compatible with a wider range of PHPUnit versions, preferably without BC-break in the setup as quite a lot of external standards use the PHPCS native abstract test classes for their tests as well.

jrfnl:
While the PHP 8.1 issues have been solved by now, I'm going to leave this issue open, as IMO a refactor of the test suite setup may still be a good idea and make the setup more sustainable in the long run.

The setup as is, with the `AllTests`, the dynamic test suite creation etc seems to be based on a really old PHPUnit version and while it still works, the framework as-is, is creaking more and more with each new PHP version.

jrfnl:
As things are at this time, if PHPCS wants to allow running tests on PHPUnit 10.x (which it should), at least two things needs to happen (maybe more):
1. Rename abstract base test cases. Those classes are no longer allowed to have the `Test` suffix.
    I'd recommend renaming the suffix from `UnitTest` to `TestCase`.
    See: https://github.com/sebastianbergmann/phpunit/issues/5132
2. Changing the test suite set up to remove the use of the old style set up with `AllTests.php` and `AllSniffs.php` containing a static `suite` method, which is no longer supported.
    > PHPUnit no longer invokes a static method named suite on a class that is declared in a file that is passed as an argument to the CLI test runner
    Ref: https://github.com/sebastianbergmann/phpunit/blob/10.0.19/ChangeLog-10.0.md#1000---2023-02-03 (Changed section)
    Also related:
    * https://github.com/sebastianbergmann/phpunit/issues/3213
    * https://github.com/sebastianbergmann/phpunit/issues/4460

If losing the line created via the bootstrap `printPHPCodeSnifferTestOutput()` method would not be considered a big deal, changing things round to remove the `AllTests` and `AllSniffs` files and letting PHPUnit self-discover the tests in a test suite would probably work.
[...]

PR #3400, opened 2021-07-23
PHP 8.1: prevent deprecation notices about missing return types

jrfnl:
A plain run against PHP 8.1.0-beta1 currently results in the following deprecation notices:
```
Deprecated: Return type of PHP_CodeSniffer\Files\FileList::current() should either be compatible with Iterator::current(): mixed, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in path/to/PHP_CodeSniffer/src/Files/FileList.php on line 186
Deprecated: Return type of PHP_CodeSniffer\Files\FileList::next() should either be compatible with Iterator::next(): void, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in in path/to/PHP_CodeSniffer/src/Files/FileList.php on line 217
Deprecated: Return type of PHP_CodeSniffer\Files\FileList::key() should either be compatible with Iterator::key(): mixed, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in in path/to/PHP_CodeSniffer/src/Files/FileList.php on line 204
Deprecated: Return type of PHP_CodeSniffer\Files\FileList::valid() should either be compatible with Iterator::valid(): bool, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in in path/to/PHP_CodeSniffer/src/Files/FileList.php on line 230
Deprecated: Return type of PHP_CodeSniffer\Files\FileList::rewind() should either be compatible with Iterator::rewind(): void, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in in path/to/PHP_CodeSniffer/src/Files/FileList.php on line 173
Deprecated: Return type of PHP_CodeSniffer\Files\FileList::count() should either be compatible with Countable::count(): int, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in in path/to/PHP_CodeSniffer/src/Files/FileList.php on line 247
Deprecated: Return type of PHP_CodeSniffer\Filters\Filter::getChildren() should either be compatible with RecursiveFilterIterator::getChildren(): ?RecursiveFilterIterator, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in in path/to/PHP_CodeSniffer/src/Filters/Filter.php on line 135
Deprecated: Return type of PHP_CodeSniffer\Filters\Filter::accept() should either be compatible with FilterIterator::accept(): bool, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in path/to/PHP_CodeSniffer/src/Filters/Filter.php on line 93
```

The run will still succeed.

These deprecation notices relate to the [Return types for internal methods RFC](https://wiki.php.net/rfc/internal_method_return_types) in PHP 8.1 and in particular, the change made in [PHP PR #7239](https://github.com/php/php-src/pull/7239), which adds return types to the various SPL Iterator interface methods.

Basically, as of PHP 8.1, these methods in classes which implement the `Iterator` interface are expected to have a return type declared.
The return type can be the same as used in PHP itself or a more specific type. This complies with the Liskov principle of covariance, which allows the return type of a child overloaded method to be more specific than that of the parent.

The problem with this is that return types were only introduced in PHP 7.0 and therefore cannot be used as PHP_CodeSniffer 3.x has a minimum PHP version of 5.4.

Luckily an attribute has been added to silence the deprecation warning.

While attributes are a PHP 8.0+ feature, due to the choice of the `#[]` syntax, in PHP < 8.0, attributes will just be ignored and treated as comments, so there is no drawback to using the attribute.

**Note**: fixing these issues by adding the attributes exposed problems with attributes in the various `Commenting` sniffs. Fixes for those have been pulled separately.


jrfnl:
Note: due to the issues in the commenting sniffs, the builds for this PR will fail on the CS check not being able to handle the attributes correctly.

PR #3396, #3397 and PR #3398 should be merged first. Once those are merged, this PR can be rebased and the builds will pass (well, all except for PHP 8.1, but that's a different matter - see #3395 )

jrfnl:
Rebased now the attribute/comment PRs have been merged.

jrfnl:
Note: the current failures still on PHP 8.1 are unrelated to this PR and the build against this PR shows that the slew of deprecation notices at the start of a run is now gone, which is exactly what this PR is about.

afilina (approved):

PR #3401, opened 2021-07-25
Generic/One*PerFile sniffs: efficiency tweak

jrfnl:
Each of these sniffs would start walking the rest of the file from the token next to the OO keyword to the end of the file in search of the next (class/interface/trait) keyword.

As nested OO structure declarations (with the exception of anonymous classes, but those are outside the realm of these sniffs) are not supported in PHP, we can make these sniffs a lot faster by starting the search _after_ the scope closer for the current OO structure.

### DIFF
--- a/src/Standards/Generic/Sniffs/Files/OneClassPerFileSniff.php
+++ b/src/Standards/Generic/Sniffs/Files/OneClassPerFileSniff.php
@@ -39,7 +39,13 @@ class OneClassPerFileSniff implements Sniff
      */
     public function process(File $phpcsFile, $stackPtr)
     {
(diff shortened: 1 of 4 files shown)


jrfnl:
Note: the build failure on PHP 8.1 is unrelated to this PR. PRs #3394 and #3400 need to be merged to solve the runtime issues, the tokenizer issues are a whole other matter.

jrfnl:
Rebased to get a passing build on PHP 8.1.

gsherwood:
Sorry about missing this - was a nice easy merge and a good change.

jrfnl:
No worries and thanks for getting it merged.

PR #3411, opened 2021-08-16
PHP 8.1: fix retokenization of "&" character

jrfnl:
In PHP < 8.1, the ampersand was tokenized as a simple token, basically just a plain string "&".
As of PHP 8.1, due to the introduction of intersection types, PHP is introducing two new tokens for the ampersand.

This PR proposes to "undo" the new PHP 8.1 tokenization of the ampersand in favour of the pre-existing tokenization of the character as `T_BITWISE_AND` as it has been in PHPCS since forever.

This fixes a lot of current test failures on PHP 8.1, though not all as there are more tokenizer changes which still need to be accounted for.

We may want to consider adding an extra `'is_reference'` array key index to the token array for these tokens, which would allow the `File::isReference()` method to resolve tokens on PHP 8.1 much more quickly and more easily.

We also may want to have a think about whether we want to move to the PHP 8.1 tokenization in PHPCS 4.x. All the same, this PR should not be held back by a decision like that as, for now, it just needs to be fixed for PHPCS 3.x.

Props also to @afilina (Anna Filina) for doing this together with me during pair programming.


jrfnl:
Related PHP RFC link: https://wiki.php.net/rfc/pure-intersection-types and the related PR: https://github.com/php/php-src/pull/6799

afilina (approved):

PR #3417, opened 2021-08-27
Modification so that the file path is read

jhonymiler:
This modification was the simplest way I found to get php_codesniffer to work on my computer. I had a problem where the call to the file was being made with
--stdin-path=

but the correct should be this way:
C:\user\my-user\vendor\bin\phpcs.bat --report=json -q --encoding=UTF-8 --standard=PSR1 --error-severity=5 --warning-severity=5 C :\xampp\htdocs\file.php

Since I couldn't find the place to modify this argument, I found an XGH way to work.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -962,7 +962,13 @@ class Config
                 if ($this->stdinPath === false) {
                     $this->stdinPath = trim(substr($arg, 11));
                 }
-
+                /**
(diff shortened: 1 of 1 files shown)


gsherwood:
As @jrfnl has said, stdin path is not guaranteed to be a real file path and should not be processed as one, so closing this.

PR #3426, opened 2021-09-05
GH Actions: various tweaks

jrfnl:
### GH Actions: minor matrix simplification

We can remove the `experimental` key in favour of checking against the PHP version in `continue-on-error`.

### GH Actions: stablelize the error_reporting

What is included in `E_ALL` can vary across PHP versions. Setting `error_reporting` to `-1` will ensure ALL errors are always shown, no matter what.

### GH Actions: minor tweaks to the conditions

The `matrix.php` setting is a text string, so let's compare against it as a text string.


gsherwood:
Thanks for these changes - all made perfect sense

PR #3429, opened 2021-09-06
Generic/[Upper|Lower]CaseConstant: ignore type declarations

### LINKED ISSUE #3332: Generic.PHP.UpperCaseConstant: don't check null and false types
**Describe the bug**
Sniff `Generic.PHP.UpperCaseConstant` checks all occurences of FALSE/NULL even if they are used in type.

**Code sample**
```php
function debug(string|null $message): void|false {
        echo "DEBUG: $message\n";
}
```

**Custom ruleset**
```xml
<?xml version="1.0"?>
  <rule ref="Generic.PHP.UpperCaseConstant" />
</ruleset>
```

**To reproduce**
2. Run `phpcs test.php ...`
```
-------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------
 3 | ERROR | [x] TRUE, FALSE and NULL must be uppercase; expected "NULL" but found "null"
 3 | ERROR | [x] TRUE, FALSE and NULL must be uppercase; expected "FALSE" but found "false"
```

**Expected behavior**
No error is reported for null/false used as type.

 - OS: Ubuntu Linux 20.10
 - PHP: 8.0.5
 - PHPCS: 3.6.0
 - Standard: custom

jrfnl:
This commit adjusts the `Generic.PHP.UpperCaseConstant` and the `Generic.PHP.LowerCaseConstant` sniffs to ignore property, parameter and return type declarations.

Type declarations have their own rules and should not be treated the same as other `false`, `true` and `null` uses.

Includes unit tests.

Fixes #3332


jrfnl:
Rebased to get a passing build on PHP 8.1.

gsherwood:
Thanks for changing these sniffs. I ended up having upper extend lower to remove much of the duplicate logic and to force these to always be in sync.

jrfnl:
> Thanks for changing these sniffs. I ended up having upper extend lower to remove much of the duplicate logic and to force these to always be in sync.

Makes sense as these sniffs will (should) never be used within the same standard as they contradict each other.

I've become a bit wary of sniffs extending each other after seeing too many problems with that in external standards, but in this case it shouldn't be problematic.

PR #3433, opened 2021-09-11
FunctionCommentThrowTagSniff: added support to multiple class definition in @throws

WinterSilence:
Adds support to multiple definition like as `@throws Exception|Exception2`


jrfnl:
@WinterSilence Just wondering: isn't it more common to have multiple `@throws` tags if multiple exceptions are thrown ? Why would you use `Type1|Type2` ?
Also see: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#514-throws

WinterSilence:
@jrfnl 1. it's not PSR standard and multiple definition is allowed 2. sniff not support `@throws Exception @throws Exception`

PR #3435, opened 2021-09-11
Update example in class `AbstractScopeSniff`

WinterSilence:


jrfnl (approved):

jrfnl (changes requested):
Hang on.. I approved the original change, but I don't approve of the current version. While some of the current changes are okay, some are not.

The example code now given no longer matches the textual documentation above it, so this can't go in as-is.

WinterSilence:
@jrfnl I'm missed abstract method `processTokenOutsideScope()` in first commit, current example don't contain any extra code.

WinterSilence (commented):
Add missed method processTokenOutsideScope

PR #3442, opened 2021-09-30
GH Actions: split PHAR build test from other tests

jrfnl:
As things were in the CI script, the `phpcs` and `phpcbf` Phars would be build against each supported PHP version and the tests run with the `phpcs` Phar would use the Phar as build on the same PHP version as the test was being run on.

This test was not representative of the reality as with each release, only one `phpcs` and one `phpcbf` phar is released.

The changes in this commit:
* Split the building of the Phar off from the main test script.
* This new `build` job:
    - Tests building the Phar against all supported PHP versions as an early detection system for problems in the Phar extension/Phar building script.
    - Uploads both phars as build against PHP 7.4.
        The uploaded Phars match the Phars as would be included on a release as they are build against the same PHP version as used for releases.
        **_These Phars will now also be available for PRs and merges to `master` to allow for easier testing of change proposals by issue reporters who may not have a git clone of the repo._**
        The uploaded Phars will be available for 28 days (default 90).
    - Does a cursory test with both the `phpcs.phar` as well as the `phpcbf.phar` to test that the build phar files are _functional_.
* In the `test` job, which now depends on the `build` job, the Phar will now no longer be build, but the Phar as uploaded in the `build` job - i.e. the Phar build against the same PHP version as used when releasing Phars - is downloaded and used to run the Phar test.

The uploaded Phar files can be found on the "Summary" page of the `test` workflow after the build has finished for each build run.

---

### 🆕 GH Actions: change the building of the (release) PHARs to PHP 8.0

The PHP native `php_strip_whitespace()` function used in the PHAR build script is PHP version sensitive.

The function strips comments and whitespace out of the code, but in PHP 7.4, attributes are seen as comments due to the `#[]` syntax, which means that if the PHAR files were being generated on PHP 7.4, the PHP 8.1 `#[ReturnTypeWillChange]` attributes as put in via PR #3400 would be stripped out.

In other words: for the PHAR files to be cross-version compatible, they *MUST* be generated on PHP 8.0 or higher.

This fixes this for the CI part of things. This partially addresses #3497 .

Ref: https://www.php.net/manual/en/function.php-strip-whitespace.php

---

### 🆕 Build Phar: show more debug information

* Number of files added for each PHAR.
* Total time the script took to run.

### 🆕 Build Phar: add custom "strip whitespace and comments" function

This implements option 3 as per the discussion in the thread for PR #3442.

A custom `stripWhitespaceAndComments()` function has been created and added to the script used to build the PHAR files and replaces the use of the PHP native `php_strip_whitespace()` function, which does not allow for creating PHP cross-version compatible PHAR files if attributes are used anywhere in the code.

The problem with the PHP native `php_strip_whitespace()` function is this:
* When run on PHP <= 7.4, attributes would be stripped from the code as they are seen as `#` comments. This undoes the deprecation silencing for methods for which no return type can be added yet (as needed for full PHP 8.1 compatibility).
* When run on PHP >= 8.0, attributes are _not_ stripped, but recognized correctly, however, the function strips **all** new lines, turning the file effectively into one long line of code. This is problematic when the PHAR would subsequently be run on PHP < 8.0, as any code after the first attribute would then be seen as "commented out", leading to the PHAR not running with a parse error.

To solve this, the new `stripWhitespaceAndComments()` function emulates the PHP native function, with two important differences:
* New lines are very selectively left in the regenerated content of the files. By ensuring that there is always a new line after an attribute closer, we can prevent code from being seen as commented out in PHP < 8.0.
* As the PHPCS native PHP tokenizer is used to interpret the file content, the token stream will be the same PHP cross-version, meaning that attributes will be recognized on all supported PHP versions and the script can now be run again on any supported PHP version and the generated PHAR files will be the same.

As an additional performance tweak, `xml` files will no longer be passed to the whitespace/comment stripping. This had no effect previously and as XML files would tokenize as 100% `T_INLINE_HTML`, passing these to the new function would have no effect either (other than slowing down the script).
Same goes for the `license.txt` file.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Upload the PHPCS phar
         uses: actions/upload-artifact@v2
-        if: ${{ success() && matrix.php == '7.4' }}
+        if: ${{ success() && matrix.php == '8.0' }}
         with:
           name: phpcs-phar
           path: ./phpcs.phar
@@ -49,7 +49,7 @@ jobs:
 
       - name: Upload the PHPCBF phar
         uses: actions/upload-artifact@v2
-        if: ${{ success() && matrix.php == '7.4' }}
+        if: ${{ success() && matrix.php == '8.0' }}
         with:
           name: phpcbf-phar
           path: ./phpcbf.phar
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,8 +13,60 @@ on:
   workflow_dispatch:
 
 jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php: ['5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2']
+
+    name: "Build Phar on PHP: ${{ matrix.php }}"
+
+    continue-on-error: ${{ matrix.php == '8.2' }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+          ini-values: phar.readonly=Off, error_reporting=-1, display_errors=On
+
+      - name: Build the phar
+        run: php scripts/build-phar.php
+
+      - name: Upload the PHPCS phar
+        uses: actions/upload-artifact@v2
+        if: ${{ success() && matrix.php == '7.4' }}
+        with:
+          name: phpcs-phar
+          path: ./phpcs.phar
+          if-no-files-found: error
+          retention-days: 28
+
+      - name: Upload the PHPCBF phar
+        uses: actions/upload-artifact@v2
+        if: ${{ success() && matrix.php == '7.4' }}
+        with:
+          name: phpcbf-phar
+          path: ./phpcbf.phar
+          if-no-files-found: error
+          retention-days: 28
+
+      # Both the below only check a few files which are rarely changed and therefore unlikely to have issues.
+      # This test is about testing that the phars are functional, *not* about whether the code style complies.
+      - name: 'PHPCS: check code style using the Phar file to test the Phar is functional'
+        run: php phpcs.phar ./scripts
+
+      - name: 'PHPCBF: fix code style using the Phar file to test the Phar is functional'
+        run: php phpcbf.phar ./scripts
+
   test:
     runs-on: ubuntu-latest
+    needs: build
 
     strategy:
       # Keys:
@@ -45,11 +97,11 @@ jobs:
           # Set the "short_open_tag" ini to make sure specific conditions are tested.
           # Also turn on error_reporting to ensure all notices are shown.
           if [[ ${{ matrix.custom_ini }} == true && "${{ matrix.php }}" == '5.5' ]]; then
-            echo '::set-output name=PHP_INI::phar.readonly=Off, error_reporting=-1, display_errors=On, date.timezone=Australia/Sydney, short_open_tag=On, asp_tags=On'
+            echo '::set-output name=PHP_INI::error_reporting=-1, display_errors=On, date.timezone=Australia/Sydney, short_open_tag=On, asp_tags=On'
           elif [[ ${{ matrix.custom_ini }} == true && "${{ matrix.php }}" == '7.0' ]]; then
-            echo '::set-output name=PHP_INI::phar.readonly=Off, error_reporting=-1, display_errors=On, date.timezone=Australia/Sydney, short_open_tag=On'
+            echo '::set-output name=PHP_INI::error_reporting=-1, display_errors=On, date.timezone=Australia/Sydney, short_open_tag=On'
           else
-            echo '::set-output name=PHP_INI::phar.readonly=Off, error_reporting=-1, display_errors=On'
+            echo '::set-output name=PHP_INI::error_reporting=-1, display_errors=On'
           fi
 
       - name: Install PHP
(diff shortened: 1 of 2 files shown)


jrfnl:
Grr... sometimes.... I _hate_ PHP...

Okay, the reason the builds are failing is because all the whitespace + comments are being stripped, in other words: each file is now just one big line of code.

While that is fine on PHP 8.0 where attributes are a recognized syntax, for PHP < 8.0, this means that all code after an attribute is seen as a comment.....

Unfortunately the `php_strip_whitespace()` function doesn't have an option to _not_ strip new lines from the code, so now we have a conundrum.

* Either the use of the `php_strip_whitespace()` function has to be removed and then the PHAR can be build on any PHP version and will run fine on any PHP version, but the PHAR will be significantly larger;
* Or we need to create separate PHAR files for PHP 8.0+ and PHP < 8.0.
* Or we need to create a custom version of `php_strip_whitespace()` which strips comments, but does not remove new lines.
* Or we move the PHAR building to a dedicated tool like [humbug/box](https://github.com/box-project/box), which already has such a custom function, though in that case, we'd still need to generate the PHARs on PHP 8.0 for the time being as the last release of Box does not yet contain a fix for the same issue (attributes being stripped) which was also present in the custom function code in Box.
    Also see: https://github.com/box-project/box/issues/567 and https://github.com/box-project/box/pull/568
    
@gsherwood Which solution would you prefer ?

jrfnl:
Update: `humbug/box` has just released [version 3.14.0](https://github.com/box-project/box/releases/tag/3.14.0) which includes the fixes needed to generate the PHAR on any PHP version and allow it to work on any PHP version (within the confines of a project PHP version constraints).

gsherwood:
If I rebuild the phars without whitespace and comments stripped, the file size of each increases from 1.3M to 2.4M. I don't think that's a disaster, so that's probably the short term plan. I will re-upload the 3.6.2 phars so they work on 8.1.

Longer term:
I looked at the box project and it's tokenzing the contents looking for attributes. PHPCS already does the same, so will likely play around with just getting PHPCS to tokenize the files and rebuild them while removing the bits we don't need. Should provide enough control over compression and I'm already very confident that PHPCS can tokenize it's own files correctly.

gsherwood:
> I will re-upload the 3.6.2 phars so they work on 8.1

That's done now.

jrfnl:
> Longer term:
> I looked at the box project and it's tokenzing the contents looking for attributes. PHPCS already does the same, so will likely play around with just getting PHPCS to tokenize the files and rebuild them while removing the bits we don't need. Should provide enough control over compression and I'm already very confident that PHPCS can tokenize it's own files correctly.

Okay, sounds like, option 3 it is.

Also sounds like you're already looking into this yourself ?

gsherwood:
> Also sounds like you're already looking into this yourself ?

I will eventually, but I'll keep building the phars on 8.1 until I do so nothing is blocked by this change.

jrfnl:
> I will eventually, but I'll keep building the phars on 8.1 until I do so nothing is blocked by this change.

I'd be happy to prepare the changes and add them to this PR if you like ?

gsherwood:
> I'd be happy to prepare the changes and add them to this PR if you like ?

I haven't started anything, so if you're able to give it a go I'd be very grateful.

jrfnl:
> I haven't started anything, so if you're able to give it a go I'd be very grateful.

I may have some time to dig into this next week. I do expect a PHPCS native solution to be significantly slower than using Box, as the tokenizer processing which would be done by the PHPCS tokenizer has a lot of overhead which isn't needed for this task.

gsherwood:
> I do expect a PHPCS native solution to be significantly slower than using Box, as the tokenizer processing which would be done by the PHPCS tokenizer has a lot of overhead which isn't needed for this task.

I expect the same, but I'm not that concerned given the nature of the task.

jrfnl:
Hmm.. my brain was working it out already, so got a working proof of concept. Want to do some more testing with it, but I expect to add the extra commits to this PR tomorrow.

jrfnl:
I've updated the PR with the additional changes. The second commit (changing the artifact build PHAR to PHP 8.0) is no longer needed, but also doesn't do any harm.

The changes which were already in this PR to change the GH Actions tasks related to the PHARs will safeguard the changes made to the build script and demonstrate that the proposed fix is working correctly.

I've also done some test runs to benchmark the changes using PHP 7.4.26:

#### Original script as per `3.6.2` + debug commit (problematic on PHP 8.1)
```
Building phpcs phar
        => /path/to/PHP_CodeSniffer/phpcs.phar
        ** file exists, removing **
        => adding files... done
           Added 439 files
        => adding stub... done
Building phpcbf phar
        => /path/to/PHP_CodeSniffer/phpcbf.phar
        ** file exists, removing **
        => adding files... done
           Added 439 files
        => adding stub... done
DONE in 28.77 secs

Filesize generated phpcs.phar file: 1333949 bytes
Filesize generated phpcs.phar file: 1333953 bytes
```

#### Current script on `master` + debug commit (yields double-size PHAR files)
```
Building phpcs phar
        => /path/to/PHP_CodeSniffer/phpcs.phar
        ** file exists, removing **
        => adding files... done
           Added 439 files
        => adding stub... done
Building phpcbf phar
        => /path/to/PHP_CodeSniffer/phpcbf.phar
        ** file exists, removing **
        => adding files... done
           Added 439 files
        => adding stub... done
DONE in 63.18 secs

Filesize generated phpcs.phar file: 2514695 bytes
Filesize generated phpcs.phar file: 2514699 bytes
```

#### Version of the fix currently proposed where new lines would not be removed at all (still larger file size than necessary)
```
Building phpcs phar
        => /path/to/PHP_CodeSniffer/phpcs.phar
        ** file exists, removing **
        => adding files... done
           Added 439 files
        => adding stub... done
Building phpcbf phar
        => /path/to/PHP_CodeSniffer/phpcbf.phar
        ** file exists, removing **
        => adding files... done
           Added 439 files
        => adding stub... done
DONE in 53.08 secs

Filesize generated phpcs.phar file: 1917725 bytes
Filesize generated phpcs.phar file: 1917729 bytes
```

#### Script with the fix as pulled
```
Building phpcs phar
        => /path/to/PHP_CodeSniffer/phpcs.phar
        ** file exists, removing **
        => adding files... done
           Added 439 files
        => adding stub... done
Building phpcbf phar
        => /path/to/PHP_CodeSniffer/phpcbf.phar
        ** file exists, removing **
        => adding files... done
           Added 439 files
        => adding stub... done
DONE in 34.86 secs

Filesize generated phpcs.phar file: 1387969 bytes
Filesize generated phpcs.phar file: 1387973 bytes
```

All in all, filesize is very similar to what it originally was and runtime-wise, the impact is minimal.

gsherwood:
Thanks so much. I love everything about this PR - the whole Phar build and test process is so much cleaner and better.

jrfnl:
Thanks for getting this merged @gsherwood. With these changes in place we now have an extra safeguard in CI for the PHARS as released not being broken and the PHAR artifacts being available with each PR build should also make it easier for reporters to test PRs intened to fix their reported issue.

PR #3446, opened 2021-10-06
Tokenizer/PHP: fix ? tokenization after attribute

### LINKED ISSUE #3445: Nullable parameter after attribute incorrectly tokenized as ternary operator
**Describe the bug**
On a method with a PHP 8 attribute before a nullable parameter, cs if forcing to have a space after the ? of a nullable the nullable parameter

**Code sample**
```php
public function setDefault(#[ImportValue(
        column: 'Material by default',
        transformer: YesNoBooleanTransformer::class,
        constraints: [
            [
                Assert\Type::class,
                ['type' => 'bool'],
            ],
        ]
    )] ?bool $value = null): void
    {
        $this->setBool('default', $value);
    }
```

**To reproduce**
2. Run `phpcs test.php ...`
```
 451 | ERROR | [x] Expected at least 1 space after "?"; 0 found
 451 | ERROR | [x] Parameter $value has null default value, but is not marked as nullable.
     |       |     (SlevomatCodingStandard.TypeHints.NullableTypeForNullDefaultValue.NullabilityTypeMissing)
```

**Expected behavior**
We should be able to put a nullable parameter after a PHP 8 parameter attribute. Here cs is confusing the PHP8 attribute and the bool as a ternary operator, thus forcing the space after the "?" causing the "Parameter $value has null default value" because this work fine : 

```
public function setDefault(#[ImportValue(
        column: 'Material by default',
        transformer: YesNoBooleanTransformer::class,
        constraints: [
            [
                Assert\Type::class,
                ['type' => 'bool'],
            ],
        ]
    )] ? bool $value): void
    {
        $this->setBool('default', $value);
    }
```

 - PHP: 8.0
 - PHPCS: 3.5
 - Standard: PSR2, PSR12, MySource, PSR1, Squiz, PEAR, Zend, Doctrine and SlevomatCodingStandard

jrfnl:
In a select set of circumstance, when a `?` would follow an attribute with a nested array in it, the `?` would be tokenized as `T_INLINE_THEN`, not `T_NULLABLE`.
Fixed now.

Includes unit test via the `PSR12.Operators.OperatorSpacing` sniff.

Note: Ternary `?` vs nullable (and ternary `:` vs colon) should really get a full set of unit tests, but I don't currently have the time to set that up.

Fixes #3445

### DIFF
--- a/src/Standards/PSR12/Tests/Operators/OperatorSpacingUnitTest.inc
+++ b/src/Standards/PSR12/Tests/Operators/OperatorSpacingUnitTest.inc
@@ -63,3 +63,15 @@ $fn = fn(array & $one) => 1;
 $fn = static fn(DateTime $a, DateTime $b): int => -($a->getTimestamp() <=> $b->getTimestamp());
 
 function issue3267(string|int ...$values) {}
+
+function setDefault(#[ImportValue(
+        constraints: [
+            [
+                Assert\Type::class,
+                ['type' => 'bool'],
+            ],
+        ]
+    )] ?bool $value = null): void
+    {
+        // Do something
+    }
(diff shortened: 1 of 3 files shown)


jrfnl:
Felt like an icky one... an thanks for including it in 3.6.1!

PR #3447, opened 2021-10-06
Squiz/ScopeClosingBrace: bug fix - prevent removing inline HTML when fixing

### LINKED ISSUE #3422: Squiz.WhiteSpace.ScopeClosingBrace fixer removes HTML content when fixing closing brace alignment
**Describe the bug**
Div end tag removed when reformatting using phpcbf to reformat code to PSR-12 standard

**Code sample**
```php
// before reformatter
    </div> <?php endif; ?>

// after reformatter
    <?php endif; ?>
```

**Custom ruleset**
```xml
<?xml version="1.0"?>
  <description>If you are using a custom ruleset, please enter it here.</description>

  <description>Besides the below, have some exclude patterns that shouldn't affect the outcode.</description>

    <rule ref="PSR12"/>
    <arg name="extensions" value="php" />

</ruleset>
```

**To reproduce**
2. Run `php phpcbf.phar --standard=ruleset.xml test.php -v `
3. See code reformatted with the end div removed

**Expected behavior**
Should not remove HTML

 - OS:  MacOS 11.5.2
 - PHP: 8.0.2
 - $ php phpcbf.phar --version
PHP_CodeSniffer version 3.6.0 (stable) by Squiz (http://www.squiz.net)
 - Standard:  PSR12

jrfnl:
When non-empty inline HTML is on the same line, before the scope closing, the sniff would incorrectly throw the `Closing brace indented incorrectly; expected %s spaces, found %s` error, instead of the `Closing brace must be on a line by itself` error.

This would ultimately lead to the fixer removing the non-empty inline HTML, potentially breaking HTML display/structure.

Fixed now. Includes unit test.

Fixes #3422

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/ScopeClosingBraceSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/ScopeClosingBraceSniff.php
@@ -65,11 +65,20 @@ class ScopeClosingBraceSniff implements Sniff
 
         // Check that the closing brace is on it's own line.
         $lastContent = $phpcsFile->findPrevious([T_INLINE_HTML, T_WHITESPACE, T_OPEN_TAG], ($scopeEnd - 1), $scopeStart, true);
-        if ($tokens[$lastContent]['line'] === $tokens[$scopeEnd]['line']) {
+        for ($lineStart = $scopeEnd; $tokens[$lineStart]['column'] > 1; $lineStart--);
+
+        if ($tokens[$lastContent]['line'] === $tokens[$scopeEnd]['line']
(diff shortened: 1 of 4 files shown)


gsherwood:
Thanks a lot for fixing this. Assuming inline HTML is always empty was not some of my best work...

jrfnl:
> Thanks a lot for fixing this. Assuming inline HTML is always empty was not some of my best work...

Well, the original sniff is pretty ancient (from the initial import), so from well before the fixer functionality was available, so not that surprising that something like this slipped through the cracks.

What **_is_** surprising is that no other bug reports about this had been filed in the seven years since the fixers got added 😂

PR #3449, opened 2021-10-12
Fix deprecation for PHP 8.1

### LINKED ISSUE #3448: PHP 8.1 deprecation notice while generating running time value
**Describe the bug**
Running sniffer on [Drupal core](https://www.drupal.org/project/drupal/issues/3242889) I'm getting 
```
PHP Deprecated:  Implicit conversion from float 235309.85403060913 to int loses precision in /var/www/html/web/vendor/squizlabs/php_codesniffer/src/Util/Timing.php on line 67

Deprecated: Implicit conversion from float 235309.85403060913 to int loses precision in /var/www/html/web/vendor/squizlabs/php_codesniffer/src/Util/Timing.php on line 67
Time: 3 mins, 55.31 secs; Memory: 158MB
```

**Code sample**
No idea, just a precision https://php.watch/versions/8.1/deprecate-implicit-conversion-incompatible-float-string
```php
$t=152815.288066864; 
var_dump(((int)$t % 60000));
var_dump(($t % 60000));
```

**To reproduce**
1. Sniffers should run longer 60 seconds
2. https://3v4l.org/flZrS
```
Deprecated: Implicit conversion from float 152815.288066864 to int loses precision in .../vendor/squizlabs/php_codesniffer/src/Util/Timing.php on line 67
```

**Expected behavior**
Clean output

 - OS: Ubuntu 21.10
 - PHP: 8.1.0 rc3
 - PHPCS: 3.6.1
 - Standard: few sets

**Additional context**
I think this just needs to use different method to get integer part of timings

andypost:
Closes #3448

With this fix I got clean reports

Steps https://3v4l.org/flZrS

### DIFF
--- a/src/Util/Timing.php
+++ b/src/Util/Timing.php
@@ -64,7 +64,7 @@ class Timing
 
         if ($time > 60000) {
             $mins = floor($time / 60000);
-            $secs = round((($time % 60000) / 1000), 2);
+            $secs = round((fmod($time, 60000) / 1000), 2);
             $time = $mins.' mins';
             if ($secs !== 0) {
                 $time .= ", $secs secs";


jrfnl (changes requested):
Okay, finally had a chance to have a closer look, but this fix is incorrect as it doesn't fix anything.

Division with floats works perfectly fine without deprecation notice in PHP 8.1, this line (73) is not the "problem" line.

The actual problem is in line 67:
```php
$secs = round((($time % 60000) / 1000), 2);
```

As per the [PHP manual on arithmetic operators](https://www.php.net/manual/en/language.operators.arithmetic.php):
> Operands of modulo are converted to int before processing. For floating-point modulo, see `fmod()`.

And as `fmod()` has been available since PHP 4.2.0, it can safely be used to maintain the behaviour as it was previously.
Ref: https://www.php.net/manual/en/function.fmod.php

In other words, line 67 should be changed to the below to solve the actual problem:
```php
$secs = round((fmod($time, 60000) / 1000), 2);
```

Also see:
https://3v4l.org/r3DZZ (without fix)
https://3v4l.org/9FelR (with correct fix)

andypost:
@jrfnl thank you for review, fixed feedback... it's strange how I mixed lines

jrfnl:
@andypost No worries, happens to the best of us, but it does very much explain my confusion when I initially looked at the PR.

jrfnl (approved):

gsherwood:
@jrfnl Thanks a lot for the review and guidance on the fix
@andypost Thanks for the bug report and PR

PR #3455, opened 2021-10-20
Auto-add fullstop to @throws if rule is used

pascalheidmann:
Automatic fix for cases like

```php
/**
 * @throws FoobarException throws
*/
```

will be

```php
/**
 * @throws FoobarException throws.
*/
```


gsherwood:
Thanks for the PR, but this is not the kind of auto-fix I like PHPCS to do as I can't be sure a full stop is required here. Changing comments is something I think is best left to manual code review.

PR #3457, opened 2021-10-21
PSR12/ClassInstantiation: bug fix for attributes with anonymous classes

### LINKED ISSUE #3456: PSR12.Classes.ClassInstantiation.MissingParentheses false positive using attributes on anonymous class
**Describe the bug**

When attributes are set on anonymous classes, phpcs inappropriately flag `PSR12.Classes.ClassInstantiation.MissingParentheses` violation

**Code sample**
```php
<?php

$class1 = new class {
    #[SomeAttribute('summary')]
    public const SOME_STUFF = 'foo';
};

$class2 = new #[SomeAttribute('summary')] class {
    public const SOME_STUFF = 'foo';
};
```

**Custom ruleset**

PSR12

**To reproduce**
2. Run `./vendor/bin/phpcs --standard=PSR12 test.php ...`
```
-------------------------------------------------------------------------
-------------------------------------------------------------------------
 8 | ERROR | [x] Parentheses must be used when instantiating a new class
-------------------------------------------------------------------------
-------------------------------------------------------------------------
```

**Expected behavior**

1. Should'nt report an error at Line 8
2. `phpcbf` shouldn't attempt to fix by producing invalid code

 - OS: Linux Fedora 34
 - PHP: 8.0
 - PHPCS: 3.6.1
 - Standard: PSR12

**Additional context**

phpcbf will produce invalid code:
```
./src/vendor/bin/phpcbf --standard=PSR12 test.php 

PHPCBF RESULT SUMMARY
/home/mvacelet/workspace/tuleap/test.php             1      0
A TOTAL OF 1 ERROR WERE FIXED IN 1 FILE
```
and resulting file:
```php
<?php

$class1 = new class {
    #[SomeAttribute('summary')]
    public const SOME_STUFF = 'foo';
};

$class2 = new() #[SomeAttribute('summary')] class {
    public const SOME_STUFF = 'foo';
};
```

jrfnl:
The `class` keyword was correctly tokenized as `T_ANON_CLASS`, even when there is an attribute between the `new` and the `class` keyword, however, the `PSR12.Classes.ClassInstantiation` sniff did not take that possibility into account, leading to false positives.

This commit:
* Adds code to skip over attributes.
    Note: I've elected to skip over attributes, but as these attributes AFAICS can only be added in combination with an anonymous class and anonymous classes are ignored by this sniff, it would also be a valid option to bow out of the sniff when an attribute is encountered. As this would be inconsistent with how the `T_ANON_CLASS` token was handled so far, I've not gone with that option.
* Adds some additional extra guard code to prevent future false positives for when the class name could not be determined.

Includes unit test.

Fixes #3456

### DIFF
--- a/src/Standards/PSR12/Sniffs/Classes/ClassInstantiationSniff.php
+++ b/src/Standards/PSR12/Sniffs/Classes/ClassInstantiationSniff.php
@@ -63,6 +63,14 @@ class ClassInstantiationSniff implements Sniff
                 continue;
             }
 
+            // Skip over potential attributes for anonymous classes.
+            if ($tokens[$i]['code'] === T_ATTRIBUTE
+                && isset($tokens[$i]['attribute_closer']) === true
+            ) {
+                $i = $tokens[$i]['attribute_closer'];
+                continue;
+            }
+
             if ($tokens[$i]['code'] === T_OPEN_SQUARE_BRACKET
                 || $tokens[$i]['code'] === T_OPEN_CURLY_BRACKET
             ) {
@@ -72,7 +80,7 @@ class ClassInstantiationSniff implements Sniff
 
             $classNameEnd = $i;
             break;
-        }
+        }//end for
 
         if ($classNameEnd === null) {
             return;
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks for fixing this

jrfnl:
You're welcome.

PR #3464, opened 2021-10-30
Generic/MultipleStatementAlignment: bug fix for closure params with defaults

### LINKED ISSUE #3460: Generic.Formatting.MultipleStatementAlignment false positive on closure with parameters
**Describe the bug**

**Code sample**
```php
<?php declare(strict_types = 1);

namespace Bugs;

use Mockery;

class Bug {
    public function bug(): mixed {
        $a = static function (array $variables = []) {
            return $variables;
        };
        $b = $a;

        return $b;
    }
}
```

**Custom ruleset**
```xml
<?xml version="1.0"?>
<ruleset name="Bug"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="./vendor/squizlabs/php_codesniffer/phpcs.xsd">
    <file>./packages</file>
    <rule ref="Generic.Formatting.MultipleStatementAlignment"/>
</ruleset>
```

**To reproduce**
2. Run `phpcs test.php ...`
```
---------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
---------------------------------------------------------------------------------------------------------------
 12 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 37 spaces but found 1 space
---------------------------------------------------------------------------------------------------------------
---------------------------------------------------------------------------------------------------------------
```

**Expected behavior**

No error.

OS: [e.g., Windows 10, MacOS 10.15]: Ubuntu Latest
PHP: [e.g., 7.2, 7.4]: 8.0
PHPCS: [e.g., 3.5.5, master]: master
Standard: [e.g., PSR2, PSR12, Squiz]: none

jrfnl:
For use within the loop in the `checkAlignment()` method, closures are explicitly removed for the `$scopeOpeners` array as they can be (and often are) assigned to a variable.

However, a closure which contained a parameter with a default value would take the equal sign for the default value as the start of a new block.

The `process()` method already contained a safeguard against this, but it's the `checkAlignment()` method which is being called recursively, so this check was not executed for those recursive calls.

Calling the `process()` method recursively would be troublesome as the returned stackPtr is incremented by one in the `process()` method, so moving the initial check from the `process()` method to the `checkAlignment()` method seemed a reasonable solution.

There will probably be other ways in which this could be solved and I'm not 100% sure this is the best solution, but it is a solution which works without breaking any of the existing unit tests.

Includes a few additional unit tests to further safeguard against this issue and make sure nothing new is broken.

Fixes #3460

### DIFF
--- a/src/Standards/Generic/Sniffs/Formatting/MultipleStatementAlignmentSniff.php
+++ b/src/Standards/Generic/Sniffs/Formatting/MultipleStatementAlignmentSniff.php
@@ -80,25 +80,6 @@ class MultipleStatementAlignmentSniff implements Sniff
      */
     public function process(File $phpcsFile, $stackPtr)
     {
-        $tokens = $phpcsFile->getTokens();
-
-        // Ignore assignments used in a condition, like an IF or FOR.
-        if (isset($tokens[$stackPtr]['nested_parenthesis']) === true) {
-            // If the parenthesis is on the same line as the assignment,
-            // then it should be ignored as it is specifically being grouped.
-            $parens    = $tokens[$stackPtr]['nested_parenthesis'];
-            $lastParen = array_pop($parens);
-            if ($tokens[$lastParen]['line'] === $tokens[$stackPtr]['line']) {
-                return;
-            }
-
-            foreach ($tokens[$stackPtr]['nested_parenthesis'] as $start => $end) {
-                if (isset($tokens[$start]['parenthesis_owner']) === true) {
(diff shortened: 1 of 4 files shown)


gsherwood:
Thanks for fixing this. Very happy to accept a straightforward fix that doesn't break anything :)

Issue #3468, opened 2021-11-11
do/while loops are double-counted in Generic.Metrics.CyclomaticComplexity


MarkBaker:
Both while loops and do/while loops should increment the Cyclomatic Complexity score by one, because there are two paths through the code in both cases.

In the case of a `while` loop, the `while` code block will only be executed if the `while` condition is met, otherwise code execution will branch over the `while` code block. This gives just two branches in the code.
```php
while ($condition) {            |\
    ...                         | \
    ...                         | /
}                               |<
                                v
```
When calculating the score, the `T_WHILE` token is recognised, and increases the CYC value correctly by one.

For a `do/while` loop, the `do` code block will always be executed at least once, and code execution may then branch back to repeat execution of that block, or may continue on through the code based on the `while` condition. This also has only two branches in the code.
```php
do {                            |<
    ...                         | \
    ...                         | /
} while ($condition);           |/
                                |
                                v
```

However, when calculating the score, the sniff checks for both the `T_DO` ___and___ the terminating `T_WHILE` tokens, incrementing CYC by two (once for each token) in the `do/while` loop, rather than just by one (as it should).

This double-count error can easily be eliminated by removing the increment for the `T_DO` token.

Note that some the existing tests also erroneously assume that do/while should increment the CYC count by two, so they also require modification.

PR #3470, opened 2021-11-11
Resolve double-increment of CYC count for do/while loop in Generic.Metrics.CyclomaticComplexity sniff

MarkBaker:
Fix for Issue #3468

Resolve double-increment of CYC count for do/while loop


MarkBaker:
Replaced by [PR #3495](https://github.com/squizlabs/PHP_CodeSniffer/pull/3495)

Issue #3472, opened 2021-11-11
PHP 8 match() expression is not counted in Generic.Metrics.CyclomaticComplexity


MarkBaker:
PHP8 Introduced `match()` as a new expression syntax, similar to a `switch` statement but with a rather different structure, using an arrow (`=>`) to separate the `case` from the expression that it would return. This should also be factored into the CYC calculation for the `Generic.Metrics.CyclomaticComplexity` sniff.

When calculating the CYC, this should be evaluated like a `switch` statement, with each `T_MATCH_ARROW` token corresponding to a `case` and increasing the CYC by one. There is no need to add one for the `T_MATCH_DEFAULT` token, because this is followed by a `T_MATCH_ARROW` token, and to do so would be to double-count that entry.

PR to follow.

PR #3471, opened 2021-11-11
Factor ternary operator, null coalescence (and assignment) operators into CYC calculation

MarkBaker:
Issue #3469

Add ternary operator, null coalescence operator and null coalescence assignment operator to tokens that trigger an increase in cyclomatic complexity

PR #3473, opened 2021-11-12
Calculate CYC for PHP 8 match() expression in Cyclomatic Complexity sniff

MarkBaker:
Issue #3472

Calculate CYC for PHP 8 match() expression in Cyclomatic Complexity sniff


gsherwood:
I'm not sure what happened to the test files here, but you seem to have added a new `.3` test file and removed the main test file.

It looks like all that needs doing is to include your new test cases in the main test file and ensure the additional warning line is included.

Thanks for the fix though. I'd be happy to merge this if the tests can be corrected. If you don't have time, I can make the adjustments during the merge.

MarkBaker:
> I'm not sure what happened to the test files here, but you seem to have added a new `.3` test file and removed the main test file.
> 
> It looks like all that needs doing is to include your new test cases in the main test file and ensure the additional warning line is included.
> 
> Thanks for the fix though. I'd be happy to merge this if the tests can be corrected. If you don't have time, I can make the adjustments during the merge.
>
> I'm not sure what happened to the test files here, but you seem to have added a new `.3` test file and removed the main test file.

I split the test file into 3 separate files because I did three separate PRs, one for a fix to prevent the the do/while loop being double-counted in the calculation ([#3470](https://github.com/squizlabs/PHP_CodeSniffer/pull/3470)); one for the ternary and null coalescence operators ([#3471](https://github.com/squizlabs/PHP_CodeSniffer/pull/3471)); and this one.... without splitting the tests for each into separate files for each PR; I'd get problems with the error list and warning list line numbers not matching up.
It seemed a better approach separating the PRs, because they're different issues with the calculation, even though that meant separate test files; but if you want a single test file; then the 3 separate PRs all need to be merged into a single PR as well to ensure that those warning/error line numbers match up.

jrfnl:
> I split the test file into 3 separate files because I did three separate PRs, one for a fix to prevent the the do/while loop being double-counted in the calculation ([PR #3470](https://github.com/squizlabs/PHP_CodeSniffer/pull/3470)); one for the ternary and null coalescence operators ([PR #3471](https://github.com/squizlabs/PHP_CodeSniffer/pull/3471)); and this one.... without splitting the tests for each into separate files for each PR; I'd get problems with the error list and warning list line numbers not matching up

I understand where you are coming from, but in that case, the "rename original test file" should have been done in each of the PRs.

As things are, the three PRs cannot be viewed individually anyway as 1 PR renames the test case file and sets the test file up to expect multiple test files (PR #3470), the other two do a hard remove of the old case test file without bringing it back and remove the expectations for the original case file too.
Have a good look at the diffs to see what I mean.

As there is no (sniff) technical reason for the separate test case files (other than sending in the PRs at the same time), having three test case files also doesn't follow the test best practices for sniff tests.

What about having one PR with three commits ?

MarkBaker:
> As there is no (sniff) technical reason for the separate test case files (other than sending in the PRs at the same time), having three test case files also doesn't follow the test best practices for sniff tests.
> 
> What about having one PR with three commits ?

I'll close the three separate PRs; and resubmit as a single PR

MarkBaker:
Replaced by [PR #3495](https://github.com/squizlabs/PHP_CodeSniffer/pull/3495)

Issue #3474, opened 2021-11-15
PHP 8.1 `enum` not recognized


fefas:
**Describe the bug**
PHP 8.1 `enum` not recognized.

**Code sample**
```php
<?php

enum MyEnum
{
    case CONST1;
    case CONST2;
}
```

**To reproduce**

Just run the PHPCS with the `PSR12` ruleset and the output will be something like:
```
FILE: /var/www/code/src/MyEnum.php
-----------------------------------------------------------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
-----------------------------------------------------------------------------------------------------------------------
 5 | ERROR | [x] Line indented incorrectly; expected 0 spaces, found 4
   |       |     (Generic.WhiteSpace.ScopeIndent.IncorrectExact)
 6 | ERROR | [x] Line indented incorrectly; expected 0 spaces, found 4
   |       |     (Generic.WhiteSpace.ScopeIndent.IncorrectExact)
 7 | ERROR | [x] Line indented incorrectly; expected at least 4 spaces, found 0
   |       |     (Generic.WhiteSpace.ScopeIndent.Incorrect)
-----------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 3 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-----------------------------------------------------------------------------------------------------------------------
```

By running the auto fix, I get:

```php
<?php

enum MyEnum
{
case CONST1;
case CONST2;
    }
```

**Expected behavior**
The following code should cause no CS error:
```php
<?php

enum MyEnum
{
    case CONST1;
    case CONST2;
}
```

**Versions (please complete the following information):**
 - OS: Docker on MacOS (image tag `php:8.1.0RC5-cli-alpine3.14`)
 - PHP: 8.1 RC5 
 - PHPCS: 3.6.1
 - Standard: PSR12

jrfnl:
PHPCS currently doesn't contain any syntax support yet for new syntaxes added in PHP 8.1. Work on this still has to start as all maintainers are overloaded.

fefas:
No pressure :) just reporting and trying to contribute somehow

kukulich:
I plan to do something to PHP 8.1 support this weekend.

iansltx:
To clean this up, ScopeIndentSniff will need to be updated as well, looks like.

michnovka:
[...]

PR #3478, opened 2021-11-19
PHP 8.1: Added support for `enum` keyword

kukulich:
Fixes https://github.com/squizlabs/PHP_CodeSniffer/issues/3474

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -152,6 +152,13 @@ class PHP extends Tokenizer
             'shared' => false,
             'with'   => [],
         ],
+        T_ENUM          => [
+            'start'  => [T_OPEN_CURLY_BRACKET => T_OPEN_CURLY_BRACKET],
+            'end'    => [T_CLOSE_CURLY_BRACKET => T_CLOSE_CURLY_BRACKET],
+            'strict' => true,
+            'shared' => false,
+            'with'   => [],
+        ],
         T_USE           => [
             'start'  => [T_OPEN_CURLY_BRACKET => T_OPEN_CURLY_BRACKET],
             'end'    => [T_CLOSE_CURLY_BRACKET => T_CLOSE_CURLY_BRACKET],
@@ -339,6 +346,7 @@ class PHP extends Tokenizer
         T_ENDIF                    => 5,
         T_ENDSWITCH                => 9,
         T_ENDWHILE                 => 8,
+        T_ENUM                     => 4,
         T_EVAL                     => 4,
         T_EXTENDS                  => 7,
         T_FILE                     => 8,
@@ -467,6 +475,7 @@ class PHP extends Tokenizer
         T_CLASS                    => true,
         T_INTERFACE                => true,
         T_TRAIT                    => true,
+        T_ENUM                     => true,
         T_EXTENDS                  => true,
         T_IMPLEMENTS               => true,
         T_ATTRIBUTE                => true,
@@ -950,6 +959,42 @@ class PHP extends Tokenizer
 
                 // Continue, as we're done with this token.
                 continue;
+            }//end if
+
+            /*
+                Enum keyword for PHP < 8.1
+            */
+
+            if ($tokenIsArray === true
+                && $token[0] === T_STRING
+                && strtolower($token[1]) === 'enum'
+            ) {
+                // Get the next non-empty token.
+                for ($i = ($stackPtr + 1); $i < $numTokens; $i++) {
+                    if (is_array($tokens[$i]) === false
+                        || isset(Util\Tokens::$emptyTokens[$tokens[$i][0]]) === false
+                    ) {
+                        break;
+                    }
+                }
+
+                if (isset($tokens[$i]) === true
+                    && is_array($tokens[$i]) === true
+                    && $tokens[$i][0] === T_STRING
+                ) {
+                    $newToken            = [];
+                    $newToken['code']    = T_ENUM;
+                    $newToken['type']    = 'T_ENUM';
+                    $newToken['content'] = $token[1];
+                    $finalTokens[$newStackPtr] = $newToken;
+
+                    if (PHP_CODESNIFFER_VERBOSITY > 1) {
+                        echo "\t\t* token $stackPtr changed from T_STRING to T_ENUM".PHP_EOL;
+                    }
+
+                    $newStackPtr++;
+                    continue;
+                }
             }//end if
 
             /*
--- a/src/Util/Tokens.php
+++ b/src/Util/Tokens.php
@@ -167,6 +167,10 @@ if (defined('T_READONLY') === false) {
     define('T_READONLY', 'PHPCS_T_READONLY');
 }
 
+if (defined('T_ENUM') === false) {
+    define('T_ENUM', 'PHPCS_T_ENUM');
+}
+
 // Tokens used for parsing doc blocks.
 define('T_DOC_COMMENT_STAR', 'PHPCS_T_DOC_COMMENT_STAR');
 define('T_DOC_COMMENT_WHITESPACE', 'PHPCS_T_DOC_COMMENT_WHITESPACE');
@@ -194,6 +198,7 @@ final class Tokens
         T_CLASS               => 1000,
         T_INTERFACE           => 1000,
         T_TRAIT               => 1000,
+        T_ENUM                => 1000,
         T_NAMESPACE           => 1000,
         T_FUNCTION            => 100,
         T_CLOSURE             => 100,
@@ -419,6 +424,7 @@ final class Tokens
         T_ANON_CLASS => T_ANON_CLASS,
         T_INTERFACE  => T_INTERFACE,
         T_TRAIT      => T_TRAIT,
+        T_ENUM       => T_ENUM,
         T_NAMESPACE  => T_NAMESPACE,
         T_FUNCTION   => T_FUNCTION,
         T_CLOSURE    => T_CLOSURE,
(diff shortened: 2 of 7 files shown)


src/Tokenizers/PHP.php
```diff
@@ -870,6 +879,35 @@ protected function tokenize($string)
...
+
+                if (isset($tokens[$i]) === true
+                    && is_array($tokens[$i]) === true
+                    && $tokens[$i][0] === T_STRING
+                ) {
+                    $tokens[$stackPtr][0] = T_ENUM;
+                    $token[0] = T_ENUM;
+                    if (PHP_CODESNIFFER_VERBOSITY > 1) {
+                        echo "\t\t* token $stackPtr changed from T_STRING to T_ENUM".PHP_EOL;
+                    }
```
jrfnl, line 907: Just checking: why not assign the token to the `$finalTokens` array and `continue` onto the next token ?

src/Util/Tokens.php
```diff
@@ -629,6 +635,7 @@ final class Tokens
         T_ANON_CLASS => T_ANON_CLASS,
         T_INTERFACE  => T_INTERFACE,
         T_TRAIT      => T_TRAIT,
+        T_ENUM       => T_ENUM,
```
jrfnl, line 638: I agree with this change in principle, but this **_will_** affect all sniffs looking for the `$ooScopeTokens` tokens and as the type declaration in the enum declaration is unique, I suspect that a number of sniffs referring to this token array will need updating to handle this correctly.

tests/Core/Tokenizer/EnumTest.inc
```diff
@@ -0,0 +1,55 @@
...
+
+        /* testEnumUsedAsPropertyName */
+        $this->enum = 'foo';
+    }
+}
+
+/* testEnumUsedAsFunctionName */
+function enum()
+{
+}
```
jrfnl, line 55: Could tests be added for the following ?

```php
/* testDeclarationContainingComment */
enum /* comment */ Name
{
    case SOME_CASE;
}

/* testParseErrorMissingname */
enum {
    case SOME_CASE;
}

/* testParseErrorLiveCoding */
// This must be the last test in the file.
enum
```

I also wonder about the below, not sure if it _could_ be valid, but would be interesting to ensure the tokenization across PHP versions is consistent:
```php
enum Enum {}
```

And for the non-enum test cases - what about adding these ? (note: this is more to safeguard that these are tokenized as `T_STRING` across all supported PHP versions than to check the `T_ENUM` backfill):
```php
namespace My\Enum\Collection;
$obj = new Enum;
$var = enum($a, $b);
var = namespace\enum();
echo Enum::CONSTANT;
echo ClassName::ENUM;
// etc
```

jrfnl (commented):
@kukulich 👋🏻 Thank you for creating this PR. Looking good!

I've left some comments inline to think over. Hope it helps.

src/Tokenizers/PHP.php
kukulich: Thanks, modified.

tests/Core/Tokenizer/EnumTest.inc
kukulich: Thanks, added.

src/Util/Tokens.php
kukulich: Yes, but I think it should be better to solve it in other PRs. This PR should be merged first.

src/Tokenizers/PHP.php
```diff
@@ -870,6 +879,42 @@ protected function tokenize($string)
...
+                }
+
+                if (isset($tokens[$i]) === true
+                    && is_array($tokens[$i]) === true
+                    && $tokens[$i][0] === T_STRING
+                ) {
+                    $newToken            = [];
+                    $newToken['code']    = T_ENUM;
+                    $newToken['type']    = 'T_ENUM';
+                    $newToken['content'] = 'enum';
```
jrfnl, line 906: I think the `content` should be taken from `$tokens[$i][1]` as by doing it this way, you may inadvertently be changing the case of the keyword.

jrfnl:
Thanks for the update @kukulich, left one more comment. Other that that, 👍🏻

src/Tokenizers/PHP.php
kukulich: Fixed.

jrfnl (approved):
I haven't verified against the RFC, I trust @kukulich will have done that, but PR-wise, this looks to cover all the bases 💯

tests/Core/Tokenizer/EnumTest.php
```diff
@@ -0,0 +1,163 @@
...
+     * @dataProvider dataNotEnums
+     * @covers       PHP_CodeSniffer\Tokenizers\PHP::tokenize
+     *
+     * @return void
+     */
+    public function testNotEnums($testMarker)
+    {
+        $tokens = self::$phpcsFile->getTokens();
+
+        $target = $this->getTargetToken($testMarker, [T_ENUM, T_STRING]);
```
jrfnl, line 126:
```suggestion
        $target = $this->getTargetToken($testMarker, [T_ENUM, T_STRING], 'enum');
```

The test as-is may currently "find" a `T_STRING` token _before_ the `enum` target, which could lead to the wrong token being tested.

Also means that the data provider may need to provide an optional second parameter to provide the keyword in the case as used in the test for those tests not using lowercase.

Example - in this test:
```php
namespace My\Enum\Collection;
```

The test as it currently is, is testing `My`, not `Enum`.

tests/Core/Tokenizer/EnumTest.php
```diff
@@ -0,0 +1,163 @@
...
+     * @param string $testMarker              The comment which prefaces the target token in the test file.
+     * @param int    $expectedScopeOpenerLine Expected line of scope opener.
+     * @param int    $expectedScopeCloserLine Expected line of scope closer.
+     *
+     * @dataProvider dataEnums
+     * @covers       PHP_CodeSniffer\Tokenizers\PHP::tokenize
+     *
+     * @return void
+     */
+    public function testEnums($testMarker, $expectedScopeOpenerLine, $expectedScopeCloserLine)
```
jrfnl, line 30: Just wondering - why test the _line_ and not the actual token position (in relation to the keyword token) ?
I realize the latter is more fiddly for setting up the tests, but it's also much more precise.

In other tests, this has been done by the data provider providing an `offset` and the token position then being calculated by `$expectedOpener = $enum + $openerOffset;`.

src/Util/Tokens.php
```diff
@@ -154,6 +154,10 @@
     define('T_ATTRIBUTE', 'PHPCS_T_ATTRIBUTE');
 }
 
+if (defined('T_ENUM') === false) {
+    define('T_ENUM', 'PHPCS_T_ENUM');
+}
+
```
jrfnl, line 160: This needs to be moved to below the `8.1 tokens` marker.

jrfnl (changes requested):
Sorry, still found some more issues to be addressed.

tests/Core/Tokenizer/EnumTest.php
kukulich: Fixed.

src/Util/Tokens.php
kukulich: Fixed.

tests/Core/Tokenizer/EnumTest.php
kukulich: Modified.

tests/Core/Tokenizer/EnumTest.php
jrfnl: By hard-coding the `enum` in the `testEnums()` test, no test can be added to ensure that the `enum` keyword not lowercase, is handled correctly.
This should be adjusted in a similar way as done for `testNotEnums()`, though it's perfectly fine to make the `$content` parameter for the test function optional with `'enum'` being the default value (for both test methods).

tests/Core/Tokenizer/EnumTest.php
jrfnl: Looks like you modified it for scope opener, but removed the check for scope closer ?

kukulich: Added now.

tests/Core/Tokenizer/EnumTest.php
kukulich: Added case insensitive test.

tarlepp:
What is the current state of this one?

kukulich:
Rebased on master.

tests/Core/Tokenizer/BackfillEnumTest.php
```diff
@@ -0,0 +1,229 @@
...
+     * @covers       PHP_CodeSniffer\Tokenizers\PHP::tokenize
+     *
+     * @return void
+     */
+    public function testNotEnums($testMarker, $testContent)
+    {
+        $tokens = self::$phpcsFile->getTokens();
+
+        $target = $this->getTargetToken($testMarker, [T_ENUM, T_STRING], $testContent);
+        $this->assertSame(T_STRING, $tokens[$target]['code']);
```
gsherwood, line 144: Due to the way namespaces are tokenized in 4.0 (they become `T_NAME_RELATIVE` or `T_NAME_QUALIFIED`) I had to reverse this logic in the 4.0 branch to use `assertNotSame(T_ENUM...)`. I didn't make the same change in master.

gsherwood:
Thanks a lot for this PR. I really appreciate the effort you're putting in to add support for 8.1 features to PHPCS.

PR #3480, opened 2021-11-20
PHP 8.1: Added support for `readonly` keyword

kukulich:


src/Tokenizers/PHP.php
```diff
@@ -945,6 +946,56 @@ protected function tokenize($string)
...
+                // Get the next non-empty token.
+                for ($i = ($stackPtr + 1); $i < $numTokens; $i++) {
+                    if (is_array($tokens[$i]) === false
+                        || isset(Util\Tokens::$emptyTokens[$tokens[$i][0]]) === false
+                    ) {
+                        break;
+                    }
+                }
+
+                $shouldBeReadonly = isset($tokens[$i]) === false || (is_array($tokens[$i]) === true && in_array($tokens[$i][0], [T_STRING, T_PUBLIC, T_PROTECTED, T_PRIVATE]) === true);
```
jrfnl, line 968: I have a niggly feeling that `T_VAR`, `T_STATIC` and `T_VARIABLE` should also be included in this list.... (and in the tests: `readonly var $property`, `readonly static $property`, `public readonly $property`)

FYI: IIRC, `readonly` did become a full-keyword with only one very specific exception: it's not a keyword when followed by an open parenthesis.

tests/Core/Tokenizer/ReadonlyTest.inc
```diff
@@ -0,0 +1,53 @@
...
+    private readonly int $privateReadonlyProperty;
+    /* testPublicReadonlyPropertyWithReadonlyFirst */
+    readonly public int $publicReadonlyProperty;
+    /* testProtectedReadonlyPropertyWithReadonlyFirst */
+    readonly protected int $protectedReadonlyProperty;
+    /* testPrivateReadonlyPropertyWithReadonlyFirst */
+    readonly private int $privateReadonlyProperty;
+    /* testReadonlyWithCommentsInDeclaration */
+    private /* Comment */ readonly /* Comment */ int /* Comment */ $readonlyPropertyWithCommentsInDeclaration;
+
```
jrfnl, line 21: Could some tests be added with:
* The "old-school" `var` keyword.
* The `static` keyword in varying order (with the keywords in vaying orders).
* No type declaration.
* And could the types used be varied a bit ? - I'd particularly like to see the `array` type be included in the tests, both as singular type as well as the first in a union type, as that's been problematic before (and a check that that `array` type is still re-tokenized to `T_STRING` when used in combination with `readonly).
    You may (also) want to add those to the existing `ArrayKeywordTest` and possibly add some tests with `readonly` to the `BitwiseOrTest` as well.

src/Util/Tokens.php
```diff
@@ -154,6 +154,10 @@
     define('T_ATTRIBUTE', 'PHPCS_T_ATTRIBUTE');
 }
 
+if (defined('T_READONLY') === false) {
+    define('T_READONLY', 'PHPCS_T_READONLY');
+}
+
```
jrfnl, line 160: This needs to be moved down to below the `// Some PHP 8.1 tokens, replicated for lower versions.` marker.

src/Tokenizers/PHP.php
```diff
@@ -945,6 +946,56 @@ protected function tokenize($string)
                 continue;
             }//end if
 
+            /*
+                "readonly" keyword support
+                PHP < 8.1: Converts T_STRING to T_READONLY - Name tokens (T_NAME_*) have to be converted to T_STRING first
```
jrfnl, line 951: Note: If I read this correctly, this means that this PR can't be forward-ported to the 4.x branch without changes. This may need some extra attention.

jrfnl (commented):
Thanks for taking this on @kukulich ! I've left some notes and suggestions in-line.

tests/Core/Tokenizer/ReadonlyTest.inc
kukulich: - `readonly` cannot be used with `var`
- It cannot be used in `static` property
- The type is required for readonly properties

However I have to add more types, eg. `parent`, `self`, `?int`, `int|null`, `array`.

src/Util/Tokens.php
kukulich: Fixed.

src/Tokenizers/PHP.php
kukulich: Yes, I currently expend there are no `T_NAME_*` tokens.

src/Tokenizers/PHP.php
kukulich: https://github.com/squizlabs/PHP_CodeSniffer/pull/3480#discussion_r753674931

tests/Core/Tokenizer/ReadonlyTest.inc
jrfnl: > * `readonly` cannot be used with `var`
>     * It cannot be used in `static` property
>     * The type is required for readonly properties

The thing is: whether or not it can _legally_ be used like that, users new to PHP 8.1 _may_ still try. If the tokenization of this is not adjusted to take these "incorrect" uses into account:
* Sniffs will not be able to _flag_ those incorrect uses.
* The `readonly` keyword would be grouped in with the type declaration or seen _as_ the type declaration when using the `File::getMemberProperties()` method.
* etc...

All in all, handling `readonly` consistently by allowing for typical mistakes devs may make, will make life a lot easier on sniff writers.

Keep in mind: sniff analysis is token based, not parser based, so parse errors (including incomplete code during live coding) should not break sniffs.

jrfnl: Note: no matter what, the tests should be added anyway to document how the PHPCS tokenizer handles these situations.

kukulich: > Keep in mind: sniff analysis is token based, not parser based, so parse errors (including incomplete code during live coding) should not break sniffs.

Yes, but I think it's better to not convert string to invalid `readonly`. In this way it will break less things.

kukulich: However, it looks to copy the PHP 8.1 behavier it's necessary to support it: https://3v4l.org/oEbgM

kukulich: Done.

src/Tokenizers/PHP.php
kukulich: Rewritted.

gsherwood:
Thanks a lot for this contribution.

I made a couple of little changes, including adding support for `readonly const`. While it is a syntax error in 8.1, it still tokenizes as `T_READONLY`...

PR #3481, opened 2021-11-20
Support for PHP 8.1 Explicit Octal Notation in the Tokenizer

MarkBaker:
Added support for Explicit Octal Notation (introduced in PHP 8.1) in the tokenizer.

When running phpcs against PHP < 8.1, an expression like 
```php
$foo = 0o777;
```
separates the `0o777` into two tokens; a `T_LNUMBER` with a value `"0"` and a `T_STRING` with a value `"o777"`.

These need to be treated as  single `T_LNUMBER` token with a value of `"0o777"` to be processed correctly by phpcs


tests/Core/Tokenizer/BackfillNumericSeparatorTest.inc
```diff
@@ -34,6 +34,9 @@ $foo = 0b0101_1111;
 /* testOctal */
 $foo = 0137_041;
 
+/* testExplicitOctal (Introduced in PHP 8.1) */
```
gsherwood, line 37: No need to include the text `(Introduced in PHP 8.1)` in the test comment - it's clearer to read without it

tests/Core/Tokenizer/BackfillNumericSeparatorTest.php
```diff
@@ -132,6 +132,13 @@ public function dataTestBackfill()
                     'value'  => '0137_041',
                 ],
             ],
+            [
+                [
+                    'marker' => '/* testExplicitOctal (Introduced in PHP 8.1) */',
```
gsherwood, line 137: This needs changing if test file comment is changed as well

tests/Core/Tokenizer/ExplicitOctalNotationTest.inc
```diff
@@ -0,0 +1,4 @@
+<?php
```
gsherwood, line 1: Given this is a test of the backfill functionality, the test files would be better named `BackfillExplicitOctalNotationTest.php|inc`

src/Tokenizers/PHP.php
```diff
@@ -553,6 +553,23 @@ protected function tokenize($string)
                 $lastNotEmptyToken = ($newStackPtr - 1);
             }
 
+            /*
+                For Explicit Octal Notation prior to PHP 8.1 we need to combine the
+                T_NUMBER and T_STRING token values into a single token value, and
```
gsherwood, line 558: Minor correction to comment: the token type is `T_LNUMBER` and not `T_NUMBER`

tests/Core/Tokenizer/ExplicitOctalNotationTest.inc
```diff
@@ -0,0 +1,4 @@
...
+
+/* testExplicitOctal declaration */
```
gsherwood, line 3: I'd remove `declaration` from this comment for consistency with other test files

tests/Core/Tokenizer/ExplicitOctalNotationTest.php
```diff
@@ -0,0 +1,63 @@
...
+     * @dataProvider dataExplicitOctalNotation
+     * @covers       PHP_CodeSniffer\Tokenizers\PHP::tokenize
+     *
+     * @return void
+     */
+    public function testExplicitOctalNotation($testData)
+    {
+        $tokens = self::$phpcsFile->getTokens();
+
+        $number = $this->getTargetToken($testData['marker'], [T_LNUMBER, T_DNUMBER, T_STRING]);
```
gsherwood, line 32: Are we only looking for T_LNUMBER tokens here or are there test cases to add where T_DNUMBER and T_STRING would be valid to find? Might need some tests to ensure some strings that look octal are not actually being touched ?

gsherwood (changes requested):

src/Tokenizers/PHP.php
jrfnl: IMO this code block also needs to be moved down after the initial comment processing bits and pieces.

I'm also wondering why the `$token` and `$tokens` variables are being adjusted and the loop is allowed to continue instead of the setting the token in the `$finalTokens` and `continue`-ing the loop.

tests/Core/Tokenizer/BackfillNumericSeparatorTest.inc
```diff
@@ -34,6 +34,9 @@ $foo = 0b0101_1111;
...
+$foo = 0o137_041;
```
jrfnl, line 38: Please also include a test with the `o` in uppercase.

tests/Core/Tokenizer/ExplicitOctalNotationTest.inc
```diff
@@ -0,0 +1,4 @@
...
+$foo = 0o137041;
```
jrfnl, line 4: Again: what about a number with uppercase `o` ?

jrfnl: Also wondering if any test could be conceived where the token stream should _not_ be changed (i.e. guard against false positives).

src/Tokenizers/PHP.php
```diff
@@ -553,6 +553,23 @@ protected function tokenize($string)
...
+                then ignore the T_STRING token.
+            */
+
+            if (PHP_VERSION_ID < 80100
+                && $tokenIsArray === true && $token[1] === '0'
+                && (isset($tokens[($stackPtr + 1)]) === true
+                && is_array($tokens[($stackPtr + 1)]) === true
+                && $tokens[($stackPtr + 1)][0] === T_STRING
+                && $tokens[($stackPtr + 1)][1][0] === 'o')
```
jrfnl, line 567: This does not account for uppercase `O`, which I'm pretty sure is valid (it is for the other explicit annotations).

gsherwood:
Thanks a lot for the contribution and for making those requested changes.

I made a couple of minor changes to the tests but didn't add anything about a case where the token should not be changed (can't think of a case off the top of my head). If anyone has any, please post here and I'll add the test case in.

PR #3483, opened 2021-11-20
PHP 8.1: Added `T_ENUM_CASE`

kukulich:
Based on https://github.com/squizlabs/PHP_CodeSniffer/pull/3478

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -477,6 +477,7 @@ class PHP extends Tokenizer
         T_INTERFACE                => true,
         T_TRAIT                    => true,
         T_ENUM                     => true,
+        T_ENUM_CASE                => true,
         T_EXTENDS                  => true,
         T_IMPLEMENTS               => true,
         T_ATTRIBUTE                => true,
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -347,6 +347,7 @@ class PHP extends Tokenizer
         T_ENDSWITCH                => 9,
         T_ENDWHILE                 => 8,
         T_ENUM                     => 4,
+        T_ENUM_CASE                => 4,
         T_EVAL                     => 4,
         T_EXTENDS                  => 7,
         T_FILE                     => 8,
@@ -982,6 +983,9 @@ class PHP extends Tokenizer
                     && is_array($tokens[$i]) === true
                     && $tokens[$i][0] === T_STRING
                 ) {
+                    // Modify $tokens directly so we can use it later when converting enum "case".
+                    $tokens[$stackPtr][0] = T_ENUM;
+
                     $newToken            = [];
                     $newToken['code']    = T_ENUM;
                     $newToken['type']    = 'T_ENUM';
@@ -990,6 +994,65 @@ class PHP extends Tokenizer
 
                     if (PHP_CODESNIFFER_VERBOSITY > 1) {
                         echo "\t\t* token $stackPtr changed from T_STRING to T_ENUM".PHP_EOL;
+                    }
+
+                    $newStackPtr++;
+                    continue;
+                }
+            }//end if
+
+            /*
+                Convert enum "case" to T_ENUM_CASE
+            */
+
+            if ($tokenIsArray === true
+                && $token[0] === T_CASE
+                && isset($this->tstringContexts[$finalTokens[$lastNotEmptyToken]['code']]) === false
+            ) {
+                $isEnumCase = false;
+                $scope      = 1;
+
+                for ($i = ($stackPtr - 1); $i > 0; $i--) {
+                    if ($tokens[$i] === '}') {
+                        $scope++;
+                        continue;
+                    }
+
+                    if ($tokens[$i] === '{') {
+                        $scope--;
+                        continue;
+                    }
+
+                    if (is_array($tokens[$i]) === false) {
+                        continue;
+                    }
+
+                    if ($scope !== 0) {
+                        continue;
+                    }
+
+                    if ($tokens[$i][0] === T_SWITCH) {
+                        break;
+                    }
+
+                    if ($tokens[$i][0] === T_ENUM || $tokens[$i][0] === T_ENUM_CASE) {
+                        $isEnumCase = true;
+                        break;
+                    }
+                }//end for
+
+                if ($isEnumCase === true) {
+                    // Modify $tokens directly so we can use it as optimisation for other enum "case".
+                    $tokens[$stackPtr][0] = T_ENUM_CASE;
+
+                    $newToken            = [];
+                    $newToken['code']    = T_ENUM_CASE;
+                    $newToken['type']    = 'T_ENUM_CASE';
+                    $newToken['content'] = $token[1];
+                    $finalTokens[$newStackPtr] = $newToken;
+
+                    if (PHP_CODESNIFFER_VERBOSITY > 1) {
+                        echo "\t\t* token $stackPtr changed from T_CASE to T_ENUM_CASE".PHP_EOL;
                     }
 
                     $newStackPtr++;
--- a/tests/Core/Tokenizer/EnumCaseTest.inc
+++ b/tests/Core/Tokenizer/EnumCaseTest.inc
@@ -1,0 +1,78 @@
+<?php
+
+enum Foo
+{
+    /* testPureEnumCase */
+    case SOME_CASE;
+}
+
+enum Boo: int {
+    /* testBackingIntegerEnumCase */
+    case ONE = 1;
+}
+
+enum Hoo: string
+{
+    /* testBackingStringEnumCase */
+    case ONE = 'one';
+}
+
+enum ComplexEnum: int implements SomeInterface
+{
+    use SomeTrait {
+        traitMethod as enumMethod;
+    }
+
+    const SOME_CONSTANT = true;
+
+    /* testEnumCaseInComplexEnum */
+    case ONE = 1;
+
+    /* testEnumCaseIsCaseInsensitive */
+    CaSe TWO = 2;
+
+    public function someMethod(): bool
+    {
+        switch (true) {
+            /* testCaseWithSemicolonIsNotEnumCase */
+            case CONSTANT;
+        }
+    }
+
+    /* testEnumCaseAfterSwitch */
+    case THREE = 3;
+
+    public function someOtherMethod(): bool
+    {
+        switch (true):
+            case false:
+        endswitch;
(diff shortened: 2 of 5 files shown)


tests/Core/Tokenizer/EnumCaseTest.php
```diff
@@ -0,0 +1,142 @@
...
+     * @dataProvider dataEnumCases
+     * @covers       PHP_CodeSniffer\Tokenizers\PHP::processAdditional
+     *
+     * @return void
+     */
+    public function testEnumCases($testMarker, $testContent)
+    {
+        $tokens = self::$phpcsFile->getTokens();
+
+        $enum = $this->getTargetToken($testMarker, [T_ENUM_CASE, T_CASE], $testContent);
```
jrfnl, line 33: `$testContent` isn't needed for this one as for both `T_ENUM_CASE` as well as `T_CASE` would have the same content.

It's mostly needed when one of the tokens being searched for is `T_STRING` and there are multiple of the same tokens in the statement after the test marker.

Then again, maybe `T_STRING` should be added to the token array ?

tests/Core/Tokenizer/EnumCaseTest.php
```diff
@@ -0,0 +1,142 @@
...
+
+        $this->assertSame(T_ENUM_CASE, $tokens[$enum]['code']);
+        $this->assertSame('T_ENUM_CASE', $tokens[$enum]['type']);
```
jrfnl, line 36: I believe it would be good idea to ensure that the `scope_owner`, `scope_opener` and `scope_closer` keys are **_not_** set for the found token.
Again, this is similar to what is done in the `DefaultKeywordTest::testMatchDefault()` test method.

Might it be an idea to transform this complete test along the same lines as the `DefaultKeywordTest` ?
I.e. `CaseKeywordTest` and have detailed tests for both the `enum` `case` keyword (including making sure no scope is set) as well as the `switch` `case` keyword (including making sure scope **is** set)  and have a separate set of tests in the same file for the `case` keyword being used when not a keyword ?

src/Tokenizers/PHP.php
```diff
@@ -466,6 +475,7 @@ class PHP extends Tokenizer
         T_CLASS                    => true,
         T_INTERFACE                => true,
         T_TRAIT                    => true,
+        T_ENUM                     => true,
```
jrfnl, line 478: I have a feeling that `T_ENUM_CASE` could also be added here as the first non-whitespace token after `T_ENUM_CASE` should always be tokenized as `T_STRING` or am I overlooking something ?

Might also be a good idea to cover that via the tests, i.e.:
```php
enum {
    /* testKeywordAsCaseNameShouldBeTstring1 */
    case FUNCTION = 'function';
    /* testKeywordAsCaseNameShouldBeTstring2 */
    case FALSE = 'false';
    /* testKeywordAsCaseNameShouldBeTstring3 */
    case DEFAULT = 'default'; // Should maybe also be added to the `DefaultKeywordTest` file ?
    /* testKeywordAsCaseNameShouldBeTstring4 */
    case ARRAY = 'array';
}
```

src/Tokenizers/PHP.php
```diff
@@ -870,6 +880,96 @@ protected function tokenize($string)
...
+
+                    if (is_array($tokens[$i]) === false) {
+                        continue;
+                    }
+
+                    if ($scope === 0 && $tokens[$i][0] === T_SWITCH) {
+                        break;
+                    }
+
+                    if ($tokens[$i][0] === T_ENUM) {
```
jrfnl, line 951: Should this also check for `$scope === 0` ? I.e. can `enum`s be declared within a `switch` ?
Looks like it's a supported syntax: https://3v4l.org/riI63

```php
switch ($x) {
    case 'a': {
        enum Foo {}
        break;
    }

    case 'b';
        enum Bar {}
        break;
}
```

src/Tokenizers/PHP.php
```diff
@@ -870,6 +880,96 @@ protected function tokenize($string)
...
+                }
+            }//end if
+
+            /*
+                Convert enum "case" to T_ENUM_CASE
+            */
+
+            if ($tokenIsArray === true
+                && $token[0] === T_CASE
+            ) {
```
jrfnl, line 928: Should this have a similar check as with the `T_DEFAULT` retokenization to prevent definitely _not_ case keywords from being retokenized ?

```php
function case() {}
```

See:
https://github.com/squizlabs/PHP_CodeSniffer/blob/5fb9b643da9e0420f979b0f87fdac4667225d621/src/Tokenizers/PHP.php#L1425
https://github.com/squizlabs/PHP_CodeSniffer/blob/5fb9b643da9e0420f979b0f87fdac4667225d621/src/Tokenizers/PHP.php#L1466-L1479

tests/Core/Tokenizer/EnumCaseTest.php
```diff
@@ -0,0 +1,142 @@
...
+     * @dataProvider dataNotEnumCases
+     * @covers       PHP_CodeSniffer\Tokenizers\PHP::processAdditional
+     *
+     * @return void
+     */
+    public function testNotEnumCases($testMarker, $testContent)
+    {
+        $tokens = self::$phpcsFile->getTokens();
+
+        $enum = $this->getTargetToken($testMarker, [T_ENUM_CASE, T_CASE], $testContent);
```
jrfnl, line 99: `$testContent` isn't needed for this one as for both `T_ENUM_CASE` as well as `T_CASE` would have the same content.

It's mostly needed when one of the tokens being searched for is `T_STRING` and there are multiple of the same tokens in the statement after the test marker.

Then again, maybe `T_STRING` should be added to the token array ?

tests/Core/Tokenizer/EnumCaseTest.php
```diff
@@ -0,0 +1,142 @@
...
+
+
+    /**
+     * Test that the enum "case" is converted to T_ENUM_CASE.
+     *
+     * @param string $testMarker  The comment which prefaces the target token in the test file.
+     * @param string $testContent The token content to expect.
+     *
+     * @dataProvider dataEnumCases
+     * @covers       PHP_CodeSniffer\Tokenizers\PHP::processAdditional
```
jrfnl, line 25:
```suggestion
     * @covers       PHP_CodeSniffer\Tokenizers\PHP::tokenize
     * @covers       PHP_CodeSniffer\Tokenizers\Tokenizer::recurseScopeMap
```

tests/Core/Tokenizer/EnumCaseTest.php
```diff
@@ -0,0 +1,142 @@
...
+
+
+    /**
+     * Test that "case" that is not enum case is still tokenized as `T_CASE`.
+     *
+     * @param string $testMarker  The comment which prefaces the target token in the test file.
+     * @param string $testContent The token content to expect.
+     *
+     * @dataProvider dataNotEnumCases
+     * @covers       PHP_CodeSniffer\Tokenizers\PHP::processAdditional
```
jrfnl, line 91:
```suggestion
     * @covers       PHP_CodeSniffer\Tokenizers\PHP::tokenize
```

tests/Core/Tokenizer/EnumCaseTest.inc
```diff
@@ -0,0 +1,65 @@
...
+switch (true) {
+    /* testCaseWithConstantIsNotEnumCase */
+    case CONSTANT:
+    /* testCaseWithConstantAndIdenticalIsNotEnumCase */
+    case CONSTANT === 1:
+    /* testCaseWithAssigmentToConstantIsNotEnumCase */
+    case CONSTANT = 1:
+    /* testIsNotEnumCaseIsCaseInsensitive */
+    cAsE CONSTANT:
+}
```
jrfnl, line 65: I'd recommend adding far more extensive tests for varying combinations of `enum` and `switch`, as well as covering as many situations as one can think of where `case` can be used and should be tokenized as `T_STRING` instead of `T_CASE` or `T_ENUM_CASE`.

See the test cases for the `DefaultKeywordTest` for inspiration.

jrfnl (commented):
@kukulich 👋🏻 I've had an initial look at this and left some comments for you to think over.

src/Tokenizers/PHP.php
```diff
@@ -870,6 +880,96 @@ protected function tokenize($string)
...
+                    $newStackPtr++;
+                    continue;
+                }
+            }//end if
+
+            /*
+                Convert enum "case" to T_ENUM_CASE
+            */
+
+            if ($tokenIsArray === true
```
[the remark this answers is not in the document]
kukulich: > This feels very inefficient.

Yes, but it took me couple of hours to create this "final" version that solve all situations I found.

I've just figured out one small optimisation for enum case.

tests/Core/Tokenizer/EnumCaseTest.php
kukulich: > $testContent isn't needed for this one as for both T_ENUM_CASE as well as T_CASE would have the same content.

You're right. Removed.

tests/Core/Tokenizer/EnumCaseTest.php
kukulich: > I believe it would be good idea to ensure that the scope_owner, scope_opener and scope_closer keys are not set for the found token.

Added.

src/Tokenizers/PHP.php
kukulich: Thanks, added.

src/Tokenizers/PHP.php
kukulich: https://github.com/squizlabs/PHP_CodeSniffer/pull/3484

src/Tokenizers/PHP.php
kukulich: Needs https://github.com/squizlabs/PHP_CodeSniffer/pull/3484

kukulich:
Rebased and ready for review.

src/Tokenizers/PHP.php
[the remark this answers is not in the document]
gsherwood: I did some testing and it's definitely faster doing the work in `Tokenizer::recurseScopeMap()`, and a less code as well. It does require adding `T_CASE` to the `tstringContexts` member var otherwise the CASE values are not converted to T_STRING using the new method.

gsherwood:
Thanks a lot for this PR, and for keeping it rebased.

I've merged it in without the performance improvements that come from using `Tokenizer::recurseScopeMap()`. If real-world performance suffers due to this change, I can merge that in.

PR #3488, opened 2021-11-27
GH Actions: PHP 8.1 has been released

jrfnl:
PHP 8.1 has been released, so builds against it should no longer be allowed to fail.

Also adding PHP 8.2 as the new "nightly" build (which is allowed to fail).


gsherwood:
Thanks for doing this

jrfnl:
You're welcome ;-) (as always)

PR #3494, opened 2021-12-03
PHP 7.4: Array and string offset access syntax with curly braces is deprecated

julien-:


jrfnl:
PHPCS 2.x is no longer maintained. The last release was in 2018 (before PHP 7.4 was even under development). Please upgrade to use PHPCS 3.x.

Issue #3437, opened 2021-09-17
PSR12 does not forbid blank lines at the start of the class body


InvisibleSmiley:
**Describe the bug**
In certain situations, PSR12 does not forbid blank lines at the start of class bodies, **e.g. with class constants at the top**.
This is in contrast to:
- the definition of PSR-12 (https://www.php-fig.org/psr/psr-12/) that states "Opening braces MUST be on their own line and MUST NOT be preceded or followed by a blank line."
- the comment above `<rule ref="PSR2.Classes.ClassDeclaration"/>` in file PSR12/ruleset.xml which states: "Opening braces MUST be on their own line and MUST NOT be preceded or followed by a blank line."

In some other cases, blank lines at the start of class bodies are rejected by specific rules, e.g. those for trait imports (use), properties and methods (functions).

**Code sample**
```php
<?php

namespace Foo;

class Bar
{


    public const X = 1;
}
```

**Expected behavior**
Blank lines at the start of class bodies should always result in some sort of error with PSR12.

**Versions (please complete the following information):**
 - OS: Windows 10
 - PHP: 7.4
 - PHPCS: 3.6.0
 - Standard: PSR12

**Additional context**
I think the ClassDeclaration sniff neeeds to be adjusted for PSR12 so that it does not matter which kind of element is at the start of the class body (constants, property, method, ...). Leaving this logic to specific sniffs might result in future breakages as PHP evolves.

scorgn:
I am seeing this as well, it seems that it is also not notifying of a blank line after a method opening brace.

> 5. Control Structures
> [...]
> The body MUST be on the next line after the opening brace

I am on
- OS: MacOS Catalina 10.15.7
- PHP: 7.3.30
- PHPCS: 3.6.0
- Standard: PSR12

gsherwood:
This looks to be a missing check from the start. PSR2 allows as many blank lines as you want after the opening brace, so using that sniff alone was not enough to enforce the PSR12 rule. I'll add a new sniff to specifically enforce no blank lines after the opening brace.

gsherwood:
> it seems that it is also not notifying of a blank line after a method opening brace.

I think that's just for control structures, not functions and methods. I can't see a rule under methods that enforces no blank lines at the start of the method body

gsherwood:
I've added `PSR12.Classes.OpeningBraceSpace` to enforce this. The sample code provided now produces this error:
```
FILE: temp.php
[...]

PR #3499, opened 2021-12-13
PSR12: Ensure there are no empty line before and after function body

pbnPierre:
If I follow this issue https://github.com/squizlabs/PHP_CodeSniffer/issues/3437 resolvment
Should we also trim spaces on functions for PSR-12 standard

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1136,6 +1136,8 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="OpenTagSniff.php" role="php" />
        </dir>
        <dir name="Functions">
+        <file baseinstalldir="PHP/CodeSniffer" name="FunctionClosingBraceSpaceSniff.php" role="php" />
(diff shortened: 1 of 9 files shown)


jrfnl:
PSR-12 has no explicit rules about blank lines at the start of the function.

It does contain the phrase:
> the closing brace MUST go on the next line following the body

... which is in direct contrast to the sniff being proposed here.

To satisfy the above, the `PSR2.Methods.FunctionClosingBrace` exists and is included in the PSR12 standard.

pbnPierre:
Every examples in PSR-12 does not contain any blank lines before and after functions braces.
Ex:
```
    final public static function bar()
    {
        // method body
    }
```

For me 
this
```
    final public static function bar()
    {

        // method body
    }
```

or this
```
    final public static function bar()
    {
        // method body

    }
```
Should not be allowed in PSR-12 standards

Anyway  i did not managed to code the fixers it so i'm giving up.
If anyone wants to do it feel free.

Jeroen-G:
I used `FunctionOpeningBraceSpace`

Issue #3505, opened 2021-12-17
The nullsafe operator is not counted in Generic.Metrics.CyclomaticComplexity


MarkBaker:
As highlighted in discussion of [Issue #3501](https://github.com/squizlabs/PHP_CodeSniffer/issues/3501), the NullSafe Operator (`?->`) is not taken into consideration when calculating complexity in the Generic.Metrics.CyclomaticComplexity sniff.

This operator should increment the Cyclomatic Complexity score by 1, because there are two paths through the code each cases.

```
$result = $object->getX()?->getY();
```
can be re-written as an if/else statement
```
if ($object->getX() !== null) {
    $result = $object->getX()->getY();
} else {
    $result = null;
}
```
with two branches through the code, which should increment the CYC count by one.

gsherwood:
Fixed by attached PR

PR #3506, opened 2021-12-17
Include NullSafe operator in calculation for Cyclomatic Complexity Sniff

MarkBaker:
Issue [#3505](https://github.com/squizlabs/PHP_CodeSniffer/issues/3505)

Add nullsafe object operator to tokens that trigger an increase in cyclomatic complexity


gsherwood:
Thanks for the PR

PR #3507, opened 2021-12-18
PHP 8.1 | Tokenizer/PHP: readonly bug fix

jrfnl:
Follow up on #3480

While going through some sniffs to add support, I realized that parameters with a reference were not accounted for correctly. Didn't think of this before either.

Fixed now, including test.

/cc @kukulich


gsherwood:
Nice find, thanks for fixing this.

PR #3513, opened 2021-12-18
PHP 8.1 | Tokenizer/PHP: readonly vs union types bug fix

jrfnl:
More follow up on #3480.

Done some more stress testing and found that the retokenization of `|` to `T_TYPE_UNION` was broken when combined with the `readonly` keyword.

Fixed now, includes test.

/cc @kukulich

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -2710,7 +2710,8 @@ class PHP extends Tokenizer
 
                     if ($suspectedType === 'property or parameter'
                         && (isset(Util\Tokens::$scopeModifiers[$this->tokens[$x]['code']]) === true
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks for fixing this

jrfnl:
@gsherwood I'm fairly certain I found one more case where `T_READONLY` should be added to a condition in the PHP Tokenizer class, but I haven't been able to come up with a failing test case for that one yet. Will keep an eye on it:

https://github.com/jrfnl/PHP_CodeSniffer/blob/d5c1cf755070447305e6e847db0299bc4e4c0190/src/Tokenizers/PHP.php#L1577

gsherwood:
> but I haven't been able to come up with a failing test case for that one yet

I tried a few code snippets but couldn't come up with any failed cases either.

PR #3515, opened 2021-12-20
File::getMemberProperties(): minor tweaks after readonly merge

jrfnl:
Follow up on #3480

I realized after the merge that I had not published a few follow up comments. Addressing those here.

* The new `is_readonly` key still needed to be added to the documentation for the `File::getMemberProperties()` method.
* For consistency in how the method is tested, adding the `is_readonly` property to all other test expectations, which makes the `testPHP81NotReadonly` test case redundant.
* Adding two additional test cases. In particular I wanted to verify (and safeguard) that the retokenization of the `?` to `T_NULLABLE` after a `readonly` keyword would not be broken (which it partially was, but that has now been fixed in #3513).


jrfnl:
Rebased to get past imaginary merge conflicts.

@gsherwood As this is a follow-up on a PR which has gone into PHPCS 3.7.0 and updates the documentation for that change, can this still be merged in 3.7.0 ?

gsherwood:
Thank you

PR #3516, opened 2021-12-20
PHP 8.1 | File::getMethodParameters(): allow for readonly keyword

jrfnl:
... in constructor property promotion.

Note: the `property_readonly` key will always be added when constructor property promotion has been detected. The `readonly_token` only when the `readonly` token has actually been found.

Includes unit tests.


gsherwood:
Thanks a lot for adding this support.

Issue #3144, opened 2020-10-22
parallel processes that do nothing but need CPU 


voku:
**Describe the bug**
I don't know if this is a bug or not, but it looks strange:

Some processes seem to do nothing ```0
wait4(18694, 0x7ffe6ddd6cbc, WNOHANG, NULL) = 0``` (strace) but still need max CPU power (top). 

**Code sample**
Sorry, but the bug can't be shown in a small code sample, you need many files, so that one process is ready before other processes when we call PHP CodeSniffer via ```--parallel=X```.

**Custom ruleset**
<details>
<summary>Custom ruleset</summary>
```xml
<?xml version="1.0"?>
<ruleset name="StandardVdmg">
    <!-- 1. Intro -->

    <description>The VDMG coding standard.</description>
    <arg name="tab-width" value="4"/>
    <arg name="encoding" value="UTF-8" />
    <arg name="extensions" value="php"/>
    <arg name="colors"/>

    <!-- show progress of the run and show sniff names -->
    <arg value="sp"/>

    <!-- Don't hide tokenizer exceptions -->
    <rule ref="Internal.Tokenizer.Exception">
        <type>error</type>
    </rule>

    <autoload>./StandardVdmg/SniffsHelper/autoload.php</autoload>

    <rule ref="./StandardVdmg/Sniffs/VdmgClassFileNameSniff.php" />
    <rule ref="./StandardVdmg/Sniffs/VdmgConditionalExpressionOrderSniff.php" />
    <rule ref="./StandardVdmg/Sniffs/VdmgDisallowImplicitMicrotimeAsStringSniff.php" />
    <rule ref="./StandardVdmg/Sniffs/VdmgDisallowUniqidWithoutMoreEntropySniff.php" />
    <rule ref="./StandardVdmg/Sniffs/VdmgDocBlockReturnSelfSniff.php" />
    <rule ref="./StandardVdmg/Sniffs/VdmgEmptyLinesSniff.php" />
    <!-- <rule ref="./StandardVdmg/Sniffs/VdmgExecutableRegExSniff.php" /> -->
    <rule ref="./StandardVdmg/Sniffs/VdmgForbiddenCommentsSniff.php" />
    <rule ref="./StandardVdmg/Sniffs/VdmgForbiddenGlobalSniff.php" />
[...]

Issue #3518, opened 2021-12-20
PHP_CodeSniffer gets inifinitely stuck on PHP 8.1 with parallel processing enabled


morozov:
**Describe the bug**
If `phpcs` is run on PHP 8.1 with parallel processing enabled, the process never ends.

**Custom ruleset**
N/A

**To reproduce**
Run `phpcs --parallel=2`

**Expected behavior**
PHP_CodeSniffer exits as soon as all files have been checked.

**Versions (please complete the following information):**
 - OS: MacOS 11.6.2
 - PHP:  8.1.1
 - PHPCS: 3.6.2
 - Standard: N/A

**Additional context**
Quick debugging shows that although the child process exits with the exit code equal to its PID, `pcntl_waitpid()` keeps being executed in an infinite loop returning `0`.
https://github.com/squizlabs/PHP_CodeSniffer/blob/7eef328f4703620a6e7ebecb66b5fab9ca5015e3/src/Runner.php#L723-L724
When the same build runs on PHP 8.0, `pcntl_waitpid` returns the value of the child PID as soon as it exits.

According to the PHP [documentation](https://www.php.net/manual/en/function.exit.php),
> Exit statuses should be in the range 0 to 254, the exit status 255 is reserved by PHP and shall not be used.

Using PID as the exit code seems to be a violation of the API and may be the root cause.
https://github.com/squizlabs/PHP_CodeSniffer/blob/7eef328f4703620a6e7ebecb66b5fab9ca5015e3/src/Runner.php#L539

morozov:
UPD: it doesn't look related to the actual code. The same issue is reproducible without the exit code being PID:
```php
<?php

$pid = pcntl_fork();
if ($pid === -1) {
    echo 'Failed to create child process', PHP_EOL;
} elseif ($pid !== 0) {
    echo 'Waiting for child child PID ', $pid, PHP_EOL;
    $result = pcntl_waitpid($pid, $status);
    echo 'Received ', $result, PHP_EOL;
} else {
    echo 'In child', PHP_EOL;
}
```
Produces:
```
$ php -v
PHP 8.0.14 (cli) (built: Dec 16 2021 11:20:34) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.14, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.14, Copyright (c), by Zend Technologies
[...]

PR #3519, opened 2021-12-21
Get rid of busy-waiting in the parent process

morozov:
This patch simplifies the logic of waiting for child processes during parallel processing and may address #3144.

Change summary:

1. Instead of busy-waiting using `pcntl_waitpid(WNOHANG)` in a loop, let `pcntl_waitpid()` block until a child exits. This should address the following symptom:
   > Some processes seem to do nothing `0 wait4(18694, 0x7ffe6ddd6cbc, WNOHANG, NULL) = 0` (strace) but still need max CPU power (top).
2. Index child processes by PID to simplify the logic of unsetting exited children.
3. Do not use `$pid` as the exit code. This is unnecessary and violates the [API](https://www.php.net/manual/en/function.exit.php),
   > Exit statuses should be in the range 0 to 254, the exit status 255 is reserved by PHP and shall not be used.

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -460,10 +460,7 @@ class Runner
                 if ($pid === -1) {
                     throw new RuntimeException('Failed to create child process');
                 } else if ($pid !== 0) {
-                    $childProcs[] = [
-                        'pid' => $pid,
-                        'out' => $childOutFilename,
-                    ];
+                    $childProcs[$pid] = $childOutFilename;
                 } else {
                     // Move forward to the start of the batch.
                     $todo->rewind();
@@ -536,7 +533,7 @@ class Runner
 
                     $output .= ";\n?".'>';
                     file_put_contents($childOutFilename, $output);
-                    exit($pid);
+                    exit();
                 }//end if
             }//end for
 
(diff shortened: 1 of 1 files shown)


src/Runner.php
```diff
@@ -719,54 +716,51 @@ private function processChildProcs($childProcs)
...
-                        if (isset($childOutput) === false) {
-                            // The child process died, so the run has failed.
-                            $file = new DummyFile('', $this->ruleset, $this->config);
-                            $file->setErrorCounts(1, 0, 0, 0);
-                            $this->printProgress($file, $totalBatches, $numProcessed);
-                            $success = false;
-                            continue;
-                        }
+            $pid = pcntl_waitpid(0, $status);
+            $out = $childProcs[$pid];
```
gsherwood, line 720: Should at least run an `isset()` here in case `pcntl_waitpid` returns an error code and not a PID we know about.

morozov: Thanks. I didn't notice the possible error in the return value.

 I believe instead of checking for `isset()`, it would make more sense to check for the error and assume that the returned PID is always known to the runner. At lest if it happens to be not true, we'll see a notice instead of this issue silently ignored.

morozov:
@gsherwood is there anything else needed for this patch to be accepted?

gsherwood:
Thanks for this change. I can't replicate #3144 but it makes sense that this change should fix that condition. I've left this out of the changelog for now but will come back and address that before release. Either I'll mark it as a bug fix or just include it as a possible performance improvement.

@voku Are you able to test out this change by any chance to see if your bug report has been addressed now?

morozov:
IIRC, I stumbled upon this issue, while trying to troubleshoot https://github.com/squizlabs/PHP_CodeSniffer/issues/3518. Essentially, if a child gets stuck doing nothing, the parent will busy-wait for it consuming the entire CPU. So installing the gRPC extension may help reproduce the bug.

The combination of these two bugs was exacerbated if PHP_CodeSniffer was [integrated into PhpStorm](https://www.jetbrains.com/help/phpstorm/using-php-code-sniffer.html). The IDE might spawn multiple PHP_CodeSniffer processes which in a few minutes would end up consuming all available CPU cores.

voku:
> Thanks for this change. I can't replicate #3144 but it makes sense that this change should fix that condition. I've left this out of the changelog for now but will come back and address that before release. Either I'll mark it as a bug fix or just include it as a possible performance improvement.
> 
> @voku Are you able to test out this change by any chance to see if your bug report has been addressed now?

I tested the latest master and I do not see any `wait4(30150, 0x7ffc2b7f9e5c, WNOHANG, NULL) = 0` calls anymore. :+1: The process is killed when it's done and do not need more resources. Thanks @morozov for the fix! :deciduous_tree:

DavidGoodwin:
I've upgraded via composer to dev-master (dd8a709) and when running something like this ...

```
php7.4 ./vendor/bin/phpcbf --parallel=16 --standard=PSR12 --exclude=Generic.Files.LineLength ./lib/app ./tests
```

I see it hanging. 

If I stick an 'strace' infront of the above command, the last few lines are : 

```
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7fc7c76fbc50) = 1301408
getcwd("/home/david/src/something/something", 4096) = 45
clock_gettime(CLOCK_MONOTONIC, {tv_sec=692607, tv_nsec=368218623}) = 0
openat(AT_FDCWD, "/tmp/phpcs-childLUQLZV", O_RDWR|O_CREAT|O_EXCL, 0600) = 3
close(3)                                = 0
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7fc7c76fbc50) = 1301409
getcwd("/home/david/src/something/something", 4096) = 45
clock_gettime(CLOCK_MONOTONIC, {tv_sec=692607, tv_nsec=373677576}) = 0
openat(AT_FDCWD, "/tmp/phpcs-child2vkt1g", O_RDWR|O_CREAT|O_EXCL, 0600) = 3
close(3)                                = 0
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7fc7c76fbc50) = 1301410
getcwd("/home/david/src/something/something", 4096) = 45
clock_gettime(CLOCK_MONOTONIC, {tv_sec=692607, tv_nsec=377547515}) = 0
openat(AT_FDCWD, "/tmp/phpcs-childdXvAyS", O_RDWR|O_CREAT|O_EXCL, 0600) = 3
close(3)                                = 0
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7fc7c76fbc50) = 1301411
wait4(0, 
```

```
$ ./vendor/bin/phpcbf --version
PHP_CodeSniffer version 3.7.0 (stable) by Squiz (http://www.squiz.net)
```
```
$ composer show | grep squiz
squizlabs/php_codesniffer             dev-master dd8a709 PHP_CodeSniffer tokenizes PHP, JavaScript and CSS files and detects violations of a defined set of coding standards.
```

If I continue to run it with the PHP grpc flags ( `@php -d grpc.enable_fork_support=1 -d grpc.poll_strategy=epoll1 ./vendor/bin/phpcbf --parallel=$(nproc --all) --standard=PSR12 --exclude=Generic.Files.LineLength ./lib/app ./tests` ) it's happy.

PR #3520, opened 2021-12-23
Tokenizer/PHP: fix incorrect condition order

jrfnl:
Follow up on #3350

As things were, `$tokens[i]` may not exist and the current condition order would lead to a `PHP Warning:  Undefined array key 517 in php_codesniffer/src/Tokenizers/PHP.php on line 2101` notice if the end of the file was reached.

Fixed now.


gsherwood:
Thanks

PR #3523, opened 2021-12-31
Close popen handle with pclose

jtojnar:


jrfnl (approved):
Good catch! ✔️

gsherwood:
Agreed, good catch :) Thanks.

PR #3525, opened 2022-01-04
FileComment: update year in test file

jrfnl:
The regular annual update to make sure the build still passes ;-)

Issue #3479, opened 2021-11-20
PHP 8.1 support


kukulich:
Issue to track what needs to be done:

- [x] Support of `enum` keyword https://github.com/squizlabs/PHP_CodeSniffer/pull/3478
- [x] `enum` support to sniffs https://github.com/squizlabs/PHP_CodeSniffer/pull/3482
- [x] New token `T_ENUM_CASE` https://github.com/squizlabs/PHP_CodeSniffer/pull/3483
- [x] Support of `readonly` keyword https://github.com/squizlabs/PHP_CodeSniffer/pull/3480
- [x] `never` buildin type https://github.com/squizlabs/PHP_CodeSniffer/pull/3582
- [x] Intersection types - new token `T_TYPE_INTERSECTION` and  update `File::getMethodParameters()`, `File::getMethodProperties()` and `File::getMemberProperties()` https://github.com/squizlabs/PHP_CodeSniffer/pull/3581
- [x] Explicit octal notation https://github.com/squizlabs/PHP_CodeSniffer/pull/3481
- [x] Final constants - reported in PSR12.Properties.ConstantVisibility false positive when using public final const syntax #3526, fixed in PHP 8.1 | PSR12/ConstantVisibility: allow for class constants to be final #3527. A cursory review of existing sniffs showed no other obvious sniffs which would need adujsting.

jrfnl:
**Proposal**: 

For **intersection types**, I'd like to propose to re-tokenize to a new PHPCS native `T_TYPE_INTERSECTION` to prevent sniffs looking for `T_BITWISE_AND` from getting confused and starting to try and handle the token.

This is in-line with the changes made for PHP 8.0 union types (see #3032 and #3302).

Note: the `File::getMethodParameters()`, `File::getMethodProperties()` and `File::getMemberProperties()` methods will also need adjusting.

---

Also note that the `File::getMemberProperties()` also still needs adjusting for the `readonly` keyword.

jrfnl:
~~Check sniffs that use Tokens::$ooScopeTokens~~

For all `Tokens` arrays being updated with new tokens:
- [ ] Check every single PHPCS native sniff using them for either handling the new syntax correctly or bowing out.
- [ ] No matter what the decision: add a couple of tests to each of these sniffs to guard against the sniff throwing false positives for the new syntax/fixing things incorrectly etc.

[...]

PR #3527, opened 2022-01-05
PHP 8.1 | PSR12/ConstantVisibility: allow for class constants to be `final`

### LINKED ISSUE #3526: PSR12.Properties.ConstantVisibility false positive when using public final const syntax
**Describe the bug**
False positive found with the example below
**Code sample**
```php
class SampleEnum
{
    public final const SAMPLE_CONST = 'SAMPLE';
}

```

**To reproduce**
1. Create a file called `SampleEnum.php` with the code sample above...
2. Run `phpcs SampleEnum.php ...`
```
Visibility must be declared on all constants if your project supports PHP 7.1 or later
    |         | (PSR12.Properties.ConstantVisibility.NotFound)
```

**Expected behavior**
False positive. It shouldn't show a warning.

 - OS: Linux
 - PHP: 8.1
 - PHPCS: 3.6.2
 - Standard: PSR12

jrfnl:
PHP 8.1 introduces the ability to declare class constants as `final`.

The tokenization of the `final` keyword in this context appears to be consistent PHPCS cross-version, so AFAICS no changes are needed to the PHP tokenizer.

However, the `PSR12.Properties.ConstantVisibility` sniff does need to allow for them.

Includes unit tests.

Ref: https://wiki.php.net/rfc/final_class_const

Fixes #3526

Related to #3479

---

Notes:
* The test failures are unrelated to this PR and PR #3525 is already open to fix those.
* I've done a cursory scan of all sniffs referring to the `T_CONST` token and could not find any other sniffs which would need adjusting for this PHP change.

### DIFF
--- a/src/Standards/PSR12/Sniffs/Properties/ConstantVisibilitySniff.php
+++ b/src/Standards/PSR12/Sniffs/Properties/ConstantVisibilitySniff.php
@@ -47,7 +47,10 @@ class ConstantVisibilitySniff implements Sniff
             return;
         }
 
-        $prev = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($stackPtr - 1), null, true);
+        $ignore   = Tokens::$emptyTokens;
+        $ignore[] = T_FINAL;
+
+        $prev = $phpcsFile->findPrevious($ignore, ($stackPtr - 1), null, true);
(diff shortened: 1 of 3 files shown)


jrfnl:
Rebased to get a passing build.

gsherwood:
Thanks for fixing this

PR #3532, opened 2022-01-12
Add support for diacritics in camel caps names

julienfalque:


jrfnl (commented):
@julienfalque Sorry it's taken so long before this review.

I really like the principle of this chance, however, I do have some questions.

* If I remember correctly, it was possible at some point to compile the PCRE extension without UTF8 support. Should that be taken into account as this package still supports PHP 5.4 ?
* Is there a specific reason to use the POSIX character classes instead of the [escape sequences for Unicode properties](https://www.php.net/manual/en/regexp.reference.unicode.php) ?
    IIRC (again) the POSIX character classes may not always support Unicode depending on the PCRE version. Should that be taken into account as this package still supports PHP 5.4 ?
* Will using the Unicode PCRE option for all uses of this function (while the majority of code is likely using ASCII based names) have an impact on performance ?

julienfalque:
> * If I remember correctly, it was possible at some point to compile the PCRE extension without UTF8 support. Should that be taken into account as this package still supports PHP 5.4 ?

I would say this should be taken into account but I'm not sure there's a reliable way to detect the (lack of) UTF-8 support.

> * Is there a specific reason to use the POSIX character classes instead of the [escape sequences for Unicode properties](https://www.php.net/manual/en/regexp.reference.unicode.php) ?
    IIRC (again) the POSIX character classes may not always support Unicode depending on the PCRE version. Should that be taken into account as this package still supports PHP 5.4 ?

No specific reason, I can use `\p{Lu}`/`\p{Ll}` if you prefer.

> * Will using the Unicode PCRE option for all uses of this function (while the majority of code is likely using ASCII based names) have an impact on performance ?

To be honest I don't know. I guess it will have a negative impact but barely noticeable.

PR #3533, opened 2022-01-17
fix: skip processing of Squiz.Commenting.FunctionComment for @inheritDoc

### LINKED ISSUE #2770: Squiz.Commenting.FunctionComment.MissingReturn @inheritDoc not resolved?

PavelSavushkinMix:
It's supported to use {@inheritDoc}. Add support of @inheritDoc (w/o brackets).
Original PR number - #3051

Closes #2770

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -737,7 +737,7 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
      *                                                  in the stack passed in $tokens.
      * @param int                         $commentStart The position in the stack where the comment started.
      *
-     * @return boolean TRUE if the docblock contains only {@inheritdoc} (case-insensitive).
+     * @return boolean TRUE if the docblock contains {@inheritdoc} or @inheritdoc (case-insensitive).
      */
     protected function checkInheritdoc(File $phpcsFile, $stackPtr, $commentStart)
     {
@@ -751,12 +751,12 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
         for ($i = $commentStart; $i <= $tokens[$commentStart]['comment_closer']; $i++) {
             if (in_array($tokens[$i]['code'], $allowedTokens) === false) {
                 $trimmedContent = strtolower(trim($tokens[$i]['content']));
+	            $allowedInheritDocs = [
+		            '{@inheritdoc}',
+		            '@inheritdoc',
+	            ];
 
-                if ($trimmedContent === '{@inheritdoc}') {
-                    return true;
-                } else {
-                    return false;
-                }
+	            return in_array($trimmedContent, $allowedInheritDocs);
             }
         }
 


jrfnl:
For context:

phpDocumentor states that inheritance is implicit when there is no docblock, so the tag should only be used for partial inheritance via an inline tag `{@inheritDoc}`

Refs:
* https://docs.phpdoc.org/3.0/guide/guides/inheritance.html
* https://docs.phpdoc.org/3.0/guide/references/phpdoc/inline-tags/index.html#inline-tag-reference

The draft PSR-19 reference does allow for the tag without brackets, but then again, is still subject to change as still in draft.

Refs:
* https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#41-making-inheritance-explicit-using-the-inheritdoc-tag
* https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#42-using-the-inheritdoc-inline-tag-to-augment-a-description

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -751,12 +751,12 @@ protected function checkInheritdoc(File $phpcsFile, $stackPtr, $commentStart)
...
+		            '{@inheritdoc}',
+		            '@inheritdoc',
+	            ];
 
-                if ($trimmedContent === '{@inheritdoc}') {
-                    return true;
-                } else {
-                    return false;
-                }
+	            return in_array($trimmedContent, $allowedInheritDocs);
```
jrfnl, line 759: Performance-wise, a straight (strict) comparison against each of the two values will probably be more performant than using `in_array()`.
If `in_array()` stays, the third parameter `$strict` should be set to `true` for a type-safe comparison.

jrfnl (changes requested):
This PR is missing tests. Please add some.

PavelSavushkinMix:
> This PR is missing tests. Please add some.

Thanks for the fast reply.
Added tests.

PavelSavushkinMix:
So, @jrfnl, would you like me to close the PR due to provided references?

Or just keep it till the PSR-19 is released?

src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -1042,7 +1052,7 @@ public function ignored() {
 
 // phpcs:set Squiz.Commenting.FunctionComment specialMethods[] __construct,__destruct
 
-/**
```
jrfnl, line 1045: This space may be there on purpose to test something specific. Might be better to leave it as is.

jrfnl:
> So, @jrfnl, would you like me to close the PR due to provided references?
> 
> Or just keep it till the PSR-19 is released?

Not really up to me. I added the references so the maintainer can make an informed decision.

I'd say, leave it open for now and wait for a response from someone with commit rights.
Be aware that it may be a while before you will get such a response.,

src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
PavelSavushkinMix: Well, I've run `phpunit` w/o any errors.
Could you apply a suggestion, pls, if you still think we need this space?

Thanks.

src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -1042,7 +1052,7 @@ public function ignored() {
 
 // phpcs:set Squiz.Commenting.FunctionComment specialMethods[] __construct,__destruct
 
-/** 
+/**
```
PavelSavushkinMix, line 1055:
```suggestion
/** 
```

Issue #3161, opened 2020-11-09
Change T_GOTO_LABEL to not include colon


gsherwood:
The colon after a goto label does not have to be the very next token; there can be whitespace and comments between them. PHP_CodeSniffer will not tokenize the label correctly unless the colon immediately follows the label.

To fix this, `T_GOTO_LABEL` should be changed to include the label only, and exclude the colon.

So `a:` becomes:
```
T_GOTO_LABEL => a
T_COLON => :
```
Instead of:
```
T_GOTO_LABEL => a:
```
---
And `a /*comment here*/ :` becomes:
```
T_GOTO_LABEL => a
T_WHITESPACE => ·
T_COMMENT => /*comment·here*/
T_WHITESPACE => ·
T_COLON => :
```
Instead of:
```
T_STRING => a
T_WHITESPACE => ·
T_COMMENT => /*comment·here*/
T_WHITESPACE => ·
T_COLON => :
```

PR #3534, opened 2022-01-18
Name of typed enum tokenized as T_GOTO_LABEL

jrfnl:
Follow up on #3478.

When an `enum` would be backed by a type and the colon would directly follow the name of the `enum`, the name would be incorrectly tokenized (including the colon) as `T_GOTO_LABEL`.

Discovered while reviewing #3482

Also related to #3161

/cc @kukulich 

---

### Tests/BackfillEnum: update token calculations

... after the bugfix for `T_GOTO_LABEL`.

Also added a minor tweak to one test case to make sure that spacing around the type colon is handled correctly.


jrfnl:
@gsherwood Could this PR please be milestoned to 3.7.0 ?

gsherwood:

PR #3538, opened 2022-01-25
CodeSniffer.conf: end code with semicolon

jrfnl:
While not problematic, it just feels very untidy for the `CodeSniffer.conf` file not to end the code block with a semicolon.


gsherwood:
Very picky :) but it bugged me when I saw it in the PR too.

jrfnl:
One of those files we never really look at, until we do... 😁

Issue #2898, opened 2020-03-10
Revise exit codes for PHPCS and PHPCBF commands


gsherwood:
Version 4 is a good time to review the exit codes that both scripts produce to ensure that success cases all use `0` and that all failure cases have codes that make sense. Documentation in the wiki to describe the new exist codes should be an output of this work.

jrfnl:
Loosely related open issues:
* #1818 
* #2412 (discoverability of availability of exit code related CLI arguments)

nelson6e65:
Param `--runtime-set ignore_warnings_on_exit 1` works on `phpcs`, but it does not for `phpcbf`.

dfelton:
Is there any documentation somewhere that lists all non `0` or `1` exit codes that exist, and the meaning of them?

Today was the first time I discovered an exit code of `3` is used for when and invalid standard is given. It'd be nice to be able to read a compiled list of what exit codes there are, and I cannot seem to find any information on the wiki on this topic.

jrfnl:
@dfelton I think this is what you are looking for: https://github.com/squizlabs/PHP_CodeSniffer/issues/930#issuecomment-229562566

XedinUnknown:
Those exist codes should probably be part of the docs, not in some obscure comment.

PR #3541, opened 2022-01-30
allow stdin fix to exit sucessfully

PierreAntoineGuillaume:
When submitting a file to phpcbf from stdin, phpcbf always fails.

I propose to make it succeed (exit 0) when no errors were found, given that cbf has no direct concept of "non-fixable errors".

It would allow tools to swap content of a buffered file when exit code is non-zero, and discard if nothing has been fixed.


gsherwood:
Changing a return value would need to target a new major version (4.0 in this case), and that is already being tracked on issue #2898.

Thanks for taking the time to write this PR, but I'm going to close it as a proper plan of the exit codes need to be made so they all make sense together. If you have a proposal, please drop it on that other issue to keep the discussion going.

PR #3542, opened 2022-02-01
Ruleset::explain(): fix plural vs singular phrasing

jrfnl:
When running the `phpcs --standard=Name -e` command, the line at the top of the output would always presumes that a standard contains more than one sniff.
The sniff count for the standards within the standard - `StandardName (# sniff[s])` - already handled this correctly.

Fixed the top line now.

Output without this fix:
```
The DummySubDir standard contains 1 sniffs

DummySubDir (1 sniff)
----------------------
  DummySubDir.Demo.Demo
```

Output with this fix:
```
The DummySubDir standard contains 1 sniff

DummySubDir (1 sniff)
----------------------
  DummySubDir.Demo.Demo
```

### DIFF
--- a/src/Ruleset.php
+++ b/src/Ruleset.php
@@ -249,7 +249,12 @@ class Ruleset
         // one last time and clear the output buffer.
         $sniffs[] = '';
 
-        echo PHP_EOL."The $this->name standard contains $sniffCount sniffs".PHP_EOL;
+        $summaryLine = PHP_EOL."The $this->name standard contains 1 sniff".PHP_EOL;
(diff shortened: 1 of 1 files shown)


jrfnl:
@gsherwood If you're looking for another simple PR and  merge.... 😇

gsherwood:
Bad assumption that a standard would be more than 1 sniff :) Thanks

jrfnl:
> Bad assumption that a standard would be more than 1 sniff :) Thanks

But such an easy assumption to make, as it will be true in the fast majority of cases 😀

PR #3543, opened 2022-02-06
QA: minor code simplification

jrfnl:
Replace a nested function call with a singular function call, which does the same.


gsherwood:
Thanks

PR #708, opened 2015-09-17
Specified array and callback docblock types

scato:
In reaction to #685, I've looked around the code base, but I haven't found any other sniffs that relate to array type hints.

I have found another inconsistency in callable/callback. I've asked the phpDocumenter team for clarification: https://github.com/phpDocumentor/phpDocumentor2/issues/1645

On top of that, I've added tests.

(This PR would also fix #601 and #642.)

PR #3544, opened 2022-02-08
Support generics in docblock

tm1000:
This supports the following:

```
    /**
	 * @param Collection<int, TelephoneNumber> $telephoneNumbers
	 */
	public function myFunction(
		Collection $telephoneNumbers,
```

Where-as before it would say that the typehint needs to be `Collection<int, TelephoneNumber>` now it will correctly deduce that the type hint needs to be `Collection` and will not return an error

I based this off of https://github.com/squizlabs/PHP_CodeSniffer/pull/708 which is similar in scope. 

~~Unfortunately I was unable to run the unit tests because I'm on PHP 8.1 and something in the code continues to try to use the function `each` which does not exist in php 8.0~~

Edit: There is now a test!


jrfnl:
> Unfortunately I was unable to run the unit tests because I'm on PHP 8.1 and something in the code continues to try to use the function `each` which does not exist in php 8.0

Please add tests for this feature.

To run the tests locally on PHP 8.1:
* Run `composer install --ignore-platform-reqs` to get PHPUnit 7.5.
* Then run `vendor/bin/phpunit tests/AllTests.php --no-configuration --bootstrap=tests/bootstrap.php --dont-report-useless-tests`

See: https://github.com/squizlabs/PHP_CodeSniffer/blob/3e6ce10d7f9092d0e7b9cd0a959aba3529952a27/.github/workflows/test.yml#L121-L142

tm1000:
@jrfnl I already did that it does not work because of this library's usage of "each"

```shell
PHP Fatal error:  Uncaught Error: Call to undefined function each() in /usr/src/PHP_CodeSniffer/vendor/phpunit/phpunit/src/Util/Getopt.php:38
Stack trace:
#0 /usr/src/PHP_CodeSniffer/vendor/phpunit/phpunit/src/TextUI/Command.php(245): PHPUnit_Util_Getopt::getopt()
#1 /usr/src/PHP_CodeSniffer/vendor/phpunit/phpunit/src/TextUI/Command.php(120): PHPUnit_TextUI_Command->handleArguments()
#2 /usr/src/PHP_CodeSniffer/vendor/phpunit/phpunit/src/TextUI/Command.php(109): PHPUnit_TextUI_Command->run()
#3 /usr/src/PHP_CodeSniffer/vendor/phpunit/phpunit/phpunit(47): PHPUnit_TextUI_Command::main()
#4 {main}
  thrown in /usr/src/PHP_CodeSniffer/vendor/phpunit/phpunit/src/Util/Getopt.php on line 38
```

tm1000:
@jrfnl got it to work! I had to delete the composer.lock file and vendor dir then run with `--ignore-platform-reqs`

Thanks for pointing that out. much appreciated. There is now a test!

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -419,7 +419,9 @@ protected function processParams(File $phpcsFile, $stackPtr, $commentStart)
 
                 // Check type hint for array and custom type.
                 $suggestedTypeHint = '';
-                if (strpos($suggestedName, 'array') !== false || substr($suggestedName, -2) === '[]') {
+                if (preg_match('/^(.*)<.*>$/', $suggestedName, $matches)) {
```
jrfnl, line 422: 1. For performance optimization, I'd recommend having this condition as the _last_ condition in the if/elseif chain, not the first.
2. I'd also suggest optimizing the regex a little more, like so:
    ```suggestion
                    if (preg_match('/^([^<]+)<[^>]+>$/', $suggestedName, $matches)) {
    ```

src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -1042,7 +1042,19 @@ public function ignored() {
...
+ * Using generic as a type hint should satisfy a specified object parameter type.
+ * @see https://phpstan.org/blog/generics-in-php-using-phpdocs
+ *
+ * @param Collection<int, string> $values An object with int, string pairs.
+ *
+ * @return void
+ */
+public function specifiedArray1(Collection $values) {
+
+}// end specifiedArray1()
```
jrfnl, line 1060: Missing new line at end of file (also for fixed).

jrfnl (commented):
Looks good, just two small notes.

tm1000:
@jrfnl Did as you asked however I could not make the if/then last. I was able to make it second to last. The reasoning is because the last if/then is a "if not in list then do this". The suggestedName before the expression will never be in that array.

jrfnl (approved):

jrfnl:
Linking issue #3589 for visibility.

tm1000:
@jrfnl @gsherwood Now that its been about 9 months what is the process to get this merged?

jrfnl:
@tm1000 Probably not what you want to hear, but patiently wait until Greg has some time to review this. I can review (which I did), but I don't have commit, so there's nothing more I can do.

bsandstrom (approved):

WillemHoman:
@jrfnl would love to see this merged, is that going to be possible?

tm1000:
Ref: https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/750

PR #3546, opened 2022-02-10
Tokenizer/PHP: bug fix - parent/static keywords in class instantiations

jrfnl:
Follow up on #3484 /cc @kukulich 

Just like `new class`, `new parent`, `new self` and `new static` should also be preserved and with the `parent`, `self` and `static` keywords remaining as their dedicated token.

* For `new static`, the tokenization changed due to the context sensitive keywords change. This has now been fixed.
* `new self` was fine before and is still fine.
* `new parent` apparently wasn't handled correctly, even before the change. The condition which was in place for handling the same situation for `self` has now been updated to also handle `parent`.

Includes unit tests.

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -608,8 +608,11 @@ class PHP extends Tokenizer
             ) {
                 $preserveKeyword = false;
 
-                // `new class` should be preserved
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks for fixing this

PR #3551, opened 2022-02-25
PSR2/SwitchDeclaration: bug fix when determining terminating statement

### LINKED ISSUE #3550: False positive from PSR2.ControlStructures.SwitchDeclaration.TerminatingComment when using trailing comment
**Describe the bug**
A false positive is reported if a `switch` case contains an `if`/`else` structure where all branches exit the switch, and one of the branches ends with a comment. Minimal example is below; the error is also reported if one or more `elseif` branches exist, if the comment is on a different branch, if the comment is on the line after the `break` statement, or if `return` is used in place of `break`. No error is reported if the comment is removed, or is on the line before the `break`.

**Code sample**
```php
<?php

switch (rand()) {
    case 1:
        if (rand() === 1) {
            break;
        } else {
            break; // comment
        }
    default:
        break;
}
```

**Custom ruleset**
```xml
<?xml version="1.0"?>
  <rule ref="PSR2.ControlStructures.SwitchDeclaration"/>
</ruleset>
```

**To reproduce**
2. Run `phpcs test.php ...`
```
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
 4 | ERROR | There must be a comment when fall-through is intentional in a
   |       | non-empty case body
   |       | (PSR2.ControlStructures.SwitchDeclaration.TerminatingComment)
--------------------------------------------------------------------------------
```

**Expected behavior**
No error should be reported.

**Versions:**
 - OS: macOS 12.2.1
 - PHP: 8.3.1
 - PHPCS: 3.6.2
 - Standard: -

**Additional context**
-

jrfnl:
Trailing comments within control structures nested within a switch case would break the determination of whether or not there is a terminating statement within the nested control structure.

Making the `findNestedTerminator()` method look for non-empty instead of non-whitespace tokens fixes that and shouldn't break the `TerminatingComment` check as that has it's own check whether the last token in the case statement is a comment.

Includes unit tests.

Fixes #3550

### DIFF
--- a/src/Standards/PSR2/Sniffs/ControlStructures/SwitchDeclarationSniff.php
+++ b/src/Standards/PSR2/Sniffs/ControlStructures/SwitchDeclarationSniff.php
@@ -247,7 +247,7 @@ class SwitchDeclarationSniff implements Sniff
     {
         $tokens = $phpcsFile->getTokens();
 
-        $lastToken = $phpcsFile->findPrevious(T_WHITESPACE, ($end - 1), $stackPtr, true);
+        $lastToken = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($end - 1), $stackPtr, true);
         if ($lastToken === false) {
             return false;
         }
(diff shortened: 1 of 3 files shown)


zorac:
I've tested this with a number of variations on my original test case, and it works perfectly. Many thanks for the speedy patch!

jrfnl:
@zorac Thanks for testing & confirming the fix!

gsherwood:
Thanks for the fix

PR #3552, opened 2022-02-28
PHP 8.1 | Tokenizer/PHP: bug fix for overeager explicit octal notation backfill

jrfnl:
Follow up on #3481. /cc @MarkBaker 

Just like for all other type of integer notations, if a numeric literal separator is used, it is not allowed between the prefix and the actual number.
```php
// This is fine.
$b = 0b1_0;
$o = 0o6_3;

// This is an invalid use of the numeric literal separator.
$b = 0b_10;
$o = 0o_63;
```

This PR fixes the backfill for explicit octal notation to NOT backfill these type of invalid sequences as the inconsistent tokenization across PHP versions which that causes, can create havoc in sniffs.

Includes adding additional unit tests.

@gsherwood Greg - can this PR please be earmarked for PHPCS 3.7.0 to prevent the incorrect tokenization getting into a released version ?

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -728,7 +728,8 @@ class PHP extends Tokenizer
                 && (isset($tokens[($stackPtr + 1)]) === true
                 && is_array($tokens[($stackPtr + 1)]) === true
                 && $tokens[($stackPtr + 1)][0] === T_STRING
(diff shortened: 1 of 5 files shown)


gsherwood:
Thanks a lot for finding and fixing this before release.

jrfnl:
Thanks for merging this!

PR #3566, opened 2022-03-23
Fixed return types for ForbiddenFunctionsUnitTest

### LINKED ISSUE #3565: Wrong types for tests of forbidden functions
**Describe the bug**
In PHPStorm by JetBrains jetbrains.com code highlights red, because he scanning vendor directory and find  global functions in ForbiddenFunctionsUnitTest.inc. Exclude directory not helping. 

Example:
Screenshot 2022-03-23 at 19 06 59

**Code sample**
<?php

function test(int $b) {}

$a = count([1]);
test($a);
```

ilyapishchulin:
Fixed #3565

### DIFF
--- a/src/Standards/Generic/Tests/PHP/ForbiddenFunctionsUnitTest.inc
+++ b/src/Standards/Generic/Tests/PHP/ForbiddenFunctionsUnitTest.inc
@@ -19,7 +19,9 @@ $class->unset($filepath);
 
 function delete() {}
 function unset() {}
+/** @return int */
 function sizeof() {}
+/** @return int */
 function count() {}
 
 trait DelProvider {


jrfnl (commented):
This is test code from PHPCS which should not be adjusted for misconfigured IDEs of projects not-PHPCS.

Also: you've just broken these tests as the the test is based on line numbers and those have now jumped.

ilyapishchulin:
It’s not about misconfigured IDEs. I can fix tests, it’s will be ok?

jrfnl:
> Thanks for answer. It’s not about misconfigured IDEs. I can fix tests, it’s will be ok?

IMO no as, as I said before, you are changing a test file for PHPCS and adding something to it which is completely unrelated to the sniff test.

I know there are ways to tell PHPStorm to exclude certain directories from the code completion. I suggest you look into that.

ilyapishchulin:
Like I already said, exclude directory won't work. Ok, thanks!

PR #3573, opened 2022-03-31
PHP 8.1 | Add `enum` support to various additional sniffs

jrfnl:
Follow up on #3482 as per https://github.com/squizlabs/PHP_CodeSniffer/pull/3482#issuecomment-1016007608

### Squiz/ScopeClosingBrace: add tests for enum support


gsherwood:
Thanks!

PR #3574, opened 2022-03-31
PHP 8.1 | Generic/LowerCaseKeyword: simplify registered tokens + add `enum` support

jrfnl:
This is a PR to simplify maintenance of this sniff.

Since PR #3484, the `Tokens` class contains a `$contextSensitiveKeywords` property which largely overlaps with the list of tokens registered for this sniff.

All tokens in that list should always be included in the targets for this sniff anyway, so we may as well leverage the new token array.

Includes adding the new custom `T_ENUM_CASE` token to the list of additional tokens to take into account.

Includes unit test.


gsherwood:
Thanks for enum case support and taking the time for the cleanup

PR #3575, opened 2022-03-31
Squiz.Scope.MethodScope misses visibility keyword on previous line

jrfnl:
The `Squiz.Scope.MethodScope` sniff is intended to check whether each method has visibility declared, but would in actual fact also enforce that the visibility should be on the same line as the `function` keyword as it stopped searching for the visibility keyword once it reached the start of the line.

Fixed now by using the `File::getMethodProperties()` method for determining visibility instead of searching with sniff specific logic.

Includes unit test.


gsherwood:
A nice clean fix to the problem. Thanks for fixing it.

jrfnl:
Happy to help.

PR #3577, opened 2022-03-31
PHP 8.1 | File::getMemberProperties(): handle `enum` properties same as interface properties

jrfnl:
Neither enums nor interfaces support properties being declared on them.

This adjusts the `File::getMemberProperties()` method to handle properties declared on an `enum` in the same way as properties declared on an `interface` were already being handled.

Includes unit test.


gsherwood:
Thanks

PR #3578, opened 2022-04-01
PHP 8.1 | File::findImplementedInterfaceNames(): add support for enums implementing interfaces

jrfnl:
An `enum` declaration can implement one or more interfaces.

This commit updates the `File::findImplementedInterfaceNames()` method to allow it to find the interfaces implemented by an `enum` as well.

Includes unit tests.

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -2765,11 +2765,11 @@ class File
 
 
     /**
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks

PR #3582, opened 2022-04-22
PHP 8.1: Support of `never` type

kukulich:


tests/Core/File/GetMethodPropertiesTest.inc
```diff
@@ -126,3 +126,10 @@ interface FooBar {
 /* testPHP8DuplicateTypeInUnionWhitespaceAndComment */
 // Intentional fatal error - duplicate types are not allowed in union types, but that's not the concern of the method.
 function duplicateTypeInUnion(): int | /*comment*/ string | INT {}
+
+/* testPHP81NeverType */
+function never(): never {}
+
+/* testPHP81NulltableNeverType */
```
jrfnl, line 133: 👍🏻  on adding this test.

```suggestion
/* testPHP81NullableNeverType */
```

tests/Core/File/GetMethodPropertiesTest.php
```diff
@@ -728,6 +728,52 @@ public function testPHP8DuplicateTypeInUnionWhitespaceAndComment()
...
+
+    }//end testPHP81NeverType()
+
+
+    /**
+     * Verify recognition of PHP8.1 type "never"  with (illegal) nullability.
+     *
+     * @return void
+     */
+    public function testPHP81NulltableNeverType()
```
jrfnl, line 759:
```suggestion
    public function testPHP81NullableNeverType()
```

jrfnl (approved):
Thanks a lot for adding these extra tests to the File and Tokenizer test classes too!

Changes look A-okay to me, tested and confirmed working.

@gsherwood Would be great if this could still go into 3.7.0 🙏🏻

tests/Core/File/GetMethodPropertiesTest.inc
kukulich: Thanks.

gsherwood:
Thanks

PR #3595, opened 2022-05-18
PHP 8.1: Added missing tests of intersection types for `LowerCaseTypeSniff`

kukulich:
Removed during rebase by mistake...


gsherwood:
Thanks for getting those back in

PR #3608, opened 2022-06-12
[Hot fix] Tokenizer/PHP: 2 bug fixes for improved context sensitive keyword support

### LINKED ISSUE #3607: Method with name `eval` is always reported as error
**Describe the bug**

PHPCS considers methods with name `eval` to be incorrect

**Code sample**

```php
<?php

declare(strict_types=1);

namespace App;

class Foo
{
    public function eval()
    {
        return;
    }
}
```

**Custom ruleset**
No custom rules. Just PSR12 standrad

**To reproduce**

2. Run `phpcs --standard=PSR12 test.php`

```
 9 | ERROR | Expected "function abc(...)"; found "function abc(...)"
```

If you change method name to basically anything else - `eval1`, `evaluate` etc - there is no error

**Expected behavior**
No complains.

 - OS: Ubuntu 20.04
 - PHP: 7.4.3
 - PHPCS: master (5e27199)
 - Standard: PSR12

Same thing happens with PHP 8.1.6

jrfnl:
### Tokenizer/PHP: bug fix in improved context sensitive keyword support [1]

As reported in #3607, the `eval` keyword was not included in the list of context sensitive keyword. This is a regression compared to PHPCS 3.6.2.

Fixed now, including unit tests.

Fixes #3607

### Tokenizer/PHP: bug fix in improved context sensitive keyword support [2]

While investigating #3607, I figured adding a test with a function declared to return by reference would also not be amiss and found that that situation was not accounted for.

This commit fixes that.
Includes unit test.

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -607,38 +607,57 @@ class PHP extends Tokenizer
 
             if ($tokenIsArray === true
                 && isset(Util\Tokens::$contextSensitiveKeywords[$token[0]]) === true
-                && isset($this->tstringContexts[$finalTokens[$lastNotEmptyToken]['code']]) === true
+                && (isset($this->tstringContexts[$finalTokens[$lastNotEmptyToken]['code']]) === true
+                || $finalTokens[$lastNotEmptyToken]['content'] === '&')
             ) {
-                $preserveKeyword = false;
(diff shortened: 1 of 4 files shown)


jrfnl:
@gsherwood Both of the above fixes are fixes for regressions against PHPCS 3.6.2. With that in mind, I believe this PR should still be included in PHPCS 3.7.0.

gsherwood:
Thanks so much for fixing this, and so quickly!

PR #3610, opened 2022-06-13
Tokenizer/PHP: bug fix in improved context sensitive keyword support

### LINKED ISSUE #3609: Method with name `empty` is always reported as error
**Describe the bug**
PHPCS considers methods with name `empty` to be incorrect

**Code sample**
```php
<?php

declare(strict_types=1);

namespace App;

class Foo
{
    public function empty()
    {
        return;
    }
}
```

**Custom ruleset**
No custom rules. Just PSR12 standard

**To reproduce**
2. Run `phpcs --standard=PSR12 test.php`
```
 9 | ERROR | Expected "function abc(...)"; found "function abc(...)"
```

If you change method name to basically anything else - `empty1`, `createEmpty` etc - there is no error

**Expected behavior**
No complains.

 - OS: Ubuntu 20.04
 - PHP: 8.1.6
 - PHPCS: 3.7.0
 - Standard: PSR12

jrfnl:
As reported in #3609, the `empty` keyword, as well as the `isset` and `unset` keywords, was not included in the list of context sensitive keyword. This is a regression compared to PHPCS 3.6.2.

Fixed now, including unit tests.

Note: I've now done a full comparison with the [reserved keyword list as per the PHP manual](https://www.php.net/manual/en/reserved.keywords.php). This should hopefully fix the remaining stragglers.

Fixes #3609


jrfnl:
And merged 45 minutes later, we're on a roll 🛼

kukulich:
@jrfnl Thank you. I used the list from https://wiki.php.net/rfc/context_sensitive_lexer and it looks it was not complete...

PR #3625, opened 2022-06-30
Squiz/FunctionOpeningBraceSpace: handle class-like structures

jtojnar:


jtojnar:
Maybe the Sniff should be renamed?

gsherwood:
That's not the purpose of this sniff. There are other that do OO structures.

If you want the brace on the same line, include the rule `Generic.Classes.OpeningBraceSameLine.BraceOnNewLine`
If you want the brace of the next line, include the rule `PSR2.Classes.ClassDeclaration.OpenBraceNewLine`

Closing as I'm not going to merge this into the target sniff.

jtojnar:
@gsherwood This is not about positioning the brace, I already use PSR 2.

But it looks like this is already covered by `PHP_CodeSniffer\Standards\PSR12\Sniffs\Classes\OpeningBraceSpaceSniff`. Only it is impossible to find.

PR #3629, opened 2022-07-09
PHP 8.2 | Ruleset: prevent notices about dynamic properties being set

### LINKED ISSUE #3489: PHP 8.2: dynamic properties are deprecated
Recently, the [Deprecate dynamic properties RFC](https://wiki.php.net/rfc/deprecate_dynamic_properties) has been approved for PHP 8.2 and as I expected it will cause havoc, I've started doing some test runs against various repos, including PHPCS.

The test run against PHPCS only went to proof my suspicion as this deprecation (or rather the fatal error this will become in PHP 9.0) breaks currently supported functionality in PHPCS.

### Context

The deprecation of dynamic properties basically means that any property which is not explicitly declared in a class, but _is_ being get/set will now yield a "Creation of dynamic property ClassName::$propertyName is deprecated" deprecation notice, which will become a fatal error in PHP 9.0.

There are three exceptions to this:
* Dynamic properties are still supported when the class implements the magic `__get()` and `__set()` methods.
* Dynamic properties are still supported when a class extends `stdClass` (which implements the magic `__get()` and `__set()` methods).
* Dynamic properties are still supported when a class is given the `#[AllowDynamicProperties]` attribute, though this attribute is expected to only be supported for a limited time (until ~PHP 9.0) as the intention is to eventually remove support for dynamic properties completely from PHP (with the exception of the above two situations).

### What problem does this cause in PHPCS ?

In PHPCS, the value of `public` properties can be adjusted from within a ruleset and from within (test) files using the `phpcs:set` annotation.

While this feature is _mostly_ used for adjusting the properties for individual sniffs, there are a number of (external) standards which use the same property in a multitude of sniffs.

As things are, PHPCS currently explicitly supports setting the value for a (`public`) property for all sniffs in a category/standard in one go, like so:
```xml
    <rule ref="PSR1">
        <properties>
            <property name="setforallsniffs" value="true" />
        </properties>
    </rule>
```

This is also documented and safeguarded as such via the `RuleInclusionTest`.

The deprecation/removal of support for dynamic properties breaks this functionality.

### What are our options ?

1. **Add the `#[AllowDynamicProperties]` attribute to every sniff**.
    **Impact**: **HIGH** for sniff maintainers, no impact for users.
    **Breaking change**: **No**
    **Stable**: **No.** The attribute would also need to be added to every single sniff out in the wild in external standards, so this will break as soon as an external standard does not have the attribute. It will also break (again) when support for the attribute is removed from PHP.
2. **Add the magic `__get()` and `__set()` methods to the `Sniff` interface**.
    **Impact**: **HIGH** for sniff maintainers, no impact for users.
    **Breaking change**: **Yes** as every single sniff class, both internal and external, would need to implement these methods.
    This change could only be made for PHPCS 4.0, but will allow for standards to be cross-version compatible with PHPCS 3.x and 4.x without extra effort.
    **Stable**: **Yes**, for those standards which choose to support PHPCS 4.x and make this change.
3. **Add a new `AbstractSniff` base class which includes the magic `__get()` and `__set()` methods**.
    **Impact**: **HIGH** for sniff maintainers, no impact for users.
    **Breaking change**: **Yes** as the class declaration for every single sniff, both internal and external would need to be updated from `SniffName implements Sniff` to `SniffName extends Sniff`. This change could only be made for PHPCS 4.0 and will make it a lot more complicated for standards to be cross-version compatible with PHPCS 3.x and 4.x (for those who desire this).
    **Stable**: **Yes**, for those standards which choose to support PHPCS 4.x and make this change.
4. **Make all sniffs extend `stdClass`**.
    **Impact**: **HIGH** for sniff maintainers, no impact for users.
    **Breaking change**: **No** The class declaration for every single (abstract) sniff, both internal and external would need to be updated from `SniffName implements Sniff` to `SniffName extends stdClass implements Sniff`, but this is not enforced by PHPCS, so standard maintainers can implement this whenever they are getting ready to support PHP 8.2. This change would also be cross-version compatible with both PHPCS 3.x as well as 4.x (for those who desire this).
    **Stable**: **Yes**, for those standards which choose to make this change.
5. **Remove support for setting properties for categories/standards in one go**.
    **Impact**: none for sniff maintainers, **MEDIUM** impact for users.
    Any custom ruleset/standard which currently uses this feature will need to be adjusted and for those standards where this feature is commonly used, this will mean fiddly ruleset adjustments - the property would need to be explicitly set for every sniff using it, new sniffs added to a standard would not automatically receive the property anymore etc
    While this feature may not be used in a huge amount of standards, for those standards - and their users - where it is used, it will cause a lot of churn in continuous ruleset adjustments now and in the future.
    **Breaking change**: **Yes** A feature which has always been supported would now be removed.
    **Stable**: **No**, the need to continuously update rulesets whenever new sniffs using common properties come out, make this unstable for end-users.
6. **Only set properties received from a ruleset when the property exists on a sniff**.
    ... and throw an informative error when a ruleset attempts to set a property on a sniff which doesn't have that property.
    The error message should only be thrown when the property is being set on an individual sniff, not when the property is (attempted to be) set for a standard/category of sniffs.
    **Impact**: **LOW/MEDIUM** for sniff maintainers, **LOW** impact for users.
    **Breaking change**: **Yes, sort of**.
    While probably rare, there may be a few standards out there relying on "magic dynamic" properties which may or may not be available to the sniff depending on the ruleset under which a sniff is run. Those standards would break with this change, but that break can be mitigated by the standard maintainers by either adding the magic `__get()`/`__set()`/`__isset()` methods or making the property/properties explicit on each sniff.
    **Stable**: **Yes, sort of** and the informative error message for typos in properties would be in line with the "goal" of the PHP RFC.
  

### What will not work

* Adding the `#[AllowDynamicProperties]` attribute to the `Sniff` interface, for it to be inherited by implementing classes, is not an option as it will result in a `Fatal error:  Cannot apply #[AllowDynamicProperties] to interface`.

### Proposal

All things considered, I'd like to propose implementing option 6 for the following reasons:
* The change would not be bound to PHPCS 4.x, which would put pressure on the roadmap for PHPCS 4.x as it would then be closely tied to the PHP 8.2 release schedule,
* The change will yield improved error messages for users with typos in property names in custom ruleset, which is in line with the PHP RFC and would help end-users of PHPCS.
* While it does have the potential to break (a limited set of) external standards, I suspect this will be exceedingly rare and that far more standards will benefit from the continued support for setting properties for a category/standard in one go.

Opinions ?

jrfnl:
### :warning: This PR contains two - albeit small - BC-breaks! :warning:

The BC-breaks can be broken down as follows:

### For end-users

If a property is set via a ruleset for an **_individual_** sniff and that sniff does not have the property explicitly declared, nor does the sniff declare the PHP magic `__set()` method or `extend stdClass`, the ruleset will now be regarded as invalid.

In practice, this means the PHPCS run will halt and will show an informative notice to the end-user about the invalid property setting and the PHPCS run will exit with exit code `3`.

image

For properties being set for a complete category of sniffs or a complete standard, the PHPCS run will not halt, nor show an error when the property isn't supported for all sniffs in the category/standard.
In that case, the property will only be set on those sniffs which support it (i.e. either have the property explicitly declared on the sniff, or have the magic `__set()` method declared on the sniff, or `extend stdClass`) and will be silently ignored for all other sniffs.

:point_right: Aside from possibly surfacing some incidental typos in property settings in rulesets, I expect this change to go mostly unnoticed by end-users.

### For developers of PHPCS standards/integrations

1. The format of the `$ruleset['sniff code']['properties']['property name']` sub-array has changed from a `array|string` type property value to a sub-array containing two entries:
    1. `'value'` - containing the `array|string` type property value.
    2. `'scope'` - containing a string, either `sniff` or `standard`, to indicate whether the ruleset property directive was for an individual sniff or for a whole standard/category of sniffs.
2. The third parameter for the `Ruleset::setSniffProperty()` method has been changed to expect the above mentioned array of `$settings` instead of just a property value.

This change is necessary to allow for throwing the error notice when an invalid property is being set on an invalid sniff versus silently ignoring the invalid property if the ruleset specified it on a standard/category.

:point_right: The impact of this BC-break is expected to be small, as it would require for an external standard or integration of PHPCS to:
* `extend` the `Ruleset` class **and** overload the `setSniffProperty()` method to have any impact;
* or to call the `setSniffProperty()` method on the `Ruleset` object directly (which _may_ happen in custom test frameworks for external standards).

Note:
* The change to the `processRule()` method cannot be overloaded as it is a `private` method.
* The change in the `populateTokenListeners()` method is only cosmetic and has no functional impact.

---

### Technical choices made:

For consistent handling of properties set in a (custom) ruleset across PHP versions, I have chosen to only support setting properties when:
* A property is explicitly declared on a sniff class.
* A sniff class explicitly declares (or inherits) the magic `__set()` method.
* A sniff class `extends` `stdClass`.

Note: no additional check has been added to verify that a declared property is `public`. If that is not the case a PHP native error would be thrown previously and still will be now. This behaviour has not changed.

I have chosen **not** to respect the `#[\AllowDynamicProperties]` attribute as it is not possible (without tokenizing the sniff files which would create a lot of overhead) to determine whether that attribute exists on a class when PHPCS runs on PHP < 8.0 as the Reflection method needed to detect the attribute is only available on PHP 8.0+.

In other words, adding support for the attribute would introduce an inconsistency in how properties set in a ruleset are handled based on the PHP version on which PHPCS is being run.
It would also mean that the error notice for invalid properties set on individual sniffs would only be available on PHP 8.2+, which would greatly reduce the usefulness of the error notice.

In my opinion, consistency here is more important than supporting the attribute, especially as there are three fully supported ways to allow for supporting properties to be set from a ruleset.
The three above mentioned ways to allow for setting properties from a ruleset for a sniff are all fully supported on all PHP versions currently supported by PHPCS, so there is no compelling reason to also allow for the attribute.

### Suggested changelog entry

> PHP 8.2: prevent deprecation notices for properties set in a (custom) ruleset for complete standards/complete sniff categories
> - Invalid properties set for individual sniffs will now result in an error and halt the execution of PHPCS.
> - Properties set for complete standards/complete sniff categories will now only be set on sniffs which explicitly support the property.
>       The property will be silently ignored for those sniffs which do not support the property.
> - For sniff developers: it is strongly recommended for sniffs to explicitly declare any user-adjustable `public` properties.
>    If dynamic properties need to be supported for a sniff, either declare the magic `__set()`/`__get()`/`__isset()`/`__unset()` methods on the sniff or let the sniff extend `stdClass`.
>    Note: The `#[\AllowDynamicProperties]` attribute will have **no effect** for properties which are being set in rulesets.
> - Sniff developers/integrators of PHPCS may need to make some small adjustments to allow for changes to the PHPCS internals. See the PR description of PR #3629 for full details.

### Other notes

The changes are accompanied by a full set of tests covering the change.
There is a little overlap between the `RuleInclusionTest` class and the new `SetSniffPropertyTest` class, but for the most part, the tests compliment each other.

Includes:
* Correct handling of properties being set via `phpcs:set` annotations.
* The `RuleInclusionTest` XML fixture and test expectations have been adjusted to allow for these changes.
    1. As we now need "settable" properties to exist on sniffs to use for the `testSettingProperties()` test, the `PSR1` ruleset is not sufficient as the sniffs in that ruleset don't have public properties.
        For that reason, the "set property for complete standard" test has been changed to use `PSR2` instead and set the `indent` property.
        The test expectations has been adjusted to match.
    2. Along the same lines, the `Zend.NamingConventions` sniffs don't have public properties, so the "setting property for complete category" test has been switched to `PSR12.Operators` and set the `ignoreSpacingBeforeAssignments` property.
        The test expectations has been adjusted to match.
    3. In both these cases, the XML `<rule>` now contains both a property which is valid for at least some sniffs in the standard/category and a property which is invalid in for all sniffs.
        For the _invalid_ properties, a separate `testSettingInvalidPropertiesOnStandardsAndCategoriesSilentlyFails()` test with data provider has been added to verify these will not be set.
    4. As the ruleset has changed, the expectations for the `testHasSniffCodes()` test and the `testRegisteredSniffCodes()` test have been updated to match.
    5. The test descriptions for the test cases in the `dataSettingProperties()` have also been updated to make it more explicit what each test case is testing.
* The PHPCS ruleset has been updated to exclude sniffs which function as test fixtures from the PHPCS check for this repo.
* A minor tweak has been made to the `ValidatePEARPackageXML` script to prevent a warning about sniffs which function as test fixtures.
    These test fixtures should receive the `role="test"` attribute, not a `role="php"` attribute.

Fixes #3489

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -138,6 +138,16 @@ http://pear.php.net/dtd/package-2.0.xsd">
       </dir>
      </dir>
      <dir name="Ruleset">
+      <dir name="Fixtures">
+       <dir name="Sniffs">
+        <dir name="Category">
+         <file baseinstalldir="" name="SetPropertyAllowedAsDeclaredSniff.php" role="test" />
+         <file baseinstalldir="" name="SetPropertyAllowedViaMagicMethodSniff.php" role="test" />
+         <file baseinstalldir="" name="SetPropertyAllowedViaStdClassSniff.php" role="test" />
+         <file baseinstalldir="" name="SetPropertyNotAllowedViaAttributeSniff.php" role="test" />
+        </dir>
+       </dir>
+      </dir>
       <file baseinstalldir="" name="RuleInclusionTest.php" role="test" />
       <file baseinstalldir="" name="RuleInclusionTest.xml" role="test" />
       <file baseinstalldir="" name="RuleInclusionTest-include.xml" role="test" />
@@ -145,6 +155,15 @@ http://pear.php.net/dtd/package-2.0.xsd">
       <file baseinstalldir="" name="RuleInclusionAbsoluteLinuxTest.xml" role="test" />
       <file baseinstalldir="" name="RuleInclusionAbsoluteWindowsTest.php" role="test" />
       <file baseinstalldir="" name="RuleInclusionAbsoluteWindowsTest.xml" role="test" />
+      <file baseinstalldir="" name="SetSniffPropertyTest.php" role="test" />
+      <file baseinstalldir="" name="SetPropertyAllowedAsDeclaredTest.xml" role="test" />
+      <file baseinstalldir="" name="SetPropertyAllowedViaMagicMethodTest.xml" role="test" />
+      <file baseinstalldir="" name="SetPropertyAllowedViaStdClassTest.xml" role="test" />
+      <file baseinstalldir="" name="SetPropertyAppliesPropertyToMultipleSniffsInCategoryTest.xml" role="test" />
+      <file baseinstalldir="" name="SetPropertyThrowsErrorOnInvalidPropertyTest.xml" role="test" />
+      <file baseinstalldir="" name="SetPropertyNotAllowedViaAttributeTest.xml" role="test" />
+      <file baseinstalldir="" name="SetPropertyDoesNotThrowErrorOnInvalidPropertyWhenSetForStandardTest.xml" role="test" />
+      <file baseinstalldir="" name="SetPropertyDoesNotThrowErrorOnInvalidPropertyWhenSetForCategoryTest.xml" role="test" />
      </dir>
      <dir name="Sniffs">
       <file baseinstalldir="" name="AbstractArraySniffTest.inc" role="test" />
@@ -2128,6 +2147,19 @@ http://pear.php.net/dtd/package-2.0.xsd">
    <install as="CodeSniffer/Core/Ruleset/RuleInclusionAbsoluteLinuxTest.xml" name="tests/Core/Ruleset/RuleInclusionAbsoluteLinuxTest.xml" />
    <install as="CodeSniffer/Core/Ruleset/RuleInclusionAbsoluteWindowsTest.php" name="tests/Core/Ruleset/RuleInclusionAbsoluteWindowsTest.php" />
    <install as="CodeSniffer/Core/Ruleset/RuleInclusionAbsoluteWindowsTest.xml" name="tests/Core/Ruleset/RuleInclusionAbsoluteWindowsTest.xml" />
+   <install as="CodeSniffer/Core/Ruleset/SetSniffPropertyTest.php" name="tests/Core/Ruleset/SetSniffPropertyTest.php" />
+   <install as="CodeSniffer/Core/Ruleset/SetPropertyAllowedAsDeclaredTest.xml" name="tests/Core/Ruleset/SetPropertyAllowedAsDeclaredTest.xml" />
+   <install as="CodeSniffer/Core/Ruleset/SetPropertyAllowedViaMagicMethodTest.xml" name="tests/Core/Ruleset/SetPropertyAllowedViaMagicMethodTest.xml" />
+   <install as="CodeSniffer/Core/Ruleset/SetPropertyAllowedViaStdClassTest.xml" name="tests/Core/Ruleset/SetPropertyAllowedViaStdClassTest.xml" />
+   <install as="CodeSniffer/Core/Ruleset/SetPropertyAppliesPropertyToMultipleSniffsInCategoryTest.xml" name="tests/Core/Ruleset/SetPropertyAppliesPropertyToMultipleSniffsInCategoryTest.xml" />
+   <install as="CodeSniffer/Core/Ruleset/SetPropertyNotAllowedViaAttributeTest.xml" name="tests/Core/Ruleset/SetPropertyNotAllowedViaAttributeTest.xml" />
+   <install as="CodeSniffer/Core/Ruleset/SetPropertyDoesNotThrowErrorOnInvalidPropertyWhenSetForCategoryTest.xml" name="tests/Core/Ruleset/SetPropertyDoesNotThrowErrorOnInvalidPropertyWhenSetForCategoryTest.xml" />
+   <install as="CodeSniffer/Core/Ruleset/SetPropertyDoesNotThrowErrorOnInvalidPropertyWhenSetForStandardTest.xml" name="tests/Core/Ruleset/SetPropertyDoesNotThrowErrorOnInvalidPropertyWhenSetForStandardTest.xml" />
+   <install as="CodeSniffer/Core/Ruleset/SetPropertyThrowsErrorOnInvalidPropertyTest.xml" name="tests/Core/Ruleset/SetPropertyThrowsErrorOnInvalidPropertyTest.xml" />
+   <install as="CodeSniffer/Core/Ruleset/Fixtures/Sniffs/Category/SetPropertyAllowedAsDeclaredSniff.php" name="tests/Core/Ruleset/Fixtures/Sniffs/Category/SetPropertyAllowedAsDeclaredSniff.php" />
+   <install as="CodeSniffer/Core/Ruleset/Fixtures/Sniffs/Category/SetPropertyAllowedViaMagicMethodSniff.php" name="tests/Core/Ruleset/Fixtures/Sniffs/Category/SetPropertyAllowedViaMagicMethodSniff.php" />
+   <install as="CodeSniffer/Core/Ruleset/Fixtures/Sniffs/Category/SetPropertyAllowedViaStdClassSniff.php" name="tests/Core/Ruleset/Fixtures/Sniffs/Category/SetPropertyAllowedViaStdClassSniff.php" />
+   <install as="CodeSniffer/Core/Ruleset/Fixtures/Sniffs/Category/SetPropertyNotAllowedViaAttributeSniff.php" name="tests/Core/Ruleset/Fixtures/Sniffs/Category/SetPropertyNotAllowedViaAttributeSniff.php" />
    <install as="CodeSniffer/Core/Sniffs/AbstractArraySniffTest.inc" name="tests/Core/Sniffs/AbstractArraySniffTest.inc" />
    <install as="CodeSniffer/Core/Sniffs/AbstractArraySniffTest.php" name="tests/Core/Sniffs/AbstractArraySniffTest.php" />
    <install as="CodeSniffer/Core/Sniffs/AbstractArraySniffTestable.php" name="tests/Core/Sniffs/AbstractArraySniffTestable.php" />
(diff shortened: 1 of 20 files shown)


jrfnl:
> or to call the `setSniffProperty()` method on the `Ruleset` object directly (which may happen in custom test frameworks for external standards).

I could possibly make this BC-break smaller if so desired.

My thoughts on this:
* Check if the received param is an array in the expected (new) format or a string/array in another format (= value).
* If not the expected (new) format, use the parameter value as the "value" and set the "scope" to `standard` + throw a deprecation notice to alert integrators.
* The deprecation notice + code handling the "old" parameter format could then be removed in PHPCS 4.0.0.

Would this be desired ?

A [code search](https://sourcegraph.com/search?q=context:global+-%3EsetSniffProperty+lang:php+-file:%28%5E%7C/%29vendor/+fork:no+archived:no&patternType=regexp) shows me that at least the following external standards/integrations would/will be impacted by the change:
* [VariableAnalysis](https://github.com/sirbrillig/phpcs-variable-analysis/) /cc @sirbrillig
* [ImportDetection](https://github.com/sirbrillig/phpcs-import-detection) /cc @sirbrillig
* [HumanMade](https://github.com/humanmade/coding-standards)
* [Inpsyde](https://github.com/inpsyde/php-coding-standards) /cc @gmazzap
* [Commit Hook](https://github.com/andkirby/commithook/) /cc @andkirby
* [Neutron Coding Standard](https://github.com/automattic/phpcs-neutron-standard) /cc @sirbrillig

gmazzap:
Thank you for the excellent work on this and the ping.

As per the change, this will affect us on the _tests_ for the custom sniffs, so it will not impact CI for production libraries. I will monitor this issue and fix the tests when this is merged.

Thanks again.

jrfnl:
Ha! Managed to still find one bug in the original PR. (support for array properties with `extend="true"` was broken)

As it looks like no-one has started reviewing this PR yet, I've fixed that up in the original commit and made a small tweak to the tests to make sure that situation will always be checked as well.

benjaminprojas:
Any traction on this? I'm not 100% sure if this will fix it, but VS Code no longer runs PHPCS because of the PHP 8.2 errors thrown through these files. My specific error is:

```
PHP Deprecated:  Creation of dynamic property WordPressCS\WordPress\Sniffs\WP\I18nSniff::$check_translator_comments is deprecated in /Users/benjamin/Work/Awesome Motive/AIOSEO/git/aioseo/vendor/squizlabs/php_codesniffer/src/Ruleset.php on line 1331
```

jrfnl:
> Any traction on this? I'm not 100% sure if this will fix it, but VS Code no longer runs PHPCS because of the PHP 8.2 errors thrown through these files. My specific error is:

This PR would turn that notice into a ruleset error as described in the "For end-users" section above. 

**_Off-topic for this PR:_**
I presume you are getting that error because you are using the `develop` branch of WPCS. If you use `develop`, you need to keep yourself informed of changes as it is by nature an "unstable" branch.
In this case, you are getting the notification as you are setting a property in your ruleset which was always discouraged to be set and which has now been removed, so you need to fix your ruleset.
See the commit message here: https://github.com/WordPress/WordPress-Coding-Standards/pull/2214/commits/31064cf7a49931c0cf165b923ac0a86c20b2e4b4

benjaminprojas:
Thanks for the input on that. Looks like I was confused by the purpose of this PR. 

But your comment pointed me in the right direction. Turns out this is added by the HM-Minimum ruleset we are using to extend the WPCS rules: https://github.com/humanmade/coding-standards/blob/e8ec095e7128a3819e254a238902746ec00ece0d/HM-Minimum/ruleset.xml#L160-L163

We already have a forked copy for PHP 8.x purposes so the fix was easy for us. I'll open an issue there as well.

jrfnl:
@gsherwood and me discussed this and I'll make the BC break for extenders smaller as per comment https://github.com/squizlabs/PHP_CodeSniffer/pull/3629#issuecomment-1179536571

jrfnl:
I've rebased the original commit without changes and added a second commit to this PR to implement the BC-layer as per https://github.com/squizlabs/PHP_CodeSniffer/pull/3629#issuecomment-1179536571.

**⚠️ Note: this new, second commit should **NOT** be ported to PHPCS 4.x.**


### Ruleset::setSniffProperty(): add BC-layer for old format property values

This commit adds a BC layer to handle property values passed to `Ruleset::setSniffProperty()` in the old (mixed) format.

This BC-layer will never be hit when PHPCS is used from the CLI/with a ruleset. This BC-layer is only in place for integrations with PHPCS which may call the `Ruleset::setSniffProperty()` method directly.

The `Ruleset::setSniffProperty()` will still handle properties passed in the old format correctly, but will also throw a deprecation notice to allow the maintainers of the integration to update their code.

Includes dedicated tests to ensure this BC-layer works as intended.

### Updated suggested changelog entry

> PHP 8.2: prevent deprecation notices for properties set in a (custom) ruleset for complete standards/complete sniff categories
> - Invalid properties set for individual sniffs will now result in an error and halt the execution of PHPCS with a descriptive error message to allow users to fix their ruleset.
> - Properties set for complete standards/complete sniff categories will now only be set on sniffs which explicitly support the property.
>       The property will be silently ignored for those sniffs which do not support the property.
> - For sniff developers: it is strongly recommended for sniffs to explicitly declare any user-adjustable `public` properties.
>    If dynamic properties need to be supported for a sniff, either declare the magic `__set()`/`__get()`/`__isset()`/`__unset()` methods on the sniff or let the sniff extend `stdClass`.
>    Note: The `#[\AllowDynamicProperties]` attribute will have **no effect** for properties which are being set in rulesets.
> - Sniff developers/integrators of PHPCS may need to make some small adjustments to allow for changes to the PHPCS internals. A deprecation notice will be thrown for those situations. See the PR description of PR #3629 for full details.

jrfnl:

PR #3630, opened 2022-07-10
Runner: show actionable information when PHPCS runs out of memory

### LINKED ISSUE #3621: Catch and show informative error message for out of memory errors
Depending on the size of a code base, the amount of sniffs run and the `memory_limit` set for PHP, PHPCS can run out of memory.

Depending on the point at which it runs out of memory, it may show the PHP native "out of memory" error or it may exit without any further information (seen that a number of times as well).

Catching out of memory errors and displaying a user-friendly error message isn't straight-forward, as one can't allocate anymore memory in PHP by that time.

However, my friend @schlessera has worked out a way to do it for another CLI tool and [shared the code with me](https://github.com/wp-cli/package-command/blob/afac00c12612ba13ed42bb0da2c76002b9a5d9e8/src/Package_Command.php#L1346-L1392).

@gsherwood Would you be interested in me implementing a user friendly error message for out of memory errors for PHPCS ?

jrfnl:
As discussed in #3621, this PR adds a new function to show a "friendly" error message if/when PHPCS runs out of memory.

This functionality can't be tested via the unit/integration test suite as the PHP shutdown can't be mocked.

To test the functionality, I used a test run of PHPCS over its own code base with the following command:
```bash
phpcs -d memory_limit 32M
```

This should yield something along the lines of the below screenshot:
image

Fixes #3621

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -53,6 +53,8 @@ class Runner
      */
     public function runPHPCS()
     {
+        $this->register_out_of_memory_shutdown_message('phpcs');
+
         try {
             Util\Timing::startTiming();
             Runner::checkRequirements();
@@ -153,6 +155,8 @@ class Runner
      */
     public function runPHPCBF()
     {
+        $this->register_out_of_memory_shutdown_message('phpcbf');
+
         if (defined('PHP_CODESNIFFER_CBF') === false) {
             define('PHP_CODESNIFFER_CBF', true);
         }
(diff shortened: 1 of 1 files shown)


gsherwood:
Thanks for this patch. I couldn't help but rename the function though.

Interestingly, this doesn't work as reliably in the 4.0 branch. It seems to depend on how much memory is allocated. But it does work sometimes, and at specific memory limits for me, so I've left it in there unchanged.

jrfnl:
> Thanks for this patch. I couldn't help but rename the function though.

Oh dear... sorry about that... I completely forgot to do that. Thanks for fixing that up!
(should we maybe activate a camelCase method name sniff for this repo ?)

> Interestingly, this doesn't work as reliably in the 4.0 branch. It seems to depend on how much memory is allocated. But it does work sometimes, and at specific memory limits for me, so I've left it in there unchanged.

I'm at a conference at the moment. I'll have a look at the 4.x branch and see if I can make any improvements to get it working there once I'm back home.

PR #3632, opened 2022-07-14
Short list not tokenized correctly in control structures without braces

jrfnl:
Originally reported in sirbrillig/phpcs-variable-analysis#263.

When the short list syntax would be used as the first contents within a control structure without braces, the square brackets would be tokenized as plain square brackets, not as short array brackets.

```php
if ( true )
    [ $a ] = [ 'hi' ]; // The first `[` in this line was tokenized incorrectly.

return $a ?? '';
```

Fixed now by checking whether the preceding parenthesis closer has an owner and if that owner could be a scope owner.

Includes unit tests.
Includes updating the data providers to use named data sets to allow for easier debugging.


gsherwood:
Thank you

Issue #3200, opened 2021-01-22
Unexpected spacing for anonymous classes with constructor arguments


TimWolla:
**Describe the bug**

When creating an anonymous class while passing constructor arguments phpcs requests a space between `class` and the `(`. I did not expect this and it looks off.

**Code sample**
```php
<?php

namespace foo;

class Foo
{
    protected function foo()
    {
        new class(1) extends Bar { // line 9
        };
    }
}
```

**Custom ruleset**
```xml
<?xml version="1.0"?>
<ruleset>
	<file>./test/</file>
	<arg name="extensions" value="php" />
	<arg value="p"/>
	<arg name="basepath" value="."/>

	<rule ref="PSR12"/>
</ruleset>
```

**To reproduce**

1. Create a file called `test.php` with the code sample above.
2. Run `phpcs test.php`
3. See error message displayed

```
FILE: test/test.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 9 | ERROR | [x] Expected 1 space after class keyword; 0 found
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

**Expected behavior**

I expected that error not to be reported, because I believe there should not be a space between `class` and `(`.

**Versions (please complete the following information):**
 - OS: Docker container (Debian based) on Ubuntu host.
 - PHP: 7.4.12
 - PHPCS: 3.5.8 (stable)
 - Standard: PSR12

**Additional context**

[PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) with the PSR2 ruleset expects the space to not be there.

TimWolla:
Looking at this: https://github.com/php-fig/fig-standards/pull/1206 the consensus of the standard appears to be that the space should not be there, even if the current version of PSR-12 does not say anything regarding this topic.

gsherwood:
If the errata is accepted then I'd change the PSR-12 standard included with PHPCS. But right now the PSR-12 standard says:
> Anonymous Classes MUST follow the same guidelines and principles as closures in the above section.

And in the closures section, all the examples clearly show a space between the function keyword and the opening parenthesis of the argument list. So that's all I have to go on until a change is made to clarify things.

TimWolla:
[...]

PR #3636, opened 2022-07-23
PEAR/PSR2/FunctionCallSignature: support anonymous classes

jrfnl:
The function call spacing for anonymous class instantiations was so far not checked by these ~~or any other PHPCS native~~ sniffs.

In my opinion, object instantiations of anonymous classes should be treated the same an object instantiations of non-anonymous classes.

The `PEAR.Functions.FunctionCallSignature` and the `PSR2.Methods.FunctionCallSignature` sniffs check the object instantiation spacing for non-anonymous classes, so seem like the logical place to also check the spacing for anonymous class object instantiations.

To add this support, the `T_ANON_CLASS` token has been added to the `Tokens::$functionNameTokens` array.

Notes:
* As PSR12 does not specify the spacing between the `class` keyword and the open parenthesis (or rather is unclear about it), I am explicitly excluding anonymous classes from the "space before open parenthesis" check.
    Related: squizlabs/PHP_CodeSniffer#3200
* I have verified all other uses of the `Tokens::$functionNameTokens` array within PHPCS.
    - The `Generic.WhiteSpace.ArbitraryParenthesesSpacing` sniff is not affected by the change and already contains a test to verify this.
    - The `Squiz.Operators.ComparisonOperatorUsage` sniff also is not affected by the change. I have added tests to confirm this in a separate commit.
* Obviously external standards using the token array _may_ be affected by the change, but a scan of the most popular external standards showed me that the token array is rarely used and when it is used, is mostly used incorrectly.
    The only sniff using the array, which looks to be using it correctly and which may be affected, is the `WebImpressCodingStandard.WhiteSpace.ScopeIndent` sniff. Whether this is positive or negative is up to @michalbundyra to determine.

Includes unit tests for both the `PEAR.Functions.FunctionCallSignature` and the `PSR2.Methods.FunctionCallSignature` sniffs .

### DIFF
--- a/src/Standards/PEAR/Sniffs/Functions/FunctionCallSignatureSniff.php
+++ b/src/Standards/PEAR/Sniffs/Functions/FunctionCallSignatureSniff.php
@@ -120,7 +120,9 @@ class FunctionCallSignatureSniff implements Sniff
 
         $closeBracket = $tokens[$openBracket]['parenthesis_closer'];
 
-        if (($stackPtr + 1) !== $openBracket) {
+        if ($tokens[$stackPtr]['code'] !== T_ANON_CLASS
+            && ($stackPtr + 1) !== $openBracket
+        ) {
             // Checking this: $value = my_function[*](...).
             $error = 'Space before opening parenthesis of function call prohibited';
             $fix   = $phpcsFile->addFixableError($error, $stackPtr, 'SpaceBeforeOpenBracket');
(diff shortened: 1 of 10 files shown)


jrfnl:
Note: this does cause some overlap with the `PSR12.Classes.AnonClassDeclaration` sniff in the PSR12 standard. Not sure what to do about this, though it shouldn't be a cause for fixer conflicts (as the sniffs expect the same thing).

jrfnl:
Rebased without any real changes to solve the merge conflicts.

DannyvdSluijs (approved):
The tests in this PR function as a proof the added functionality works as expected including the undefined number of spaces between the `class` keyword and opening parenthesis.

jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/47

jrfnl:

PR #3640, opened 2022-07-29
Generic.WhiteSpace.DisallowTabIndent not reporting errors for PHP 7.3 flexible heredoc/nowdoc syntax

jrfnl:
Since PHP 7.3, heredoc/nowdoc closers may be indented.
This indent can use either tabs or spaces and the indent is included in the `T_END_HEREDOC`/`T_END_NOWDOC` token contents as received from the PHP native tokenizer.

However, these tokens where not included in the tokens to look at for the `Generic.WhiteSpace.DisallowTabIndent` sniff, which could lead to false negatives.

Fixed now, includes tests.


gsherwood:
Thanks a lot for fixing this.

Issue #3158, opened 2020-11-04
Keeping the test-utils class while removing the tests.


VincentLanglet:
I saw multiple issue about adding or removing the tests from the gitattributes
https://github.com/squizlabs/PHP_CodeSniffer/issues/548
https://github.com/squizlabs/PHP_CodeSniffer/issues/1908

Imho there is two king of files in the tests/ dirs
- Tests which are used to test a specific sniff/file
- Utils files which are used to run the test

I think the best would be to exclude the first ones but keep the second one (for developper which use them to run test

For instance, it's right to export-ignore this in the next major.
```
/src/Standards/Generic/Tests export-ignore
/src/Standards/MySource/Tests export-ignore
/src/Standards/PEAR/Tests export-ignore
/src/Standards/PSR1/Tests export-ignore
/src/Standards/PSR2/Tests export-ignore
/src/Standards/PSR12/Tests export-ignore
/src/Standards/Squiz/Tests export-ignore
/src/Standards/Zend/Tests export-ignore
```

But I'm more concern about this line
```
/tests export-ignore
```

When we're writing on our sniff,
The class `tests\Standards\AbstractMethodUnitTest` is useful. 
Same for `tests\bootstrap.php` and maybe others (and all the dependencies).

Maybe there's something to do about it, like not excluding everything or moving some files somewhere else.

gsherwood:
As `/tests` still contains quite a lot of test code in .inc files, what if I instead excluded `/tests/Core` and left the rest of the `/tests` directory alone. Would that work?

VincentLanglet:
> As `/tests` still contains quite a lot of test code in .inc files, what if I instead excluded `/tests/Core` and left the rest of the `/tests` directory alone. Would that work?

This seems like a great solution indeed.

The only files which won't work is the `test/AllTest.php` file, since it require `'Core/AllTests.php'`
```
require_once 'Core/AllTests.php';
require_once 'Standards/AllSniffs.php';
require_once 'TestSuite.php';

class PHP_CodeSniffer_AllTests
{
    public static function suite()
    {
        $GLOBALS['PHP_CODESNIFFER_STANDARD_DIRS'] = [];
        $GLOBALS['PHP_CODESNIFFER_TEST_DIRS']     = [];

        $suite = new TestSuite('PHP CodeSniffer');

        $suite->addTest(Core\AllTests::suite());
        $suite->addTest(Standards\AllSniffs::suite());

        return $suite;

    }
[...]

PR #3643, opened 2022-08-05
Expore ignore tests directory

jack-worman:
Tests are useless outside the actual project.
It also gets in the way of IDE's because it suggests global test functions like `function variable($var) {}` from `GetMethodParametersTest`


jrfnl:
This has been discussed multiple times and this change has already been made for the 4.x branch, but will not be accepted for the 3.x branch.

See #548, #1908, #3158, #3176, #3204 (amongst others)

jack-worman:
Thanks for the response and sorry for the noise.

PR #3653, opened 2022-08-24
False positives for `match()` in `OperatorSpacingSniff`

kukulich:


jrfnl (changes requested):
Nice catch! Just one remark.

src/Standards/Squiz/Sniffs/WhiteSpace/OperatorSpacingSniff.php
```diff
@@ -77,12 +77,14 @@ public function register()
...
-            T_YIELD    => T_YIELD,
-            T_FN_ARROW => T_FN_ARROW,
+            T_RETURN        => T_RETURN,
+            T_ECHO          => T_ECHO,
+            T_EXIT          => T_EXIT,
+            T_PRINT         => T_PRINT,
+            T_YIELD         => T_YIELD,
+            T_FN_ARROW      => T_FN_ARROW,
+            T_MATCH_ARROW   => T_MATCH_ARROW,
+            T_MATCH_DEFAULT => T_MATCH_DEFAULT,
```
jrfnl, line 87: `T_MATCH_DEFAULT` shouldn't be needed as the sniff only looks at the token directly adjacent, which will be the `T_MATCH_ARROW` between the `default` keyword and the `-`.

(oops.. apparently I'd forgotten to save my remark.... sorry)

kukulich: Thanks.

jrfnl (approved):
👍🏻

gsherwood:
Thanks for fixing this. I found a similar error in Squiz.Formatting.OperatorBracket as a result.

jrfnl:
Nice!

PR #3655, opened 2022-09-04
Add "static analysis" Composer keyword

GaryJones:
As per https://getcomposer.org/doc/04-schema.md#keywords by including "static analysis" as a keyword in the `composer.json` file, Composer 2.4.0-RC1 and later will prompt users if the package is installed with `composer require` instead of `composer require --dev`. See https://github.com/composer/composer/pull/10960 for more info.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,8 @@
     "type": "library",
     "keywords": [
         "phpcs",
-        "standards"
+        "standards",
+        "static analysis",
     ],
     "homepage": "https://github.com/squizlabs/PHP_CodeSniffer",
     "license": "BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


GaryJones:
Errant comma on the first commit, so these can be rebased/squashed on merge.

jrfnl (approved):
Nice one! Looks like a good improvement.

jrfnl:
Note: (new) external standards which `require` PHPCS will also get prompted whether `squizlabs/php_codesniffer` was deliberately requested as `require`, not `require-dev`. IMO, that's fine. Maintainers of external standards can select to keep the package in `require` on the command-line prompt.

gsherwood:
Thanks for this. I did end up squashing them, hence no merged tag.

PR #3660, opened 2022-09-09
Generic/LowerCaseType: examine types in arrow function declarations

jrfnl:
PHP 7.4 arrow functions were not yet being taken into account for this sniff.

Fixed now.
Includes unit tests.


gsherwood:
Thanks for adding this

Issue #3666, opened 2022-09-15
PEAR.Functions.FunctionCallSignature incorrect indent fix when checking mixed HTML/PHP files


giorgosbotis:
running phpcbf on a file containing the following code:
``` 
<?php include get_theme_file_path(
        '/theme_extra/test_block.php'
      ); ?>   
```

I get the following error:
`Fatal error: Uncaught PHP_CodeSniffer\Exceptions\RuntimeException: Undefined array key -1 in /app/myproject/vendor/squizlabs/php_codesniffer/src/Fixer.php on line 510 in /app/myproject/vendor/squizlabs/php_codesniffer/src/Runner.php:604`

gsherwood:
I can't replicate this issue. What coding standard are you using and with what PHPCS version?

giorgosbotis:
Sorry for not adding info:
phpcs 3.7.1
php 8.0.19
`phpcbf --standard=PEAR ./test.php`

gsherwood:
Thanks, have managed to replicate. Doesn't appear to be tab related, but the indent is important.

gsherwood:
So the issue here is the indent before the open PHP tag. If it's divisible by the indent size (normally 4 characters) then you wont get an error message and nothing will try to be fixed.

If it's not (I believe the example code has 6 spaces on indent) then it will try and bring the indent back in line with a tab stop. When it did this with inline HTML (anything outside the PHP tags) it did the fix incorrectly and removed content. If the content was the first thing in the file, you get the error message in this bug report.

The fix I'm committing will still report the same error message, but the auto-fixer in PHPCBF will now treat inline HTML differently so it can apply the fix properly.

gsherwood:
I've committed a fix for this issue, which will be in the 3.7.2 release. Thanks a lot for reporting it.

jrfnl:
Looks like the OP also reported this (with a different description) in https://github.com/WordPress/WordPress-Coding-Standards/issues/2083 for which I opened PR #3667....

giorgosbotis:
It was reported with a different description, as there were two different issues I encountered. 

The one reported here was getting an exception running cbf with PEAR standard only, while the other one was Worpdress-Core standard producing wrong code (no exception there).

I am sorry but I had no clue they were connected or the same.

PR #3667, opened 2022-09-17
PEAR/FunctionCallSignature: minor tweaks + extra test

jrfnl:
Originally reported in https://github.com/WordPress/WordPress-Coding-Standards/issues/2083

The fixer for the `OpeningIndent` error for multi-line function calls could inadvertently remove a PHP close tag (or other inline HTML content) on the previous line if the first non-whitespace token on a line was `T_INLINE_HTML`. In the case of a PHP close tag, this would cause a parse error in the file.

This parse error would then result in the file being incorrectly tokenized for the second fixer loop, with the `<?php` after the inline HTML being tokenized as below, which causes problems with other sniffs:
```
 13 | L1 | C 43 | CC 1 | ( 0) | T_WHITESPACE               | [  3]: ⸱⸱⸱
 14 | L1 | C 46 | CC 1 | ( 0) | T_LESS_THAN                | [  1]: <
 15 | L1 | C 47 | CC 1 | ( 0) | T_INLINE_THEN              | [  1]: ?
 16 | L1 | C 48 | CC 1 | ( 0) | T_STRING                   | [  3]: php
```

Fixed now.

Includes unit test.

Includes minor defensive coding fix (`$first !== false`) on line 345 and adding of an inline comment to clarify the code within the fixer.

### DIFF
--- a/src/Standards/PEAR/Sniffs/Functions/FunctionCallSignatureSniff.php
+++ b/src/Standards/PEAR/Sniffs/Functions/FunctionCallSignatureSniff.php
@@ -340,7 +340,8 @@ class FunctionCallSignatureSniff implements Sniff
         // call itself is, so we can work out how far to
         // indent the arguments.
         $first = $phpcsFile->findFirstOnLine(T_WHITESPACE, $stackPtr, true);
-        if ($tokens[$first]['code'] === T_CONSTANT_ENCAPSED_STRING
+        if ($first !== false
+            && $tokens[$first]['code'] === T_CONSTANT_ENCAPSED_STRING
             && $tokens[($first - 1)]['code'] === T_CONSTANT_ENCAPSED_STRING
         ) {
             // We are in a multi-line string, so find the start and use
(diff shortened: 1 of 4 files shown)


jrfnl:
Note: the test failure on PHP 8.2 is unrelated to this PR and also happening on `master`.

Also note that this PR will conflict (on the tests) with PR #3636. I will rebase and fix the conflict depending on which PR gets merged first.

jrfnl:
> Note: the test failure on PHP 8.2 is unrelated to this PR and also happening on master.

This is related to the PHP 8.2 disjunctive normal types feature which changed the tokenization - see: https://github.com/php/php-src/pull/9512

This needs to be fixed when support for DNF is added to the PHPCS tokenizer.

GaryJones (commented):
Tested the fix for the original WPCS report, and all seems well. ✅

jrfnl:
Fixed by commit https://github.com/squizlabs/PHP_CodeSniffer/commit/9445108a57b46f4e84a890788de5d2388346460a in response to issue #3666, though this PR still contains some minimal differences (defensive coding, comment + minor difference in the fix) which may be worth having a look at.

gsherwood:
> This is related to the PHP 8.2 disjunctive normal types feature which changed the tokenization

Found myself here while looking into the failure. This looks to caused by tokenization of function calls changing when the name is `readonly`:
```php
$var = readonly()
```
So where the string `readonly` used to be the token `T_STRING`, it's now `T_READONLY`. The code still runs fine so I'm not sure if the change was intentional.

jrfnl:
(and yeah, DNT is not going to be fun to sort out in the tokenizer.... I have a feeling we'll probably need to add `T_OPEN_TYPE_PARENTHESIS` and `T_CLOSE_TYPE_PARENTHESIS` tokens or something as otherwise the chances of sniffs which check spacing around "normal" parentheses will all start triggering on DNT, while I imagine people may want to special case those for the spacing around and inside the parentheses.)

jrfnl:
> Fixed by commit [9445108](https://github.com/squizlabs/PHP_CodeSniffer/commit/9445108a57b46f4e84a890788de5d2388346460a) in response to issue #3666, though this PR still contains some minimal differences (defensive coding, comment + minor difference in the fix) which may be worth having a look at.

Rebased to make the remaining changes mergable.

jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/134

PR #3669, opened 2022-09-20
PSR12/ClassInstantiation: fix regression for `new parent`

### LINKED ISSUE #3668: PSR12.Classes.ClassInstantiation.MissingParentheses false positive when instantiating parent classes
**Describe the bug**
When instantiating a parent class using the keyword "parent", PHPCS flags this as wrong and PHPCBF inserts brackets after the `new` thus breaking code.

**Code sample**
see also: https://3v4l.org/5WDl4
```php
class SomeClass extends ParentClass
{
//...
    public function foo(): static
    {
        $class = new parent();
        //...
    }
}
```

**To reproduce**
2. Run `phpcs test.php ...`
```

--------------------------------------------------------------------------------------------------------------------------------
 line | ERROR | [x] Parentheses must be used when instantiating a new class
      |       |     (PSR12.Classes.ClassInstantiation.MissingParentheses)

```

**Expected behavior**
No error report and no changed code from PHPCBF.

 - OS: Linux
 - PHP: 8.0
 - PHPCS: 3.7.1
 - Standard: PSR12

jrfnl:
Related to #3546 which fixed an inconsistency after #3484.

The change of the tokenization from `T_STRING` to `T_PARENT` for the `parent` keyword in `new parent` caused a regression in the `PSR12.Classes.ClassInstantiation` sniff.

Fixed now.

Includes unit tests.

Fixes #3668

### DIFF
--- a/src/Standards/PSR12/Sniffs/Classes/ClassInstantiationSniff.php
+++ b/src/Standards/PSR12/Sniffs/Classes/ClassInstantiationSniff.php
@@ -48,6 +48,7 @@ class ClassInstantiationSniff implements Sniff
             T_NS_SEPARATOR             => T_NS_SEPARATOR,
             T_SELF                     => T_SELF,
             T_STATIC                   => T_STATIC,
+            T_PARENT                   => T_PARENT,
             T_VARIABLE                 => T_VARIABLE,
             T_DOLLAR                   => T_DOLLAR,
             T_OBJECT_OPERATOR          => T_OBJECT_OPERATOR,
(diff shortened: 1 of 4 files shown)


jrfnl:
FYI: I've done a code search on all sniffs for `T_SELF` and have verified that any sniffs using `T_SELF` also take `T_PARENT` into account (if needed). This has resulted in two more PRs: #3670 and #3671. The other sniffs using `T_SELF` already handled `T_PARENT` correctly.

PR #3671, opened 2022-09-20
Squiz/OperatorBracket: fix regression for `new parent`

jrfnl:
Related to #3546 which fixed an inconsistency after #3484.

The change of the tokenization from `T_STRING` to `T_PARENT` for the `parent` keyword caused a regression in the `Squiz.Formatting.OperatorBracket` sniff.

Fixed now.

Includes unit tests.


jrfnl:
Looks like something wonky is going on with Packagist I suppose retriggering the build at a later moment in time should fix this.

Issue #3585, opened 2022-04-22
Add documentation schema


dingo-d:
In WordPress Coding Standards I've started to work on adding [xsd schema](https://github.com/WordPress/WordPress-Coding-Standards/pull/2037) for the phpcs documentation.

If you're up for it I can move the PR on this repo and we can then work on it here.

Let me know if you would like to have this in the current repo, or if we should just offload it to PHPCSUtils so that external standards ca use it if they want to.

Current xsd schema (passing the validation on WPCS) looks like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified">
    <xs:element name="documentation">
        <xs:complexType>
            <xs:sequence maxOccurs="unbounded">
                <xs:group ref="rulegroup"/>
            </xs:sequence>
            <xs:attribute name="title" use="required" type="titleType"/>
            <xs:anyAttribute processContents="lax"/>
        </xs:complexType>
    </xs:element>

    <xs:group name="rulegroup">
        <xs:sequence>
            <xs:element name="standard" type="standardType"/>
            <xs:element name="code_comparison" type="code_comparisonType" maxOccurs="unbounded"/>
            <xs:any minOccurs="0"/>
        </xs:sequence>
    </xs:group>

    <xs:simpleType name="titleType">
        <xs:restriction base="xs:string">
            <xs:maxLength value="58"/>
        </xs:restriction>
    </xs:simpleType>

    <xs:complexType name="code_comparisonType">
        <xs:sequence>
            <xs:element name="code" type="codeType" maxOccurs="2" minOccurs="2"/>
        </xs:sequence>
    </xs:complexType>

    <xs:complexType name="codeType">
        <xs:simpleContent>
            <xs:extension base="xs:string">
                <xs:attribute name="title" use="required" type="xs:string"/>
            </xs:extension>
        </xs:simpleContent>
    </xs:complexType>

    <xs:complexType name="standardType">
        <xs:simpleContent>
            <xs:extension base="xs:string">
                <xs:anyAttribute/>
            </xs:extension>
        </xs:simpleContent>
    </xs:complexType>
</xs:schema>
```

gsherwood:
The documentation format isn't a high priority for me, so best to keep this external. I may consider removal of the whole doc schema in the future.

jrfnl:
[...]

PR #2872, opened 2020-02-16
Travis: add check to make sure that sniffs are "feature complete"

jrfnl:
This is a follow-up/replacement PR for #2364.

The tooling which was previously pulled (and never merged) has been enhanced and been released in a separate package `PHPCSDevTools` and can be used both by PHPCS itself as well as by external standards.

Full details of the features this offers: https://github.com/PHPCSStandards/PHPCSDevTools#checking-whether-all-sniffs-in-a-phpcs-standard-are-feature-complete

Notes:
* This check only needs to be run on one build as the results won't change between PHP versions.
    For that reason, I've elected to add it to the fastest PHP version run to balance out the runs some more.
* The script is currently being run in `quiet` mode which will only error on missing unit tests and will not report on missing documentation.
    In "normal" mode, the build would fail as there are currently 104 sniffs without documentation in the standards being checked (excluding `MySource`).
    If/when the missing documentation files have been added, the `-q` flag should be removed and having unit tests as well as documentation should be enforced for all sniffs.
* The `MySource` standard is not checked for two reasons:
    1. The `MySource.Channels.IncludeOwnSystem` sniff is missing unit tests and would fail the build.
    2. The `MySource` standard is being removed in PHPCS 4.0 anyway, so making it feature complete is redundant.
* I've not added the tooling as a dependency to the `composer.json` file and am using a `global` install in the Travis script on purpose.
    For another feature within the package, the package `require`s PHPCS itself, so this circular dependency would be troublesome for composer.
    As the _feature complete_ script doesn't depend on PHPCS, this isn't a real issue and using a `global` install allows the script to run without problems.

---
<details>
  <summary>List of sniffs missing documentation (generated by the feature complete tool)</summary>

This is the current output of the script if the `-q` (quiet) flag would be removed. Includes the `MySource` standard.

```
PHPCSDevTools: Sniff feature completeness checker version 1.0.0                                                                
by Juliette Reinders Folmer                                                                                                    
                                                                                                                               
............................................................  60 / 244 ( 25%)                                                  
............................................................ 120 / 244 ( 49%)                                                  
............................................................ 180 / 244 ( 74%)                                                  
............................................................ 240 / 244 ( 98%)                                                  
....                                                         244 / 244 (100%)                                                  
                                                                                                                               
WARNING: Documentation missing for /src/Standards/Generic/Sniffs/Arrays/ArrayIndentSniff.php.                                  
[...]

PR #3674, opened 2022-09-23
Docs: fix invalid XML docs for three sniffs

jrfnl:
Each of these three sniffs had a `<code_comparison>` block containing _three_ `<code>` elements instead of two.

Discovered by validating the docs against the XSD schema created by the amazing @dingo-d. The XSD schema is now available in [PHPCSDevTools](https://github.com/PHPCSStandards/PHPCSDevTools) 1.2.0 as discussed in #3585.

If you're interested, I could refresh (a variation of) PR #2872 to add XML doc validation against the schema to the GH Actions checks.

### DIFF
--- a/src/Standards/Generic/Docs/Classes/OpeningBraceSameLineStandard.xml
+++ b/src/Standards/Generic/Docs/Classes/OpeningBraceSameLineStandard.xml
@@ -18,6 +18,14 @@ class Foo
 }
         ]]>
         </code>
+    </code_comparison>
+    <code_comparison>
+        <code title="Valid: Opening brace is the last thing on the line.">
+        <![CDATA[
+class Foo <em>{</em>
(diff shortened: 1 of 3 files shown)


gsherwood:
Thanks. All those docs changes make perfect sense.

jrfnl:
> If you're interested, I could refresh (a variation of) PR https://github.com/squizlabs/PHP_CodeSniffer/pull/2872 to add XML doc validation against the schema to the GH Actions checks.

Should I prepare that GH Actions addition to safeguard this for all XML docs from now on ?

The GH Action change doesn't require a dependency on PHPCSDevTools, it could just download the XSD file using `curl`, though having the PHPCSDevTools `dev` dependency would make it easier for contributors to run the same check locally.

PR #3675, opened 2022-09-26
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -12,6 +12,9 @@ on:
   # Allow manually triggering the workflow.
   workflow_dispatch:
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   phpstan:
     name: "PHP: 7.4 | PHPStan"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,9 @@ on:
   # Allow manually triggering the workflow.
   workflow_dispatch:
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


jrfnl:
@sashashura No. Sorry, but a first time contribution to a repo from someone who's never interacted with the repo before, who has no (public) history on Git (and IIRC when I first looked when the PR was originally pulled, a public history, with no contributions) and who claims something is a security improvement raises all sorts of questions and suspicions for me.

I will not review this PR. I will do my own research when I find the time and then I'll see.

jrfnl:
Yeah so ? If you understand open source, you should know not to whine about that (aside from that: I'm not the maintainer).
Also links provided by an OP I don't trust, means I don't trust those links either.

TimWolla:
> who claims something is a security improvement raises all sorts of questions and suspicions for me.

FWIW: The claims are correct. The same effective change was also applied to PHP itself in https://github.com/php/php-src/commit/1d45ca58c86ac103ea624c3af3c81c20863950ad.

> means I don't trust those links either.

Those go directly to GitHub's documentation, not some third party resource.

PR #3676, opened 2022-09-26
Fix php 8.1 error for trim() function

Kleinast:
This is a quick fix, but I don't know if it's normal that some $this->files array keys are 'null'

### DIFF
--- a/src/Files/LocalFile.php
+++ b/src/Files/LocalFile.php
@@ -29,7 +29,7 @@ class LocalFile extends File
      */
     public function __construct($path, Ruleset $ruleset, Config $config)
     {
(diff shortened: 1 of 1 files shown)


jrfnl:
@Kleinast Could you please elaborate how you got this error ? If PHPCS is used as intended, that error shouldn't be possible...

jrfnl:
Hang on - just now seeing issue #3677. Will continue there.

jrfnl:
This has now been fixed via https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/1150 and the fix will be included in the 3.13.3 release.

PR #3687, opened 2022-10-07
Added support for array multi line function comments

martinssipenko:
This change adds the ability to define array structures over several lines.

### Background:
PHPStan allows using [array shapes](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes) that can quickly turn into very long comment lines and this change allows them to be broken over several lines for readability purposes without having to disable the given Sniff.

```php
/**
 * @param array<array{
 *            foo: array{foo: int},
 *            bar: int
 *        }> $bb B.
 */
public function multiLineArray(array $bb)
{
}
```

#### Before:
Would result in `Squiz.Commenting.FunctionComment.MissingParamName` or `Squiz.Commenting.FunctionComment.MissingParamTag` or similar errors.

And the doc block would need to be written as:
```php
/**
 * @param array<array{foo: array{foo: int}, bar: int}> $bb B.
 */
```

#### After:
No errors are being reported.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -306,16 +306,47 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
             $varSpace     = 0;
             $comment      = '';
             $commentLines = [];
+            $isMultiLine  = false;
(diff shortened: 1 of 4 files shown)


jrfnl:
@martinssipenko I haven't looked at this PR in detail, but if you are still interested in contributing this to PHPCS, I'd be happy to look at this if pulled to the new repo: https://github.com/PHPCSStandards/PHP_CodeSniffer

PR #3691, opened 2022-10-14
GH Actions: fix use of deprecated `set-output`

jrfnl:
GitHub has deprecated the use of `set-output` (and `set-state`) in favour of new environment files.

This commit updates workflows to use the new methodology.

Refs:
* https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files

👉🏻 Note: this won't get rid of all warning yet as a lot of predefined action runners also use `set-output`, but most of those are in the process of updating and/or have released a new version already, so the other warnings should automatically disappear over the next few weeks.


gsherwood:
Thanks a lot for making this change.

PR #3692, opened 2022-10-14
GH Actions: update the xmllint-problem-matcher

jrfnl:
The `xmllint-problem-matcher` action runner has released a new version which updates it to use node 16. This gets rid of a warning which was shown in the action logs.

Note: I've [suggested to the author to use long-running branches for the action runner instead](https://github.com/korelstar/xmllint-problem-matcher/pull/7), which would make this update redundant, but no telling if or when they'll respond to that, let alone if they will follow my suggestion.

Refs:
* https://github.com/korelstar/xmllint-problem-matcher/releases/tag/v1.1

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -34,7 +34,7 @@ jobs:
 
       # Show XML violations inline in the file diff.
       # @link https://github.com/marketplace/actions/xmllint-problem-matcher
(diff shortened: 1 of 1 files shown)


jrfnl:
Closing - the action runner branch has become a long-running branch, so this update is no longer needed.

PR #3694, opened 2022-10-16
Generic.WhiteSpace.SpreadOperatorSpacingAfter does not ignore spread operator in PHP 8.1 first class callables

jrfnl:
... as in that case the spread operator does not apply to a variable/function call and the likelihood of the sniff conflicting with other sniffs which examine the spacing on the inside of parenthesis is high.

Includes unit tests.

Ref:
* https://wiki.php.net/rfc/first_class_callable_syntax
* https://3v4l.org/YTTO1


gsherwood:
Thanks for fixing this.

jrfnl:
yvw

HEXYEBO0 (approved):

PR #3696, opened 2022-10-18
Add constants for exit return codes

BafS:
Inspired by https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Console/Command/Command.php#L37-L40, to make the code more readable and to avoid miss-using exit codes.

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -23,6 +23,9 @@ use PHP_CodeSniffer\Util\Standards;
 
 class Runner
 {
+    const SUCCESS = 0;
+    const FAILURE = 1;
+    const INVALID = 2;
 
     /**
      * The config data for the run.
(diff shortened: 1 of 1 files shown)


gsherwood:
This is a good idea in general, but the exit codes don't map like that. PHPCS and PHPCBF use the same exit codes to mean different things, which has resulted in a suggestion like:
```php
if ($this->reporter->totalFixable === 0) {
    // PHPCBF fixed all fixable errors.
    return self::FAILURE;
}
```
That's obviously not a failed run for PHPCS given it fixed all errors.

So this would need a bit of work before it could be merged. Likely different constants for PHPCS and PHPCBF, with appropriate names and documentation of the constant itself.

Linking the 4.0 issue to change the exit codes as this would be helpful there: https://github.com/squizlabs/PHP_CodeSniffer/issues/2898

BafS:
I see that return codes are not consistent, so maybe that could be done for v4 only yes.

PR #3698, opened 2022-10-20
NamedFunctionCallArgumentsTest: add test with PHP 8.1 enum keyword

jrfnl:
(this is to safeguard that the `enum` retokenization doesn't interfere with the parameter label retokenization)


gsherwood:
Thanks

PR #3700, opened 2022-10-26
Tokenizer/PHP: add tests for consistent tokenization heredocs with interpolated strings

jrfnl:
Similar to previous PR #3604 which fixed the tokenization for complex double quoted strings with interpolated variables/expressions and added tests for it.

Context: I ran into a tokenizer issue with heredocs and interpolated variables elsewhere, so I wrote some tests for it. Turned out all was okay with the tokenizer, but the test elsewhere had a parse error in PHP < 7.3. Either way, I figured I may as well pull the tests to safeguard things stay okay ;-)


gsherwood:
Thanks

PR #3703, opened 2022-10-30
Updated Outdated PHP Docblocks

AndrewDawes:
I saw that quite a few PHP Docblocks were showing the wrong return type and one with a throw tag which was no longer relevant. I updated the comments, please let me know if I missed something here or made some other mistake and I'm happy to fix it 🙂 

I also noticed that the codebase uses a mix short and long names for types e.g.  int vs integer, bool vs boolean. Which is the correct one to use? I'm happy to create another PR to make them the same throughout the codebase!

### DIFF
--- a/src/Files/FileList.php
+++ b/src/Files/FileList.php
@@ -199,7 +199,7 @@ class FileList implements \Iterator, \Countable
     /**
      * Return the file path of the current file being processed.
      *
-     * @return void
+     * @return int|string|null
      */
     #[ReturnTypeWillChange]
     public function key()
--- a/src/Reports/Gitblame.php
+++ b/src/Reports/Gitblame.php
@@ -28,7 +28,7 @@ class Gitblame extends VersionControl
      *
      * @param string $line Line to parse.
      *
-     * @return mixed string or false if impossible to recover.
+     * @return array|false|string|string[]|null string or false if impossible to recover.
      */
     protected function getAuthor($line)
     {
--- a/src/Tokenizers/Tokenizer.php
+++ b/src/Tokenizers/Tokenizer.php
@@ -80,8 +80,7 @@ abstract class Tokenizer
      * @param string                         $eolChar The EOL char used in the content.
      *
      * @return void
-     * @throws \PHP_CodeSniffer\Exceptions\TokenizerException If the file appears to be minified.
-     */
+	 */
     public function __construct($content, $config, $eolChar='\n')
     {
         $this->eolChar = $eolChar;
--- a/src/Util/Common.php
+++ b/src/Util/Common.php
@@ -35,7 +35,7 @@ class Common
      *
      * @param string $path The path to use.
      *
-     * @return mixed
+     * @return boolean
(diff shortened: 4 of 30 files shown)


BafS (approved):

src/Files/FileList.php
```diff
@@ -199,7 +199,7 @@ public function current()
     /**
      * Return the file path of the current file being processed.
      *
-     * @return void
+     * @return int|string|null
```
jrfnl, line 202:
```suggestion
     * @return string|null Path name or `null` when the end of the iterator has been reached.
```

src/Reports/Gitblame.php
```diff
@@ -28,7 +28,7 @@ class Gitblame extends VersionControl
      *
      * @param string $line Line to parse.
      *
-     * @return mixed string or false if impossible to recover.
+     * @return array|false|string|string[]|null string or false if impossible to recover.
```
jrfnl, line 31:
```suggestion
     * @return string|false string or false if impossible to recover.
```

What are the other types based on ? I can't find a code path which would lead to those....

src/Tokenizers/Tokenizer.php
```diff
@@ -80,8 +80,7 @@ abstract class Tokenizer
      * @param string                         $eolChar The EOL char used in the content.
      *
      * @return void
-     * @throws \PHP_CodeSniffer\Exceptions\TokenizerException If the file appears to be minified.
-     */
+	 */
```
jrfnl, line 83: Something looks off about the alignment here....
```suggestion
     */
```

jrfnl: Actually, on second thought, I'd leave this one (for now) as two out of three of the Tokenizer sub-classes throw that Exception from the overload `__construct()` method.

src/Util/Common.php
```diff
@@ -83,7 +83,7 @@ public static function isReadable($path)
      *
      * @param string $path The path to use.
      *
-     * @return mixed
+     * @return array|false|string|string[]
```
jrfnl, line 86:
```suggestion
     * @return string|false
```

What are the other types based on ? I can't find a code path which would lead to those....

jrfnl (changes requested):
@AndrewDawes Thank you for this PR and sorry that it took a while before this review.

Most fixes look good, thanks for those!

There are a few, however, which leave me wondering. Would be great if you could have another look at those.

jrfnl:
@AndrewDawes I'm sorry to see you closed this PR instead of responding to the provided feedback as I would have liked to be able to merge the majority of these fixes.

I will copy the correct fixes in another branch I have open with other docfixes now and will give you credit for the fixes from this PR.

AndrewDawes:
@jrfnl Oh sorry about that, it wasn't my intention. I had forgotten about this PR and I was just going through removing some forks yesterday not realising I still had a PR on this one :(

Thanks very much for taking care of it though! :)

PR #3704, opened 2022-11-02
Generic/CharacterBeforePHPOpeningTag: prevent false positive in mixed PHP/HTML files

jrfnl:
When a file would contain a short open echo tag `<?=` before a full PHP tag `<?php`, the sniff would throw a "The opening PHP tag must be the first content in the file" error, which IMO is a false positive.

Fixed now.

Includes additional tests demonstrating the issue and safeguarding the fix.


BafS (approved):

gsherwood:
I don't agree this is a false positive, and I think the proposed change is actually a bit confusing for this specific sniff.

If the short echo represents the start of PHP code, then test file 4 is an error due to the HTML before it. If it doesn't, then test file 5 is an error as the PHP tag isn't the first thing in the file.

What you're proposing is probably a logical rule for projects, but it's not what this sniff was contributed to do. It exists to ensure accidental output of HTML before PHP begins, like when a blank line is added before the opening tag accidentally and you get the error about output already having started. Maybe another sniff that enforces specific rules for mixed HTML/PHP files is required.

PR #3715, opened 2022-11-19
Generic/UnusedFunctionParameter: bug fix x 2 (magic methods and error codes)

jrfnl:
### Generic/UnusedFunctionParameter: ignore class context for closures/arrow functions

As things were, if a closure or arrow function declared within a class which extends or implements would have an unused function parameter, it would get the `InExtendedClass` or `InImplementedInterface` addition in the error code.

Those additions were only intended for function declarations where the declaration would potentially be overloading a method from a parent and would have to comply with the method signature of the method in the parent class/interface.

This could lead to underreporting if a standard explicitly excludes the error codes contain `InExtendedClass` and/or `InImplementedInterface`.

Fixed now.

Includes additional unit test, though the tests don't safeguard this much as they don't check the error codes of the messages thrown.

The change can be tested manually by running the new tests against `master`, which will show:
```
 163 | WARNING | The method parameter $d is never used (Generic.CodeAnalysis.UnusedFunctionParameter.FoundInExtendedClassAfterLastUsed)
 172 | WARNING | The method parameter $d is never used
     |         | (Generic.CodeAnalysis.UnusedFunctionParameter.FoundInImplementedInterfaceAfterLastUsed)
```

... while with the change in this commit, this will be fixed to:
```
 163 | WARNING | The method parameter $d is never used (Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed)
 172 | WARNING | The method parameter $d is never used (Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed)
```

### Generic/UnusedFunctionParameter: ignore magic methods

The function signature of magic methods - with the exception of `__construct()` and `__invoke()` - is dictated by PHP and unused parameters cannot be removed, which means that the warnings for these can never be resolved, only ignored via annotations.

This commit fixes this by checking whether a function is a magic method and if so, bowing out.

Includes unit tests.

Note: while not all magic methods take arguments, I'm still including the (nearly) full list of magic methods in the property as the other magic methods can be ignored anyway (no arguments).

### DIFF
--- a/src/Standards/Generic/Sniffs/CodeAnalysis/UnusedFunctionParameterSniff.php
+++ b/src/Standards/Generic/Sniffs/CodeAnalysis/UnusedFunctionParameterSniff.php
@@ -69,14 +69,17 @@ class UnusedFunctionParameterSniff implements Sniff
         $errorCode  = 'Found';
         $implements = false;
         $extends    = false;
-        $classPtr   = $phpcsFile->getCondition($stackPtr, T_CLASS);
-        if ($classPtr !== false) {
-            $implements = $phpcsFile->findImplementedInterfaceNames($classPtr);
-            $extends    = $phpcsFile->findExtendedClassName($classPtr);
-            if ($extends !== false) {
-                $errorCode .= 'InExtendedClass';
-            } else if ($implements !== false) {
-                $errorCode .= 'InImplementedInterface';
+
+        if ($token['code'] === T_FUNCTION) {
+            $classPtr   = $phpcsFile->getCondition($stackPtr, T_CLASS);
+            if ($classPtr !== false) {
+                $implements = $phpcsFile->findImplementedInterfaceNames($classPtr);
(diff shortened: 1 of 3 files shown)


jrfnl:
> > The function signature of magic methods is dictated by PHP and unused parameters cannot be removed
> 
> Aren't `__construct` and `__invoke` exceptions to this?

@gsherwood Ow! Good catch. Thanks for that!

I've updated the commit to exclude `__construct()` and `__invoke()` now and I've added two extra sets of tests at the bottom of the test file to safeguard that those will continue to trigger errors.

jrfnl:
(also updated the commit description above now)

DannyvdSluijs (approved):

jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/50

jrfnl:

PR #3716, opened 2022-11-21
Runner::processFile(): make internal error more informative

jrfnl:
When sniffs use traits/utility classes/helper functions, the message for the `Internal.Exception` can be unclear as it may point to the trait/utility classes/helper function as the culprit without any indication of which sniff caused the error.

This can cause bugs in sniffs to be reported by end-users in the wrong repository, but even when reported in the right repository, it makes the report unclear and means the maintainer will need to do more work to try and reproduce the issue.

This PR attempts to make the `Internal.Exception` error message more informative by pinpointing the sniff which caused the issue to occur. While this _may_ result in duplicate information when no traits/utility classes/helper functions are involved, for those cases when those are involved, I believe this will be a useful change.

Example output for a (simulated) error _without_ this change:
```
------------------------------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------------------------------
 1 | ERROR | An error occurred during processing; checking has been aborted. The error message was: Undefined variable
   |       | $tabWidth in
   |       | path/to/PHP_CodeSniffer/src/Standards/Generic/Sniffs/WhiteSpace/DisallowSpaceIndentSniff.php
   |       | on line 64 (Internal.Exception)
------------------------------------------------------------------------------------------------------------------------
```

Example output for the same (simulated) error _with_ this change:
```
------------------------------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------------------------------
 1 | ERROR | An error occurred during processing; checking has been aborted. The error message was: Undefined variable
   |       | $tabWidth in
   |       | path/to/PHP_CodeSniffer/src/Standards/Generic/Sniffs/WhiteSpace/DisallowSpaceIndentSniff.php
   |       | on line 64
   |       | The error originated in the Generic.WhiteSpace.DisallowSpaceIndent sniff on line 64.
   |       | (Internal.Exception)
------------------------------------------------------------------------------------------------------------------------
```

Example output for a (simulated) error _without_ this change for a sniff using a helper function:
```
------------------------------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------------------------------
 1 | ERROR | An error occurred during processing; checking has been aborted. The error message was: The
   |       | PHPCSUtils\Tokens\Collections::functionDeclarationTokensBC() method is deprecated since PHPCSUtils
   |       | 1.0.0-alpha4. Use the PHPCSUtils\Tokens\Collections::functionDeclarationTokens() method instead. in
   |       | path/to/PHPCSUtils/PHPCSUtils/Tokens/Collections.php on line 753
   |       | (Internal.Exception)
------------------------------------------------------------------------------------------------------------------------
```

Example output for the same (simulated) error _with_ this change for a sniff using a helper function:
```
------------------------------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------------------------------
 1 | ERROR | An error occurred during processing; checking has been aborted. The error message was: The
   |       | PHPCSUtils\Tokens\Collections::functionDeclarationTokensBC() method is deprecated since PHPCSUtils
   |       | 1.0.0-alpha4. Use the PHPCSUtils\Tokens\Collections::functionDeclarationTokens() method instead. in
   |       | path/to/PHPCSUtils/PHPCSUtils/Tokens/Collections.php on line 753
   |       | The error originated in the PHPCompatibility.Interfaces.NewInterfaces sniff on line 201.
   |       | (Internal.Exception)
------------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -652,6 +652,39 @@ class Runner
             }
         } catch (\Exception $e) {
             $error = 'An error occurred during processing; checking has been aborted. The error message was: '.$e->getMessage();
+
+            // Determine which sniff caused the error.
+            $sniffStack = null;
+            $nextStack  = null;
+            foreach ($e->getTrace() as $step) {
+                if (isset($step['file']) === false) {
+                    continue;
+                }
+
+                if (empty($sniffStack) === false) {
+                    $nextStack = $step;
+                    break;
+                }
+
+                if (substr($step['file'], -9) === 'Sniff.php') {
+                    $sniffStack = $step;
+                    continue;
+                }
+            }
+
+            if (empty($sniffStack) === false) {
+                if (empty($nextStack) === false
+                    && isset($nextStack['class']) === true
+                    && substr($nextStack['class'], -5) === 'Sniff'
+                ) {
+                    $sniffCode = Common::getSniffCode($nextStack['class']);
+                } else {
+                    $sniffCode = substr(strrchr(str_replace('\\', '/', $sniffStack['file']), '/'), 1);
+                }
+
+                $error .= sprintf(PHP_EOL.'The error originated in the %s sniff on line %s.', $sniffCode, $sniffStack['line']);
+            }
+
             $file->addErrorOnLine($error, 1, 'Internal.Exception');
         }//end try
 


src/Runner.php
```diff
@@ -652,6 +652,39 @@ public function processFile($file)
...
+                if (isset($step['file']) === false) {
+                    continue;
+                }
+
+                if (substr($step['file'], -9) === 'Sniff.php') {
+                    $sniffStack = $step;
+                    continue;
+                }
+
+                if (empty($sniffStack) === false) {
```
BafS, line 669: why not?
```suggestion
                if (!empty($sniffStack)) {
```

or even more simply

```suggestion
                if ($sniffStack) {
```

both are more readable

BafS: Oh I see, I was not aware of this sorry, it's a bit strange because it's less readable and less efficient.

BafS (approved):

src/Runner.php
jrfnl: 🤷🏻‍♀️

jrfnl:
Found a logic error in the original code. Fixed it and force pushed the commit again.

jrfnl:

PR #3717, opened 2022-11-21
Fix `Squiz.Commenting.FunctionComment.InvalidNoReturn` false positive when return type is `never`

axlon:
The `Squiz.Commenting.FunctionComment.InvalidNoReturn` causes false positives when it encounters the `never` return type.

Tested this locally with:

```xml
<rule ref="Squiz.Commenting.FunctionComment.InvalidNoReturn"/>
```

```php
/**
 * @return never
 */
private function foo()
{
    throw new \RuntimeException('This method never returns');
}
```

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -143,9 +143,12 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
                             }
                         }
                     }//end if
-                } else if ($returnType !== 'mixed' && in_array('void', $typeNames, true) === false) {
-                    // If return type is not void, there needs to be a return statement
-                    // somewhere in the function that returns something.
+                } else if ($returnType !== 'mixed'
+                    && $returnType !== 'never'
+                    && in_array('void', $typeNames, true) === false
+                ) {
+                    // If return type is not void, never, or mixed, there needs to be a
+                    // return statement somewhere in the function that returns something.
                     if (isset($tokens[$stackPtr]['scope_closer']) === true) {
                         $endToken = $tokens[$stackPtr]['scope_closer'];
                         for ($returnToken = $stackPtr; $returnToken < $endToken; $returnToken++) {
--- a/src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
+++ b/src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
@@ -1129,3 +1129,8 @@ public function variableCaseTest(
 public function variableOrderMismatch($bar, $baz, $foo) {
     return;
 }
+
+/**
+ * @return never
+ */
+function foo() {}
--- a/src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc.fixed
+++ b/src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc.fixed
@@ -1129,3 +1129,8 @@ public function variableCaseTest(
 public function variableOrderMismatch($bar, $baz, $foo) {
     return;
 }
+
+/**
+ * @return never
+ */
+function foo() {}


axlon:
@jrfnl I managed to add the test, please let me know if anything needs changing 👍

jrfnl (approved):
@axlon Thanks for adding that test. All looks good!

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -143,7 +143,7 @@ protected function processReturn(File $phpcsFile, $stackPtr, $commentStart)
                             }
                         }
                     }//end if
-                } else if ($returnType !== 'mixed' && in_array('void', $typeNames, true) === false) {
+                } else if ($returnType !== 'mixed' && array_intersect($typeNames, ['void', 'never']) === []) {
```
jrfnl, line 146: I wonder if the `array_intersect()` could be avoided ?

src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -1006,6 +1006,11 @@ function foo($foo) {}
  */
 public function foo($a, $b) {}
 
+/**
```
jrfnl, line 1009: Why did you place this test in the middle of the `inheritDoc` test instead of at the end of the file ?

src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.php
```diff
@@ -115,7 +115,7 @@ public function getErrorList()
             997  => 1,
             1004 => 2,
             1006 => 1,
-            1029 => 1,
+            1034 => 1,
```
jrfnl, line 118: If the tests would stay in the current place, you will also need to adjust the line number referenced below on line 154 (test which only expects an error on PHP < 8.0).

jrfnl (changes requested):
@axlon I'm sorry, I spoke too soon. Looks like the tests are failing now. Could you please have a look ?

axlon:
@jrfnl I made the changes you requested, hope everything is ok, could you start the workflow for tests? Currently don't have a PHP 7.0 environment available to test on.

jrfnl (changes requested):
> one thing I changed from before is that an error should now be emitted if a function has a documented return type of `never` but has a return statement (which already happened for `void` functions). I couldn't figure out how write a test for this, could you point me in the right direction?
> 
> ```
> -------------------------------------------------------------------------------------------
> FOUND 1 ERROR AFFECTING 1 LINE
> -------------------------------------------------------------------------------------------
>  182 | ERROR | Function return type is never, but function contains return statement
> -------------------------------------------------------------------------------------------
> ```

This change evolves this PR from a bug fix to an enhancement and will definitely need more tests.

It also looks like this new part is unfinished as it doesn't seem to work the way it should ?

Could you please have another look at this ?

axlon:
@jrfnl 

> This change evolves this PR from a bug fix to an enhancement and will definitely need more tests.

I think you are right, it definitely became a lot larger than I intended and I will need to immerse myself a bit more into writing tests for this library. I think I will revert this to only being a bugfix for now, and will follow up with a full fledged PR later. 👍

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -138,13 +138,16 @@ protected function processReturn(File $phpcsFile, $stackPtr, $commentStart)
...
+                                $error = "Function return type is {$returnType}, but function contains return statement";
                                 $phpcsFile->addError($error, $return, 'InvalidReturnVoid');
                             }
                         }
                     }//end if
-                } else if ($returnType !== 'mixed' && in_array('void', $typeNames, true) === false) {
-                    // If return type is not void, there needs to be a return statement
+                } else if ($returnType !== 'mixed'
+                    && in_array('never', $typeNames) === false
+                    && in_array('void', $typeNames) === false
```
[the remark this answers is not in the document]
axlon: Yeah, I agree. The name of `never` says it all; the function never returns. So it couldn't be part of a union because that would mean it _does_ return something at least sometimes.

Similarly, I think `void` in a union also never makes sense (and should be replaced with `null`), but in the context of this PR I think it should be left as is.

axlon:
@jrfnl I think the PR is ready for review now

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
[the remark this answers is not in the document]
jrfnl: > Similarly, I think `void` in a union also never makes sense (and should be replaced with `null`), but in the context of this PR I think it should be left as is.

Agreed.

jrfnl:
@axlon Would you mind rebasing this PR (and possibly squashing the commits) to get it past the merge conflict ?

jrfnl (approved):
Thanks @axlon ! The isolated bug fix, which this PR is now (again), looks good.

Just checking: do you intend to submit a PR for the follow-up which was discussed in the comments as well ?

axlon:
@jrfnl Thanks! Yes, I still intend to create a followup PR whenever I find the time, but if someone else picks it up before me that is fine by me as well

jrfnl:

PR #3720, opened 2022-11-24
Generic/RequireStrictTypes: various bugfixes

jrfnl:
### Generic/RequireStrictTypes: add extra tests

These tests safeguard the following, which is already handled correctly by the sniff:
* Execution directives are case-insensitive in PHP.
* The sniff should ignore docblocks between a PHP open tag and a declare statement.

### Generic/RequireStrictTypes: bug fix - limit token search to the statement

While hopefully unlikely in real life, the sniff should handle parse errors correctly.

Along the same lines, live coding should be handled correctly and a `declare()` statement should only be examined once the statement has been completed.

This updates the sniff to handle both situations correctly.
Without these fixes, no error would be thrown in test case file 5 or 6 (false negative), while an error would be thrown in test case file 7 (false positive).

Includes unit tests.

Includes minor efficiency tweak to only start looking for a "next" token _after_ the current token.

### Generic/RequireStrictTypes: bug fix - allow for multi directive statements

PHP allows for multiple directives to be passed in one `declare()` statement.

The sniff, however, did not allow for this, which could lead to false positives.

Fixed now.

Includes unit tests.

### Generic/RequireStrictTypes: add warning for when value is 0

Strict types is disabled when the value for the `strict_types` directive is `0`.

As this sniff is supposed to be about enforcing the use of `strict_types`, `strict_types` declarations which turn the feature off should be flagged as well.

Implemented with a separate error code to allow for selectively turning this warning off.

Includes dedicated tests.

---

## Future Scope

This sniff isn't very efficient as it will walk the whole file until a `declare()` statement is found or until the end of the file is reached.

Declare statements for `strict_types` MUST be the first statement in a file, so IMO there are two options:
* Either refactor the sniff to look for the first statement in a file instead of walking a complete file.
* Or alternatively, keep the token walking, but don't stop on the first `declare()` statement and flag any `strict_types` declarations which are not the first statement in a file with a separate error code.

In both cases, "view" files containing a mix of PHP and HTML should be taken into account, similarly, hashbang lines should be taken into account.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -771,6 +771,15 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="RequireStrictTypesUnitTest.8.inc" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="RequireStrictTypesUnitTest.9.inc" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="RequireStrictTypesUnitTest.10.inc" role="test" />
+        <file baseinstalldir="PHP/CodeSniffer" name="RequireStrictTypesUnitTest.11.inc" role="test" />
+        <file baseinstalldir="PHP/CodeSniffer" name="RequireStrictTypesUnitTest.11.inc.fixed" role="test" />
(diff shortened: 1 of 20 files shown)


jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/51

jrfnl:

PR #3724, opened 2022-12-05
restart phpcs/phpcbf when xdebug is loaded

### LINKED ISSUE #3337: Codesniffer slow when Xdebug is enabled
I don't know if this is an issue under Code Sniffer or Xdebug, but when i run code sniffers latest version with Xdebug enabled,
if becomes very slow, until even the point that its nor working anymore. i saw in previous comments/ reports that PHPCS is not using xdebug. But never the less when xdebug is toggled its a difference

OS is windows 10 Pro in combination with IIS:
I have upgraded PHP NTS from 8.0.3 to 8.0.6 same behavior
I have tested with xdebug 3.0.3 and also 3.0.4(latest version)

When switching between Code sniffer 3.6.0 and the previous version 3.5.8 its makes a massive difference on execution time, this is when xdebug is enabled.
I have 1435 classes, and files to process and when cache turned off it takes around: Time: 3 mins, 24 secs; Memory: 438MB

But when I run the same set under 3.6.0 then I never been able to finish yet (again when xdebug is enabled).
If I disable xdebug, then 3.6.0 is working fine.

I don't know if someone have the same issue, or can replicate this problem? Again, its possible that the issue is xdebug related, but i only experience it with PHPCS.

Xdebug config from the php.ini file
```
; XDEBUG may only be loaded on the development servers and are used to get more information regarding exceptions and errors that happen.
[PHP_XDEBUG]
zend_extension = xdebug

; Additional configuration for some of the extensions
[xdebug]
xdebug.mode = develop
xdebug.var_display_max_depth = -1
xdebug.var_display_max_children = -1
xdebug.var_display_max_data = -1
xdebug.scream = 0
xdebug.force_error_reporting = 1
xdebug.force_display_errors = 1
xdebug.output_dir = "D:\Inetpub\logs\xdebug\"
; Enable profiler to analyses slow pages
;xdebug.mode = profile
xdebug.profiler_output_name = "callgrind.out.%H_%R.%u"
; Enable trace system for memory
; xdebug.mode=trace
xdebug.trace_output_name = "trace.out.%u.%H_%R"

```

staabm:
closes https://github.com/squizlabs/PHP_CodeSniffer/issues/3337

running `phpcs` on the project itself takes 11-12 seconds when xdebug is loaded.
with the added restart script it takes 6-7 seconds. so its roughly 40-50% faster.

running `phpcs` with a primed cache takes ~1.8 seconds without this PR and ~0.9 seconds with the PR.

test on mac m1 pro 
```
php -v
PHP 8.1.13 (cli) (built: Nov 24 2022 15:58:42) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.13, Copyright (c) Zend Technologies
    with Xdebug v3.1.5, Copyright (c) 2002-2022, by Derick Rethans
    with Zend OPcache v8.1.13, Copyright (c), by Zend Technologies
```

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -73,6 +73,8 @@ if (class_exists('PHP_CodeSniffer\Autoload', false) === false) {
                     return;
                 }
 
+                require_once __DIR__.'/vendor/autoload.php';
+
                 if (strpos(__DIR__, 'phar://') !== 0
                     && @file_exists(__DIR__.'/../../autoload.php') === true
                 ) {
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
         "php": ">=5.4.0",
         "ext-tokenizer": "*",
         "ext-xmlwriter": "*",
-        "ext-simplexml": "*"
+        "ext-simplexml": "*",
+        "composer/xdebug-handler": "^3.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.0 || ^5.0 || ^6.0 || ^7.0"
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -12,6 +12,7 @@
 
 namespace PHP_CodeSniffer;
 
+use Composer\XdebugHandler\XdebugHandler;
 use PHP_CodeSniffer\Exceptions\DeepExitException;
 use PHP_CodeSniffer\Exceptions\RuntimeException;
 use PHP_CodeSniffer\Files\DummyFile;
@@ -53,6 +54,10 @@ class Runner
      */
     public function runPHPCS()
     {
+        $xdebug = new XdebugHandler('phpcs');
+        $xdebug->check();
+        unset($xdebug);
+
         $this->registerOutOfMemoryShutdownMessage('phpcs');
 
         try {
@@ -155,6 +160,10 @@ class Runner
      */
     public function runPHPCBF()
     {
+        $xdebug = new XdebugHandler('phpcbf');
+        $xdebug->check();
+        unset($xdebug);
+
         $this->registerOutOfMemoryShutdownMessage('phpcbf');
 
         if (defined('PHP_CODESNIFFER_CBF') === false) {


jrfnl (commented):
I very much like the principle of this change, however I do see some potential problems.

1. I know this project is not keen on adding additional run-time dependencies.
    If this would be accepted, I believe the script to generate the PHAR file - `scripts/build-phar.php` - will need to be adjusted too.
    I also think that this will probably not play nice with a PEAR install, which this package still supports, though PEAR support will be dropped in PHPCS 4.x.
2. Loading the `vendor/autoload.php` file may have side-effects for the functioning of certain sniffs.
    Most sniffs are written to  presume a "clean" environment with only PHPCS loaded in memory.
    Loading the `vendor/autoload.php` file can result in other packages loading files into memory, especially packages which are intended as polyfills have a habit of doing this.
    Any sniff which includes a presumption of a "clean" environment may now start to throw false positives/negatives if they, for instance, use `function_exists()` or `class_exists()` to check for PHP native functions versus userland functions - or when they want to flag PHP functions not available on the PHP version on which PHPCS is being run.
   
What could help, would be if the `xdebug-handler` package contained its own autoloader and to only include that autoloader instead of including `vendor/autoload.php`. Still not sure if that's enough though.

What would be even better, would be a PHPCS native implementation of the same, though I haven't looked in detail at the package to know how much work (and code) would be involved in that.

Either way, just my two pennies.

staabm:
> 2\. Loading the `vendor/autoload.php` file may have side-effects for the functioning of certain sniffs.

the external package added [consists only of 4 classes](https://github.com/composer/xdebug-handler/tree/main/src). we could include these 4 files manually so we don't need to bootstrap the whole autoloader.

> 1. If this would be accepted, I believe the script to generate the PHAR file - `scripts/build-phar.php` - will need to be adjusted too.

thanks for the hint. I can try to take care of it, if required.

staabm:
Ping @gsherwood

PR #3738, opened 2023-01-04
Update LineLengthSniff.php

Stepa4man:


jrfnl (commented):
These limits can be changed from a ruleset - see: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#genericfileslinelength

Changing the limits of the sniff itself would be a HUGE breaking change, which would affect every single project using this sniff, so IMO, this is not an acceptable change.

PR #3741, opened 2023-01-08
$data variable should be array

juanpgarciac:
in some analysis, I was receiving this error "Uncaught TypeError: vsprintf(): Argument #2 ($values) must be of type array, string given" on line 1056

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -1053,7 +1053,7 @@ class File
         }
 
         if (empty($data) === false) {
-            $message = vsprintf($message, $data);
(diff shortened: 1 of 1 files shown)


jrfnl:
@juanpgarciac Thanks foŕyour willingness to contribute.

In my opinion, this is the wrong fix. This should be fixed at the point where this function is being passed incorrect information by a call to `File::addError()` or `File::addWarning()`, not here.
If anything, the `$data` parameter here should probably get an `array` type declaration.

juanpgarciac:
Ok, agreed. I will check on your suggestion later on. Thanks!

PR #3751, opened 2023-02-03
Allow auto-fixing Generic.Commenting.DocComment

fredden:
See also #3752 where this same change is being applied to the `master` branch.


fredden:
Per the contribution guidelines, this can be safely closed in favour of #3752.

PR #3737, opened 2023-01-04
FileComment: update year in test files

jrfnl:
The regular annual update to make sure the build still passes ;-)

Happy 2023!

PR #3753, opened 2023-02-03
Fix failing test

fredden:
This file includes the current year, and is now failing. I noticed this while working on #3751 / #3752.


jrfnl:
Duplicate of #3737....

fredden:
So it is. Thanks for highlighting.

PR #3756, opened 2023-02-07
Allow multiple spaces in PropertyDeclaration sniff to respect more carefully the PSR-12.

### LINKED ISSUE #3729: PSR12 / There must be 1 space after the property type
**Describe the bug**

**Code sample**
```php
class test 
{
    public FlightHeader  $flight_header;
    public Carbon        $flight_date;
    public BookingNumber $booking_number;
}
```

**Custom ruleset**
```xml
<?xml version="1.0"?>
<ruleset name="schauinsland" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/squizlabs/PHP_CodeSniffer/master/phpcs.xsd">
    <rule ref="PSR12"/>
    <rule ref="Squiz.NamingConventions.ValidVariableName.NotCamelCaps"/>
    <rule ref="Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps"/>
    <rule ref="Squiz.NamingConventions.ValidVariableName.StringNotCamelCaps"/>
    <rule ref="Generic.Arrays.DisallowLongArraySyntax">
        <type>warning</type>
    </rule>
        <properties>
            <property name="forbiddenFunctions" type="array">
                <element key="sizeof" value="count"/>
                <element key="delete" value="unset"/>
                <element key="print" value="echo"/>
                <element key="create_function" value="null"/>
            </property>
        </properties>
    </rule>

    <exclude-pattern>/vendor/</exclude-pattern>
</ruleset>
```

**To reproduce**
```
 11 | ERROR | [x] There must be 1 space after the property type
    |       |     declaration; 2 found
 12 | ERROR | [x] There must be 1 space after the property type
    |       |     declaration; 8 found
```

**Expected behavior**
no error

**Reason**
The PSR12 does not disalow multiple spaces after the declaration:

> There MUST be a space between type declaration and property name.

it just says there must be a space, not one space

niconoe-:
Fix #3729 

Note to maintainers: I wasn't in capacity of unit test this before any push. I'm having PHP 8.0 on my machine and when I run `php vendor/bin/phpunit` I have a fatal error due to a usage of the `each()` internal function by PHPUnit which is now removed from the PHP lib. I really think a more recent version of PHPUnit should be used, but that's another story.

### DIFF
--- a/src/Standards/PSR2/Sniffs/Classes/PropertyDeclarationSniff.php
+++ b/src/Standards/PSR2/Sniffs/Classes/PropertyDeclarationSniff.php
@@ -85,7 +85,9 @@ class PropertyDeclarationSniff extends AbstractVariableSniff
                 if ($tokens[$next]['line'] !== $tokens[$typeToken]['line']) {
                     $found = 'newline';
                 } else {
-                    $found = $tokens[($typeToken + 1)]['length'];
+                    // According to PSR-12: "There MUST be a space between type declaration and property name."
+                    // This does not mean a single space only is expected, but at least 1. So, if many, ignore sniff.
+                    return;
                 }
 
                 $data = [$found];
@@ -96,17 +98,13 @@ class PropertyDeclarationSniff extends AbstractVariableSniff
                 } else {
                     $fix = $phpcsFile->addFixableError($error, $typeToken, 'SpacingAfterType', $data);
                     if ($fix === true) {
-                        if ($found === 'newline') {
-                            $phpcsFile->fixer->beginChangeset();
-                            for ($x = ($typeToken + 1); $x < $next; $x++) {
-                                $phpcsFile->fixer->replaceToken($x, '');
-                            }
-
-                            $phpcsFile->fixer->addContent($typeToken, ' ');
-                            $phpcsFile->fixer->endChangeset();
-                        } else {
-                            $phpcsFile->fixer->replaceToken(($typeToken + 1), ' ');
+                        $phpcsFile->fixer->beginChangeset();
+                        for ($x = ($typeToken + 1); $x < $next; $x++) {
+                            $phpcsFile->fixer->replaceToken($x, '');
                         }
+
+                        $phpcsFile->fixer->addContent($typeToken, ' ');
+                        $phpcsFile->fixer->endChangeset();
                     }
                 }
             }//end if
--- a/src/Standards/PSR2/Tests/Classes/PropertyDeclarationUnitTest.inc.fixed
+++ b/src/Standards/PSR2/Tests/Classes/PropertyDeclarationUnitTest.inc.fixed
@@ -38,7 +38,7 @@ class ABC {
     public static $propD;
     protected static
         $propE;
-    private static /*comment*/   $propF;
+    private static    /*comment*/   $propF;
 }
 
 class MyClass
@@ -65,7 +65,7 @@ class MyClass
 {
     public string $var = null;
     protected ?Folder\ClassName $var = null;
-    public int $var = null;
+    public int   $var = null;
     public static int /*comment*/$var = null;
 }
 
@@ -74,7 +74,7 @@ class ReadOnlyProp {
         $bar,
         $var = null;
 
-    protected readonly ?string $foo;
+    protected readonly ?string    $foo;
 
     readonly array $foo;
 }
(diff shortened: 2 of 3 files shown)


niconoe-:
@gsherwood I'm not actually a fan of allowing multiple spaces here. I took this issue a little bit because of the state of art, and also mainly because I wish I could contribute more to PHPCS in the future, and I needed a small thing to start, but I'm not as convinced as OP about how to interpret the PSR statement about it.

If I can help you in another way without taking care of something drown into its own complexity for my first contributions, I would love to do it.
Thanks a lot.

jrfnl:
Re: the PR - as this is a PSR-2 sniff, both the text in PSR2 as well as PSR12 will need to be taken into account. It also may need clarification for the PSR group (may already be available in the errata, but I haven't checked).

>  I have a fatal error due to a usage of the each() internal function by PHPUnit

IIRC running the test on PHP 8.0 will work fine with PHPUnit 7.x. I usually use the PHPUnit Phar, but you can also install via Composer using `--ignore-platform-reqs`.
Also see: https://github.com/squizlabs/PHP_CodeSniffer/blob/add95a74551c3ba8fc99ef7651ad05f553b3fbbf/.github/workflows/test.yml#L115-L142

niconoe-:
> Re: the PR - as this is a PSR-2 sniff, both the text in PSR2 as well as PSR12 will need to be taken into account. It also may need clarification for the PSR group (may already be available in the errata, but I haven't checked).

Here are the definitions of [PSR-12](https://www.php-fig.org/psr/psr-12/#43-properties-and-constants) and [PSR-2](https://www.php-fig.org/psr/psr-2/#42-properties). I wasn't able to find any erratum on any of those PSR. 

The [PER](https://www.php-fig.org/per/coding-style/#43-properties-and-constants) that extends and replaces [PSR-12] is a copy/paste of PSR-12 content on the section about the properties and constants style, so it doesn't help much more.

> I usually use the PHPUnit Phar, but you can also install via Composer using --ignore-platform-reqs.
I tried the PHAR way but it requires me to change my $PATH which I don't want to, as it will affect the way I'm dealing with the projects of my company.
I tried also with the `--ignore-platform-reqs` but I got the same error message.

TBH, I was expecting GitHub to trigger some tests based on its workflow files and if any issue came this way, I would have been able to fix them, but a maintainer has to approve running the workflows instead of being automatic…

I'll try with a docker container in PHP 5.4 if I can, but I just hope it won't trigger other issues like absolute paths usages in unit tests that can't exist in my container as well.

niconoe-:
I'm sorry, but I tried to run `./vendor/bin/phpunit` from:
- docker container php:5.4-cli
- docker container php:5.6-cli
- docker container php:7.4-cli
- docker container composer:2 (using PHP 8.1 internally)

And no matter what, I always have either the `undefined each() function` error, or the `Class 'PHP_Timer' not found` error.

Without a better way to run PHPUnit on this repository, I won't be able to contribute anymore.

jrfnl:
> And no matter what, I always have either the `undefined each() function` error, or the `Class 'PHP_Timer' not found` error.

I'm not familiar with those Docker containers, but that sounds like the wrong PHPUnit version is being used (PHPUnit 4.x before a PHP requirement was included in the PHPUnit `composer.json`) and that those Docker containers still run the tests against PHP 8.0 (as otherwise you wouldn't get the notice about `each()`).

> Without a better way to run PHPUnit on this repository, I won't be able to contribute anymore.

IIRC, once at least one PR has been approved to run and the PR has been accepted, CI will run automatically for PRs to the repo by the same contributor submitted after that.

In the mean time, if you want to see what CI will spit out, you can turn on the GH Actions runs in your own fork of this repo.
This is not enable by default by GitHub, but IIRC, if you go to the "Actions" tab in the GH webinterface for your fork, you should be able to turn actions on. If you then force push this PR without changes, the actions should run.

niconoe-:
I finally found a docker image (jakzal/phpqa:php7.4) that is having composer and PHP 7.4 so I was able to update my dependencies and use PHPUnit 7.5 instead of PHPUnit 5 which is bugged. This made me fix the unit tests regarding my changes, so now, it passes.

This PR now needs review and approval. Thanks a lot for your feedbacks @jrfnl ! :slightly_smiling_face:

PR #3757, opened 2023-02-10
Squiz/VariableComment: allow union types to be detected

fcool:
If one uses a union type for an object property, the doc-type-hint could not be related to the variable type.

Fixed now.

Includes unit test.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/VariableCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/VariableCommentSniff.php
@@ -40,6 +40,7 @@ class VariableCommentSniff extends AbstractVariableSniff
             T_STRING       => T_STRING,
             T_NS_SEPARATOR => T_NS_SEPARATOR,
             T_NULLABLE     => T_NULLABLE,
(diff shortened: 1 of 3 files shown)


fcool:
The unit test will fail, without the adjustment made in this PR, because it detects a missing @var annotation - which is incorrect ;)

jrfnl:
@fcool In light of #3932, I'd like to invite you to recreate this PR in the https://github.com/PHPCSStandards/PHP_CodeSniffer repository.

There's a related issue open in that repo (https://github.com/PHPCSStandards/PHP_CodeSniffer/issues/178) and it may be a good idea to add the `T_TYPE_INTERSECTION` token (+ test) to the PR as well.

If you don't have time to do so, I can cherrypick this commit if need be. Just let me know.

fcool:
So if I understood everything correct, I need just to take over this PR to the new repository? 

Glad you stepped up, by the way.

jrfnl:
> So if I understood everything correct, I need just to take over this PR to the new repository? And of course, adding `T_TYPE_INTERSECTION`?

That would be great, yes!

PR #3761, opened 2023-02-16
PHP 8.1 | Config: fix passing null to non-nullable notice

### LINKED ISSUE #3760: PHP Deprecated:  preg_match(): Passing null to parameter #2
**Describe the bug**
When running the phpcs on a WordPress plugin, I get this error:

```
PHP Deprecated:  preg_match(): Passing null to parameter #2 ($subject) 
of type string is deprecated in ../squizlabs/php_codesniffer/src/Config.php on line 228

Deprecated: preg_match(): Passing null to parameter #2 ($subject) 
of type string is deprecated in ../squizlabs/php_codesniffer/src/Config.php on line 228
....... 7 / 7 (100%)
```

**Expected behavior**
No error showing up

 - OS: MacOS
 - PHP: 8.2
 - PHP_CodeSniffer version 3.7.1 (stable) by Squiz (http://www.squiz.net)
 - Standard: WordPress

**Additional context**
```
stty --version
stty (GNU coreutils) 8.23
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

jrfnl:
As per issue #3760, if the `stty size` command errors out, `shell_exec()` will return `null`. Per the [documentation](https://www.php.net/manual/en/function.shell-exec.php#refsect1-function.shell-exec-returnvalues), the `shell_exec()` command can also return `false`, which would also be an unusable value (`preg_match()` will not match, so `'auto'` would be cast to int, resulting in `0` being set as the value).

If the output of `shell_exec()` would be `null`, PHP as of PHP 8.1 will throw a "passing null to non-nullable" notice.

Additionally, if the output from `shell_exec()` would be `false`, the original input would be an invalid non-numeric value or the original input would be `auto`, but the `shell_exec()` command is disabled, the code as-it-was, would set the `reportWidth` to `0`, which is not that helpful. See: https://3v4l.org/GE7sK

~~I've now changed the logic to only update the `reportWidth` property value when the received value is `auto` AND a valid CLI width could be determined. This solves the originally reported issue + two of the above mentioned additional issues. The only one _not_ solved is "_the original input would be an invalid non-numeric value _", but in that case, it's clear user-error and not something which PHPCS needs to handle.~~

I've now changed the logic to:
* If the received value is `auto`, check if a valid CLI width can be determined and only then use that value.
* In all other cases, including when the value is set directly on the object, the value will be validated and if not valid, a default value of width `80` will be used, which is in line with the default width as stated in [the documentation](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Configuration-Options#setting-the-default-report-width).

Fixes #3760

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -90,6 +90,13 @@ class Config
     const STABILITY = 'stable';
 
     /**
+     * Default report width when no report width is provided and 'auto' does not yield a valid width.
+     *
+     * @var int
+     */
+    const DEFAULT_REPORT_WIDTH = 80;
+
+    /**
      * An array of settings that PHPCS and PHPCBF accept.
      *
      * This array is not meant to be accessed directly. Instead, use the settings
@@ -223,13 +230,20 @@ class Config
         switch ($name) {
         case 'reportWidth' :
             // Support auto terminal width.
-            if ($value === 'auto'
-                && function_exists('shell_exec') === true
-                && preg_match('|\d+ (\d+)|', shell_exec('stty size 2>&1'), $matches) === 1
-            ) {
-                $value = (int) $matches[1];
-            } else {
+            if ($value === 'auto' && function_exists('shell_exec') === true) {
+                $dimensions = shell_exec('stty size 2>&1');
+                if (is_string($dimensions) === true && preg_match('|\d+ (\d+)|', $dimensions, $matches) === 1) {
+                    $value = (int) $matches[1];
+                    break;
+                }
+            }
+
+            if (is_int($value) === true) {
+                $value = abs($value);
+            } else if (is_string($value) === true && preg_match('`^\d+$`', $value) === 1) {
                 $value = (int) $value;
+            } else {
+                $value = self::DEFAULT_REPORT_WIDTH;
             }
             break;
         case 'standards' :


jrfnl:
Note: the build failure is unrelated to this PR and will be fixed once #3737 has been merged.

src/Config.php
```diff
@@ -223,11 +223,13 @@ public function __set($name, $value)
...
-            ) {
-                $value = (int) $matches[1];
+            if ($value === 'auto') {
+                if (function_exists('shell_exec') === true) {
+                    $dimensions = shell_exec('stty size 2>&1');
+                    if (is_string($dimensions) === true && preg_match('|\d+ (\d+)|', $dimensions, $matches) === 1) {
+                        $value = (int) $matches[1];
+                    }
+                }
             } else {
```
fredden, line 233: Do we need any 'else' clauses here? If the `$value` is `auto` and we can't detect a sensible value, should we hard-code a fall-back width?
```suggestion
                        $value = (int) $matches[1];
                    } else {
                        $value = self::DEFAULT_REPORT_WIDTH;
                    }
                } else {
                    $value = self::DEFAULT_REPORT_WIDTH;
                }
            } else {
```

jrfnl: IIRC, the default value is `80`, but yeah, might make sense having that encoded here. Still, I'd like to hear @gsherwood's opinion on it before implementing it.

jrfnl: @gsherwood and me discussed this and I'll update this PR to include a default with value.

jrfnl:
I've updated this PR to include a default value of `80`, which is in line with the [documentation about report width in the wiki](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Configuration-Options#setting-the-default-report-width).

I've also updated the commit message and the PR description to be in line with this additional change.

Compared to the original PR, the main difference is that the "the original input would be an invalid non-numeric value" case is now also handled and will silently ignore the invalid value and fall back to the default width of `80`.

### Suggested changelog entry

> The "reportWidth" will now consistently fall back to a width of `80` when either an invalid report width was passed or the width of the terminal could not be determined when report width was set to `'auto'`.
>    This will also prevent a potential "passing null to non-nullable" deprecation notice on PHP 8.1+.

jrfnl:

PR #3770, opened 2023-02-27
Squiz/NonExecutableCode: fix false positives when code goes in and out of PHP/HTML

fredden:
The following code snippet is currently flagging as `Squiz.PHP.NonExecutableCode.Unreachable`, but probably shouldn't.

```php
<?php if (true): ?>
    <?php foreach ([] as $item): ?>
        <?php continue; ?>
    <?php endforeach; ?>
<?php endif; ?>
```

This pull request adds a check to avoid this snippet from being flagged as problematic.

### DIFF
--- a/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
@@ -258,6 +258,16 @@ class NonExecutableCodeSniff implements Sniff
                 continue;
             }
 
+            // Skip HTML whitespace.
+            if ($tokens[$i]['code'] === T_INLINE_HTML && \trim($tokens[$i]['content']) === '') {
+                continue;
+            }
+
+            // Skip PHP re-open tag (eg, after inline HTML).
+            if ($tokens[$i]['code'] === T_OPEN_TAG) {
+                continue;
+            }
+
             $line = $tokens[$i]['line'];
             if ($line > $lastLine) {
                 $type    = substr($tokens[$stackPtr]['type'], 2);
--- a/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
@@ -204,8 +204,8 @@ class NonExecutableCodeSniff implements Sniff
             $end = ($phpcsFile->numTokens - 1);
         }//end if
 
-        // Find the semicolon that ends this statement, skipping
-        // nested statements like FOR loops and closures.
+        // Find the semicolon or closing PHP tag that ends this statement,
+        // skipping nested statements like FOR loops and closures.
         for ($start = ($stackPtr + 1); $start < $phpcsFile->numTokens; $start++) {
             if ($start === $end) {
                 break;
@@ -225,11 +225,7 @@ class NonExecutableCodeSniff implements Sniff
                 continue;
             }
 
-            if ($tokens[$start]['code'] === T_SEMICOLON) {
-                break;
-            }
-
-            if ($tokens[$start]['code'] === T_CLOSE_TAG) {
+            if ($tokens[$start]['code'] === T_SEMICOLON || $tokens[$start]['code'] === T_CLOSE_TAG) {
                 break;
             }
         }//end for
--- a/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
@@ -228,6 +228,10 @@ class NonExecutableCodeSniff implements Sniff
             if ($tokens[$start]['code'] === T_SEMICOLON) {
                 break;
             }
+
+            if ($tokens[$start]['code'] === T_CLOSE_TAG) {
+                break;
+            }
         }//end for
 
         if (isset($tokens[$start]) === false) {
--- a/package.xml
+++ b/package.xml
@@ -1857,6 +1857,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
         <file baseinstalldir="PHP/CodeSniffer" name="LowercasePHPFunctionsUnitTest.php" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="NonExecutableCodeUnitTest.1.inc" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="NonExecutableCodeUnitTest.2.inc" role="test" />
+        <file baseinstalldir="PHP/CodeSniffer" name="NonExecutableCodeUnitTest.3.inc" role="test" />
         <file baseinstalldir="PHP/CodeSniffer" name="NonExecutableCodeUnitTest.php" role="test" />
        </dir>
        <dir name="Scope">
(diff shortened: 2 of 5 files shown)


src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
```diff
@@ -258,6 +258,12 @@ public function process(File $phpcsFile, $stackPtr)
                 continue;
             }
 
+            // Skip close PHP tag, followed by whitespace, followed by open PHP tag.
+            if ($tokens[$i]['code'] === T_INLINE_HTML && \trim($tokens[$i]['content']) === '') {
+                $i = $phpcsFile->findNext(T_OPEN_TAG, $i, $end);
+                continue;
+            }
```
jrfnl, line 265: I believe this block of code will not work the way you think it does.

Please add the following code samples to the tests:
```php
<?php if (true): ?>
    <?php foreach ([] as $item): ?>
        <?php continue; ?>
        <div>non-executable</div>
    <?php endforeach; ?>
<?php endif; ?>

<?php if (true): ?>
    <?php foreach ([] as $item): ?>
        <?php continue; ?>

        <div>non-executable</div>

    <?php endforeach; ?>
<?php endif; ?>
```

The first one will still be reported (correctly), the second one will no longer report an issue, while it should.

fredden: Yes, I can see this now. Thanks for highlighting the issue. I'll add these test cases and rework the suggestion to make it report correctly.

fredden: I think I've fixed this now.

src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
```diff
@@ -258,6 +258,16 @@ public function process(File $phpcsFile, $stackPtr)
                 continue;
             }
 
+            // Skip HTML whitespace.
+            if ($tokens[$i]['code'] === T_INLINE_HTML && \trim($tokens[$i]['content']) === '') {
+                continue;
+            }
+
+            // Skip PHP re-open tag (eg, after inline HTML).
+            if ($tokens[$i]['code'] === T_OPEN_TAG) {
```
jrfnl, line 267: What about if the "re-open tag" is a PHP open tag with echo `<?=` ?

Might be good to add some tests for that as well.

jrfnl (changes requested):
Not done a full review yet, but still noticed some things when taking a second look which I believe need to be addressed before this PR can progress.

Just thinking: PHP will automatically add a semi-colon if one is missing before a PHP close tag. Will this fix work correctly with code missing the semi-colon ?

```php
<!-- no problem here -->
<?php if (true): ?>
    <?php foreach ([] as $item): ?>
        <?php continue ?>
    <?php endforeach; ?>
<?php endif; ?>

<!-- should detect an error here -->
<?php if (true): ?>
    <?php foreach ([] as $item): ?>
        <?php continue ?>
        <div>non-executable</div>
    <?php endforeach; ?>
<?php endif; ?>
```

src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
fredden: Yes, the sniff already handles this case correctly. I've added a test to confirm.

jrfnl: I've just tested this, but as far as I can see, the sniff does not handle this correctly, but the test used for this is flawed.

Try this:
```php
<!-- no problem here -->
<?php if (true): ?>
    <?php foreach ([] as $item): ?>
        <?php continue; ?>
        <?= ?>
    <?php endforeach; ?>
<?php endif; ?>

<!-- should detect an error here -->
<?php if (true): ?>
    <?php foreach ([] as $item): ?>
        <?php continue; ?>
        <?=
        'unreachable'
        ?>
    <?php endforeach; ?>
<?php endif; ?>
```

While an empty `<?= ?>` isn't very useful, that's not the concern of the sniff.
If the `<?=` was handled correctly, the first and only error for the above code would be on the line containing the `'unreachable'`. As things are, both lines containing a `<?=` also throw errors.

jrfnl (commented):
@fredden Reviewed again and left some comments.

> While technically out of scope for this pull request, I will investigate adding code to solve this.

Thanks for looking into this. In my opinion, it's not actually out of scope as the situation you are trying to address - code moving in an out of PHP/HTML - is the only time this comes into play.

src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
```diff
@@ -228,6 +228,10 @@ public function process(File $phpcsFile, $stackPtr)
             if ($tokens[$start]['code'] === T_SEMICOLON) {
                 break;
             }
+
+            if ($tokens[$start]['code'] === T_CLOSE_TAG) {
```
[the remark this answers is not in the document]
fredden: I've made this change in 0924523cd.

src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
fredden: Thanks for the feedback. I've looked into both of these suggested test cases.

The sniff will complain about any _line of code_ which is unreachable. (See https://github.com/squizlabs/PHP_CodeSniffer/blob/d148febc2a2eb82972121d7f962883f7a5697b55/src/Standards/Squiz/Tests/PHP/NonExecutableCodeUnitTest.2.inc#L1-L12 and https://github.com/squizlabs/PHP_CodeSniffer/blob/d148febc2a2eb82972121d7f962883f7a5697b55/src/Standards/Squiz/Tests/PHP/NonExecutableCodeUnitTest.php#L89-L92 - one complaint for each of lines 7-10 from the `exit` on line 3, and an additional complaint for line 10 from the `break` on line 9.)

I didn't see an existing test showing the expected behaviour when there are multiple unreachable statements on one line. I've added a test showing the existing behaviour in 34ba861a8.

Looking at these two suggested sniffs:
- `<?= ?>` is a parse error, so I don't think it needs any special treatment in this sniff. https://3v4l.org/MXtbr
- One statement split across multiple lines also isn't covered by existing tests; I've added a test showing the existing behaviour in 2067e7c6e.
- We could add some logic to detect something like `<?= ''; ?>` which isn't a parse error, and will produce no output, but such a line of code seems like something that should be flagged to developers for review, so I'm comfortable with this sniff reporting that as unreachable. I don't think it's feasible for this sniff to detect output of a variable which happens to be empty and alert differently. A human can review the line of code in that case and make a sensible judgement.

@jrfnl do these cover the concerns raised here, or would you like me to add these specific tests as well?

jrfnl: Just to confirm - the tests added in the `2` case file just document existing behaviour. The behaviour regarding these is not changed by this PR. Correct ?

fredden: Yes, the changes in `src/Standards/Squiz/Tests/PHP/NonExecutableCodeUnitTest.2.inc` (and the corresponding additions in `src/Standards/Squiz/Tests/PHP/NonExecutableCodeUnitTest.php`) are documenting existing behaviour. I can move these to a separate pull request if you would prefer.

jrfnl (approved):
@fredden Sorry it took me a while to get back to this PR. I've done some extensive testing with it now and IMO this is good to go. Thank you for your work on this!

jrfnl:

PR #3777, opened 2023-03-15
Squiz/NonExecutableCode: slew of bug fixes, mostly related to modern PHP

### LINKED ISSUE #2857: Squiz.PHP.NonExecutableCode.Unreachable when using ternary operator
Getting Squiz.PHP.NonExecutableCode.Unreachable for following (valid) code:

```php
$var = 'hello';
echo $var == 'foo' ? 'bar' : die( 'world' );

echo 'example';
```

### LINKED ISSUE #3592: PHP 8 throw expressions not supported by `Squiz.PHP.NonExecutableCode` rule
**Describe the bug**
The [throw expression](https://wiki.php.net/rfc/throw_expression) added in PHP 8 is not supported by the `Squiz.PHP.NonExecutableCode` rule.

**Code sample**
```php
<?php declare(strict_types=1);

function doStuff(?string $thing): string {
	$foo = $thing ?? throw new \Exception();

	return $foo;
}
```

**Custom ruleset**
```xml
<?xml version="1.0"?>
  <rule ref="Squiz.PHP.NonExecutableCode" />
</ruleset>
```

**To reproduce**
2. Run `phpcs test.php ...`
```
6 | WARNING | Code after the THROW statement on line 4 cannot be executed
```

**Expected behavior**
Throw expressions should be recognized as expressions, not statements.

 - OS: KDE Neon Linux
 - PHP: 8.1
 - PHPCS: 9.5.20
 - Standard: Squiz

jrfnl:
### Squiz/NonExecutableCode: remove redundant line

`$prev` is already defined in exactly the same way on line 66. No need to re-define it.

### Squiz/NonExecutableCode: only ignore inline statements when PHP allows them

Aside from `T_EXIT`, PHP (prior to PHP 8.0) does not allow for the inline use of the other tokens this sniff registers.
While using those tokens inline would result in a parse error, that is not the concern of this sniff.

By ignoring these, the sniff causes false negatives.

This commit makes the sniff more selective and only ignores inline statements for tokens which are allowed in inline statements.

Includes unit test.

### Squiz/NonExecutableCode: bug fix - allow for all logic operators

The sniff, as it was, only allowed for `or` and `||` logical operators before a termination expression, not for `and`, `&&` or `xor`, while PHP allows these too.

See: https://3v4l.org/OWS4n#veol

Fixed now.

Includes unit tests.

### Squiz/NonExecutableCode: bug fix - expressions in ternary

The `T_EXIT` token can be used in ternaries without affecting code after the ternary expression.

See: https://3v4l.org/oMWuA#veol

Fixed now.

Includes unit test.

Fixes #2857

### Squiz/NonExecutableCode: bug fix - expressions after PHP 7.0/7.4 null coalesce (equals)

PHP 7.0 introduced the null coalesce operator.
PHP 7.4 introduced the null coalesce equals operator.

The `T_EXIT` token can be used in combination with those without affecting the code directly following it.

See: https://3v4l.org/C218d#veol

Fixed now.

Includes unit test.

Related to #2857

### Squiz/NonExecutableCode: bug fix - expressions in PHP 7.4 arrow functions

PHP 7.4 introduced arrow functions. The `T_EXIT` token can be used in those without affecting the code following it.

See: https://3v4l.org/gSrt4#veol

Fixed now.

Includes unit test.

### Squiz/NonExecutableCode: bug fix - PHP 8.0 inline throw expressions

PHP 8.0 introduced the ability to use `throw` as an expression instead of as a statement.
Ref: https://wiki.php.net/rfc/throw_expression

When used as an expression, the `throw` does not necessarily affect the code after it.

See: https://3v4l.org/AmMf2

Fixed now.

Includes unit tests.

Fixes #3592

### DIFF
--- a/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
@@ -68,6 +68,11 @@ class NonExecutableCodeSniff implements Sniff
             if (isset(Tokens::$booleanOperators[$tokens[$prev]['code']]) === true) {
                 return;
             }
+
+            // Expressions are allowed in the `else` clause of ternaries.
+            if ($tokens[$prev]['code'] === T_INLINE_THEN || $tokens[$prev]['code'] === T_INLINE_ELSE) {
+                return;
+            }
         }
 
         // Check if this token is actually part of a one-line IF or ELSE statement.
--- a/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/NonExecutableCodeSniff.php
@@ -22,9 +22,15 @@ class NonExecutableCodeSniff implements Sniff
      * This is in contrast to terminating statements, which cannot be used inline
      * and would result in a parse error (which is not the concern of this sniff).
      *
+     * `throw` can be used as an expression since PHP 8.0.
+     * {@link https://wiki.php.net/rfc/throw_expression}
+     *
      * @var array
      */
-    private $expressionTokens = [T_EXIT => T_EXIT];
+    private $expressionTokens = [
+        T_EXIT  => T_EXIT,
+        T_THROW => T_THROW,
+    ];
 
 
     /**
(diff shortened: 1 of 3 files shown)


jrfnl:
@gsherwood Could this PR please be earmarked for PHPCS 3.8.0 ?

jrfnl:

PR #3785, opened 2023-03-29
Fix PHP Fatal error in `Squiz.Commenting.FunctionComment`

fredden:
Today I have seen an invalid `@param` tag in a docblock, and `Squiz.Commenting.FunctionComment` returned a PHP Fatal error while parsing the file. The error was:

```
There was 1 error:

1) PHP_CodeSniffer\Standards\Squiz\Tests\Commenting\FunctionCommentUnitTest::testSniff
Uninitialized string offset 0

/path/to/squizlabs-php-codesniffer/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php:413
/path/to/squizlabs-php-codesniffer/src/Standards/PEAR/Sniffs/Commenting/FunctionCommentSniff.php:152
/path/to/squizlabs-php-codesniffer/src/Files/File.php:498
/path/to/squizlabs-php-codesniffer/src/Files/LocalFile.php:92
/path/to/squizlabs-php-codesniffer/tests/Standards/AbstractSniffUnitTest.php:173
/path/to/squizlabs-php-codesniffer/tests/TestSuite7.php:28
```

I have tracked this down, added a test case to cover this, and fixed the bug.

**Update 2023-06-08**: by fixing the bug with a different technique (as suggested by @jrfnl on a call - thanks!), the sniff now complains about this case correctly and can auto-fix it.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -406,7 +406,7 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
 
             foreach ($typeNames as $typeName) {
                 // Strip nullable operator.
-                if (strlen($typeName) > 1 && $typeName[0] === '?') {
+                if (isset($typeName[0]) === true && $typeName[0] === '?') {
                     $typeName = substr($typeName, 1);
                 }
 
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -406,7 +406,7 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
 
             foreach ($typeNames as $typeName) {
                 // Strip nullable operator.
-                if ($typeName[0] === '?') {
+                if (strlen($typeName) > 1 && $typeName[0] === '?') {
                     $typeName = substr($typeName, 1);
                 }
 
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -405,8 +405,12 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
             $suggestedTypeNames = [];
 
             foreach ($typeNames as $typeName) {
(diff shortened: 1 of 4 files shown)


src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -406,7 +406,7 @@ protected function processParams(File $phpcsFile, $stackPtr, $commentStart)
 
             foreach ($typeNames as $typeName) {
                 // Strip nullable operator.
-                if ($typeName[0] === '?') {
+                if (strlen($typeName) > 1 && $typeName[0] === '?') {
```
jrfnl, line 409: Would it make sense to simplify this to the below ?
```suggestion
                if (isset($typeName[0]) === true && $typeName[0] === '?') {
```

Note: the `=== true` is unnecessary, but a required coding convention in this repo.

fredden: Both options will work. I don't see either as "more simple." I find `strlen()` to be a little more clear, but am happy to swap this for `isset()` for its better performance.

jrfnl: Well, the string length is irrelevant if only looking for the first byte and yes, performance. Either way, your PR.

fredden: I'm happy to receive feedback. Please don't be discouraged from providing useful insight as you have been. :+1:

jrfnl: No worries, I didn't take it like that.

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -406,7 +406,7 @@ protected function processParams(File $phpcsFile, $stackPtr, $commentStart)
 
             foreach ($typeNames as $typeName) {
                 // Strip nullable operator.
```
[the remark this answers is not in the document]
fredden: Thanks for this suggestion. I've updated the code to use this technique instead, and the sniff now detects the problem and offers to auto-fix it too. This is a nice outcome.

jrfnl (approved):
Reviewed & tested. All good ✅

jrfnl:

PR #3795, opened 2023-04-06
Add new sniff to ban the use of compact() function

pbogut:
Adds new generic sniffer for banning usage of `compact()` function. 

Using this function often is not playing nice with static analysers and LSPs and makes refactoring harder.

I've added sniffer to detect if `compact()` is used to create array. If function call contains only strings that could be valid variable names then it can be replace automatically by array use `compact('a')` will be fixed as `['a' => $a]`.

Not sure if this should be added to `package.xml`? And I'm also not sure if `Arrays` is good namespace for it. It is function after all, but it is just used to build array. I'm happy to change whatever is needed.


src/Standards/Generic/Tests/Arrays/DisallowCompactArrayBuilderUnitTest.1.inc
```diff
@@ -0,0 +1,33 @@
...
+
+view('some.view', compact("a", 'b', 'c'));
+view('some.view', compact(
+    'a',
+    'b',
+    'c'
+));
+
+$var = compact('aa', 'invalid-var.name');
+$var = Bazz::compact('a', 'b');
```
jrfnl, line 33: I see quite some code in this test case file which doesn't seem to be testing anything in the sniff... ?

Here are some more test cases to take into account:
```php
compact(...$names);
compact( 'prefix' . $name, '$name' . 'suffix', "some$name");
compact(...get_names('category1', 'category2'));

// Last test in the file.
// Live coding/parse error.
compact( 'a', 'b'
```

See this as an invitation to get a little more creative with the tests and to make sure the sniff handles those situations correctly.

src/Standards/Generic/Tests/Arrays/DisallowCompactArrayBuilderUnitTest.2.inc
```diff
@@ -0,0 +1,22 @@
+<?php
+$var = compact('s', 'b');
+
+// The following function has a simulated git conflict for testing.
+// This is not a merge conflict - it is a valid test case to test handling of
+// compact function call without associated closer.
+// Please do not remove.
+function test()
```
jrfnl, line 8: I'm not sure why this test is included ?

src/Standards/Generic/Sniffs/Arrays/DisallowCompactArrayBuilderSniff.php
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+
+        $content = $tokens[$stackPtr]['content'];
+
+        if ($content !== 'compact') {
```
jrfnl, line 48: I suggest adding tests with the below (function names in PHP are case-INsensitive):
```php
COMPACT('a');
Compact('a');
```

jrfnl (changes requested):
I've done a cursory review of the sniff without an opinion on whether it should be included in PHPCS.

I would note that I'm not sure it's a good idea to include a fixer in this sniff unless you are certain that very possible code situation has been covered. As things are, I would consider this a pretty risky fixer.

> Not sure if this should be added to `package.xml`?

Yes, all files need to be listed in `package.xml`.

> And I'm also not sure if `Arrays` is good namespace for it.

See my suggestion inline.

pbogut:
Thank you for review, will add more tests for cases you listed and check code suggestions. As for the fixer, it is checking if there is:

1. Open parenthesis after function name
2. Close parenthesis after open one
3. Anything in between is ONLY string (that should be [valid PHP variable name](https://www.php.net/manual/en/language.variables.basics.php)), empty token (white spaces, comments) or comma `,`.

If any of the above is not true then error would not be fixable.

src/Standards/Generic/Tests/Arrays/DisallowCompactArrayBuilderUnitTest.2.inc
pbogut: Fair enough, will remove. (I've copied some testes from different sniff as I started to work on it).

src/Standards/Generic/Tests/Arrays/DisallowCompactArrayBuilderUnitTest.1.inc
pbogut: Most of these were already covered as sniff requires only strings and commas within parenthesis. I've added them to test case. Last one was actually causing error so I fixed that as well.

src/Standards/Generic/Sniffs/Arrays/DisallowCompactArrayBuilderSniff.php
```diff
@@ -0,0 +1,131 @@
...
+            return;
+        }
+
+        $error = 'Array must not be created with compact() function';
+
+        $fix      = true;
+        $toExpand = [];
+        $openPtr  = $next;
+        $closePtr = null;
+        $validVarNameRegExp = '/^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$/';
```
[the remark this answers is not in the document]
pbogut: I did that, thanks for suggestion.

src/Standards/Generic/Sniffs/Arrays/DisallowCompactArrayBuilderSniff.php
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        $prev = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($stackPtr - 1), null, true);
+
+        // Make sure its not method of an object or class.
+        if ($tokens[$prev]['code'] === T_OBJECT_OPERATOR || $tokens[$prev]['code'] === T_DOUBLE_COLON) {
+            // Not a builtin function call.
+            return;
+        }
+
```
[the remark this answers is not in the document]
pbogut: I've added them to the test case, they should not produce sniff errors. 
I've also added `\compact()` as this is valid function call and should produce error.

src/Standards/Generic/Sniffs/Arrays/DisallowCompactArrayBuilderSniff.php
```diff
@@ -0,0 +1,131 @@
...
+        $next = $phpcsFile->findNext(Tokens::$emptyTokens, ($stackPtr + 1), null, true);
+        if ($next === false || $tokens[$next]['code'] !== T_OPEN_PARENTHESIS) {
+            // Not a function call.
+            return;
+        }
+
+        $prev = $phpcsFile->findPrevious(Tokens::$emptyTokens, ($stackPtr - 1), null, true);
+
+        // Make sure its not method of an object or class.
+        if ($tokens[$prev]['code'] === T_OBJECT_OPERATOR || $tokens[$prev]['code'] === T_DOUBLE_COLON) {
```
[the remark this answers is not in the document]
pbogut: Yes, thanks. I've added that, plus few other things. 
I'm now questioning if I should look back for things that indicates it's not actually builtin function call, or maybe should change approach and list things that indicates it is function call.

So what I mean by that is if its better to check (as it is now) :

```
 if ($tokens[$prev]['code'] === T_OBJECT_OPERATOR ... ) {
     return;
 }
 ````
 Or maybe would be better something like:
 ```
 if ($tokens[$prev]['code'] !== T_EQUAL ...) {
     return;
 }
 ```
 It might be more future proof if changed, but not sure if I wont end up with long list of allowed tokens.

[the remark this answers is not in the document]
jrfnl: In my experience the current approach of bowing out of specific tokens is the correct one.
Changing the logic around would make the sniff much less stable as new tokens get added regularly (either via PHP itself or in PHPCS to account for changes in PHP) and this sniff would need to be updated each time that happens.

jrfnl:
> As for the fixer, it is checking if there is:
> 
>     1. Open parenthesis after function name
>     2. Close parenthesis after open one
>     3. Anything in between is ONLY string (that should be [valid PHP variable name](https://www.php.net/manual/en/language.variables.basics.php)), empty token (white spaces, comments) or comma `,`.
> 
> If any of the above is not true then error would not be fixable.

So that means that only plain strings would be taken into account, while per the documentation of `compact()`, the function allows for receiving arrays as well ?

> compact() takes a variable number of parameters. Each parameter can be either a string containing the name of the variable, or an array of variable names. The array can contain other arrays of variable names inside it; compact() handles it recursively. 

Source: https://www.php.net/manual/en/function.compact.php#refsect1-function.compact-parameters

pbogut:
> So that means that only plain strings would be taken into account, while per the documentation of `compact()`, the function allows for receiving arrays as well ?

For fixing it yes. It will still be reported if its compact containing an array, but it would not be fixed automatically.

My reasoning was that simpler fixer the better, and most cases in the wild I've seen were just using simple string arguments.  Otherwise would have to check for arrays (short and long ones) and array values (that may have keys specified). And also nested arrays. For example this: `compact(['c' => array(['b' => 'a'])])` is valid compact  call and would produce `['a' => $a]` array.

Do you think its wrong way of looking at it?

jrfnl:
> Do you think its wrong way of looking at it?

No, I concur that is probably the sensible way of doing it to prevent the fixer from getting overly complicated/making incorrect fixes.

PR #3797, opened 2023-04-13
Tokenizer/PHP: more context sensitive keyword fixes

jrfnl:
PHPCS re-tokenizes the `self`, `parent`, `true`, `false` and `null` keywords to a PHPCS native token.

This re-tokenization did not take the following situations into account:
* Those keywords being used as function names when the function is declared to return by reference.
* Those keywords being used as a function call.

Additionally, the PHP native `T_STATIC` token would not be (re-)tokenized to `T_STRING` when used as a function call, though it was tokenized correctly when used as a method call. While using the `static` keyword for a global function declaration is illegal in PHP, the tokenization in PHPCS should still be consistent.

This commit fixes those issues.

These issues were discovered while investigating issue PHPCompatibility/PHPCompatibility#1489

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -682,6 +682,36 @@ class PHP extends Tokenizer
             }//end if
 
             /*
+                Special case for `static` used as a function name, i.e. `static()`.
+            */
+
+            if ($tokenIsArray === true
+                && $token[0] === T_STATIC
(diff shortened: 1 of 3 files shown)


jrfnl:
Note: I've updated the commit to contain yet more unit tests.

jrfnl:

PR #3801, opened 2023-04-15
File::getMethodParameters(): allow for readonly promoted properties without visibility

jrfnl:
Follow up on PR #3516 which was included in PHPCS 3.7.0.

Turns out that constructor property promotion also allows for declaring properties with the `readonly` keyword, but without explicit visibility set. See: https://3v4l.org/nli62

Readonly properties without explicit visibility are already handled correctly in the `File::getMemberProperties()` method, but were not handled correctly in the `File::getMethodParameters()` method.

Fixed now.

Includes updated documentation and a unit test.


jrfnl:

PR #3803, opened 2023-04-18
Tests: allow the test suite to run on PHPUnit 8.x and 9.x

### LINKED ISSUE #3490: Incompatibility with PhpUnit 9.5.10
**Describe the bug**
``PHP Fatal error:  Declaration of PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in /home/user/laminas-ldap/vendor/squizlabs/php_codesniffer/tests/Standards/AbstractSniffUnitTest.php on line 55``

**Custom ruleset**
https://github.com/laminas/laminas-coding-standard/blob/2.3.0/src/LaminasCodingStandard/ruleset.xml

**To reproduce**
1. `php7.4 vendor/bin/phpcs --standard=LaminasCodingStandard --colors -p --extensions=php .`

PHPCS output here
``PHP Fatal error:  Declaration of PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in /home/user/laminas-ldap/vendor/squizlabs/php_codesniffer/tests/Standards/AbstractSniffUnitTest.php on line 55``

 - OS: Ubuntu 21.10
 - PHP: 7.4, 8.0, 8.1
 - PHPCS: 3.6.1
 - Standard: LaminasCodingStandard

**Additional context**
```
phpunit/php-code-coverage                      9.2.9   Library that provides collection, processing, and rendering functionality for PHP code coverage information.
phpunit/php-file-iterator                      3.0.5   FilterIterator implementation that filters files based on a list of suffixes.
phpunit/php-invoker                            3.1.1   Invoke callables with a timeout
phpunit/php-text-template                      2.0.4   Simple template engine.
phpunit/php-timer                              5.0.3   Utility class for timing
phpunit/phpunit                                9.5.10  The PHP Unit Testing framework.
squizlabs/php_codesniffer                      3.6.1   PHP_CodeSniffer tokenizes PHP, JavaScript and CSS files and detects violations of a defined set of coding standards.
slevomat/coding-standard                       7.0.16  Slevomat Coding Standard for PHP_CodeSniffer complements Consistence Coding Standard by providing sniffs with additional checks.
laminas/laminas-coding-standard                2.3.0   Laminas Coding Standard
```

jrfnl:
### GetMethodParametersTest: add some missing array indexes expectations

This commit:
1. Fixes the order of a few array entries.
2. Adds some array entries which were supposed to be expected, but missing.

The use of `assertArraySubset()` hid the fact that these entries were missing. As that assertion now needs to be replaced, these issues came to light.

### GetMemberPropertiesTest: add some missing array indexes expectations

This commit adds some array entries which were supposed to be expected, but missing.

The use of `assertArraySubset()` hid the fact that these entries were missing. As that assertion now needs to be replaced, these issues came to light.

### File/Get*Tests: work round removal of assertArraySubset()

The `assertArraySubset()` method was deprecated in PHPUnit 8.x and removed in PHPUnit 9.0.0 without replacement.

The `assertArraySubset()` assertion was being used as not all token array indexes are being tested - to be specific: any index which is a token offset is not tested -.
As the `assertArraySubset()` has been removed, I'm electing to unset the token offset array indexes and replacing the assertion with a strict type `assertSame()` comparison.

### ~~GetMemberPropertiesTest~~ ++File/Get*Tests:++: work round removal of exception related annotations

Expecting exceptions via annotations was deprecated in PHPUnit 8.x and removed in PHPUnit 9.0.0 in favour of using the `expectException*()` methods.

This does need a work around for PHPUnit 4.x in which the `expectException*()` methods didn't exist yet, but as this only applies to ~~two~~ ++three++ tests, that's not a big deal.

### GotoLabelTest: work round removal of assertInternalType()

The `assertInternalType()` method was deprecated in PHPUnit 7.5.0 and removed in PHPUnit 9.0.0.
PHPUnit 7.5.0 introduced dedicated `assertIs*()` (like `assertIsInt()`) methods as a replacement.

As this is only a simple check in these two tests, a PHPUnit feature based toggle seems over the top, so I'm just replacing the assertion with an alternative which will work PHPUnit cross-version.

### RuleInclusion*Test: remove a few redundant assertions

These assertions are checking whether explicitly declared properties exist, which is redundant.

Removing the assertions does not diminish the value of the tests as there are follow-up assertions testing the value of the properties.

Removing the assertions also gets rid of a warning thrown in PHPUnit 9.6.x about the `assertObjectHasAttribute()` assertion being removed in PHPUnit 10.0.
Note: PHPUnit 10.1.0 adds these assertions back again, but under a different name `assertObjectHasProperty()`.

### ~~RuleInclusionTest~~ ++Ruleset Tests++: work round removal of assertObjectHasAttribute()

The `assertObjectHasAttribute()` method was deprecated in PHPUnit 9.6.x and removed in PHPUnit 10.0.0 without replacement.
Note: PHPUnit 10.1.0 adds the assertion back again, but under a different name `assertObjectHasProperty()`.

While only a deprecation warning is shown on PHPUnit 9.6.x and the tests will still pass, I'm electing to replace the assertion anyway with code which emulates what PHPUnit would assert.

### Tests: rename fixture methods and use annotations

As of PHPUnit 8.x, the method signature for the `setUpBeforeClass()`, `setUp()`, `tearDown()` and `tearDownAfterClass()` fixture methods has changed to require the `void` return type.

As the `void` return type isn't available until PHP 7.1, this cannot be implemented.

Annotations to the rescue.
By renaming the `setUpBeforeClass()`, `setUp()`, `tearDown()` and `tearDownAfterClass()` methods to another, descriptive name and using the `@beforeClass`, `@before`, `@after` and `@afterClass` annotations, the tests can be made cross-version compatible up to PHPUnit 9.x.

With this change, the unit tests can now be run on PHPUnit 4 - 9.

This constitutes a minor BC-break for external standards which a) extend the PHPCS native testsuite and b) overload the `AbstractSniffUnitTest::setUp()` method.
While quite a few external standards extends the PHPCS native testsuite, I very much doubt any of these overload the `setUp()` method, so IMO and taking into account that this is a test-only change, this is an acceptable change to include in the next PHPCS minor.

Ref: https://docs.phpunit.de/en/7.5/annotations.html#before

### Tests: allow the test suite to run on PHPUnit 8.x and 9.x

Includes:
* `composer.json`: widening the PHPUnit requirement to allow for PHPUnit 8.x and PHPUnit 9.x.
    Note: The recently released PHPUnit 10.x is not (yet) supported as it no longer supports the old-style test suite runner setup.
    It also would require for the abstract base test cases to be renamed as those classes are no longer allowed to end on `Test`.
    Refactoring the test suite to allow for PHPUnit 10.x is for a future PR.
* Adjusting the PHPUnit configuration to ensure the tests are run in the same way and show all notices/warnings/deprecations on all PHPUnit versions.
    The default value for a number of configuration options has changed over time.
    This makes sure they are consistently set to values which are sensible for this codebase, independently of the PHPUnit version on which the tests are run.
    Includes adding a schema annotation (set to PHPUnit 9.2 as the schema has changed in PHPUnit 9.3, though that won't prevent the tests from running correctly).
* GH Actions `test` workflow: removing work-arounds which were in place related to running PHPUnit 7.x on PHP 8.x.
* `AllTests`: Adjusting the condition which determines which `TestSuite` file to load to allow for PHPUnit 8.x and 9.x.
* Adding the `.phpunit.result.cache` file to `.gitignore`.
    PHPUnit has a caching feature build in as of PHPUnit 8, so ignore the file that generates to prevent it from being committed.

Related to (but doesn't fix) #3395

Fixes #3490


jrfnl:
@gsherwood Note: this PR should not be ported to the 4.x branch as-is. Only select commits should be included in 4.x. If so preferred, I can set up a separate PR with the cherrypicked bits and pieces which should go into PHPCS 4.x.

jrfnl:
Rebased without changes to fix merge conflict after merge of #3708.

jrfnl:
I've rebased this PR and updated it to incorporate the changes needed in the tests introduced by today's merges. Updated the PR description to match.

Note: there will now still be one warning on PHPUnit 9.x (which doesn't fail the build) due to the deprecation expectation. This is not something to worry about until PHPUnit 10.x would need to be supported (which is not needed yet as PHPUnit 9.x will run fine on PHP 8.2 and even 8.3).

jrfnl:
Note: depending on which PR gets merged first, this PR will need to update the `CONTRIBUTING.md` file and the Composer scripts which PR #3830 add.

jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/59

PR #3807, opened 2023-04-19
Make InnerFunctionsSniff detect functions inside closures

### LINKED ISSUE #3806: InnerFunctionsSniff should flag functions defined inside closures
InnerFunctionsSniff detects functions defined inside other functions, but not inside closures, meaning it won't report any violations for the following snippet:

```php
$myFunc = function () {
	function test(): int {
		return 54;
	}
};
```

For consistency, this should be treated the same as:

```php
function myFunc() {
	function test(): int {
		return 54;
	}
}
```

Daimona:
Fixes #3806

### DIFF
--- a/src/Standards/Squiz/Sniffs/PHP/InnerFunctionsSniff.php
+++ b/src/Standards/Squiz/Sniffs/PHP/InnerFunctionsSniff.php
@@ -11,6 +11,7 @@ namespace PHP_CodeSniffer\Standards\Squiz\Sniffs\PHP;
 
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
 
 class InnerFunctionsSniff implements Sniff
 {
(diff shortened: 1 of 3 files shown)


jrfnl (approved):
LGTM

jrfnl:
I agree this is a good change, but looking at this sniff now, I see more things wrong with it.

@Daimona Would you be willing to make some additional fixes ?

Additional things which I believe should be addressed:
1. There is redundant code (the check for `=`). This is due to in the older day closures being tokenized as `T_CLOSURE`. That code will never be matched anymore now.
2. The part where it checks for methods in anonymous classes should probably be expanded to check for all OO structures.
    See: https://3v4l.org/c2Cbn
3. As more checking of the `conditions` is now needed, the sniff should probably check the `conditions` array itself instead of using `getCondition()`.

Does that make sense ?

Daimona:
> 1. There is redundant code (the check for `=`). This is due to in the older day closures being tokenized as `T_FUNCTION`. That code will never be matched anymore now.

Makes sense to me! In fact, I was a bit puzzled by that check given the existence of `T_CLOSURE` (that's not listed in `register()`).

I agree with the rest as well. I've updated my PR.

jrfnl (commented):
Thanks for making these updates @Daimona !

Looking good.

I wonder if the two loops walking the `$conditions` array can be combined ?

I'm thinking that logic like this may work (very much pseudo-code):
```php
// Reverse the conditions array
foreach () {
    if ($condition === T_FUNCTION || $condition === T_CLOSURE) {
        break; // Error.
    }

    if (isset(Tokens::$ooScopeToken[$condition])) {
        return; // Found OO structure before function/closure, so this is an OO method.
    }
}
```

jrfnl:
Extra code sample you may want to add to the tests (would currently not be detected/false negative):
```php
function foo() {
    if (class_exists('Bar') === false) {
        class Bar {
            function foo() {
                function innerFunction() {} // Error.
            }
        }
    }
}
```

Daimona:
> I wonder if the two loops walking the `$conditions` array can be combined ?

I wanted to use separate loops so that we can bail out earlier if it's not a nested function. However, on second though that's simply untrue -- we'd loop through the whole array of conditions in the common case without nested functions. And I guess reversing the array of conditions is not terribly expensive, either, so I'm going to merge the loops, and make all the other suggested changes.

src/Standards/Squiz/Sniffs/PHP/InnerFunctionsSniff.php
```diff
@@ -45,39 +46,26 @@ public function process(File $phpcsFile, $stackPtr)
             return;
         }
 
-        $conditions = $tokens[$stackPtr]['conditions'];
+        $reversedConditions = array_reverse($tokens[$stackPtr]['conditions'], true);
```
jrfnl: Performance hats-up: for nested arrays, in most cases performance of array functions is better when you first slice off the bit you want to work with.

So:
``` php
$conditions = $tokens[$stackPtr]['conditions'];
$reversedConditions = array_reverse($conditions, true);
```
... should give a little better performance.

Daimona: Interesting, never heard this one before. Updated just now.

jrfnl (approved):
LGTM 🎉 

Thanks for working with me to get this sniff into shape!

src/Standards/Squiz/Sniffs/PHP/InnerFunctionsSniff.php
jrfnl: Have a look at this PR where some similar replacements in a much used sniff, lowered the memory usage of PHPCS for a complete run over a codebase by > 6%: https://github.com/squizlabs/PHP_CodeSniffer/pull/2273

jrfnl:
@Daimona I'm going to merge this PR. The changelog normally contains the full name of the contributor. Would you mind sharing yours ?

jrfnl:
> Hi :) I don't use my real name online; at least not in connection with this username. Would it be possible to just credit me as "Daimona"?

Absolutely! Just wanted to give you the chance, but only if you want it.

jrfnl:

PR #3810, opened 2023-04-24
New `Performance` report to measure sniff run time performance

### LINKED ISSUE #3784: Possible improvements for sniff performance analysis
As a sniff author, it is not uncommon to receive reports about sniffs/standards being "slow".

To analyze which sniffs could be the culprit/which sniffs to focus on to improve performance, the `-vvv` command can be used, which at the very end of the debug logs shows a "SNIFF PROCESSING REPORT" with the sniffs which have been triggered during a run and the time each sniff took, ordered from slowest to fastest sniff.

This information is super-useful, however to get a good impression of the performance of various sniffs, the best way to do so is to run the sniffs/standard against a large number of files, which makes the `-vvv` output completely unworkable as even for a simple run against 4 files, the output is easily over 60.000 lines, with a test run against 300 files, the output is well over 4 million lines...

The current available verbosity levels and their output are defined as such:
1. `-v`: file specific token and line counts with running times
2. -`vv`: the info from level 1 + verbose tokeniser output
3. `-vvv`: the info from level 1 + 2 + token processing output with sniff running times

Now there are a couple of ways I can think of to improve this situation and I'd like to hear if any of these could/would be acceptable.

1. Add an extra debugging level between the current `1` (`-v`) and `2` (`-vv`) levels, which would contain the information from level 1 + the sniff performance report.
    This would mean that `-vv` would now show the performance report, while to get the "old" `-vv` report (+ performance info), one would now need to use `-vvv`, and the "old" `-vvv` would become `-vvvv`.
    While this would be a BC-break, this type of verbosity is generally only ever used by sniff authors/PHPCS tokenizer fix contributors. So considering the very limited public this BC-break would hit, I hope this could be considered acceptable as an enhancement, which could go into a PHPCS 3.x minor.
2. Add a separate CLI option - outside of the verbosity settings - to allow for the performance information to be printed.
    Maybe using the `-r` for "(sniff) run time information" ?
    This would avoid the BC-break and would still make the information available in a more workable manner.
3. Create a `Performance` `Report` type, possibly externally hosted, like in [PHPCSDevTools](https://github.com/PHPCSStandards/PHPCSDevTools).
    To allow for this, some changes would be needed to the `File::process()` method as it currently only records "time taken by each sniff" to the `$this->listenerTimes` property if `PHP_CODESNIFFER_VERBOSITY > 2`, i.e. when the `-vvv` option is enabled, which would run counter to the problem I'm trying to solve.
    It might also make life easier if, similar to the `File::getMetrics()` method, there would also be a `File::getListenerTimes()` method.
    It would still need some work-arounds as `$this->listenerTimes` is cumulative and only the `Report::generateFileReport()` method has access to the `File` object, while the report would need to be generated in the `generate()` method which doesn't have access to the `File` object, but that is doable.

In my opinion, option 1 or 2 would be the preferred options as it would also allow standard maintainers to ask performance issue reporters to run the report, even when they can't share access to the (private) code base on which the performance issue was detected.

Having said that, option 3 would still be an improvement over the current situation.

Opinions ?

/cc @gsherwood @kukulich 

P.S.: I'd, of course, be happy to make the changes, but would like some direction before spending significant time on this.

jrfnl:
~~**_This PR depends on PR #3804 to allow for automated testing of the new report._**~~
~~Either PR #3804 should be merged first and this PR rebased to remove that commit after the merge or the EndToEnd test included in this PR should be removed from the last commit and (after this PR has been merged), the test can be added to PR #3804.~~

~~☝🏻 The above is the only reason to open this PR in draft. The actual change in the PR is ready for review.~~

---

### Util\Timing: split `printRunTime()` method

... into logical parts to allow the Performance report access to the total run time information (at the point of report creation) as well.

### File: add `getListenerTimes()` method

... to allow for access to the recorded listener times from within a report class.

### Config: add `trackTime` toggle for whether or not to track listener times

As recording the time taken by each sniff has a performance impact in and of itself on a CS run, only record the time taken by each sniff when needed. [*]

Originally, this was already done conditionally based on the `PHP_CODESNIFFER_VERBOSITY > 2` condition.
However, adding the Performance report would add a second criteria.

This commit adds a new (internal) Config setting `trackTime`, which will be set to `true` when `PHP_CODESNIFFER_VERBOSITY > 2`.

This commit paves the way for adding the second criteria in the next commit.

> [*] I've done some unscientific benchmarks for this by running PHPCS multiple times, with and without tracking the listener times, over a 300+ file codebase.
Without tracking listener times, the run time was always around 39 seconds with 56Mb memory use.
With tracking listener times, the run time was always around 54 seconds with 64Mb memory use.
>
> This, to me, shows a significant enough difference and sufficient reason to put this toggle in place to only track time when needed.

### New `Performance` report to measure sniff run time performance

The report will print the sniff name, cumulative listener run time and % of the total sniff run time for each sniff triggered during a run.

The report is ordered by cumulative listener run time in descending order.

Additionally, it will highlight sniffs which have a cumulative listener run time more than twice the average run time per sniff in orange and sniffs with a cumulative listener run time of more than three times the average run time per sniff in red.

At the bottom of the report it will also compare the total sniff relative run time with the total run time.

Fixes #3784

Includes end-to-end test (depends on PR #3804).
Includes mention of the report in the CLI help text.

The report output is as shown in the screenshots in https://github.com/squizlabs/PHP_CodeSniffer/issues/3784#issuecomment-1520203761

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -328,6 +328,7 @@ http://pear.php.net/dtd/package-2.0.xsd">
      <file baseinstalldir="PHP/CodeSniffer" name="Json.php" role="php" />
      <file baseinstalldir="PHP/CodeSniffer" name="Junit.php" role="php" />
      <file baseinstalldir="PHP/CodeSniffer" name="Notifysend.php" role="php" />
+     <file baseinstalldir="PHP/CodeSniffer" name="Performance.php" role="php" />
      <file baseinstalldir="PHP/CodeSniffer" name="Report.php" role="php" />
      <file baseinstalldir="PHP/CodeSniffer" name="Source.php" role="php" />
      <file baseinstalldir="PHP/CodeSniffer" name="Summary.php" role="php" />
--- a/src/Config.php
+++ b/src/Config.php
@@ -272,6 +272,13 @@ class Config
                 $this->settings['trackTime'] = true;
             }
             break;
+        case 'reports':
+            $reports = array_change_key_case($value, CASE_LOWER);
+            if (array_key_exists('performance', $reports) === true) {
+                $this->settings['trackTime'] = true;
+            }
+            break;
+
         default :
             // No validation required.
             break;
@@ -1420,8 +1427,8 @@ class Config
         echo ' <processes>    How many files should be checked simultaneously (default is 1)'.PHP_EOL;
         echo ' <report>       Print either the "full", "xml", "checkstyle", "csv"'.PHP_EOL;
         echo '                "json", "junit", "emacs", "source", "summary", "diff"'.PHP_EOL;
-        echo '                "svnblame", "gitblame", "hgblame" or "notifysend" report,'.PHP_EOL;
-        echo '                or specify the path to a custom report class'.PHP_EOL;
+        echo '                "svnblame", "gitblame", "hgblame", "notifysend" or "performance",'.PHP_EOL;
+        echo '                report or specify the path to a custom report class'.PHP_EOL;
         echo '                (the "full" report is printed by default)'.PHP_EOL;
         echo ' <reportFile>   Write the report to the specified file path'.PHP_EOL;
         echo ' <reportWidth>  How many columns wide screen reports should be printed'.PHP_EOL;
--- a/src/Config.php
+++ b/src/Config.php
@@ -143,6 +143,7 @@ class Config
         'stdin'           => null,
         'stdinContent'    => null,
         'stdinPath'       => null,
+        'trackTime'       => null,
         'unknown'         => null,
     ];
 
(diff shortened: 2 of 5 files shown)


jrfnl:
As per [my comment](https://github.com/squizlabs/PHP_CodeSniffer/pull/3804#issuecomment-1555459666) in PR #3804, I've removed the End2End test for the report from this PR to unblock it and am marking the PR as ready for review

jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/60

PR #3813, opened 2023-04-28
Handle `@param` in docblock for variables passed by reference

fredden:
This started as an auto-fix to remove ampersands from `@param` tags in docblocks within `Squiz.Commenting.FunctionComment`. This has evolved into "handle `@param` tags for variables passed by reference." We are also increasing test coverage of related-but-not-changed behaviour.

### DIFF
--- a/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
@@ -555,16 +555,38 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
 
             // Make sure the param name is correct.
             if (isset($realParams[$pos]) === true) {
-                $realName = $realParams[$pos]['name'];
-                if ($realName !== $param['var']) {
+                $realName     = $realParams[$pos]['name'];
+                $paramVarName = $param['var'];
+
+                if ($param['var'][0] === '&') {
+                    // Even when passed by reference, the variable name in $realParams does not have
+                    // a leading '&'. This sniff will accept both '&$var' and '$var' in these cases.
+                    $paramVarName = substr($param['var'], 1);
+
+                    // This makes sure that the 'MissingParamTag' check won't throw a false positive.
+                    $foundParams[(count($foundParams) - 1)] = $paramVarName;
+
+                    if ($realParams[$pos]['pass_by_reference'] !== true && $realName === $paramVarName) {
+                        // Don't complain about this unless the param name is otherwise correct.
+                        $error = 'Doc comment for parameter %s is prefixed with "&" but parameter is not passed by reference';
+                        $code  = 'ParamNameUnexpectedAmpersandPrefix';
+                        $data  = [$paramVarName];
+
+                        // We're not offering an auto-fix here because we can't tell if the docblock
+                        // is wrong, or the parameter should be passed by reference.
+                        $phpcsFile->addError($error, $param['tag'], $code, $data);
+                    }
+                }
+
+                if ($realName !== $paramVarName) {
                     $code = 'ParamNameNoMatch';
                     $data = [
-                        $param['var'],
+                        $paramVarName,
                         $realName,
                     ];
 
                     $error = 'Doc comment for parameter %s does not match ';
-                    if (strtolower($param['var']) === strtolower($realName)) {
+                    if (strtolower($paramVarName) === strtolower($realName)) {
                         $error .= 'case of ';
                         $code   = 'ParamNameNoCaseMatch';
                     }
@@ -572,7 +594,7 @@ class FunctionCommentSniff extends PEARFunctionCommentSniff
                     $error .= 'actual variable name %s';
 
                     $phpcsFile->addError($error, $param['tag'], $code, $data);
-                }
+                }//end if
             } else if (substr($param['var'], -4) !== ',...') {
                 // We must have an extra parameter comment.
                 $error = 'Superfluous parameter comment';
--- a/src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
+++ b/src/Standards/Squiz/Tests/Commenting/FunctionCommentUnitTest.inc
@@ -1054,3 +1054,78 @@ function throwCommentOneLine() {}
  * @return void
  */
 function doublePipeFatalError(?stdClass $object) {}
+
+/**
+ * Test for passing variables by reference
+ *
+ * This sniff treats the '&' as optional for parameters passed by reference, but
+ * forbidden for parameters which are not passed by reference.
+ *
+ * Because mismatches may be in either direction, we cannot auto-fix these.
+ *
+ * @param string $foo    A string passed in by reference.
+ * @param string &$bar   A string passed in by reference.
+ * @param string $baz    A string NOT passed in by reference.
+ * @param string &$qux   A string NOT passed in by reference.
+ * @param string &$case1 A string passed in by reference with a case mismatch.
+ * @param string &$CASE2 A string NOT passed in by reference, also with a case mismatch.
+ *
+ * @return void
+ */
+public function variablesPassedByReference(&$foo, &$bar, $baz, $qux, &$CASE1, $case2)
+{
+    return;
+}
+
+/**
+ * Test for param tag containing ref, but param in declaration not being by ref.
+ *
+ * @param string &$foo This should be flagged as (only) ParamNameUnexpectedAmpersandPrefix.
+ * @param string &$bar This should be flagged as (only) ParamNameNoMatch.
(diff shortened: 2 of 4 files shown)


jrfnl (commented):
I'm not sure what would be the right approach here.

Fixing it like you do is one way, but is the `&` there actually _wrong_ ?

An alternative way of looking at it would be to `ltrim()` a potential `&` from a param name before doing the name compare.

Not sure what is best. Maybe PSR19 can give some guidance ?

fredden:
[PSR-19](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#510-param) makes no mention of variables passed by reference. Based on this, I'll update this pull request to instead ignore this character and not auto-fix.

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -552,15 +552,34 @@ protected function processParams(File $phpcsFile, $stackPtr, $commentStart)
...
-                if ($realName !== $param['var']) {
+
+                if ($realParams[$pos]['pass_by_reference'] === true && $param['var'][0] === '&') {
+                    // When passed by reference, the variable name in $realParams does not
+                    // have a leading '&'. There is no guidance in PSR-19 to say if the
+                    // docblock SHOULD or SHOULD NOT also have a leading '&'. Therefore,
+                    // this sniff remains agnostic in this case.
+                    $paramVarName = substr($param['var'], 1);
+
+                    // This fixes a false-positive with 'MissingParamTag'.
```
jrfnl, line 563:
```suggestion
                    // This makes sure that the 'MissingParamTag' check won't throw a false positive.
```

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -552,15 +552,34 @@ protected function processParams(File $phpcsFile, $stackPtr, $commentStart)
...
+                    // either the docblock or the function / method signature needs to be
+                    // changed. We can't auto-fix this as it's not clear which of these
+                    // two is wrong.
+                    $paramVarName = $param['var'];
+                }
+
+                if ($realName !== $paramVarName) {
                     $code = 'ParamNameNoMatch';
                     $data = [
                         $param['var'],
```
jrfnl, line 577:
```suggestion
                        $paramVarName,
```

Extra test to see the issue & safeguard the fix:
```php
 * @param string &$case A string passed in by reference with a case mismatch.
...
public function variablesPassedByReference(&$foo, &$bar, $baz, $qux, &$CASE)
```

jrfnl (changes requested):
@fredden Thanks for updating this PR! Reviewed in detail now.

While this is not a PSR-19 sniff (but a Squiz one), I agree that taking guidance from PSR-19 for how to handle this is the right call. Not sure how prominent the mention of PSR-19 in the inline comments need to be though considering this is not a PSR-19 sniff.

Regarding the if/else:
* The "passed by ref + has leading `&` in param tag" case is now special cased, which solves the original issue reported. ✅ 
* The "has no leading `&` in param tag" case is also handled correctly, whether passed by ref or not. ✅ 
* I'm just wondering about the "not passed by ref but has leading `&` in param tag" case. ❓ 

I agree that flagging a "not passed by ref but has leading `&` in param tag" situation is a good thing as there is clearly mismatch. I also agree that auto-fixing this may be risking.
I'm mostly just wondering whether this should get it's own error with error code as, as things are, the following errors are thrown:
```
 1050 | ERROR | [ ] Doc comment for parameter "$qux" missing (Squiz.Commenting.FunctionComment.MissingParamTag)
 1062 | ERROR | [ ] Doc comment for parameter &$qux does not match actual variable name $qux (Squiz.Commenting.FunctionComment.ParamNameNoMatch)
```

The first of which - IMO - is incorrect as the doc comment for `$qux` is not _missing_, it just has a leading `&`.
The second of which is unclear/imprecise - the variable name does in actual fact match - as it's not the actual variable name which is incorrect, but the leading ampersand.


Code style nitpick: I personally wouldn't use an if/else for this.
The "default" for `$paramVarName` is known and needs no special handling:
```php
$paramVarName = $param['var'];
if ($realParams[$pos]['pass_by_reference'] === true && $paramVarName[0] === '&') {
    // ...
    $paramVarName = substr($paramVarName, 1);
    // ...
}

```

fredden:
`MissingParamTag` - I have adjusted the sniff to not throw this when the variable is prefixed incorrectly in the `@param` tag. Should this still be thrown when only the case mismatches? For example,

```
1) PHP_CodeSniffer\Standards\Squiz\Tests\Commenting\FunctionCommentUnitTest::testSniff
[LINE 1050] Expected 0 error(s) in FunctionCommentUnitTest.inc but found 1 error(s). The error(s) found were:
 -> Doc comment for parameter "$CASE" missing (Squiz.Commenting.FunctionComment.MissingParamTag)
[LINE 1063] Expected 0 error(s) in FunctionCommentUnitTest.inc but found 1 error(s). The error(s) found were:
 -> Doc comment for parameter &$case does not match case of actual variable name &$CASE (Squiz.Commenting.FunctionComment.ParamNameNoCaseMatch)
```

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
jrfnl: If I were making these changes, I'd keep the "param has leading & while not passed by ref" vs the "param name mismatch" checks completely separate as these are effectively checking two completely different things.

So, yes, I'd do the name compare in a way to disregard a potential leading `&` completely.

jrfnl:
> `MissingParamTag` - I have adjusted the sniff to not throw this when the variable is prefixed incorrectly in the `@param` tag. Should this still be thrown when only the case mismatches? For example,

Parameters in PHP are case-sensitive, so yes, case mismatches should IMO definitely be flagged.

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -551,24 +551,45 @@ protected function processParams(File $phpcsFile, $stackPtr, $commentStart)
...
+
+                if ($param['var'][0] === '&') {
+                    // Even when passed by reference, the variable name in $realParams does not have
+                    // a leading '&'. This sniff will accept both '&$var' and '$var' in these cases.
+                    $paramVarName = substr($param['var'], 1);
+
+                    // This makes sure that the 'MissingParamTag' check won't throw a false positive.
+                    $foundParams[(count($foundParams) - 1)] = $paramVarName;
+
+                    if ($realParams[$pos]['pass_by_reference'] !== true) {
```
jrfnl, line 565:
```suggestion
                    if ($realName === $paramVarName && $realParams[$pos]['pass_by_reference'] !== true) {
```

Extra tests to verify why I'm suggesting this change:
```php
/**
 * Test for param tag containing ref, but param in declaration not being by ref.
 *
 * @param string &$foo This should be flagged as ParamNameUnexpectedAmpersandPrefix.
 * @param string &$bar This should be flagged as ParamNameNoMatch.
 * @param string &$baz This should be flagged as ParamNameNoCaseMatch.
 *
 * @return void
 */
function passedByRefMismatch($foo, $bra, $BAZ) {
	return;
}
```

jrfnl (changes requested):
@fredden Thanks for the updates. IMO, there is one more tweak needed.

src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
fredden: This change seems fair. I had considered doing this but decided against it at the time. The reasoning was that if the variable name was incorrect *and* had a rogue leading ampersand, that both faults would be flagged. With the suggested change (which I've committed in 28ed6aec1), users will get one fault identified at a time, so fixing one fault won't make all errors go away, but instead will make the error message change. I can see how having two errors is less desirable, especially in the case where the param tags are out of order and changing the order alone would fix all errors.

jrfnl: Good points and I agree that generally speaking errors shouldn't hide behind each other, but in this case, if the variable name doesn't match, the information about whether or not the variable is passed by reference from the function signature is just plain unreliable, which is why I suggested the change.

fredden:
Thanks for your time & patience on this one @jrfnl. I appreciate the feedback etc.

jrfnl:
@fredden Would you mind rebasing this PR (and possibly squashing the commits) to get it past the merge conflict ?

jrfnl:

PR #3818, opened 2023-05-06
Update BackfillNumericSeparatorTest.php

ETM44:
I am using a symfony 6 project with php 8.1.13.
The limit of an int is 9223372036854775807.

VS Code was giving me the following error:
The integer number 10223372036854775807 is too big, converting to double insteadPHP(PHP0009)

Why did you write 10 instead of 9?

### DIFF
--- a/tests/Core/Tokenizer/BackfillNumericSeparatorTest.php
+++ b/tests/Core/Tokenizer/BackfillNumericSeparatorTest.php
@@ -57,7 +57,7 @@ class BackfillNumericSeparatorTest extends AbstractMethodUnitTest
         }
 
         $testIntMoreThanMaxType = 'T_LNUMBER';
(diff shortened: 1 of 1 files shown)


jrfnl:
> Why did you write 10 instead of 9?

This would break the test. The number used in the test is `10223372036854775807`.

https://github.com/jrfnl/PHP_CodeSniffer/blob/bb9216e5a1b3d6bbb80af20e24fe93f07f3f2ecf/tests/Core/Tokenizer/BackfillNumericSeparatorTest.inc#L37-L38

The whole point of the test is to verify and ensure that the backfill retokenizes numbers which are larger than `PHP_INT_MAX` as `T_DNUMBER` (float) instead of `T_LNUMBER` (integer).

Closing as invalid.

PR #3831, opened 2023-05-20
Config: only determine screen width if/when needed (performance improvement)

jrfnl:
### Config: only determine screen width if/when needed

I'm not sure about *nix, but on Windows, the call to `shell_exec('stty ...')` is slow.

As things were, the call to `shell_exec('stty ...')` would be:
* Made when the initial default value for `reportWidth` is being set in `restoreDefaults()`.
* Potentially made a second time if the users `CodeSniffer.conf` file contained a `report_width => 'auto'` entry.
* Potentially made a third time if any of the rulesets used for the run contained a `<config name="report-width" value="auto"/>` entry.
* Potentially made a fourth time if `--report-width=auto` would be passed from the command-line.

This is inefficient for the following reasons:
1. The output from `shell_exec('stty ...')` does not change between calls (well, providing the user doesn't resize their terminal in the microseconds between calls)
2. We don't actually need to _know_ the value `'auto'` translates to, until the `reportWidth` will be _used_.

Taking the above into account, making the same call up to four times is not desirable.

This commit moves the translation from `'auto'` to an actual terminal width from the `__set()` method to the `__get()` method and overwrites the `reportWidth` value from `'auto'` with the actual terminal width value, if available, and with the default value if the terminal width could not be determined.
This means that subsequent calls to `__get()` for the `reportWidth` will automatically use the previously determined value instead of trying to determine value again.

This removes the inefficiency and should make PHPCS runs a little bit faster (at the very least on Windows).

The only time multiple calls to `shell_exec('stty...')` could now need to be made, would be if the `reportWidth` would be changed (back to `'auto'`) between the first retrieval and a subsequent retrieval of the `reportWidth` value. As things are, this will never happen in a normal PHPCS run, though could possibly happen in a test situation.

### AbstractMethodUnitTest: take advantage of the change in reportWidth handling

For the tests using the `AbstractMethodUnitTest` class, the `reportWidth` and most other config settings are irrelevant.

This commit changes some of the set up/tear down for the test to make the use of the `Config` class more efficient.

This should make the tests using the `AbstractMethodUnitTest` class as their base significantly faster (at the very least on Windows).

While not benchmarked properly, I have done some comparisons with the test runs on my local machine on Windows.

* `phpunit --filter Core` (= the tests which use this base class + a few extra tests):
    Before: **2 minutes**.
    After: **8 seconds**.
* The same effect can be seen when running `phpunit` without a `--filter`:
    Before: **7 minutes**.
    After: **5 minutes**.
* And when I apply a similar change to the one made here to the base test class in PHPCSUtils (4621 tests):
    Before: **2.5 minutes**.
    After: **1 second**.

### Suggested changelog entry
* Runtime improvement for PHPCS CLI users. The improvement should be most noticeable for users on Windows.

Follow up on #3761 for which tests were added in #3820.

## Types of changes

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -207,6 +207,22 @@ class Config
             throw new RuntimeException("ERROR: unable to get value of property \"$name\"");
         }
 
+        // Figure out what the terminal width needs to be for "auto".
+        if ($name === 'reportWidth' && $this->settings[$name] === 'auto') {
+            if (function_exists('shell_exec') === true) {
+                $dimensions = shell_exec('stty size 2>&1');
+                if (is_string($dimensions) === true && preg_match('|\d+ (\d+)|', $dimensions, $matches) === 1) {
+                    $this->settings[$name] = (int) $matches[1];
+                }
+            }
+
+            if ($this->settings[$name] === 'auto') {
+                // If shell_exec wasn't available or didn't yield a usable value, set to the default.
+                // This will prevent subsequent retrievals of the reportWidth from making another call to stty.
+                $this->settings[$name] = self::DEFAULT_REPORT_WIDTH;
+            }
+        }
+
         return $this->settings[$name];
 
     }//end __get()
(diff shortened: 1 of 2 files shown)


jrfnl:
FYI: I've made a tiny tweak to the second commit to make the change compatible with PHP 8.3.

jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/61

jrfnl:

PR #3843, opened 2023-06-20
Allow array unpacking in ArrayDeclaration multiline Sniff

### LINKED ISSUE #3557: Unpacking associative arrays in PHP 8.1 produces `Squiz.Arrays.ArrayDeclaration.NoKeySpecified` respectively `Squiz.Arrays.ArrayDeclaration.KeySpecified`
**Describe the bug**
Since PHP 8.1 unpacking of associative arrays is supported.
Unpacking an associative array into another results in one of the errors `Squiz.Arrays.ArrayDeclaration.NoKeySpecified` or `Squiz.Arrays.ArrayDeclaration.KeySpecified` (depending on the position).

**Code sample**
```php
<?php

$arr = [
        'a' => 1,
        'b' => 2,
       ];

$merged1 = [
            'c' => 3,
            ...$arr,
           ];

$merged2 = [
            ...$arr,
            'c' => 3,
           ];
```

**Custom ruleset**
```xml
<ruleset>
    <rule ref="Squiz.Arrays.ArrayDeclaration"/>
</ruleset>
```

**To reproduce**
```
-------------------------------------------------------------------------------------------------------------------------
-------------------------------------------------------------------------------------------------------------------------
 10 | ERROR | No key specified for array entry; first entry specifies key
    |       | (Squiz.Arrays.ArrayDeclaration.NoKeySpecified)
 15 | ERROR | Key specified for array entry; first entry has no key (Squiz.Arrays.ArrayDeclaration.KeySpecified)
-------------------------------------------------------------------------------------------------------------------------
```

**Expected behavior**
This should not result in an error:
* ideally PHP_CodeSniffer can check if the unpacked array is associative
* if thats not possible, array unpacking should be ignored by this rule

**Versions:**
 - OS: Windows 10
 - PHP: 8.1
 - PHPCS: 3.6.2
 - Standard: Squiz

edorian:
See #3557

## Description

`Squiz.Arrays.ArrayDeclaration.NoKeySpecified` doesn't accept the following snippet:

```php
$x = [
  'foo' => 'bar',
  ...$baz,
];
```

It triggers an error as phpcs doesn't detect that the second entry is array unpacking.

```
Key specified for array entry; first entry has no key (Squiz.Arrays.ArrayDeclaration.KeySpecified)
``` 

The following (single line) array validates:

```php
$x = ['foo' => 'bar',  ...$baz];
```

### Suggested changelog entry

`Squiz.Arrays.ArrayDeclaration.KeySpecified`: Allow array unpacking (`...$array`) in multi-line key-value array definitions.

Fixes #3557

## Types of changes

## PR checklist

- [x] I have added tests to cover my changes.
  - I did not. I'd be happy to, if you give me a pointer on how I can do this without breaking earlier PHP versions, given it's new syntax.
- [ ] [Required for new sniffs] I have added XML documentation for the sniff.
- [ ] [Required for new files] I have added any new files to the `package.xml` file.

## Notes:

The one test failure for PHP 8.2 seems unrelated as it was already present in https://github.com/squizlabs/PHP_CodeSniffer/commit/2e59060b93b3720ff3aab2b57efda0c3ada069b0

### DIFF
--- a/src/Standards/Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php
+++ b/src/Standards/Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php
@@ -430,9 +430,22 @@ class ArrayDeclarationSniff implements Sniff
                 }
 
                 if ($keyUsed === true && $tokens[$lastToken]['code'] === T_COMMA) {
-                    $error = 'No key specified for array entry; first entry specifies key';
-                    $phpcsFile->addError($error, $nextToken, 'NoKeySpecified');
-                    return;
+                    $nextToken = $phpcsFile->findNext(Tokens::$emptyTokens, ($lastToken + 1), null, true);
+
+                    /*
+                        Allow array unpacking
+
+                        $x = [
+                          'foo' => 'bar',
+                          ...$baz,
+                        ];
+                    */
+
+                    if ($tokens[$nextToken]['code'] !== T_ELLIPSIS) {
+                        $error = 'No key specified for array entry; first entry specifies key';
+                        $phpcsFile->addError($error, $nextToken, 'NoKeySpecified');
+                        return;
+                    }
                 }
 
                 if ($keyUsed === false) {
--- a/src/Standards/Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php
+++ b/src/Standards/Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php
@@ -474,7 +474,7 @@ class ArrayDeclarationSniff implements Sniff
                         true
                     );
 
-                    $indices[]  = ['value' => $valueContent];
+                    $indices[]          = ['value' => $valueContent];
                     $usesArrayUnpacking = $phpcsFile->findPrevious(
                         Tokens::$emptyTokens,
                         ($nextToken - 2),
--- a/src/Standards/Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php
+++ b/src/Standards/Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php
@@ -475,7 +475,16 @@ class ArrayDeclarationSniff implements Sniff
                     );
 
                     $indices[]  = ['value' => $valueContent];
-                    $singleUsed = true;
+                    $usesArrayUnpacking = $phpcsFile->findPrevious(
+                        Tokens::$emptyTokens,
+                        ($nextToken - 2),
+                        null,
+                        true
+                    );
+                    if ($tokens[$usesArrayUnpacking]['code'] !== T_ELLIPSIS) {
+                        // Don't decide if an array is key => value indexed or not when PHP 7.4+ array unpacking is used.
+                        $singleUsed = true;
+                    }
                 }//end if
 
                 $lastToken = $nextToken;
--- a/src/Standards/Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php
+++ b/src/Standards/Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php
@@ -431,16 +431,7 @@ class ArrayDeclarationSniff implements Sniff
 
                 if ($keyUsed === true && $tokens[$lastToken]['code'] === T_COMMA) {
                     $nextToken = $phpcsFile->findNext(Tokens::$emptyTokens, ($lastToken + 1), null, true);
-
-                    /*
-                        Allow array unpacking
-
-                        $x = [
-                          'foo' => 'bar',
-                          ...$baz,
-                        ];
-                    */
-
+                    // Allow for PHP 7.4+ array unpacking within an array declaration.
                     if ($tokens[$nextToken]['code'] !== T_ELLIPSIS) {
                         $error = 'No key specified for array entry; first entry specifies key';
                         $phpcsFile->addError($error, $nextToken, 'NoKeySpecified');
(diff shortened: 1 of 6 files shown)


jrfnl:
> I did not. I'd be happy to, if you give me a pointer on how I can do this without breaking earlier PHP versions, given it's new syntax.

@edorian Please do add some new tests. The fact that it is new syntax is irrelevant as the PHPCS tokenizer has backported the `T_ELLIPSIS` token to PHP 5.4, so the test code should tokenize the same in all PHP versions.

(if that is not the case, we have another issue, but better to know about it than not)

edorian:
[x] Test

Without my fix:

```
PHP_CodeSniffer\Standards\Squiz\Tests\Arrays\ArrayDeclarationUnitTest::testSniff
[LINE 475] Expected 0 error(s) in ArrayDeclarationUnitTest.2.inc but found 1 error(s). The error(s) found were:
 -> No key specified for array entry; first entry specifies key (Squiz.Arrays.ArrayDeclaration.NoKeySpecified)
```

With my fix :heavy_check_mark:

src/Standards/Squiz/Sniffs/Arrays/ArrayDeclarationSniff.php
```diff
@@ -430,9 +430,22 @@ public function processMultiLineArray($phpcsFile, $stackPtr, $arrayStart, $array
...
+
+                    /*
+                        Allow array unpacking
+
+                        $x = [
+                          'foo' => 'bar',
+                          ...$baz,
+                        ];
+                    */
+
```
jrfnl, line 443:
```suggestion
                    // Allow for PHP 7.4 array unpacking with an array declaration.
```

I don't think the code sample is needed here.

src/Standards/Squiz/Tests/Arrays/ArrayDeclarationUnitTest.2.inc
```diff
@@ -468,3 +468,9 @@ yield [
 $a = [
       'a' =>
      ];
+
+// Ensure absence of NoKeySpecified error for array unpacking
```
jrfnl, line 472:
```suggestion
// Ensure absence of NoKeySpecified error for PHP 7.4+ array unpacking.
```

jrfnl (changes requested):
@edorian Hi Volker, thanks for this PR.

I can confirm the current fix is correct and working.

I've left some minor comments inline, but nothing blocking.

One thing I would like to ask - could you add the test to both the test case files ?

The `ArrayDeclarationUnitTest.1.inc` test case file is supposed to have long array syntax, while the `ArrayDeclarationUnitTest.2.2inc` test case file is supposed to have those same tests using short array syntax.

It does look like the files have diverged a little over time, but I do think it's good to have the test(s) in both files.

And looking at the test, I also wonder if there should also be a test with the array unpacking at the start of the array declaration and one where the array unpacking happens in the "middle" of an array ?

As things are, this PR addresses just one part of issue #3557 (`NoKeySpecified`), while the other part (`KeySpecified`) is not yet fixed.

edorian:
@jrfnl I've implemented your suggestions, fixed the KeySpecified error and added test cases.

I've done so by basically saying "if it's array unpacking line don't decide if it's a 'list' or a 'dict' shaped array".

Doesn't look very clean looking at the diff but it passes all tests I could come up with :)

src/Standards/Squiz/Tests/Arrays/ArrayDeclarationUnitTest.1.inc
```diff
@@ -486,3 +486,20 @@ $foo = [
 $a = array(
       'a' =>
      );
+
+// Ensure absence of NoKeySpecified error for PHP 7.4+ array unpacking.
```
jrfnl, line 490: This comment only refers to the one error code, while both are now handled. Maybe something like this ?
```suggestion
// Safeguard correct errors for key/no key when PHP 7.4+ array unpacking is encountered.
```

jrfnl (commented):
@edorian Thanks for this update. The fixes look good.

I'm just thinking that there should probably be some additional tests still to safeguard that when an array entry is ignored due to unpacking, the key/no key for literal array entries is still correct (which it is with your fixes, but safeguarding that is the idea).

What about adding these additional tests ?

```php
$x = [
      ...$a,
      'foo' => 'bar', // OK.
      'bar', // Error about no key (based on second entry).
     ];

$x = [
      ...$a,
      'bar', // OK.
      'foo' => 'bar', // Error about key (based on second entry).
     ];

$x = [
      'foo' => 'bar',
      ...$a,
      'baz' => 'bar',
      'bar', // Error about no key (based on first entry).
     ];

$x = [
      'bar',
      ...$a,
      'bar',
      'baz' => 'bar', // Error about key (based on first entry).
     ];
```

P.S.: thanks for taking on this issue, this is one of the sniffs which is unwieldy to adjust in most cases. Also see #582.

edorian:
Added your suggested tests and tried to align the two test files, but the diff of that grew to big for this PR. 

Happy to do that in a follow up when this is merged :) 

@jrfnl Thank you for the fast response an the clear directions!

src/Standards/Squiz/Tests/Arrays/ArrayDeclarationUnitTest.2.inc
```diff
@@ -464,7 +464,57 @@ yield [
        static fn () : string => '',
       ];
 
+$foo = [
+        'foo' => match ($anything) {
+            'foo' => 'bar',
+            default => null,
+        },
+       ];
+
```
edorian, line 473: This is aligning 1.inc and 2.inc

jrfnl:
@edorian Uh oh.. the tests are failing now. Could it be that the test case comment was only applied to the one test case file, not to all four ?

jrfnl (approved):
> Didn't commit/push all the changes.

@edorian Happens to the best of us.

All good now! Verified the test fail without the fix and pass with the correct expected notices with the fixes. ✅

jrfnl:

PR #3867, opened 2023-08-05
Tokenizer/PHP: bug fix for static typed properties with union/intersection types

jrfnl:
## Description
Just like the `var` keyword, the `static` keyword can also be used stand-alone with property declarations. https://3v4l.org/sbaDM

In that case, the tokenization of the `|` operator was not changed to `T_TYPE_UNION` and the `&` operator was not changed to `T_TYPE_INTERSECTION` as the `static` keyword can also be used in return type declarations, so was seen as part of the type declaration.

Fixed now by removing the `T_STATIC` token from the `$allowed` list before walking backwards from the operator.

Includes tests.

Note: this does mean that one test for the `File::getMemberProperties()` method needs to be changed, but as that was testing an illegal syntax anyway, I'm not concerned about making this change.

### Suggested changelog entry
Tokenizer/PHP: union type and intersection type operators were not correctly tokenized as such for static typed properties without explicit visibility.

## Types of changes

### DIFF
--- a/src/Tokenizers/PHP.php
+++ b/src/Tokenizers/PHP.php
@@ -2916,6 +2916,10 @@ class PHP extends Tokenizer
                     continue;
                 }
 
+                if ($suspectedType === 'property or parameter') {
+                    unset($allowed[\T_STATIC]);
+                }
+
                 $typeTokenCount = 0;
                 $typeOperators  = [$i];
                 $confirmed      = false;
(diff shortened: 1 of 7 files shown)


jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/78

jrfnl:

PR #3868, opened 2023-08-06
Ruleset::explain(): use natural sorting

jrfnl:
## Description
Use natural sorting for the list of included sniffs.

The `SORT_NATURAL` and the `SORT_FLAG_CASE` flag are available since PHP 5.4, so can be used without issue.

### Suggested changelog entry
The -e (explain) command will now list sniffs in natural order.

## Types of changes

### DIFF
--- a/src/Ruleset.php
+++ b/src/Ruleset.php
@@ -238,7 +238,7 @@ class Ruleset
     public function explain()
     {
         $sniffs = array_keys($this->sniffCodes);
-        sort($sniffs);
+        sort($sniffs, (SORT_NATURAL | SORT_FLAG_CASE));
 
         ob_start();
 


src/Ruleset.php
```diff
@@ -238,7 +238,7 @@ public function __construct(Config $config)
     public function explain()
     {
         $sniffs = array_keys($this->sniffCodes);
-        sort($sniffs);
+        sort($sniffs, (SORT_NATURAL | SORT_FLAG_CASE));
```
fredden, line 241: Would using `natcasesort()` be more legible here?

fredden (approved):

src/Ruleset.php
jrfnl: Just checked, yes, the function is available, but is still not the right choice.

`natcasesort()` maintains key association. `sort()` does not.
The `foreach()` loop used in the function uses the (numeric) key to determine whether the end of the array has been used, so the array not being re-indexed would cause it to stop the loop too early or even to display sniffs with the wrong standard.

DannyvdSluijs (approved):
Approved. The output now shows the PSR2 before the PSR12  sniffs.
```diff
--- /tmp/master.out     2023-08-28 20:52:17
+++ /tmp/diff.out       2023-08-28 20:53:08
@@ -43,16 +43,16 @@
   PEAR.WhiteSpace.ScopeClosingBrace
   PEAR.WhiteSpace.ScopeIndent
 
-PSR12 (1 sniff)
-----------------
-  PSR12.Files.OpenTag
-
 PSR2 (3 sniffs)
 ---------------
   PSR2.Classes.PropertyDeclaration
   PSR2.Files.EndFileNewline
   PSR2.Methods.MethodDeclaration
 
+PSR12 (1 sniff)
+----------------
+  PSR12.Files.OpenTag
+
 Squiz (25 sniffs)
 -----------------
   Squiz.Arrays.ArrayBracketSpacing
```

jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/79

PR #3870, opened 2023-08-09
Tokenizer::$eolChar: change default value

jrfnl:
## Description

Tokenizer::$eolChar: change default value to be a more logical one. The property is supposed to be a `string`, so having an array as a default value is confusing.

### Suggested changelog entry
_N/A_

## Types of changes

### DIFF
--- a/src/Tokenizers/Tokenizer.php
+++ b/src/Tokenizers/Tokenizer.php
@@ -27,7 +27,7 @@ abstract class Tokenizer
      *
      * @var string
      */
-    protected $eolChar = [];
+    protected $eolChar = '';
 
     /**
      * A token-based representation of the content.


src/Tokenizers/Tokenizer.php
```diff
@@ -27,7 +27,7 @@ abstract class Tokenizer
      *
      * @var string
      */
-    protected $eolChar = [];
+    protected $eolChar = '';
```
fredden, line 30: Should the default value be non-empty? What are your thoughts on using `PHP_EOL` here instead of an empty string. Or perhaps `null` or unset.

fredden: Thanks for the explanation. This all makes sense. :+1:

fredden (approved):

DannyvdSluijs (approved):

DannyvdSluijs:
Given my "approval" based on the description of the PR and the discussion about a default `EOL` character which all made sense.

jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/81

PR #3882, opened 2023-08-28
Inject the Config in the constructor

BafS:
Inject the configuration object in the constructor instead of creating the instance without having the ability to change it.

Currently it's not possible to change the configuration because it's using in a "black box" way. This PR proposes to use dependency injection to have more control over it when using the library in a programmatic way.

### Suggested changelog entry
- Add the ability to inject the Config object in the Runner class

Fixes #

## Types of changes

- [x] New feature _(non-breaking change which adds functionality)_

## PR checklist

- [x] I have added tests to cover my changes. (should not be needed in this case, already covered)
- [ ] [Required for new sniffs] I have added XML documentation for the sniff.
- [ ] [Required for new files] I have added any new files to the `package.xml` file.

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -45,6 +45,20 @@ class Runner
      */
     public $reporter = null;
 
+    /**
+     * @param \PHP_CodeSniffer\Config|null $config Config object
+     */
+    public function __construct($config = null)
+    {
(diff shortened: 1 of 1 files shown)


jrfnl:
Marking as "Close candidate" - see the discussion in #3697

PR #3905, opened 2023-10-25
Squiz/FunctionSpacing: bug fix - prevent fixer conflict with itself

### LINKED ISSUE #3904: Autofixer for Squiz.WhiteSpace.FunctionSpacing goes into "infinite" loop for method defined on the same line as a class
## Describe the bug
In the (intentionally bad) snippet of code below, I have a method defined on the same line as its class. When running PHPCS on it with the `Squiz.WhiteSpace.FunctionSpacing` sniff enabled (ruleset below), it reports that 1 blank line is expected before the method. However, if I then try to autofix that with phpcbf, it tries to add the line in the wrong place, thus going into an infinite loop because it detects that the issue still hasn't been fixed; it only aborts upon reaching the limit of 50 passes.

### Code sample
```php
<?php

class Person { public function __construct($name){} }

```

### Custom ruleset
```xml
<?xml version="1.0"?>
<ruleset>
	<arg name="tab-width" value="4" />

	<rule ref="Squiz.WhiteSpace.FunctionSpacing">
		<properties>
			<property name="spacing" value="1" />
		</properties>
	</rule>
</ruleset>
```

### To reproduce
2. Run `phpcs test.php`
```
 3 | ERROR | [x] Expected 1 blank line before function; 0 found
```
4. Run `phpcbf -vvv test.php`
5. See that it fails to fix the file. The full output shows that it's adding the line right after the PHP open tag:

```
(... all the previous passes...)
---START FILE CONTENT---
 1|<?php
 2|
 3|
 4|
 5|
 6|
 7|
 8|
 9|
10|
11|
12|
13|
14|
15|
16|
17|
18|
19|
20|
21|
22|
23|
24|
25|
26|
27|
28|class Person { public function __construct($name){} }
29|
30|
--- END FILE CONTENT ---
	Squiz.WhiteSpace.FunctionSpacing:342 replaced token 0 (T_OPEN_TAG on line 1) "<?php\n" => "<?php\n\n"
        => Fixing file: 1/1 violations remaining [made 50 passes]...            
	* fixed 1 violations, starting loop 51 *
	*** Reached maximum number of loops with 1 violations left unfixed ***
ERROR in 33ms

PHPCBF RESULT SUMMARY
(...)/test.php            FAILED TO FIX
A TOTAL OF 0 ERRORS WERE FIXED IN 1 FILE
PHPCBF FAILED TO FIX 1 FILE

```

## Expected behavior
Ideally, the issue should be fixed. Or if that's hard to do, at least PHPCS should not go into an endless loop, and instead report the issue as unfixable.

## Versions (please complete the following information)

| | |
|-|-|
| Operating System | Ubuntu 22.04
| PHP version | 8.2.11
| PHP_CodeSniffer version | master
| Standard | See above
| Install type | Composer local

## Please confirm:

- [x] I have searched the issue list and am not opening a duplicate issue.
- [x] I confirm that this bug is a bug in PHP_CodeSniffer and not in one of the external standards.
- [x] I have verified the issue still exists in the `master` branch of PHP_CodeSniffer.

jrfnl:
## Description

The `Squiz.WhiteSpace.FunctionSpacing` sniff demands # lines above and below a function declaration, but when determining the number of lines above or below, it does not care for additional code on the same line as the function declaration. I.e.: it does not demand that a function declaration starts on its own line or that the close brace is on its own line (there are other sniffs for that).

The sniff also tries to prevent "double" errors for the same issue, i.e. when two functions are each on their own line without blank lines between them, the sniff will only throw one error for "spacing after" the first function and will not throw an error for "spacing before" for the second function.

To determine whether the "spacing before" error needs to be hidden, the sniff tries to check whether the tokens on the previous line indicate the line contains a function declaration.

As things were, however, this check could _bow out_ too early as it stopped at a scope opener for a wrapping construct (class), however, that wrapping construct _could_ be declared on the same line as the function, which means that in that case, the sniff would not determine the `$foundLines` correctly, as it stops on the current line at the scope opener instead of on a non-blank line above the function line.

This commit fixes this bug by changing the `$stopAt` value to contain the wrapping scope opener applicable to the code _before_ the function line. This allows the `$foundLines` to be determined correctly and prevents the fixer conflict.

Includes additional tests.

### Suggested changelog entry

Fixed bug #3904 : Squiz/FunctionSpacing: prevent potential fixer conflict

Fixes #3904

## Types of changes

### DIFF
--- a/src/Standards/Squiz/Sniffs/WhiteSpace/FunctionSpacingSniff.php
+++ b/src/Standards/Squiz/Sniffs/WhiteSpace/FunctionSpacingSniff.php
@@ -262,21 +262,20 @@ class FunctionSpacingSniff implements Sniff
                 $prevContent = $phpcsFile->findPrevious(T_WHITESPACE, ($tokens[$prevContent]['comment_opener'] - 1), null, true);
             }
 
-            $prevLineToken = $prevContent;
-
             // Before we throw an error, check that we are not throwing an error
             // for another function. We don't want to error for no blank lines after
             // the previous function and no blank lines before this one as well.
-            $prevLine   = ($tokens[$prevContent]['line'] - 1);
-            $i          = ($stackPtr - 1);
-            $foundLines = 0;
-
             $stopAt = 0;
-            if (isset($tokens[$stackPtr]['conditions']) === true) {
-                $conditions = $tokens[$stackPtr]['conditions'];
+            if (isset($tokens[$prevLineToken]['conditions']) === true) {
+                $conditions = $tokens[$prevLineToken]['conditions'];
                 $conditions = array_keys($conditions);
                 $stopAt     = array_pop($conditions);
             }
+
+            $prevLineToken = $prevContent;
+            $prevLine      = ($tokens[$prevContent]['line'] - 1);
+            $i          = ($stackPtr - 1);
+            $foundLines = 0;
 
             while ($currentLine !== $prevLine && $currentLine > 1 && $i > $stopAt) {
                 if ($tokens[$i]['code'] === T_FUNCTION) {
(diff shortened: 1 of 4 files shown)


jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/93

jrfnl:

PR #3911, opened 2023-10-28
Various sniffs: always return EOF pointer

jrfnl:
## Description

Always return EOF pointer for files scanning the complete file in one go as if any of the error conditions are hit, it is useless to try again for the same file.

### Suggested changelog entry
_N/A_


fredden (approved):

jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/95

PR #3912, opened 2023-10-28
Tests: remove unreachable code

jrfnl:
## Description
`break` after a `return` statement is redundant.

As a side-note: I did have to chuckle when I saw that the `NonExecutableCode` test file also included this :joy:

### Suggested changelog entry
_N/A_

PR #3914, opened 2023-10-29
Respect all sniffs when reviewing PHP_CodeSniffer itself

fredden:
## Description
While reviewing https://github.com/squizlabs/PHP_CodeSniffer/pull/3912, I was surprised that `Squiz.PHP.NonExecutableCode` did not seem to complain about anything when running this over the main branch (before applying the changes from that pull request). Upon further investigation, I found that the coding standard included a parameter of `-n`, which ignores all warnings by default. Removing this allowed me to properly review #3912.

This pull request removes the `-n` parameter and makes the necessary changes to the code-base (including the rule-set) so that no warnings are omitted. This means that when we add `Squiz.PHP.NonExecutableCode` to the rule-set for PHP_CodeSniffer itself, the warnings it produces won't be ignored.

### Suggested changelog entry
Respect warnings as well as errors from sniffs within the coding standard for PHP_CodeSniffer itself.

## Types of changes
- Bug fix _(non-breaking change which fixes an issue)_

## PR checklist

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -32,6 +32,7 @@
         <exclude name="PEAR.Commenting.FileComment.MissingLinkTag"/>
         <exclude name="PEAR.Commenting.FileComment.MissingVersion"/>
         <exclude name="PEAR.Commenting.InlineComment"/>
+        <exclude name="Generic.Files.LineLength"/>
     </rule>
 
     <!-- Include some sniffs from other standards that don't conflict with PEAR -->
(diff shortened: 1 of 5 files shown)


jrfnl (commented):
@fredden I see where you are coming from, but I'm not keen on accepting this.

On the one hand, I presume the `n` was added for a reason, so I'm not keen on removing it (though I run locally with `w` to show warnings anyway).

On the other hand, the "fixes" being made are IMO wrong. The `@link` tag is the correct tag to use for the links and changing the `TODO` to `to-do`, while the tag name is `@todo` feels wrong too.

I think the better approach would be for me to update PR #3912 to change the warning from that sniff to an error.

fredden:
Superseded by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/122

PR #3915, opened 2023-10-31
File::addMessage(): do not ignore Internal errors when scanning selectively

jrfnl:
### ErrorSuppressionTest: prevent Internal errors

... about a mismatch in line endings when the code "template" is defined using a heredoc with Linux line endings, while the code snippets being inserted into the code "template" were using line endings matching the OS on which the tests were being run.

### File::addMessage(): do not ignore Internal errors when scanning selectively

When either the `--sniffs=...` CLI parameter is used, or the `--exclude=...` CLI parameter, the `File::addMessage()` method bows out when an error is passed which is not for one of the selected sniffs/is for one of the excluded sniffs.

Unfortunately, this "bowing out" did not take `Internal` errors into account, meaning those were now hidden, while those - IMO - should _always_ be thrown as they generally inform the end-user of something wrong with the file which impacts the scan results (mixed line endings/no code found etc).

Fixed now.

Includes updating two test files to allow for seeing internal errors.

### Suggested changelog entry
Internal errors will no longer be suppressed when the `--sniffs` or `--exclude` CLI arguments are used.

## Types of changes

👉🏻 This _could_ be considered a breaking change, though I'm not sure whether it should be. Opinions welcome.

Most typically, this change may impact tests for external standards which don't expect `Internal` errors.

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -909,6 +909,7 @@ class File
         // Filter out any messages for sniffs that shouldn't have run
         // due to the use of the --sniffs command line argument.
         if ($includeAll === false
+            && $parts[0] !== 'Internal'
             && ((empty($this->configCache['sniffs']) === false
             && in_array(strtolower($listenerCode), $this->configCache['sniffs'], true) === false)
             || (empty($this->configCache['exclude']) === false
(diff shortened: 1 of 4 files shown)


jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/98

PR #3916, opened 2023-10-31
Don't stop scan on invalid inline property annotation

jrfnl:
## Description
Follow up on #3629, which was merged for PHPCS 3.8.0.

PR #3629 added logic to throw a "Ruleset invalid. Property \"$propertyName\" does not exist on sniff ..." error.

This error is intended for the command-line when reading the `phpcs.xml.dist` ruleset file.

However, this error could _also_ be encountered if an inline `// phpcs:set ...` annotation would try to set a non-existent property.

While the use of `// phpcs:set` is typically reserved for sniff test case files, there is nothing stopping end-users from using the annotation.

The net-effect would be:
* The `Ruleset::setSniffProperty()` throws a `RuntimeException`.
* This exception is then passed to `File::addMessage()` where it is **not** thrown as the line on which the error is being thrown is an annotation line.
* The scan of the file stops dead in its tracks as a `RuntimeException` was encountered.
* The end-user doesn't know the file does not finish scanning as no `Internal` error is shown for the file.

To me, this is counter-intuitive and counter-productive as it may give people a false sense of security (CI is green, while in reality files are not being scanned).

To fix this, I propose the following:
* Collect all `// phpcs:set` related inline annotations encountered while scanning.
* Do **not** stop the file scan for these errors.
* Add a warning with information about the incorrect annotations on line 1 once the file has finished scanning.

Includes a test via the `Generic.PHP.BacktickOperator` sniff.

### Suggested changelog entry
I'd suggest updating the existing changelog entry for the change from PR #3629 and adding:
```
    -- Invalid sniff properties set for sniffs via inline annotations will result in an informative `Internal.PropertyDoesNotExist` errror on line, but will not halt the execution of PHPCS
```

## Types of changes

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -343,6 +343,7 @@ class File
         $listenerIgnoreTo = [];
         $inTests          = defined('PHP_CODESNIFFER_IN_TESTS');
         $checkAnnotations = $this->config->annotations;
+        $annotationErrors = [];
 
         // Foreach of the listeners that have registered to listen for this
         // token, get them to process it.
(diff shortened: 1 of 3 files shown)


jrfnl:
Closing as replaced by https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/99

jrfnl:

